PR #28, opened 2019-11-21
Implement renderables

dragly:


f3df/src/renderables.rs
```diff
@@ -0,0 +1,203 @@
...
+                instance_data.push(Face {
+                    color: [
+                        f32::from(color[0]) / 255.0,
+                        f32::from(color[1]) / 255.0,
+                        f32::from(color[2]) / 255.0,
+                    ],
+                    normal,
+                    matrix: compose_face_matrix(&center, &offset, &scale, &rotation),
+                });
+            }
```
larsmoa: Can this be a separate function accepting offset, normal, rotationAxis, rotationAngle, FaceFlags. CompressFlags or something similar?

PR #35, opened 2019-11-29
Improve support for primitives

dragly:


dragly:
The latest commits that improve build speed are also in #36, which can be reviewed separately. I added them here as well because they make the project much more pleasant to work with.

PR #36, opened 2019-11-29
Improve build speed

dragly:
This reduces the build time from 60 s to 6 s when files like
i3df/src/renderables.rs change, making the feedback loop a lot more
comfortable to work with. In addition, it reduces the build time for initial builds by extracting certain functions to reduce duplication in the generated file.

### DIFF
--- a/i3df/build.rs
+++ b/i3df/build.rs
@@ -1,13 +1,13 @@
 use heck::CamelCase;
 use i3df_specification;
+use indexmap::IndexMap;
 use quote::{format_ident, quote};
 use serde_yaml;
-use std::collections::{HashMap};
+use std::collections::HashMap;
 use std::env;
 use std::error::Error;
 use std::fs;
 use std::io::Write;
-use indexmap::IndexMap;
 
 use proc_macro2::TokenStream;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


i3df/build.rs
```diff
@@ -81,7 +84,7 @@ fn create_dtype(
             "f32" => quote! {f32},
             t => quote! { ERROR_NOT_IMPLEMENTED: #t },
```
f2cognite: Maybe nitpicking in existing code here, but would it make sense to use [`compile_error!`](https://doc.rust-lang.org/std/macro.compile_error.html) here for a better error message?
```suggestion
             t => {
               let message = format!("Type not implemented: {}", t);
               quote!{ compile_error!(#message) }
             },
```

i3df/build.rs
```diff
@@ -67,21 +67,27 @@ fn create_dtype(
...
-    // TODO implement texture
     let type_name = match &attribute._type {
         Type::Simple(t) => match t.as_str() {
             "u8" => quote! {u8},
             "u32" => quote! {u32},
             "u64" => quote! {u64},
             "f32" => quote! {f32},
-            t => quote! { ERROR_NOT_IMPLEMENTED: #t },
+             t => {
+               let message = format!("Type not implemented: {}", t);
```
dragly:
```suggestion
```

i3df/build.rs
```diff
@@ -67,21 +67,27 @@ fn create_dtype(
...
+               quote!{ compile_error!(#message) }
```
dragly:
```suggestion
               panic!("Type not implemented: {}", t);
```

PR #42, opened 2019-12-06
Update F3DF version with texture support

dragly:
And pass texture information in improved way from I3DF parser to JS.


dragly:
This includes parts of #43 which should be merged first.

PR #54, opened 2019-12-10
Move examples out of viewer and into root folder

dragly:
This allows us to reduce the size of our viewer when bundling. It also
ensures the examples are written in a way similar to how others
would write applications that depend on the viewer.


viewer/src/index.ts
```diff
@@ -8,3 +8,9 @@ export { createLocalSectorModel, createLocalPointCloudModel } from './datasource
 
 export { SectorNode } from './views/threejs';
 export { createThreeJsSectorNode, createThreeJsPointCloudNode } from './views/threejs';
+export { initializeCesiumSectorScene } from './views/cesiumjs/createCesiumSectorNode';
+
+// @ts-ignore
```
larsmoa: Why ts-ignore?

dragly: I agree. I think this was necessary because of some limitation in import -> export in TypeScript vs ECMAScript. Let us try to remove it and see what fails before merging. If necessary, we should leave a comment about this for later.

PR #58, opened 2019-12-13
Remove faulty synchronization of sector consume and discard

dragly:
This implementation currently leaves us with sectors that have both
detailed and simple data shown at the same time. It should be removed
until a better implementation is in place.


viewer/src/views/threejs/sector/createThreeJsSectorNode.ts
```diff
@@ -37,13 +37,6 @@ export async function createThreeJsSectorNode(model: SectorModel): Promise<Secto
     rootGroup.needsRedraw = true;
   };
 
-  // Sync high- and low-detail geometry
-  const [discardSectorFinal, consumeSectorFinal, consumeSectorQuadsFinal] = createSyncedConsumeAndDiscard(
```
larsmoa: Should we even bother to keep the `createSyncedConsumeAndDiscard`-implementation around?

dragly: Nope. I will remove that now.

PR #60, opened 2019-12-16
Do not override allocator

dragly:
Turns out overriding alloc_zeroed to call alloc is in fact dangerous and
leads to corrupted data. This change removes the custom allocator.


larsmoa:
This is sad.

PR #62, opened 2019-12-17
Improve performance by only consuming one sector per frame

dragly:
Also, combine fetching and parsing in the cache. For the moment, remove the ability to cancel a request in-flight. We might want to add this back later with a reactive framework instead.


viewer/src/examples/threejs/sector-with-pointcloud.ts
```diff
@@ -56,16 +56,16 @@ async function main() {
   camera.updateMatrixWorld();
 
   const clock = new THREE.Clock();
-  const render = () => {
+  const render = async () => {
     requestAnimationFrame(render);
```
larsmoa: How does an async callback to requestAnimationFrame() work?

dragly: We should probably wait with requesting another animation frame until the end of the render function - I do not think it is necessary, but should improve sanity a bit.

PR #65, opened 2019-12-19
Fix determine sector quality

dragly:
In addition, introduce the concept of a Scene to give some functions access to the list of sectors as well as the root sector. Further, add a script that creates an uploaded_sectors_simple.txt file based on a folder of .f3d files.


dragly:
@larsmoa opening this as draft in case you have some comments already. Would for instance be nice if we could keep the children and parent members on SectorMetadata readonly, while still being able to modify them during construction in the tests.

PR #75, opened 2019-12-20
Build examples in Travis builds

larsmoa:


dragly:
Could we do the nvm 10 and 12 in parallel to reduce the build time?

larsmoa:
> Could we do the nvm 10 and 12 in parallel to reduce the build time?

@dragly: In a per-build sense; probably a bit, but in practice I believe this actually would slow things down as it requires duplicated cache initialisation and cargo builds. This will be parallelised as long as there isn't any other ongoing builds, but would hog an unnecessary amount of resources.

PR #85, opened 2020-01-10
Add treeIndex to triangle meshes

larsmoa:


dragly:
As mentioned, we should consider using either indices or indexes, but otherwise this is fine :)

PR #86, opened 2020-01-11
Tree indices for all mesh types

larsmoa:


.vscode/launch.json
```diff
@@ -15,7 +15,7 @@
             "type": "chrome",
             "request": "launch",
             "name": "Launch Chrome against localhost",
-            "url": "https://localhost:8080/example-threejs-simple.html",
```
dragly: Should perhaps not be changed?

larsmoa: Oops - that's a mishap...

PR #89, opened 2020-01-14
Add simple dat.GUI based widget for inspecting scene complexity to easy performance debugging

larsmoa:


examples/src/threejs/utils/renderer-debug-widget.ts
```diff
@@ -0,0 +1,275 @@
...
+}
+type SceneInfo = ReturnType<typeof createEmptySceneInfo>;
+
+export function createRendererDebugWidget(
+  renderer: THREE.WebGLRenderer,
+  scene: THREE.Object3D,
+  parentGui?: dat.GUI,
+  intervalMs: number = 100
+): RenderOptions {
+  const gui = parentGui || new dat.GUI();
```
dragly: Will this automatically appear on screen? If not, I think `gui` needs to be returned from this function.

dragly: You can also make `parentGui` required instead. I do not think you use the optional version anywhere currently.

scripts/generate_uploaded_f3d.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 TARGET_DIR="$1"
 TARGET_FILE=$TARGET_DIR/uploaded_sectors_simple.txt
-rm -f $TARGET_FILE 2&>1 /dev/null
+rm -f $TARGET_FILE 2>&1 /dev/null
```
dragly: Nitpick: Should perhaps be in a separate PR, but feel free to leave it in

PR #91, opened 2020-01-21
Enable overriding sectors to load using regex filters in side-by-side example and introduce render and loading hints.

larsmoa:


examples/webpack.config.js
```diff
@@ -41,7 +41,7 @@ const allExamples = [
     name: "threejs-simple",
     title: "Simple",
     entry: './src/threejs/simple.ts',
-    template: 'template-example.ejs',
+    template: 'template-example.ejs',
```
dragly: Nitpick: remove

PR #93, opened 2020-01-21
toThreeVector3() now supports an optional out-parameter to avoid allocations

larsmoa:
…cations.


larsmoa:
> Should we add it as the first parameter, to align with `fromThreeVector3` and the `gl-matrix` library?

Done.

PR #95, opened 2020-01-23
Run tsc in Travis to avoid unused .ts files left behind

dragly:


dragly:
This now depends on https://github.com/cognitedata/potree-core/pull/3

PR #102, opened 2020-01-29
Support for new file format where buffers are stored in each sector and not only the root sector

larsmoa:


i3df/src/bin/dump.rs
```diff
@@ -31,21 +30,10 @@ enum Output {
...
-    let attributes = match root_sector.header.attributes {
-        Some(x) => x,
-        None => {
-            return Err(Box::new(i3df::error::Error::new(
-                "Attributes missing on root sector",
-            )))
-        }
-    };
-    let raw_sector = i3df::parse_sector(&attributes, sector_reader)?;
+    let raw_sector = i3df::parse_root_sector(sector_reader)?;
```
dragly: Let us rename the functions in i3df while we are at it. I know I will forget to do so otherwise :stuck_out_tongue:

PR #103, opened 2020-01-29
Share color and tree index arrays between meshes

dragly:
Previously, we would allocate arrays large enough to contain color and tree index information about all other meshes for each mesh.

### DIFF
--- a/viewer/src/workers/parser.worker.ts
+++ b/viewer/src/workers/parser.worker.ts
@@ -121,10 +121,15 @@ export class ParserWorker {
     const rust = await rustModule;
     // TODO handle parsing failure
     const ctm = rust.parse_ctm(buffer);
+    const indices = ctm.indices();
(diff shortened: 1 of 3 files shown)


viewer/src/workers/parser.worker.ts
```diff
@@ -108,10 +108,15 @@ export class ParserWorker {
...
+    const vertices = ctm.vertices();
+    const normals = ctm.normals();
     const result = {
-      indices: ctm.indices(),
-      vertices: ctm.vertices(),
-      normals: ctm.normals()
+      indices,
+      vertices,
+      normals,
+      colors: new Float32Array(indices.length),
```
larsmoa: Doesn't this buffer need to be 3*indices.length? Also, note that we should look into (separate from this PR) is to replace with Uin8

larsmoa: @dragly

PR #105, opened 2020-02-03
Remove support for Cesium

larsmoa:


viewer/src/index.ts
```diff
@@ -4,22 +4,19 @@
...
 // Loaders
 export { createLocalCadModel, createLocalPointCloudModel } from './datasources/local';
-export { createThreeJsSectorNode, createThreeJsPointCloudNode } from './views/threejs';
-export { initializeCesiumSectorScene } from './views/cesiumjs/createCesiumSectorNode';
 
-// Utilities
-export { toThreeVector3 } from './views/threejs/utilities';
+// ThreeJS views
+import * as threejs from './views/threejs';
+export { threejs };
```
dragly: Unfortunately, I think this style will make it impossible to have conditional imports in projects that depend on us later on. Now, I think three.js and cesium.js become hard, non-optional dependencies of our project. The reason it was in a different file was so that our users would be able to exclude three.js or cesium.js completely.

larsmoa: But is that really a problem now? If we decide to split the project into 'base', 'threejs' and 'cesiumjs' - wouldn't the import statements and the usage be identical?

PR #150, opened 2020-02-10
Implement screen space ambient occlusion

dragly:
This adds support for screen space ambient occlusion (SSAO) using the code from three.js. To achieve this, we need to pack the normals and color together when rendering. The color spectrum is reduced to 255 colors and are packed into the alpha value of the RGBA texture.

In addition, pull all materials out into a Materials interface that is stored on the CadNode to make it possible to override the drawing style (color vs. packed normal and color) from an effect.


viewer/src/views/threejs/cad/triangleMeshes.ts
```diff
@@ -5,8 +5,13 @@
 import * as THREE from 'three';
 import { TriangleMesh } from '../../../models/cad/types';
 import { sectorShaders, shaderDefines } from './shaders';
+import { RenderType } from '../materials';
```
j-bjorne: Why is this imported here?

viewer/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -2,24 +2,50 @@
 #pragma glslify: hsv2rgb = require('../color/hsv2rgb.glsl')
 #pragma glslify: packIntToColor = require('../color/packIntToColor.glsl')
 
-#ifdef COGNITE_COLOR_BY_TREE_INDEX
+const int RenderTypeColor = 1;
```
j-bjorne: Is there a way to make sure that these two separate "lists" aren't out of sync? Is it even important?

viewer/src/glsl/base/updateFragmentDepth.glsl
```diff
@@ -17,6 +17,7 @@ void updateFragmentDepth(vec3 p,mat4 projectionMatrix){
 
 void updateFragmentDepth(vec3 p,mat4 projectionMatrix){
   // Extension not available - not much we can do.
```
j-bjorne: What extension? What are we trying to do here?

viewer/src/views/threejs/cad/primitives.ts
```diff
@@ -6,20 +6,6 @@ import * as THREE from 'three';
 import { PrimitiveAttributes } from '../../../workers/types/parser.types';
```
j-bjorne: add some information about mesh.frustumCulled = false?

viewer/src/views/threejs/cad/triangleMeshes.ts
dragly: Ooops

viewer/src/glsl/base/updateFragmentColor.glsl
dragly: Not any that I know of that we can use easily. It is somewhat important, but I think we are stepping into territory of generating GLSL from code in another language. Might be something to look into, but for the time being I think we just need to try to keep these in sync.

viewer/src/glsl/base/updateFragmentDepth.glsl
dragly: This is not changed in the current PR. It is supposed to be a no-op if the fragment depth extension for WebGL is not available. Why it currently sets the color to red, I am not sure. Might have been to make it clear that something is wrong. The extension allows us to set the depth buffer to any value we like. Without it, we get the depth from the triangles in the billboards used to create the "impostors" for the different primitives. With it, we can set the depth correctly for spheres, cylinders, etc.

viewer/src/views/threejs/cad/primitives.ts
dragly: Will do in a separate PR since this was here from before

viewer/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -2,24 +2,50 @@
...
+const int RenderTypeNormal = 2;
+const int RenderTypeTreeIndex = 3;
+const int RenderTypePackColorAndNormal = 4;
 
-void updateFragmentColor(vec3 color, float treeIndex, vec3 normal) {
-    color = packIntToColor(treeIndex);
-    float amplitude = max(0.0, dot(normal, vec3(0.0, 0.0, 1.0)));
-    gl_FragColor = vec4(color * (0.4 + 0.6 * amplitude), 1.0);
-}
+uniform int renderType;
```
larsmoa: Should this uniform (and probably 'enum') be externalized and provided as an argument to upgradeFragmentColor?

viewer/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -2,24 +2,50 @@
...
+                a = (0.55 + hsv.x * 0.45);
+            } else {
+                // dark color
+                a = (0.1 + hsv.x * 0.45);
+            }
+        } else {
+            // no saturation - grayscale
+            a = hsv.z * 0.09;
+        }
+        gl_FragColor = vec4(normal.rgb, a);
```
larsmoa: Won't normal.rgb derp on negative components?  Need to use packNormalToRGB() as you do below?

viewer/src/views/threejs/cad/CadNode.ts
```diff
@@ -44,11 +45,17 @@ export class CadNode extends THREE.Object3D {
   private readonly _previousCameraMatrix = new THREE.Matrix4();
   private readonly _boundingBoxNode: THREE.Object3D;
 
+  // TODO make private + friend?
+  public readonly _materials: Materials;
```
larsmoa: This should be fixed before merge IMO.

viewer/src/views/threejs/post-processing/ssao.ts
```diff
@@ -0,0 +1,354 @@
...
+const quadCamera = new THREE.OrthographicCamera(-1, 1, 1, -1, 0, 1);
+
+export interface Pass {
+  render: (
+    renderer: THREE.WebGLRenderer,
+    scene: THREE.Scene,
+    camera: THREE.PerspectiveCamera,
+    pass?: SsaoPassType
+  ) => void;
+  setSize: (width: number, height: number) => void;
```
larsmoa: Can this be made implicit and decided by the size of the renderer provided in render()?

viewer/src/views/threejs/post-processing/ssao.ts
```diff
@@ -0,0 +1,354 @@
...
+}
+
+function traverseShaderMaterials(rootNode: THREE.Object3D, callback: (material: THREE.ShaderMaterial) => void) {
+  rootNode.traverseVisible(obj => {
+    // NOTE we cannot rely on instanceof here because THREE is not necessarily the same library
+    if (obj.type !== 'Mesh') {
+      return;
+    }
+    const mesh = obj as THREE.Mesh;
+    if (mesh.material == null || Array.isArray(mesh.material) || mesh.material.type !== 'ShaderMaterial') {
```
larsmoa: Should arrays of materials be supported?

viewer/src/views/threejs/post-processing/ssao.ts
```diff
@@ -0,0 +1,354 @@
...
+}
+
+export enum SsaoPassType {
+  Regular = 'Regular',
+  Ssao = 'Ssao',
+  SsaoFinal = 'SsaoFinal',
+  Antialias = 'Antialias'
+}
+
+export function createSsaoPass(): Pass {
```
larsmoa: Can render targets be reused to avoid allocating so many? Not sure on the memory footprint of this

viewer/src/views/threejs/post-processing/ssao.ts
```diff
@@ -0,0 +1,354 @@
...
+
+  return {
+    render,
+    setSize,
+    uniforms: ssaoUniforms
+  };
+}
+
+export class SsaoEffect {
+  private _ssaoPass: Pass;
```
larsmoa: Add readonly

viewer/src/glsl/base/updateFragmentColor.glsl
dragly: Good point! Fixed now.

viewer/src/views/threejs/post-processing/ssao.ts
dragly: Done

viewer/src/views/threejs/post-processing/ssao.ts
dragly: This function was left behind, but is no longer in use.

However, we would not have needed to support them since we are not using them ourselves and this function was only used for our materials.

viewer/src/views/threejs/post-processing/ssao.ts
dragly: Let us see if it becomes a problem first. It is very easy to mix them up by mistake once we start reusing them. If we go in that direction, we should rather rely on a system like the one postprocessing uses, with a double "buffer". That is, to only have two render targets and swap between them for each effect.

viewer/src/views/threejs/cad/CadNode.ts
dragly: Do you have any suggestions on how to fix it? I am a bit stuck with my own mindset here and cannot seem to figure out how to redesign it. Friends are not a thing in typescript (and often a bad thing to use anyways).

larsmoa: What about adding 1) get/set CadNode.RenderType as a field which traverses through the materials or 2) add a RenderType-field to the RenderingHints-struct which does the same as 1?

I think I like 1) more because RenderType (which maybe should be renamed to RenderMode?) because it's not really a hint, but a requirement.

dragly: But overriding the RenderType/Mode is really not something that should be done by an application developer - especially not if an effect is also trying to override the material. This is where it would be so much better to inject the material into the render function, but we need to do the whole traversal to change material + traversal to revert the change because of the way Three.js is designed.

Given that we have to do the forward change and back again, I really do not want the user to be able to change the material and have our "internal" effects change them too.

That being said, ideally we should not have had to change the material at all, since it will never be rendered "normally" anymore :confused: I do not know if we can achieve that, though.

dragly: And yes, 1) is better than 2) since 2) makes changing the material to the packed version seem like a common thing to do, which I do not think it will be.

dragly: 1) done

PR #151, opened 2020-02-10
Make it easier for a user to extract the underlying IO errors

dragly:
Instead of prepending our error messages with a string, just store the
error as an enum so that our users can get the underlying error if they
wish.


f3df/src/error.rs
```diff
@@ -1,32 +1,26 @@
...
     fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(fmt, "i3df::Error({:?})", self.message)
+        match self {
+            Error::IoError(x) => x.fmt(fmt),
+            Error::F3dfError(message) => write!(fmt, "i3df::Error({:?})", message)
+        }
     }
 }
 
 impl std::error::Error for Error {}
```
f2cognite: I would consider implementing [`std::error::Error::source`](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.source), which is a standard way of getting a reference to an inner error:

```
impl std::error::Error for Error {
    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
        match self {
            Error::IoError(err) => Some(err),
            _ => None,
        }
    }
}
```

f2cognite: Depending on usage, it can also be useful to add "optional accessors":

```rust
impl Error {
    pub fn io_error(&self) -> Option<&std::io::Error> {
        match self {
            Error::IoError(err) => Some(err),
            _ => None,
        }
    }
    pub fn f3df_error(&self) -> Option<&String> {
        match self {
            Error::F3dfError(message) => Some(message),
            _ => None,
        }
    }
}
```

f3df/src/error.rs
```diff
@@ -1,32 +1,26 @@
...
-
-impl Error {
-    pub fn new(message: impl ToString) -> Error {
-        Error {
-            message: message.to_string(),
-        }
-    }
+#[derive(Debug)]
+pub enum Error {
+    IoError(std::io::Error),
```
j-bjorne: Test

f3df/src/error.rs
dragly: Thanks!

PR #155, opened 2020-02-14
Add basic support for picking and overriding colors

dragly:
Also, add a simple example to demonstrate this. The picking code in this example will be moved into a separate component in a later PR.

This change also updates updateFragmentColor to properly render treeIndex.


examples/src/picking.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    requestAnimationFrame(render);
+  };
+  render();
+
+  const pickingTarget = new THREE.WebGLRenderTarget(1, 1);
+  const pixelBuffer = new Uint8Array(4);
+
+  const pick = (event: MouseEvent) => {
```
j-bjorne: Should we have a wrapper for picking for the example (change this to use the wrapper later) or keep the example as is to show a custom way of doing picking?

larsmoa: I agree that this should be wrapped in a utility

viewer/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -39,8 +55,7 @@ void updateFragmentColor(int renderMode, vec3 color, float treeIndex, vec3 norma
         gl_FragColor = vec4(packNormalToRGB(normal), 1.0);
     } else if (renderMode == RenderTypeTreeIndex) {
         color = packIntToColor(treeIndex);
-        float amplitude = max(0.0, dot(normal, vec3(0.0, 0.0, 1.0)));
-        gl_FragColor = vec4(color * (0.4 + 0.6 * amplitude), 1.0);
```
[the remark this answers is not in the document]
dragly: Yep, and it should not be applied for rendering treeIndex - unless we want wrong numbers :smile:

viewer/src/views/threejs/cad/CadNode.ts
```diff
@@ -81,11 +81,39 @@ export class CadNode extends THREE.Object3D {
...
+    this._materials.colorDataTexture.image.data[4 * treeIndex + 2] = 0;
+    this._materials.colorDataTexture.image.data[4 * treeIndex + 3] = 0;
+    this._materials.colorDataTexture.needsUpdate = true;
+  }
+
   set renderMode(mode: RenderMode) {
     this._renderMode = mode;
-    for (const material of Object.values(this._materials)) {
-      material.uniforms.renderMode.value = mode;
-    }
```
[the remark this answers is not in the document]
dragly: Because the `Materials` object now contains both materials and a texture. Iterating in this way is not very type-safe, because TypeScript doesn't care to check if all `Object.values` has proper types. So this loop failed spectacularly at runtime instead.

examples/src/picking.ts
dragly: Let us just pretend that I did not forget to push a very relevant commit on Friday and that I just fixed this to adapt to your suggestion :stuck_out_tongue:

viewer/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -2,16 +2,32 @@
...
 const int RenderTypeTreeIndex = 3;
 const int RenderTypePackColorAndNormal = 4;
+const float dataTextureWidth = 2048.0;
+const float dataTextureHeight = 2048.0;
 
 vec3 packNormalToRGB( const in vec3 normal ) {
     return normalize( normal ) * 0.5 + 0.5;
 }
 
+// TODO remove input color
```
larsmoa: Why?

viewer/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -2,16 +2,32 @@
 #pragma glslify: hsv2rgb = require('../color/hsv2rgb.glsl')
 #pragma glslify: packIntToColor = require('../color/packIntToColor.glsl')
 
+uniform sampler2D colorDataTexture;
```
larsmoa: Not sure how well this scales now as we add more inputs, but until now we have specified all inputs as arguments to ensure that shader parameters are defined one place and not scattered around. Should we consider moving this outside the subroutine as well? Should we create a struct for colorInputs to avoid having to add more and more fields?

viewer/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -2,16 +2,32 @@
...
+
 const int RenderTypeColor = 1;
 const int RenderTypeNormal = 2;
 const int RenderTypeTreeIndex = 3;
 const int RenderTypePackColorAndNormal = 4;
+const float dataTextureWidth = 2048.0;
+const float dataTextureHeight = 2048.0;
```
larsmoa: This should be provided from JS-side of things. We do not currently have this information available, but the shaders should at least be ready for it.

viewer/src/views/threejs/cad/CadNode.ts
```diff
@@ -35,6 +39,7 @@ const updateVars = {
 export class CadNode extends THREE.Object3D {
   public readonly rootSector: SectorNode;
   public readonly modelTransformation: SectorModelTransformation;
+  public readonly _shading: Shading;
```
larsmoa: Why is this field prefixed with _? Is it private?

viewer/src/views/threejs/cad/materials.ts
```diff
@@ -22,9 +23,17 @@ export interface Materials {
...
+  // Data textures
+  colorDataTexture: THREE.DataTexture;
 }
 
 export function createMaterials(): Materials {
+  const pixelCount = 2048;
+  const colorCount = pixelCount * pixelCount;
+
+  const colors = new Uint8Array(4 * colorCount);
+  const colorDataTexture = new THREE.DataTexture(colors, pixelCount, pixelCount);
```
larsmoa: How should we ensure that this function will be ready to create texture of the correct size to hold `maxTreeIndex` number of elements? Should it accept a CadMetadata-object (or whatever the current name is)?

viewer/src/views/threejs/cad/picking.ts
```diff
@@ -0,0 +1,56 @@
...
+
+type PickDelegate = (scene: THREE.Scene, camera: THREE.PerspectiveCamera, event: MouseEvent) => number;
+
+export class CadPicker {
+  private readonly _picker: {
+    pick: PickDelegate;
+  };
+
+  constructor(renderer: THREE.WebGLRenderer, cadNode: CadNode) {
+    this._picker = createCadPicker(renderer, cadNode);
```
larsmoa: Why not just implement functions inside class? #nitpick

viewer/src/views/threejs/cad/shading.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+    for (const treeIndex of treeIndices) {
+      const colorOrUndefined = options.color(treeIndex);
+      const color = colorOrUndefined ? colorOrUndefined : [0, 0, 0, 0];
+      materials.colorDataTexture.image.data[4 * treeIndex] = color[0];
+      materials.colorDataTexture.image.data[4 * treeIndex + 1] = color[1];
+      materials.colorDataTexture.image.data[4 * treeIndex + 2] = color[2];
+      materials.colorDataTexture.image.data[4 * treeIndex + 3] = color[3];
+    }
+    materials.colorDataTexture.needsUpdate = true;
```
larsmoa: Should we always do this or try to be resilient to unnecessary calls to `updateNodes()` by comparing values in the loop? Not sure what's best...

viewer/src/glsl/base/updateFragmentColor.glsl
dragly: Ooops, it was a stray comment left behind.

viewer/src/glsl/base/updateFragmentColor.glsl
dragly: Agreed. Splitting this into a new function for now, which also reduces the number of inputs to updateFragmentColor.

viewer/src/glsl/base/updateFragmentColor.glsl
dragly: Agreed, but let us keep them here for now and update the shaders once we have this ready.

viewer/src/views/threejs/cad/CadNode.ts
dragly: Ooops. Yes, let us keep this private for now.

viewer/src/views/threejs/cad/materials.ts
dragly: I think this is getting a bit too far ahead of the scope of this implementation. To create a texture of the correct size we need to also decide on/find solutions to:

- how and where we retrieve the information about the number of tree indices
- whether we can still create the materials and texture up-front or if we need to change the uniform later once data is received

I think it makes more sense to decide on how we will create a texture of the right size when we have all the necessary pieces rather than making too many assumptions about at this point.

viewer/src/views/threejs/cad/picking.ts
dragly: Just trying to avoid large classes where it is hard to see if calling a function causes mutations or not. That said, closures do not really help in this regard. I will move the implementation into a free function instead to make it more evident.

viewer/src/views/threejs/cad/shading.ts
dragly: I see what you mean, but I think I prefer the simpler version for now, until we see the need for an optimization. Since `updateNodes()` already requires treeIndices to be specified, it should reduce the temptation to call it everywhere "for good measure", which is usually when it becomes necessary to do those types of checks. The current use case is when the user clicks something, and at that point the texture will very likely change.

examples/src/picking.ts
```diff
@@ -0,0 +1,74 @@
...
+  renderer.setSize(window.innerWidth, window.innerHeight);
+  document.body.appendChild(renderer.domElement);
+
+  const { position, target, near, far } = cadNode.suggestCameraConfig();
+  const camera = new THREE.PerspectiveCamera(75, window.innerWidth / window.innerHeight, near, far);
+  const controls = new CameraControls(camera, renderer.domElement);
+  controls.setLookAt(position.x, position.y, position.z, target.x, target.y, target.z);
+  controls.update(0.0);
+  camera.updateMatrixWorld();
+  let pickingNeedsUpdate = false;
```
larsmoa: Is there a way we can avoid this flag? It feels wrong since picking works on CadNode which already returns a flag from `update`.

examples/src/picking.ts
```diff
@@ -0,0 +1,74 @@
...
+      renderer.render(scene, camera);
+      pickingNeedsUpdate = false;
+    }
+
+    requestAnimationFrame(render);
+  };
+  render();
+
+  const pick = (event: MouseEvent) => {
+    const treeIndex = reveal_threejs.pickTreeIndex({ renderer, cadNode, scene, camera, event });
```
larsmoa: Can this return position as well? And is an alternative approach to call it raypick that works either on a ray or a canvas XY position?

viewer/src/glsl/base/determineColor.glsl
```diff
@@ -0,0 +1,20 @@
...
+    float uCoord = (u + 0.5) / dataTextureWidth;
+    float vCoord = (v + 0.5) / dataTextureHeight; // invert Y axis
+    vec2 treeIndexUv = vec2(uCoord, vCoord);
+    vec4 overrideColor = texture2D(colorDataTexture, treeIndexUv);
+
+    if (overrideColor.a == 0.0) {
+        return originalColor;
+    }
+
+    return overrideColor.rgb;
```
larsmoa: `return ceil(overrideColor.a) * overrideColor.rgb + floor(overrideColor.a)*originalColor` for branch-free bliss.

viewer/src/views/threejs/cad/shading.ts
```diff
@@ -0,0 +1,37 @@
...
+export interface Shading {
+  materials: Materials;
+  updateNodes: (treeIndices: number[]) => void;
+}
+
+export interface DefaultShadingOptions {
+  color: (treeIndex: number) => number[] | undefined;
+}
+
+export function createDefaultShading(options: DefaultShadingOptions) {
```
larsmoa: Nitpick: specify returns type to 1) make it easier to see what the function returns and 2) ensure compile errors happens here, and not in calling code, when we change the Shading-interface in the future.

viewer/src/views/threejs/cad/materials.ts
```diff
@@ -22,9 +23,17 @@ export interface Materials {
   instancedMesh: THREE.ShaderMaterial;
   triangleMesh: THREE.ShaderMaterial;
   simple: THREE.ShaderMaterial;
+  // Data textures
+  colorDataTexture: THREE.DataTexture;
```
larsmoa: Is `overrideColorPerTreeIndex` or something similar more describing of the layout and purpose?

dragly:
> Approved, but please consider my comments before merging.

Will do. The auto-merge should probably be removed in these cases so Bulldozer does not merge for us :stuck_out_tongue:

examples/src/picking.ts
dragly: The flag is there because we need to update the scene after updating the colors. Think of it as "some user input changed the scene and we need to re-render". One way to integrate this more would be to add a listener in `CadNode` on some callback on `Shading` when `updateNodes` is called. What do you think?

examples/src/picking.ts
dragly: I am looking returning position now, but will do it in a later PR. After looking into returning positions, having XY or rays as input, and supporting picking together with other things, I am realizing that we might need to do picking more like it is usually done in three.js (with more manual/advanced work). The simple "wrapper" version might have to be a lot more limited (not allowed to mix with other Three.js objects). But again, I think this should be done in a separate PR.

viewer/src/glsl/base/determineColor.glsl
dragly: I do not think this works:

- if `overrideColor.a` is `0.0`, then `floor(0.0)` and `ceil(0.0)` will both return `0.0`? Then the color is `rgb(0, 0, 0)`.
- if `overrideColor.a` is `1.0`, then we get `1.0` on both, which is a mix of the two colors.

This should work, however:

```
return overrideColor.a * overrideColor.rgb + (1.0 - overrideColor.a) * originalColor;
```

PR #157, opened 2020-02-16
Move @cognite/sdk, @cognite/potree-core and three to peerDependencies.

larsmoa:


viewer/package.json
```diff
@@ -2,8 +2,8 @@
   "name": "@cognite/reveal",
   "version": "0.1.0",
   "description": "Reveal streaming component",
-  "main": "index.js",
-  "types": "index.d.ts",
+  "main": "dist/index.js",
```
dragly: Why is this being changed? The package.json is moved into the dist folder when built, so this location is wrong. 

If I recall correctly, the reason the package.json is moved to dist is because TypeScript would pick up other .ts files in src when we linked the examples to viewer. By linking to viewer/dist, we no longer picked up spurious files.

Might be that we do not need to move it after we changed the ts-loader options to

```
            options: {
              onlyCompileBundledFiles: true,
            },
```
but that seems unrelated to this change. Especially since examples still point to viewer/dist.

viewer/package.json
```diff
@@ -57,12 +60,14 @@
...
   "dependencies": {
-    "@cognite/potree-core": "^1.1.1",
-    "@cognite/sdk": "^2.13.0",
     "comlink": "^4.1.0",
     "gl-matrix": "^3.1.0",
     "glslify": "^7.0.0",
-    "glslify-import": "^3.1.0",
+    "glslify-import": "^3.1.0"
+  },
+  "peerDependencies": {
```
dragly: Why are these being made into peerDependencies?

viewer/package.json
larsmoa: Hmmm, I did this because I had to import reveal using "@cognite/reveal/dist". Not sure how or why this happened, but changing this didn't seem to break anything?

viewer/package.json
larsmoa: Three and SDK are peerDependencies because the user will decide what version they should be. They are not bundled anymore after moving them to peerDeps (i.e. no duplicated three.module.js). potree-core is maybe a stretch, what I really want it to be is an optionalDependency...

viewer/package.json
dragly: That is weird. I think it might break things when running npm pack, though. I will test on my machine.

viewer/package.json
dragly: Would not an optionalDependency work for all these then? I think `peerDependencies` are more for packages that are like plugins than ones that have optional dependencies like we do.

viewer/package.json
dragly: Ah, we removed the `cp package.json dist` line from the build. That might be why you ran into problems. However, it seems like it works now either way, so we can keep this change.

dragly:
After some testing in a simple environment, it appears that the problem was that we use a symlink between viewer and examples, which resulted in double-bundling Three.js for some reason. If we `npm pack` the viewer, it should only result in one bundle of three.js in the examples.

After testing with a simple setup with two projects, it seems like the only consequence of using a peerDependency is that the library consumer needs to make sure to install this manually. This can be useful to reduce download size for developers, but for the time being we decided to close this PR. Let us revisit this when support for other frameworks than three.js is added.

PR #134, opened 2020-02-09
Load point clouds from CDF

larsmoa:

PR #171, opened 2020-02-17
Add walkable path to model - initial commit

j-bjorne:
I added a few TODOs and comments on what I wanted to maybe do with some one the "examples", but I wanted an opinion and maybe some directions before proceeding. I'll add it as a review for now, let me know if you want things like this saved as drafts in the future.


examples/src/walkable-path.ts
```diff
@@ -0,0 +1,151 @@
...
+
+// Should find a way to add this to maybe cognite-sdk-js later?
+async function getTransitPath(client: CogniteClient, transitRequest: TransitPathRequest) {
+  const project = '3ddemo';
+  const url = `https://api.cognitedata.com/api/playground/projects/${project}/3d/pathfinder/transit`;
+  const response = await client.post<TransitPathResponse>(url, {
+    data: transitRequest
+  });
+  return response;
+}
```
larsmoa: Move to file in viewer/. For outputs, Im creating a 'CogniteClientExtensions'-class that the plan is to add hidden/non-production endpoints to. See https://github.com/cognitedata/reveal/pull/134/commits/a729c56df4b97ed941cbaed5b10ec5935d89d601#diff-ef96b80e3bb59fdb8b6ce79524587ee0 (this is by no means perfect yet).

examples/src/walkable-path.ts
```diff
@@ -0,0 +1,151 @@
...
+  controls.setLookAt(position.x, position.y, position.z, target.x, target.y, target.z);
+  controls.update(0.0);
+  camera.updateMatrixWorld();
+  scene.add(cadNode);
+
+  const diameter = 0.55;
+  const height = 1.6;
+
+  const walkablePathResponse = await getTransitPath(client, {
+    modelId: 3329507622597457,
```
larsmoa: I realize this is problematic as this is very dependent on the SDK, but how could we avoid having the modelId in the code? We should maybe add a task that allows the user to click two objects and create a path (when a model has been loaded from CDF, and not from local storage)?

examples/src/walkable-path.ts
```diff
@@ -0,0 +1,151 @@
...
+    data: transitRequest
+  });
+  return response;
+}
+
+// TODO: Should be added to reveal/viewer. The input data TransitPath response is in my
+// opinion too coupled with our sdk response, should we split it into two funcitons, one taking in a { x, y, z}[]
+// and another converting the response to said array.
+
+function createWalkablePath(pointData: TransitPathResponse, modelTransformation: SectorModelTransformation) {
```
larsmoa: I think the right way to achieve decoupling is to have `getTransitPath` above return something more friendly than 'raw JSON' as above. Also, I like when functions have defined return types as its hard to see what a function returns from the `return`-statements.

examples/src/walkable-path.ts
dragly: I think we should create a simple feature that we can reuse in other examples for this purpose: A simple box with fields for project + modelId. Then redirect to the proper URL (localhost:8080/example...?project=...&model=...). Bonus points for storing the last result in localStorage or indexedDB :)

examples/src/walkable-path.ts
```diff
@@ -0,0 +1,151 @@
...
+) {
+  const curve = new THREE.CatmullRomCurve3(points);
+  const geometry = new THREE.TubeBufferGeometry(curve, segments, radius, radiusSegments, closed);
+  return geometry;
+}
+
+async function main() {
+  const client: CogniteClient = new CogniteClient({ appId: 'WalkablePath Example' });
+  client.loginWithOAuth({
+    project: '3ddemo'
```
dragly: Same as Lars' comment below: 3ddemo is not public and should not be hard-coded in the examples. I do not think we have any public tenant that can be used, so rather add a dialog and/or use URL parameters.

examples/src/walkable-path.ts
j-bjorne: I have added this but with a default model Id. This way we have the ability to change it. In the future this should be available through the API.

examples/src/walkable-path.ts
j-bjorne: I have made it so that we have a json to threejs like parser now. I do indeed think that we should have a object that defines the segment, but I am also then thinking we want more in the response object itself before we design this object.

examples/src/walkable-path.ts
j-bjorne: Url parameters it is.

j-bjorne:
For some reason my git didn't allow me to push to this branch, and I ended up force pushing. This means that the old push is gone (since I am more used to the interface in azure I am not touching anything until you guys have had a chance to either revert to the old one or just accept this one if it looks okey)

examples/src/walkable-path.ts
j-bjorne: I moved the client to a separate function but kept it in the example. For now I think that is useful, and I would like to revisit this once we have gone through our architecture talk in the next sprint.

examples/src/walkable-path.ts
```diff
@@ -0,0 +1,374 @@
...
+      segments: [
+        {
+          path: Coordinates[];
+        }
+      ];
+    }
+  ];
+}
+
+function createWalkablePathClient(project: string) {
```
larsmoa: Add return type to make it easier to see what this returns.

examples/src/walkable-path.ts
```diff
@@ -0,0 +1,374 @@
...
+  const client: CogniteClient = new CogniteClient({ appId: 'Reveal Examples - WalkablePath' });
```
larsmoa: Instead of creating a client, this should accept  a client as argument. This is to avoid multiple SDK clients.

examples/src/walkable-path.ts
```diff
@@ -0,0 +1,374 @@
...
+  client.loginWithOAuth({
+    project
+  });
+
+  const getTransitPath = async (transitRequest: TransitPathRequest): Promise<TransitPathResponse> => {
```
larsmoa: Not available yet, but should we wait for PR https://github.com/cognitedata/reveal/pull/134 and add this to https://github.com/cognitedata/reveal/pull/134/files#diff-e9a427d01fd7ff8cea0c7429e53773a1R68?

examples/src/walkable-path.ts
```diff
@@ -0,0 +1,374 @@
...
+      const curve = new THREE.CatmullRomCurve3(path);
+      const geometry = new THREE.TubeBufferGeometry(curve, 2 * (path.length - 1), radius, radiusSegments, closed);
+      const material = new THREE.MeshPhongMaterial({ color: 0xfeafeafe });
+      const mesh = new THREE.Mesh(geometry, material);
+      mesh.position.y += heightOffset;
+      meshes.push(mesh);
+    }
+  } catch (error) {}
+  return meshes;
+}
```
larsmoa: Can we create a class that extends THREE.Object3d and place this in `viewer/src/views/threejs/cad`? (Because it's a CAD specific feature, at least for now).

j-bjorne: I would like to await moving this for now until we have ironed out exactly how we want to include these things in reveal

examples/src/walkable-path.ts
j-bjorne: Added in latest update

examples/src/walkable-path.ts
j-bjorne: Changed function name parameters to take client

examples/src/walkable-path.ts
j-bjorne: I personally think it should stay in the example until we have ironed out exactly how we want the architecture to look.

examples/src/walkable-path.ts
larsmoa: If that's the case I won't consider this done. I really think we need to finish up things and shift focus to the next item on our very long list of things we need to implement.

dragly: I would prefer to have it in a separate folder. The `cad` folder has an unfortunate name (renamed from the more unfortunate `sector`), but I think it should still be focused on streaming primitives + mesh geometry. Otherwise all kinds of features will be creeping into that folder. I would suggest it is placed in a `src/views/threejs/walkable-path` folder instead.

To me it is also okay to keep this in the example for now, but I do not mind moving it into `walkable-path` either. Just as long as it is somewhat separate from the other stuff.

dragly: To clarify: I agree that the feature itself is not done with this PR, but I am okay with merging one PR that pulls in an example and then later creating another PR that formalizes the feature. That separates a proof-of-concept example from a full-fledged feature. And I think it is nice that we have a bit of both, so that we can direct users to examples in the cases where we have not prioritized landing a complete feature.

j-bjorne: I do not think this is walkable path as a reveal feature. This is how to add walkable path together with reveal in a Three js scene if anything. I don't mind holding off merging this PR until we add it as a feature in reveal, but I also don't see any downsides with adding the example as it doesn't touch reveal and thus wont cause problems when merging and a separate PR will be done on formalizing it as a reveal feature (If it is something we want as a reveal thing, this is part of the discussion I feel is necessary to define what Reveal is exactly)

examples/src/walkable-path.ts
j-bjorne: created modelId url parameter

dragly:
What is the status of this? Can it be merged as is or is there more work to do?

larsmoa:
> What is the status of this? Can it be merged as is or is there more work to do?

Didn't realize I had to approve.

PR #179, opened 2020-02-19
Bump dependencies

dragly:
This bumps a collection of updates from dependabot.


dragly:
i just realized that this does not work perfectly because the branch will still contain all the old updates unless it is reset to master after merging. So I think the workflow needs to be:

1. create a branch named `dependency-updates` based on latest `master`
2. wait for dependabot to pull any number of PRs into `dependency-updates`
3. create a PR from `dependency-updates` to `master` (like this one)
4. optional: wait for dependabot to pull more PRs into the `dependency-updates` branch
5. merge the PR into `master`
6. delete the `dependency-updates` branch
7. start over from 1. (it is important that the `dependency-updates` branch is re-created or reset to `master`)

PR #191, opened 2020-02-20
Make dist folder when running `npm run build`

dragly:
And introduce run-script-os to handle this on all operating systems.
This is required to be able to build after deleting
`examples/package-lock.json`.

Also update package-lock.json for both `viewer` and `examples`.

Note: Pointing to `../viewer/dist` is necessary for the examples because
TypeScript cannot find the type definitions for
`@cognite/reveal/threejs` otherwise, because they cannot be listed in
`viewer/package.json`.


viewer/package.json
```diff
@@ -5,8 +5,11 @@
   "main": "index.js",
   "types": "index.d.ts",
   "scripts": {
-    "release": "mkdir -p dist && cp package.json README.md dist/ && webpack",
-    "build": "webpack --env.development",
+    "release": "npm run mkdist && webpack",
+    "build": "npm run mkdist && webpack --env.development",
+    "mkdist": "run-script-os",
+    "mkdist:win32": "if not exist 'dist' mkdir 'dist' && copy package.json README.md dist",
+    "mkdist:linux:darwin": "mkdir -p dist && cp package.json README.md dist",
```
larsmoa: This is amazing

PR #192, opened 2020-02-20
Dependency updates

dragly:


dragly:
We can wait for a few more updates before merging this. I just wanted to create the PR so bulldozer pulls master back into this branch as well.

PR #202, opened 2020-02-25
Changed build script for win32 to use powershell

j-bjorne:
Make it work on windows with powershell

### DIFF
--- a/viewer/package.json
+++ b/viewer/package.json
@@ -8,7 +8,7 @@
     "release": "npm run mkdist && webpack",
     "build": "npm run mkdist && webpack --env.development",
     "mkdist": "run-script-os",
-    "mkdist:win32": "if not exist 'dist' mkdir 'dist' && copy package.json README.md dist",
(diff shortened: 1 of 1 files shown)


viewer/package.json
```diff
@@ -8,7 +8,7 @@
     "release": "npm run mkdist && webpack",
     "build": "npm run mkdist && webpack --env.development",
     "mkdist": "run-script-os",
-    "mkdist:win32": "if not exist 'dist' mkdir 'dist' && copy package.json README.md dist",
+    "mkdist:win32": "@powershell $(if(-not(Test-Path dist -PathType Container)) { mkdir dist;$? } else { $true;$? }) -and $(Copy-Item -Path package.json, README.md -Destination dist;$?)",
```
larsmoa: Oh god, is this really an improvement? Powershell gives me the chills

dragly: We might want to just write this logic in JS and have it the same on all OSes at some point.

Or in Rust :smile:

j-bjorne: It is not an improvement, but I am using powershell in VSCode, and my terminals open as such by default. The run-script-os should be able to see if the shell running the script is powershell or cmd, then it could be like win32-cmd, win32-ps, that would be cool 😅

PR #213, opened 2020-02-26
Add support for filtering based on treeIndex

dragly:
This adds a `visible` function to the Shading delegate object which can
be defined to return a boolean depending on whether the node with a
given tree index should be visible or not.


viewer/src/views/threejs/cad/shading.ts
```diff
@@ -9,8 +9,29 @@ export interface Shading {
...
 export interface DefaultShadingOptions {
-  color: (treeIndex: number) => number[] | undefined;
+  color?: ColorDelegate;
+  visible?: VisibilityDelegate;
+}
+
+function updateColors(getColor: ColorDelegate, materials: Materials, treeIndices: number[]) {
+  for (const treeIndex of treeIndices) {
+    const colorOrUndefined = getColor(treeIndex);
+    const color = colorOrUndefined ? colorOrUndefined : [0, 0, 0, 0];
```
larsmoa: I like `const color = getColor(treeIndex) || [0,0,0,0]` better

PR #214, opened 2020-02-26
Turn initializeSectorLoader into SectorActivator

dragly:
This is part of an effort to rewrite closures into classes.


dragly:
What do you want to update in the unit test? I thought it was changed already?

larsmoa:
> What do you want to update in the unit test? I thought it was changed already?

Only the filename :)

viewer/src/models/cad/BasicSectorActivator.ts
```diff
@@ -0,0 +1,96 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+import { setUnion, setDifference } from '../../utils/setUtils';
+import { DiscardSectorDelegate, ConsumeSectorDelegate, GetSectorDelegate } from './delegates';
+
+export interface SectorActivator {
```
j-bjorne: Should we split interfaces and implementations earlier rather than later? Probably not something we need to look into before we go over architecture, but I think it is important to keep in mind going further.

dragly: "Split", as in putting them in different files?

PR #216, opened 2020-02-26
Replace use of 'uploaded_*.txt' with new 'scene.json'.

larsmoa:


viewer/src/models/cad/CadModelImpl.ts
```diff
@@ -0,0 +1,107 @@
...
+  constructor(dataRetriever: ModelDataRetriever, modelTransformation: SectorModelTransformation) {
+    this.dataRetriever = dataRetriever;
+    this.detailedParser = createParser(id => this.fetchCtm(id));
+    this.simpleParserPromise = createQuadsParser();
+    this._modelTransformation = modelTransformation;
+
+    const metadataJson = this.dataRetriever.fetchJson('scene.json');
+    this.scenePromise = parseSceneMetadata(metadataJson);
+    this._scene = undefined;
+  }
```
dragly: Make constructor private and create a static async create function. Then you can remove the initialize function and all the ensureInitialized calls:

```
export class CadModelImpl implements CadModel {
  static async create(
    dataRetriever: ModelDataRetriever,
    modelTransformation: SectorModelTransformation
  ): Promise<CadModelImpl> {
    const metadataJson = dataRetriever.fetchJson('scene.json');
    const scene = await parseSceneMetadata(metadataJson);
    const model = new CadModelImpl(dataRetriever, modelTransformation, scene);
    return model;
  }

  private readonly _scene: SectorScene;
  private readonly _modelTransformation: SectorModelTransformation;

  private readonly dataRetriever: ModelDataRetriever;
  private readonly detailedParser: ParseSectorDelegate<Sector>;
  private readonly simpleParserPromise: Promise<ParseSectorDelegate<SectorQuads>>;

  private constructor(
    dataRetriever: ModelDataRetriever,
    modelTransformation: SectorModelTransformation,
    scene: SectorScene
  ) {
    this.dataRetriever = dataRetriever;
    this.detailedParser = createParser(id => this.fetchCtm(id));
    this.simpleParserPromise = createQuadsParser();
    this._modelTransformation = modelTransformation;
    this._scene = scene;
  }
  // ...
}
```

In `loadCadModelByUrl` you can now call

```
  const model = await CadModelImpl.create(retriever, transform);
```
This makes sure `_scene` is always initialized and the `CadModelImpl` is never in an incomplete state.

viewer/src/models/cad/determineSectors.ts
```diff
@@ -104,7 +97,10 @@ export function determineSectorsQuality(scene: SectorScene, requestedDetailed: S
     if (detailed.includes(sector.id)) {
       return true;
     }
-    simple.push(sector.id);
+    // F3D file is omitted if there are no geometry in the file
```
dragly: nitpick: if there is

viewer/src/models/cad/determineSectors.ts
```diff
@@ -60,14 +60,7 @@ export async function defaultDetermineSectors(params: DetermineSectorsInput): Pr
...
     const largestAllowedQuadSize = hints.maxQuadSize * screenHeight; // no larger than x percent of the height
-    const quadSize = (() => {
-      if (!sector.simple) {
-        // Making the quad infinite in size means we will always use the detailed version instead
-        return Infinity;
-      }
-      return sector.simple.gridIncrement;
-    })();
-
+    const quadSize = sector.facesFile.quadSize;
```
dragly: Is there a quadSize even if the file does not exist?

viewer/src/models/cad/CadModelImpl.ts
larsmoa: 🥇 Fixed

viewer/src/models/cad/determineSectors.ts
larsmoa: Yes, there is now. Quadsize is deterministic based on depth in the tree, and this is now output from the optimizer. We changed this so it's always provided - filename is set to null when the file isn't present. If quadsize wasn't provided for sectors without F3D, we would always have to resort to high detail geometry as there would be no way of determining if we can use F3D.

PR #219, opened 2020-02-27
Introduce CachedRepository and remove caching details from CadNode

dragly:


viewer/src/repository/Repository.ts
```diff
@@ -0,0 +1,10 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+import { Sector, SectorQuads } from '../models/cad/types';
+
+export interface Repository {
```
larsmoa: This seems very CAD-specific - should it be a `CadRepository` which is placed under `repository/cad`? Or; should it be a `Repository<T>` with one function `get<T>(id: number, detailLevel: number)`?

dragly: Sounds like a good idea to put it under CAD for now.

PR #232, opened 2020-03-02
Change createSimpleCache into MemoryRequestCache class

dragly:


viewer/src/cache/MemoryCache.ts
```diff
@@ -0,0 +1,28 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+import { RequestCache, RequestDelegate } from './RequestCache';
+
+export class MemoryCache<T_ID, T> implements RequestCache<T_ID, T> {
```
larsmoa: MemoryRequestCache

larsmoa: Or maybe RequestMemoryCache if that makes more sense?

dragly: Done

PR #247, opened 2020-03-04
Split chunks in examples to reduce build time.

larsmoa:
New PR because old one was based on wrong branch.


larsmoa:
This PR causes examples to not load, so I'm closing for now.

PR #251, opened 2020-03-05
Bump ThreeJS to 0.114.0.

larsmoa:


examples/src/utils/renderer-debug-widget.ts
```diff
@@ -256,12 +256,14 @@ function updateSceneInfo(scene: THREE.Object3D, sceneInfo: SceneInfo) {
     materials.forEach(m => materialIds.add(m.id));
 
     if (x.name.startsWith('Primitives')) {
+      const indexAttributes = geometry.index!;
```
j-bjorne: Any reason why assigning a new const instead of calling geometry.index!.count?

larsmoa: Mja, Law of Demeter and readability. I think its more explicitly visible that we perform an explicit not-null cast when the ! is the last character.

PR #274, opened 2020-03-12
Initialize CogniteClient once in examples that load multiple models

larsmoa:
Resolves an issue where user was sent in an infinite login loop.

### DIFF
--- a/examples/src/utils/loaders.ts
+++ b/examples/src/utils/loaders.ts
@@ -51,17 +51,28 @@ export function createModelIdentifierFromUrlParams(
  * @param model Model identifier. If a number, the model is assumed to be stored in CDF.
  * @param project If model is a number (i.e. model ID), project must be provided.
  */
-export async function loadCadModelFromCdfOrUrl(model: ModelIdentifier, client?: CogniteClient): Promise<reveal.CadModel> {
+export async function loadCadModelFromCdfOrUrl(
+  model: ModelIdentifier,
+  client?: CogniteClient
+): Promise<reveal.CadModel> {
   if (isUrlModelIdentifier(model)) {
     return reveal.loadCadModelByUrl(model.modelUrl);
-  } else {
-    if (!client) {
-      client = new CogniteClient({ appId: 'cognite.reveal.example' });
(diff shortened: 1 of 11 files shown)


examples/src/utils/loaders.ts
```diff
@@ -41,13 +51,15 @@ export function createModelIdentifierFromUrlParams(
...
-export async function loadCadModelFromCdfOrUrl(model: ModelIdentifier): Promise<reveal.CadModel> {
+export async function loadCadModelFromCdfOrUrl(model: ModelIdentifier, client?: CogniteClient): Promise<reveal.CadModel> {
   if (isUrlModelIdentifier(model)) {
     return reveal.loadCadModelByUrl(model.modelUrl);
   } else {
-    const client = new CogniteClient({ appId: 'cognite.reveal.example' });
-    client.loginWithOAuth({ project: model.project });
-    await client.authenticate();
+    if (!client) {
+      client = new CogniteClient({ appId: 'cognite.reveal.example' });
```
j-bjorne: Food for thought: newing things inside functions / classes often runs the risk of the user creating said elements multiple times in code. Say you call loadCadModelFromCdfOrUrl without providing a client you technically create a client each time the user calls on said method. I don't like that, what are your thoughts on the matter?

larsmoa: I don't care tbh. It's an example. The reason it's like this is because the client doesn't need to be initialized always and this enabled to not copy code in each example.

examples/src/utils/loaders.ts
```diff
@@ -41,15 +51,28 @@ export function createModelIdentifierFromUrlParams(
...
+  return reveal.loadCadModelFromCdf(client, model.modelId);
+}
+
+export async function createClientIfNecessary(modelId: ModelIdentifier): Promise<CogniteClient | undefined> {
+  if (isUrlModelIdentifier(modelId)) {
+    // Model is not on CDF
+    return undefined;
+  }
+
+  const client = new CogniteClient({ appId: 'cognite.reveal.example' });
```
j-bjorne: Nice!

PR #275, opened 2020-03-12
Create wrappers Cognite3DModel and Cognite3DViewer for easy migration from the old viewer

larsmoa:
Create non-working wrappers for Cognite3DModel and Cognite3DViewer. Note that this PR doesn't merge to master.


viewer/src/migration/Cognite3DViewer.ts
```diff
@@ -0,0 +1,145 @@
...
+    this.controls = new ComboControls(this.camera, this.canvas);
+    this.controls.dollyFactor = 0.992;
+    this.controls.enabled = false;
+  }
+
+  dispose(): void {
+    throw new NotSupportedInMigrationWrapperError();
+  }
+
+  onClick(callback: (event: PointerEvent) => void): void {
```
dragly: Why is this not wrapped the same way as before, with `on('click'`) as the pattern?

larsmoa: I will try to change this, but I was not able to make TypeScript accept on('click') and on('cameraChange') etc with correct event type. If you manage to fix it your my hero for the day.

PR #283, opened 2020-03-14
Rename Shading to NodeAppearance

dragly:
This better reflects the purpose, which is to define the visual properties
of individual nodes, such as color, visibility, highlighting, etc.


larsmoa:
I think maybe `NodeProperties` are a bit too generic and my inital guess would be that it was a container for key->value properties assosciated to nodes. Is `NodeVisualSettings` an option?

dragly:
I agree. How about `NodeAppearance`?

PR #290, opened 2020-03-17
Add simple wrapper for Cognite3DViewer and Cognite3DModel and working example using the new wrappers

larsmoa:
The wrapper only has partial functionality, but works for visualising models.


viewer/src/migration/Cognite3DViewer.ts
```diff
@@ -0,0 +1,448 @@
...
+    this.domElement.removeChild(this.canvas);
+    this.renderer.dispose();
+    this.scene.dispose();
+    while (this.models.length > 0) {
+      const model = this.models.pop()!;
+      model.dispose();
+    }
+  }
+
+  onClick(_callback: (event: PointerEvent) => void): void {
```
dragly: Why not use the same API as the original viewer? Does this not introduce a migration to the migration? 😛

PR #308, opened 2020-03-19
Speed up Travis builds by building debug and dropping Node 10

dragly:
Unless we are building for a release branch, we will only make debug
builds, which are significantly faster than release builds. In addition,
we will only target Node 10 to reduce build times.


dragly:
We should be able to see how long this build takes to review to effect :)

PR #312, opened 2020-03-21
Add support for raypicking to Cognite3DViewer

larsmoa:


viewer/src/migration/Cognite3DViewer.ts
```diff
@@ -230,8 +231,34 @@ export class Cognite3DViewer {
...
+    });
+
+    if (results) {
+      const result = results[0]; // Nearest intersection
+      const model: Cognite3DModel = this.models.find(v => v.cadNode === result.cadNode)!;
+      const intersection: Intersection = {
+        model,
+        // TODO 2020-03-21 larsmoa: Map to nodeId,
+        nodeId: -1,
+        treeIndex: -result.treeIndex,
```
dragly: Why the negation?

viewer/src/migration/Cognite3DViewer.ts
```diff
@@ -241,9 +268,7 @@ export class Cognite3DViewer {
   private onClick(_callback: (event: PointerEvent) => void): void {
     throw new NotSupportedInMigrationWrapperError();
   }
-  private offClick(_callback: (event: PointerEvent) => void): void {
-    throw new NotSupportedInMigrationWrapperError();
-  }
+  private offClick(_callback: (event: PointerEvent) => void): void {}
```
dragly: Why is only this changed?

viewer/src/views/threejs/cad/CadNode.ts
```diff
@@ -176,33 +176,42 @@ export class CadNode extends THREE.Object3D {
...
         auditTime(100),
         fromThreeCameraConfig(),
+        suspendIfLoadingToggled(),
 
         // Determine all wanted sectors
         this._sectorCuller.determineSectors(),
 
         // Take sectors within budget
-        tap(c => console.log(`Want ${c.filter(x => x.levelOfDetail !== LevelOfDetail.Discarded).length} sectors`)),
+        // tap(c => console.log(`Want ${c.filter(x => x.levelOfDetail !== LevelOfDetail.Discarded).length} sectors`)),
```
dragly: Ooops... I did not notice these getting pulled into master. These should probably be removed altogether (or at least hidden behind a debug flag).

viewer/src/views/threejs/cad/CadNode.ts
```diff
@@ -176,33 +176,42 @@ export class CadNode extends THREE.Object3D {
   }
 
   private createLoadSectorsPipeline(): Subject<ThreeCameraConfig> {
+    const suspendLoading = this.loadingHints.suspendLoading;
+
     const loadSectorOperator = flatMap((s: WantedSector) => this._repository.loadSector(s));
     const consumeSectorOperator = flatMap((sector: ParsedSector) => this.rootSector.consumeSector(sector.id, sector));
+    function suspendIfLoadingToggled<T>(): OperatorFunction<T, T> {
+      return mergeMap<T, ObservableInput<T>>(x => (!suspendLoading ? of(x) : empty()));
```
dragly: This will create new observables that will be subscribed to for each thing that goes through the pipeline. Not too bad if it happens at the top, where we will get one or two new subscriptions every 100 ms, but this causes some serious performance issues when done further down the line. Calls to `observable.subscribe` was what took between 16 and 60 ms now and then in the first implementation, where we had something similar.

Maybe we can achieve something with a filter instead? I think those are evaluated runtime. @j-bjorne: correct me if I am wrong.

viewer/src/views/threejs/cad/CadNode.ts
```diff
@@ -176,33 +176,42 @@ export class CadNode extends THREE.Object3D {
   }
 
   private createLoadSectorsPipeline(): Subject<ThreeCameraConfig> {
+    const suspendLoading = this.loadingHints.suspendLoading;
```
dragly: This seems like an unrelated change. Or is it necessary for raypicking?

viewer/src/views/threejs/cad/CadNode.ts
```diff
@@ -176,33 +176,42 @@ export class CadNode extends THREE.Object3D {
...
         // Take sectors within budget
-        tap(c => console.log(`Want ${c.filter(x => x.levelOfDetail !== LevelOfDetail.Discarded).length} sectors`)),
+        // tap(c => console.log(`Want ${c.filter(x => x.levelOfDetail !== LevelOfDetail.Discarded).length} sectors`)),
         map(wantedSectors => this.budget.filter(wantedSectors, this._sectorScene)),
-        tap(c => {
-          console.log(`Filtered to ${c.filter(x => x.levelOfDetail !== LevelOfDetail.Discarded).length} sectors`);
-        }),
+        // tap(c => {
+        //   console.log(`Filtered to ${c.filter(x => x.levelOfDetail !== LevelOfDetail.Discarded).length} sectors`);
+        // }),
```
dragly: Same as above, should probably be removed altogether.

viewer/src/views/threejs/cad/picking.ts
```diff
@@ -38,7 +39,7 @@ export function intersectCadNodes(cadNodes: CadNode[], input: IntersectCadNodesI
       results.push(result);
     }
   }
-  return results;
+  return results.sort((l, r) => l.distance - r.distance);
```
dragly: Nice :+1:

viewer/src/views/threejs/cad/picking.ts
```diff
@@ -48,39 +49,48 @@ export function intersectCadNode(cadNode: CadNode, input: IntersectCadNodesInput
   // TODO add warning if parent has transforms
   const oldParent = cadNode.parent!;
```
dragly: We can probably remove the exclamation mark here now.

viewer/src/migration/Cognite3DViewer.ts
larsmoa: Because stoopid.

viewer/src/migration/Cognite3DViewer.ts
larsmoa: No idea :S

viewer/src/views/threejs/cad/CadNode.ts
larsmoa: @j-bjorne : bump

larsmoa: Removed suspend.

PR #318, opened 2020-03-22
Add a simple CAD sector budget

larsmoa:
Implement a simple sector budget to limit the data loaded for a given CadNode.


larsmoa:
Awaiting https://github.com/cognitedata/reveal/pull/320 (needed to fix CI build).

viewer/src/models/cad/CadBudget.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  private getCandidateSpendage(candidate: WantedSector, sectorMetadata: SectorMetadata): CadBudgetSpendage {
+    switch (candidate.levelOfDetail) {
+      case LevelOfDetail.Simple:
+        // TODO 2020-03-22 larsmoa: Need some better estimate of complexity than 1 drawCount
+        return { downloadSize: sectorMetadata.facesFile.downloadSize, drawCallCount: 1 };
+      case LevelOfDetail.Detailed:
+        return {
+          downloadSize: sectorMetadata.indexFile.downloadSize,
```
j-bjorne: Does this not need to traverse upwards and see the total cost for everything above it too?

larsmoa: We have moved the responsibility for doing the counting to the backend to avoid traversal. But I do not think so - because that would lead to counting the sectors high up in the tree many times.

j-bjorne: I'll let it go through for now. I would like a conversation about this at some point. I had some questions / ideas which I thought would make it so that upwards traversal would be needed.

PR #319, opened 2020-03-22
Do not create model twice in migration examples

dragly:
The model will created inside `addModel` anyways and the `model` reference was to an unused object.


dragly:
... which confused me for quite some time while debugging :smile:

PR #323, opened 2020-03-23
Add support for localPath in migration wrapper

dragly:


dragly:
Looking at the conflicts, I am having some second thoughts about this. The NodeIdAndTreeIndexMaps wrapper does not really work in its current state if there is nowhere to look up the unknown nodes (that is, it is usually from the API). I am actually leaning towards dropping support for local path in the wrapper for the time being.

larsmoa:
I agree. We do not need local/byUrl support in wrapper. Suggest you close, @dragly

PR #324, opened 2020-03-23
Create node ID to tree index map and use it to implement setNodeColor

dragly:
The migration wrapper uses node IDs as part of its API, which means that we need to map from node IDs to tree indices and vice versa. To do this, we can build up a map as we receive more data from the incoming sectors. We also need to refresh colors or any other properties as more data is received by looking for relevant node IDs. Most of the node ID specific code is pushed out to Cognite3DModel.


f3df/src/renderables.rs
```diff
@@ -192,17 +211,21 @@ pub fn convert_sector(sector: &crate::Sector) -> Vec<Face> {
...
                 });
             }
         }
     }
 
-    instance_data
+    Sector {
+        faces,
+        node_id_to_tree_index_map,
+        tree_index_to_node_id_map,
```
larsmoa: Is publishing two maps the most efficient approach, or should we just have one array of (treeIndex -> nodeId)?

What guarantees do we have for `treeIndices` within a sector? Will they be continous?

viewer/src/views/threejs/cad/RootSectorNode.ts
```diff
@@ -60,10 +71,18 @@ export class RootSectorNode extends SectorNode {
           return new THREE.Group();
         }
         case LevelOfDetail.Simple: {
-          return consumeSectorSimple(id, data as SectorQuads, metadata, this.materials);
+          const simpleData = data as SectorQuads;
+          this._nodeIdToTreeIndexMap = new Map([...this._nodeIdToTreeIndexMap, ...simpleData.nodeIdToTreeIndexMap]);
+          this._treeIndexToNodeIdMap = new Map([...this._treeIndexToNodeIdMap, ...simpleData.treeIndexToNodeIdMap]);
```
larsmoa: Isn't this a very slow way of updating the map? Will do a complete remake every time a sector is loaded?

viewer/src/views/threejs/cad/RootSectorNode.ts
```diff
@@ -60,10 +71,18 @@ export class RootSectorNode extends SectorNode {
...
+          this.dispatchEvent({ type: 'nodeIdToTreeIndexMapUpdated', sector });
+          return consumeSectorSimple(id, simpleData, metadata, this.materials);
         }
         case LevelOfDetail.Detailed: {
-          return consumeSectorDetailed(id, data as Sector, metadata, this.materials);
+          const detailedData = data as Sector;
+          this._nodeIdToTreeIndexMap = new Map([...this._nodeIdToTreeIndexMap, ...detailedData.nodeIdToTreeIndexMap]);
+          this._treeIndexToNodeIdMap = new Map([...this._treeIndexToNodeIdMap, ...detailedData.treeIndexToNodeIdMap]);
+          this.dispatchEvent({ type: 'nodeIdToTreeIndexMapUpdated', sector });
```
larsmoa: De-duplicate code from above.

viewer/src/models/cad/types.ts
```diff
@@ -106,6 +106,9 @@ export type InstancedMesh = {
 // TODO 2019-12-05 larsmoa: Rename to e.g. SectorGeometry to avoid
 // confusion with other Sector-class
 export interface Sector {
+  readonly nodeIdToTreeIndexMap: Map<number, number>;
+  readonly treeIndexToNodeIdMap: Map<number, number>;
```
larsmoa: Should we store this per sector? Seems like this only is intermittent data, and _only_ (?) used by the migration wrapper? Is there any way we can avoid storing the data when we don't need it? I'm thinking the map should live somewhere inside `Cognite3dModel`, but I'm open to discuss this.

larsmoa:
I'm a bit skeptic to the event as it's not type safe (?) and is not something we use elsewhere. Are there any better options?

viewer/src/views/threejs/cad/RootSectorNode.ts
```diff
@@ -25,13 +25,17 @@ export class RootSectorNode extends SectorNode {
   private readonly materials: Materials;
 
   private readonly consumeSectorCache: MemoryRequestCache<string, ParsedSector, THREE.Group>;
+  private _nodeIdToTreeIndexMap: Map<number, number>;
+  private _treeIndexToNodeIdMap: Map<number, number>;
```
larsmoa: Would an array be more efficient? I.e. Float64Array

viewer/src/views/threejs/cad/RootSectorNode.ts
dragly: I agree. Would a for-loop be better?

f3df/src/renderables.rs
dragly: I do not know, but I was thinking that building the map in JS-land would be slower than doing it in Rust? If we only pass the `tree_index_to_node_id_map`, then we need to build the other one in JS somehow.

I do not know the guarantees we have for `treeIndices`, but I am pretty sure they are not continuous, because they depend on the tree structure from the original file, while sectors depend on the spatial structure of the geometry.

viewer/src/models/cad/types.ts
dragly: Is the type you are commenting on here not also only intermittent? Where would you otherwise store it? I think we throw away this data once the sector is consumed.

dragly: We could store the final map in Cognite3dModel and not the root sector, though. That probably makes sense.

dragly:
> I'm a bit skeptic to the event as it's not type safe (?) and is not something we use elsewhere. Are there any better options?

We could register a callback in CadNodeOptions?

viewer/src/views/threejs/cad/RootSectorNode.ts
dragly: For the `treeIndexToNodeIdMap`, yes, but then you need to go from an intermittent `Map` in each sector and convert that to the Float64Array afterwards. I would strongly advise against storing `nodeIdToTreeIndexMap` as a Float64Array. It would be very sparse and potentially take up to 2^53 * 8 bytes to store. As soon as you insert a large node ID, it needs to allocate all the unused slots up to that ID.

f3df/src/renderables.rs
larsmoa: Hmm, can RS Maps be transfered without serialization to JS? Or does the map need to be rebuilt in JS-land?

viewer/src/views/threejs/cad/RootSectorNode.ts
larsmoa: Would probably be fastest

viewer/src/models/cad/types.ts
larsmoa: Do we need to have this intermittently stored in `Sector`? Gives me the creeps to have collections that are valid only for a brief period...

viewer/src/views/threejs/cad/RootSectorNode.ts
larsmoa: `nodeIdToThreeIndexMap` cannot be an array - totally agreed. However, `treeIndexToNodeIdMap` can and it would reduce allocations a lot as it can be allocated up front.

f3df/src/renderables.rs
dragly: We currently use serialization, but it should be possible to create the JS `js_sys::Map` in Rust and fill it directly. I am not sure how efficient either approach is, but it is just a bit more awkward to work with `js_sys::Map` than `std::collections::HashMap` in Rust. I was thinking to wait until we have some benchmarks before converting at least.

viewer/src/models/cad/types.ts
dragly: I do not think I understand what you prefer instead? How can we avoid storing the maps in a data structure after parsing the files? All the data in `Sector` only exists for a limited amount of time.

viewer/src/views/threejs/cad/RootSectorNode.ts
dragly: Would you prefer that one is a map and the other a Float64Array? I was thinking it would be easier to work with if both had the same type, but we can make them different (this is the way it is in the old viewer).

viewer/src/models/cad/types.ts
dragly: Closing this as we discussed it offline :)

dragly:
Although this will likely be replaced by #326, I updated it to reflect the review comments for future reference (in case we need to pull it back in).

viewer/src/migration/NodeIdAndTreeIndexMaps.ts
```diff
@@ -46,4 +54,33 @@ export class NodeIdAndTreeIndexMaps {
     });
     return result;
   }
+
+  getNodeId(treeIndex: number): number | undefined {
```
larsmoa: Why does `getNodeId()` return `number | undefined` while `getTreeIndex` returns a `Promise`? I suggest `getNodeId` is renamed to `tryGetNodeId` if it cant be a promise for some reason.

viewer/src/migration/NodeIdAndTreeIndexMaps.ts
```diff
@@ -46,4 +54,33 @@ export class NodeIdAndTreeIndexMaps {
...
+        this.updateMapsFromMap(detailedData.nodeIdToTreeIndexMap);
+        break;
+      }
+      default: {
+        break;
+      }
+    }
+  }
+
+  updateMapsFromMap(nodeIdToTreeIndexMap: Map<number, number>) {
```
larsmoa: Make private

viewer/src/migration/NodeIdAndTreeIndexMaps.ts
```diff
@@ -5,15 +5,20 @@
...
   readonly nodeIdToTreeIndexMap: Map<number, number>;
+  readonly treeIndexToNodeIdMap: Map<number, number>;
   readonly nodeIdRequestObservable: Subject<NodeIdRequest>;
   readonly nodeIdResponse: Observable<Node3D>;
   constructor(modelId: number, revisionId: number, client: CogniteClient) {
     this.nodeIdToTreeIndexMap = new Map();
+    this.treeIndexToNodeIdMap = new Map();
     this.nodeIdRequestObservable = new Subject();
     this.nodeIdResponse = this.nodeIdRequestObservable.pipe(
       bufferTime(50),
```
larsmoa: Is there any reason to wait? Or will risk choppy performance from this? I guess 50 is quite low though. Just want to make sure.

viewer/src/migration/Cognite3DModel.ts
```diff
@@ -101,8 +123,15 @@ export class Cognite3DModel extends THREE.Object3D {
     })();
   }
 
-  resetNodeColor(_nodeId: number): void {
-    throw new NotSupportedInMigrationWrapperError();
+  resetNodeColor(nodeId: number): void {
```
larsmoa: I think we should change signature to reflect that its async. It will still 'work' in old clients.

viewer/src/views/threejs/cad/RootSectorNode.ts
```diff
@@ -60,10 +60,13 @@ export class RootSectorNode extends SectorNode {
...
         }
         case LevelOfDetail.Simple: {
-          return consumeSectorSimple(id, data as SectorQuads, metadata, this.materials);
+          const simpleData = data as SectorQuads;
+          return consumeSectorSimple(id, simpleData, metadata, this.materials);
         }
         case LevelOfDetail.Detailed: {
-          return consumeSectorDetailed(id, data as Sector, metadata, this.materials);
+          const detailedData = data as Sector;
+          this.dispatchEvent({ type: 'nodeIdToTreeIndexMapUpdated', sector });
```
larsmoa: Old junk?

viewer/src/migration/NodeIdAndTreeIndexMaps.ts
dragly: Because we cannot ask the API for a node ID from a tree index currently. I agree that we can rename it to tryGetNodeId.

viewer/src/migration/NodeIdAndTreeIndexMaps.ts
dragly: Is this comment misplaced? There is no waiting here.

dragly: This is from the other PR for requesting node IDs from the API if they are not available locally. The wait time here is added in cases there are thousands of simultaneous requests. If there is no waiting, we will be sending one API request for each missing node ID, which I think we agreed that we wanted to avoid. However, there is no waiting if the data is already available after the current PR, because then you will already have the data available from the files. There is only a wait in case you are trying to set the color or hide a node that we currently do not know the tree index for.

viewer/src/migration/Cognite3DModel.ts
dragly: Fair enough. The behavior is pretty much the same, except you will get access to the promise with your suggestion.

viewer/src/views/threejs/cad/RootSectorNode.ts
dragly: Ah, yes. Good catch!

dragly:
@larsmoa updated according to your comments

PR #326, opened 2020-03-23
Implement setNodeColor by asking API for nodeId to treeIndex map

dragly:


viewer/src/migration/Cognite3DModel.ts
```diff
@@ -112,6 +153,42 @@ export class Cognite3DModel extends THREE.Object3D {
...
+  private setNodeColorByTreeIndex(treeIndex: number, r: number, g: number, b: number) {
+    this.nodeColors.set(treeIndex, [r, g, b, 255]);
+    this.cadNode.requestNodeUpdate([treeIndex]);
+  }
+
+  private setupPipeline() {
+    const nodeIdRequest = new Subject<NodeIdRequest>();
+    nodeIdRequest
+      .pipe(
+        bufferTime(3000),
```
larsmoa: Isn't this a bit over the top? E.g. if there's a loop setting a lot of colors, it can process all requests right after the function setting the colors has returned. I suggest more like ~50 ms maximum.

larsmoa: Also, should there be a safety-valve that splits requests into chunks of maximum 1000 nodeIds per request?

dragly: Ooops, yes, that should have been much lower. I forgot to change it after testing if it worked. I will set it to 50 ms.

Why would we need to limit it to 1000 per request? Because of cursors and pagination? I think this should be handled by the SDK already: https://github.com/cognitedata/cognite-sdk-js/blob/06b741b1043d2cf3df4f049020f56718e5806ae6/src/resources/3d/revealNodes3DApi.ts#L16

dragly:
> Would also be nice if the handling of the NodeId<>TreeIndex wrapper was externalized to a utility class.

I would actually prefer to have it in this class as long as this is the only place it is used, but we can change it if you want. It just seems like a class that will simply wrap the calls to Map.

larsmoa:
> > Would also be nice if the handling of the NodeId<>TreeIndex wrapper was externalized to a utility class.
> 
> I would actually prefer to have it in this class as long as this is the only place it is used, but we can change it if you want. It just seems like a class that will simply wrap the calls to Map.

Having it in a utility class enables testing :) I think it's more complex than a map since it handles 1. waiting for multiple invocations, 2. triggering requests for fetch the data and 3. applying changes whenever pending data is received.

dragly:
Good point. I will add a utility class.

dragly:
@larsmoa ping?

PR #331, opened 2020-03-24
Remove verification of package-lock.json

dragly:
It appears dependabot makes bad PRs that remove these lines anyways.
We therefore need to drop the verification to keep dependabot pushing
PRs.


dragly:
See for instance https://github.com/cognitedata/reveal/commit/f8b6d5cae676ce88c73abb97103152aec811d8ad#diff-09d6724bde495fd0c2ba310fd66df417L50

PR #354, opened 2020-03-26
Revert "Allocate textures holding color and visibility to be big enough to hold the correct number of tree indices."

dragly:
The shaders are still using fixed values. The lookups are therefore wrong and it is no longer possible to change the color or visiblity of nodes.


dragly:
No worries ;) And great! I will test it.

dragly:
#359 works. Closing.

PR #366, opened 2020-03-27
HtmlOverlayHelper class for helping with Html Overlay

j-bjorne:
Added an overlay helper class.
Added an get bounding box from api to wrapper but since examples don't have old modelId we can't use the v1 endpoint to get bounding box.

### DIFF
--- a/viewer/src/migration/Cognite3DModel.ts
+++ b/viewer/src/migration/Cognite3DModel.ts
@@ -7,7 +7,7 @@ import * as THREE from 'three';
 import { Color } from './types';
 import { NotSupportedInMigrationWrapperError } from './NotSupportedInMigrationWrapperError';
 import { CadModel } from '../models/cad/CadModel';
-import { toThreeJsBox3, CadNode } from '../views/threejs';
+import { toThreeJsBox3, CadNode, toThreeMatrix4 } from '../views/threejs';
 import { loadCadModelFromCdf } from '../datasources/cognitesdk';
 import { CadRenderHints } from '../views/CadRenderHints';
 import { NodeAppearance } from '../views/common/cad/NodeAppearance';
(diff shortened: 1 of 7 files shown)


viewer/src/migration/Cognite3DModel.ts
```diff
@@ -80,13 +80,27 @@ export class Cognite3DModel extends THREE.Object3D {
...
 
     const bounds = this.cadModel.scene.root.bounds;
     return toThreeJsBox3(box || new THREE.Box3(), bounds, this.cadModel.modelTransformation);
   }
 
+  async getBoundingBoxFromApi(nodeId: number, box?: THREE.Box3): Promise<THREE.Box3> {
+    const response = await this.client.revisions3D.retrieve3DNodes(this.modelId, this.revisionId, [{ id: nodeId }]);
+    if (response.length < 1) {
+      throw new Error('NodeId not found');
+    }
```
larsmoa: Any reason not to use `=== 0`?

viewer/src/views/threejs/HtmlOverlayHelper.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  addOverlayElement(htmlElement: HTMLElement, position3D: THREE.Vector3) {
+    this.htmlOverlays.set(htmlElement, position3D);
+  }
+
+  removeOverlayElement(htmlElement: HTMLElement) {
+    this.htmlOverlays.delete(htmlElement);
+  }
+
+  updatePositions(renderer: THREE.Renderer, camera: THREE.PerspectiveCamera) {
```
larsmoa: Suggest we replace 'renderer' with a `HTMLElement` and add an unit test :)

dragly: I think passing an `HTMLElement` makes it easier to use the API wrong. I think we specifically should require the `renderer.domElement` as most other elements will not make sense to use together with `worldToViewport`. I might be wrong about this, but are there any other HTML elements that we expect users to wish to _overlay_ their HTML elements on apart from the canvas? If you want an `HTMLElement` to make the function unit-testable, I would prefer to keep the signature of this and have it call an internal function that takes an `HTMLElement` that is unit tested.

Also regarding unit testing: This is a function I would prefer to visually test, but adding a unit test is OK. I think the important outcome of this function is the visual aspect. The actual work is done by `worldToViewport`, which should be unit tested. The fact that we use `style.top` and `style.left` are the only things you can test from this function, and I think the fact that those are used (as opposed to padding, margin, etc) is bordering to implementation details. Again, not opposed to a unit test, but I would prefer a visual test :)

viewer/src/migration/Cognite3DModel.ts
j-bjorne: No, I can change it

PR #368, opened 2020-04-02
GPU driven determineSectors

larsmoa:
GPU driven implementation of determineSectors() that takes into account how much each sector will contribute to the final, visual result.


examples/src/simple.ts
```diff
@@ -26,10 +37,58 @@ async function main() {
...
   const { position, target, near, far } = cadNode.suggestCameraConfig();
-  const camera = new THREE.PerspectiveCamera(75, window.innerWidth / window.innerHeight, near, far);
+  camera.near = near;
+  camera.far = 3 * far;
+  camera.fov = 75;
+  camera.aspect = window.innerWidth / window.innerHeight;
+  camera.updateProjectionMatrix();
+  camera.updateMatrixWorld();
+
+  document.addEventListener('keypress', event => {
```
dragly, line 61: Move to helper class

examples/webpack.config.js
```diff
@@ -230,6 +230,18 @@ module.exports = env => {
         resolve('public/'),
         resolve('dist/'),
       ],
+
+      proxy: {
```
dragly, line 234: Unrelated change?

viewer/src/data/worker/WorkerPool.ts
```diff
@@ -16,7 +16,15 @@ interface PooledWorker {
 }
 
 export class WorkerPool {
+  static get defaultPool(): WorkerPool {
```
dragly, line 19: Unrelated change?

dragly: I like it, though :)

viewer/src/glsl/base/updateFragmentDepth.glsl
```diff
@@ -1,6 +1,4 @@
-#ifdef GL_EXT_frag_depth
-
-void updateFragmentDepth(vec3 p,mat4 projectionMatrix){
+float computeFragmentDepth(vec3 p, mat4 projectionMatrix) {
```
dragly, line 1: Unrelated change

viewer/src/views/threejs/cad/CadNode.ts
```diff
@@ -182,16 +186,19 @@ export class CadNode extends THREE.Object3D {
...
+      (consumedSector: ConsumedSector) =>
+        currentLevelOfDetail.get(consumedSector.sectorId) !== consumedSector.levelOfDetail
     );
     const pipeline = new Subject<ThreeCameraConfig>();
     pipeline
       .pipe(
-        auditTime(100),
+        // TODO 2020-04-15 larsmoa: Reduce delay to something more sensible
+        // Temporary workaround to avoid flooding the GPU pipeline with readPixels.
+        debounceTime(150),
```
dragly: Will not the change to `debounceTime` cause us to potentially never update the sectors? That is, if you continue to move the camera, debounceTime will just postpone and postpone updating forever, leaving us with nothing on screen. Why not just increase the time sent to `auditTime`?

viewer/src/culling/ByVisibilityGpuSectorCuller.ts
```diff
@@ -0,0 +1,248 @@
...
+  coverageUtil?: OrderSectorsByVisibilityCoverage;
+};
+
+function assert(condition: boolean, message: string = 'assertion hit') {
+  // tslint:disable-next-line: no-console
+  console.assert(condition, message);
+}
+
+class TakenSectorMap {
+  get totalCost() {
```
dragly, line 59: Why is this a getter, while getWantedSectorCount is a member function?

viewer/src/culling/ByVisibilityGpuSectorCuller.ts
```diff
@@ -0,0 +1,248 @@
...
+
+    if (!changed) {
+      return takenSectors;
+    }
+    takenSectors.clear();
+    this.models.forEach(x => takenSectors.initializeScene(x.scene));
+
+    // Update wanted sectors
+    const prioritized = coverageUtil.orderSectorsByVisibility(this.camera);
+    this.lastUpdate.matrix = this.camera.matrix.clone();
```
dragly, line 182: Use `this.lastupdate.matrix.copy(this.camera.matrix)` instead to avoid the extra allocation.

viewer/src/culling/ByVisibilityGpuSectorCuller.ts
```diff
@@ -0,0 +1,248 @@
...
+    let debugAccumulatedPriority = 0.0;
+    const prioritizedLength = prioritized.length;
+
+    let i = 0;
+    for (i = 0; i < prioritizedLength && takenSectors.totalCost < costLimit; i++) {
+      const x = prioritized[i];
+      takenSectors.markSectorDetailed(x.scene, x.sectorId, x.priority);
+      debugAccumulatedPriority += x.priority;
+    }
+    console.log(
```
dragly, line 200: Remove logging

viewer/src/culling/TakenSectorTree.ts
```diff
@@ -0,0 +1,123 @@
...
+      `Sector ${sectorId} must be a Simple-sector, but got ${this.sectors[sectorId].lod}`
+    );
+    this.setSectorLod(sectorId, LevelOfDetail.Detailed);
+    this.markAllDiscardedChildrenAsSimple(sectorId);
+  }
+
+  private markAllDiscardedChildrenAsSimple(sectorId: number) {
+    for (const child of this.sectors[sectorId].sector.children) {
+      if (this.getSectorLod(child.id) === LevelOfDetail.Discarded) {
+        if (child.facesFile.fileName !== null) {
```
dragly, line 96: If the filename is null, should we not stop here and just use the empty sector as a simple representation rather than loading child sectors with simple data with higher number of faces?

viewer/src/glsl/sector/sectorCoverage.frag
```diff
@@ -0,0 +1,17 @@
...
+varying lowp vec2 v_seed;
+
+
+void main() {
+    float v = rand2d(gl_FragCoord.xy + v_seed);
+    if (v >= v_coverageFactor) {
+        discard;
+    }
+
+    gl_FragColor = vec4(v_color, pow(v_position.z / v_position.w, 10.0)); // Ghetto-linearization of the depth
```
dragly, line 16: Why 10.0?

dragly:
Approving, but I would prefer if the above comments are resolved.

examples/webpack.config.js
larsmoa: Yes, this was necessary in order to be able to test performance on mobile. It's cherry-picked into a separate PR (https://github.com/cognitedata/reveal/pull/391)

viewer/src/data/worker/WorkerPool.ts
larsmoa: Kinda unrelated, but its to support multiple models better by not flooding with N workers for each model added.

viewer/src/glsl/base/updateFragmentDepth.glsl
larsmoa: Yes, cherry-picked to PR #390 - will merge this first to ensure it's not part of the final commit.

viewer/src/views/threejs/cad/CadNode.ts
larsmoa: Yes, it will. I considered it the lesser evil of causing stuttering while moving around. Because we load geometry for the full model at all times (just at low detail), I considered acceptable for now. I realize it's far from optimal, but I don't see a good solution right now and the upside of the new culler outbeats the disadvantage this results in.

Note! A theory I have is that we should not do `render(); readPixels();` in the same call to `determineSectors()` as this will cause the render-pipeline to flush before `readPixels` returns.

viewer/src/culling/ByVisibilityGpuSectorCuller.ts
larsmoa: It's maybe a stupid pattern, but I tend to use functions to indicate that there is some computational cost associated with the operation. Here getWantedSectorCount() is `O(M*N)` where M is models and N is sectors. `get totalCost` is `O(M)`.

I realize I'd have a hard time to make this argument in front of a jury and the natural follow-up-question would be "Why not clearly indicate this in the function name instead?" - and hence I will rename it to `determineWantedSectorCount()`

viewer/src/culling/ByVisibilityGpuSectorCuller.ts
larsmoa: Good catch, fixed.

viewer/src/glsl/sector/sectorCoverage.frag
larsmoa: Remaints of a concept I tested, removed now.

viewer/src/culling/TakenSectorTree.ts
larsmoa: Good point, fixed.

examples/src/simple.ts
larsmoa: Created a separate example for this (short lived - will be the one and only culler when we get the new fix from J)

PR #369, opened 2020-04-03
Remove template parameter from SectorCuller and add some doc.

larsmoa:
Remove unused template parameter to make my head spin a bit slower ;)

### DIFF
--- a/viewer/src/culling/SectorCuller.ts
+++ b/viewer/src/culling/SectorCuller.ts
@@ -2,9 +2,13 @@
  * Copyright 2020 Cognite AS
  */
 
-import { OperatorFunction } from 'rxjs';
 import { WantedSector } from '../data/model/WantedSector';
+import { DetermineSectorsByProximityInput } from '../models/cad/determineSectors';
 
-export interface SectorCuller<Input> {
-  determineSectors(): OperatorFunction<Input, WantedSector[]>;
+/**
+ * Interface for implementations that are responsible for determining
+ * what sectors should be loaded.
+ */
+export interface SectorCuller {
+  determineSectors(input: DetermineSectorsByProximityInput): WantedSector[];
 }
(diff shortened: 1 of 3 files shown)


viewer/src/culling/SectorCuller.ts
```diff
@@ -2,9 +2,13 @@
  * Copyright 2020 Cognite AS
  */
 
-import { OperatorFunction } from 'rxjs';
 import { WantedSector } from '../data/model/WantedSector';
+import { DetermineSectorsByProximityInput } from '../models/cad/determineSectors';
 
-export interface SectorCuller<Input> {
```
dragly, line 8: @j-bjorne you wanted this to be generic. Any comments on whether we should keep it that way?

viewer/src/culling/SectorCuller.ts
```diff
@@ -2,9 +2,13 @@
...
-  determineSectors(): OperatorFunction<Input, WantedSector[]>;
+/**
+ * Interface for implementations that are responsible for determining
+ * what sectors should be loaded.
+ */
+export interface SectorCuller {
+  determineSectors(input: DetermineSectorsByProximityInput): WantedSector[];
```
dragly, line 13: If this only accepts the input needed by the ProximitySectorCuller, then I am not sure if we need to keep the generic interface.

viewer/src/views/threejs/cad/CadNode.ts
```diff
@@ -177,7 +176,7 @@ export class CadNode extends THREE.Object3D {
         fromThreeCameraConfig(),
 
         // Determine all wanted sectors
-        this._sectorCuller.determineSectors(),
+        map(input => this._sectorCuller.determineSectors(input)),
```
dragly, line 179: This is according to what we agreed, which was to use non-RxJS implementations as far as possible. In other words, this change should be kept even if the others are rejected.

PR #370, opened 2020-04-06
Introduce a limit on number of elements in the cache

larsmoa:
Keep a maximum number of elements in the cache and evict based on last accessed. Attempt to reduce memory footprint.

### DIFF
--- a/viewer/src/cache/MemoryRequestCache.ts
+++ b/viewer/src/cache/MemoryRequestCache.ts
@@ -4,25 +4,71 @@
 
 import { RequestCache, RequestDelegate } from './RequestCache';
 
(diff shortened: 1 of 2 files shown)


viewer/src/cache/MemoryRequestCache.ts
```diff
@@ -4,25 +4,71 @@
...
+  request(id: Key, data: Data): Result {
     const existing = this._results.get(id);
     if (existing) {
-      return existing;
+      return existing.value;
     }
-    const result: Result = this._request(id, data);
+
+    this.cleanupCache();
+    const value = ;
```
dragly, line 48: Is this allowed?

larsmoa: Dafuc - appearantly yes :s

PR #375, opened 2020-04-07
Properly discard sectors

dragly:
Do not dispose of materials, since they are shared.
Create a dummy reference to the geometry in case there are lingering
references to the mesh somewhere.

### DIFF
--- a/viewer/src/models/cad/parseSectorData.ts
+++ b/viewer/src/models/cad/parseSectorData.ts
@@ -69,7 +69,7 @@ export function createParser(fetchCtmFile: FetchCtmDelegate): ParseSectorDelegat
     (fileId: number) => {
       return loadCtmGeometry(fileId, fetchCtmFile, workerList);
     },
-    { maxElementsInCache: 10000 }
+    { maxElementsInCache: 300 }
   );
 
   async function parse(sectorId: number, sectorArrayBuffer: Uint8Array): Promise<Sector> {
--- a/viewer/src/models/cad/parseSectorData.ts
+++ b/viewer/src/models/cad/parseSectorData.ts
@@ -65,9 +65,12 @@ export function createParser(fetchCtmFile: FetchCtmDelegate): ParseSectorDelegat
   const workerList = createWorkers();
 
   // TODO define the cache outside of the createParser function to make it configurable
-  const loadCtmGeometryCache = new MemoryRequestCache((fileId: number) => {
-    return loadCtmGeometry(fileId, fetchCtmFile, workerList);
-  });
+  const loadCtmGeometryCache = new MemoryRequestCache(
+    (fileId: number) => {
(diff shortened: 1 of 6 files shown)


viewer/src/models/cad/parseSectorData.ts
```diff
@@ -65,9 +65,12 @@ export function createParser(fetchCtmFile: FetchCtmDelegate): ParseSectorDelegat
...
 
   // TODO define the cache outside of the createParser function to make it configurable
-  const loadCtmGeometryCache = new MemoryRequestCache((fileId: number) => {
-    return loadCtmGeometry(fileId, fetchCtmFile, workerList);
-  });
+  const loadCtmGeometryCache = new MemoryRequestCache(
+    (fileId: number) => {
+      return loadCtmGeometry(fileId, fetchCtmFile, workerList);
+    },
+    { maxElementsInCache: 10000 }
```
larsmoa, line 72: This seems a bit excessive?

viewer/src/views/threejs/cad/discardSector.ts
```diff
@@ -2,24 +2,17 @@
...
-    if (mesh.material && mesh.material instanceof THREE.Material) {
-      mesh.material.dispose();
-    }
-    if (mesh.material && mesh.material instanceof Array) {
-      for (const material of mesh.material) {
-        material.dispose();
-      }
+      // NOTE: Forcefully creating a new reference here to make sure
+      // there are no lingering references to the large geometry buffer
+      mesh.geometry = new THREE.Geometry();
```
larsmoa, line 15: Create a shared `emptyGeometry` object that we use rather than newing up one every time (but good idea to assign an empty to make the job easier for GC).

viewer/src/views/threejs/cad/RootSectorNode.ts
```diff
@@ -69,6 +76,8 @@ export class RootSectorNode extends SectorNode {
           throw new Error(`Unsupported level of detail ${sector.levelOfDetail}`);
       }
     })();
+    group.userData.cached = true;
+    group.userData.used = true;
```
larsmoa, line 80: Could we add some functions to `SectorNode`, e.g. `increaseRefCount()`, `isReferenced()` and `decreaseRefCount()`?

dragly: SectorNode is not known in the cache, but I can change to using a refCount in userData?

viewer/src/models/cad/parseSectorData.ts
dragly: Reduced to 300

viewer/src/views/threejs/cad/discardSector.ts
dragly: I agree

PR #376, opened 2020-04-07
Properly clean up more Rust memory in parser worker

dragly:


larsmoa:
This _looks_ right, but doesn't seem to help sadly.

PR #385, opened 2020-04-18
Add semaphore to throttle number of concurrent requests

dragly:


viewer/src/repository/cad/CachedRepository.ts
```diff
@@ -77,7 +83,21 @@ export class CachedRepository implements Repository {
         );
         return merge(
           cachedSectorObservable.pipe(flatMap(wantedSector => this._modelDataCache.get(this.cacheKey(wantedSector)))),
-          uncachedSectorObservable.pipe(this.loadSectorFromNetwork()),
+          uncachedSectorObservable.pipe(
+            flatMap(
+              async _ => this._semaphore.acquire(),
```
larsmoa, line 88: Comment on the logic of whats going on here (e.g. that acquire() returns a value that we filter on further down)

viewer/src/repository/cad/CachedRepository.ts
```diff
@@ -77,7 +83,21 @@ export class CachedRepository implements Repository {
...
+                return {
+                  wantedSector,
+                  ready
+                };
+              }
+            ),
+            filter((data: { wantedSector: WantedSector; ready: boolean }) => data.ready),
+            map((data: { wantedSector: WantedSector; ready: boolean }) => data.wantedSector),
+            this.loadSectorFromNetwork(),
+            tap(_ => this._semaphore.release())
```
larsmoa, line 99: What happens when network request fails? Slot is lost until Semaphore is cleared? Would moving the aquire to within `loadSectorFromNetwork()` and wrap in a `try { aquire(); ...} finally { release(); }` be better/clearer?

viewer/src/data/network/Semaphore.ts
```diff
@@ -0,0 +1,45 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+export class Semaphore {
```
larsmoa, line 5: Should this be moved to `src/utils/`?

viewer/src/data/network/Semaphore.ts
```diff
@@ -0,0 +1,45 @@
...
+  usedSlots() {
+    return this.slotsInUse;
+  }
+
+  clear() {
+    for (const item of this.queue) {
+      item(false);
+    }
+    this.queue = [];
+  }
```
larsmoa, line 44: These functions doesn't feel very Semaphore-ish.

viewer/src/repository/cad/CachedRepository.ts
dragly: Good point. I will see if we can just do it all in one function instead. I found the filter-map-load-tap a bit hard to read myself.

viewer/src/data/network/Semaphore.ts
dragly: Not the biggest fan of utils-folders if we can find a name that better describes something. Any thoughts @j-bjorne ?

viewer/src/data/network/Semaphore.ts
dragly: Perhaps the name of the class should be changed? How about SlotMachine? :stuck_out_tongue:
But seriously, perhaps something with queue is better?

larsmoa: `RateLimiter`? `SomethingLimier`?

viewer/src/repository/cad/CachedRepository.ts
j-bjorne, line 88: Do we need this? Can't it just return a promise instead of async _ ?

viewer/src/repository/cad/CachedRepository.ts
j-bjorne: don't use finalize, there are some changes to be made if we are to use it (namely where we move the queue logic)
And I want to move it so it works on both the i3d f3d AND ctm files

dragly: Added a catchError now. @j-bjorne you probably know best if this is used correctly.

viewer/src/data/network/Semaphore.ts
dragly: RateLimiter

viewer/src/data/network/Semaphore.ts
j-bjorne: I would want it to be called something like JobScheduler, or JobHandler / Limiter and moved into data/workers or data/utils?

PR #389, opened 2020-04-23
Add a few TODOs, types, and an integration test

dragly:
These are points that should be revisited at some point.


f3df/src/renderables.rs
```diff
@@ -87,6 +87,7 @@ pub fn convert_sector(sector: &crate::Sector) -> Sector {
             let i = cell_index_i_j % (cell_size[0]);
             let j = cell_index_i_j / (cell_size[0]);
             let k = cell_index / (cell_size_i_j);
+            // TODO this offset might be wrong
```
larsmoa, line 90: Can you add some context to why you think so?

i3df/src/lib.rs
```diff
@@ -136,7 +137,10 @@ pub fn parse_sector_header(mut input: &mut impl BufRead) -> Result<SectorHeader,
     // read attributes
     let attribute_count = input.read_u32::<LittleEndian>()?;
 
-    if attribute_count != 0 && attribute_count != ATTRIBUTE_COUNT {}
+    // TODO dragly 2020-04-22 should this not return an error?
+    if attribute_count != 0 && attribute_count != ATTRIBUTE_COUNT {
+
+    }
```
larsmoa, line 143: Shouldn't this jsn't be fixed right away?

dragly: Done

f3df/src/renderables.rs
dragly: Done

PR #390, opened 2020-04-23
Fix issues causing viewer not to work on iOS

larsmoa:
This PR fixes two issues:
1. BigUInt64Array is not supported on Safari/iOS - use u32 over u64 structs exposed by Rust
2. Avoid using gl_FragDepthEXT when WebGL extension is unavailable.

### DIFF
--- a/i3df/build.rs
+++ b/i3df/build.rs
@@ -329,7 +329,7 @@ fn main() -> Result<(), Box<dyn Error>> {
         #[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
         pub struct Texture {
             // TODO generate from YAML
-            pub file_id: u32,
+            pub file_id: f64,
             pub width: u16,
             pub height: u16,
         }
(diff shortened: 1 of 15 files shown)


i3df/build.rs
```diff
@@ -329,7 +329,7 @@ fn main() -> Result<(), Box<dyn Error>> {
         #[derive(Clone, Copy, Debug, Default, Deserialize, Serialize)]
         pub struct Texture {
             // TODO generate from YAML
-            pub file_id: u64,
+            pub file_id: u32,
```
dragly, line 332: Just a general comment:
The file_ids have previously been actual file_ids, which might require 53-bit precision. I think they must be exposed as f64 rather than u32 in that case. However, since we are moving towards files like `mesh_x.cmt` where `x` is the `file_id`, I think we could do the same for textures. That is, `texture_x.png` where `x` is in the range for `u32`.

larsmoa: I think actually we _are_ doing the same for textures. But might be a good idea to use f64 here for consistency, @dragly ?

dragly: Hm, yes, we should probably either user f64 here or change the others to u32. I agree that consistency is most important.

PR #394, opened 2020-04-27
Fix broken sector culler caused by invalid interleaved buffer

larsmoa:


viewer/src/views/threejs/OrderSectorsByVisibilityCoverage.ts
```diff
@@ -291,7 +291,7 @@ export class GpuOrderSectorsByVisibilityCoverage {
       addSector(sector.bounds, sector.id, coverageFactors);
     });
 
-    const buffer = new THREE.InstancedInterleavedBuffer(instanceValues, 5);
+    const buffer = new THREE.InstancedInterleavedBuffer(instanceValues, 4);
```
j-bjorne, line 294: Why has this been changed? Was it the wrong number in the beginning or was the data format changed?

PR #395, opened 2020-04-27
Add support for clipping planes

dragly:
This change adds support for clipping planes in all shaders. It supports setting clipping planes both on the renderer (globally) or on each CadNode (locally).


viewer/src/views/threejs/cad/MaterialManager.ts
```diff
@@ -25,29 +25,42 @@ function updateVisibility(visible: VisibilityDelegate, materials: Materials, tre
...
+
+  set clippingPlanes(clippingPlanes: THREE.Plane[]) {
+    this._clippingPlanes = clippingPlanes;
+    this.applyToAllMaterials(material => {
+      material.clippingPlanes = clippingPlanes;
+      console.log('Setting', material, clippingPlanes);
+    });
+  }
+
+  get clipIntersection(): boolean {
```
larsmoa, line 49: It's a.bit confusing what intersection means. Is there a better name for it? Or maybr using a boolean enum?

PR #396, opened 2020-04-27
Free CPU memory from BufferAttributes after upload to GPU.

larsmoa:
We never use the CPU side copies of the BufferAttributes after upload so these can safely be freed to reduce memory consumption. During testing this saved about  100 Mb of memory on a complex model consisting of 5 individual CDF models.

Also fixes a issue on iOS causing the new culler not do work optimally (accidentally committed and pushed to this branch and not its own...)


viewer/src/views/threejs/OrderSectorsByVisibilityCoverage.ts
```diff
@@ -124,7 +124,6 @@ export class GpuOrderSectorsByVisibilityCoverage {
     this.debugRenderer = new THREE.WebGLRenderer({
       antialias: false,
       alpha: false,
-      precision: 'lowp',
```
dragly, line 127: Unrelated?

dragly:
Is there no chance that the `array` property will be used by three.js after being nulled?

larsmoa:
> Is there no chance that the `array` property will be used by three.js after being nulled?

There's one exception - lost GL context, in which case all geometry must be re-uploaded. Our "fix" for this is to just reload all data which I find acceptable.

viewer/src/views/threejs/OrderSectorsByVisibilityCoverage.ts
larsmoa: Yes-ish. But just the debug view for the coverage helper...

larsmoa: It's the `Also fixes a issue on iOS causing the new culler not do work optimally (accidentally committed and pushed to this branch and not its own...)`

PR #397, opened 2020-04-28
Multi Cadmodel Hybrid Pipeline

j-bjorne:
Creating a variation of multi-model which still uses CadModel as a dataretrieves + identifier.
Still need to make it up to date with master, and also add more tests. Would like a spike to look into jest at some point to familiarize myself and maybe create mocks for the biggest classes required for a lot of testing.


examples/src/utils/renderer-debug-widget.ts
```diff
@@ -373,18 +357,19 @@ function updateWantedSectorOverride(
...
     }
     for (const node of acceptedDetailed) {
-      wanted.push({
-        sectorId: node.id,
+      /*wanted.push({
+        id: node.id,
         levelOfDetail: reveal.internal.LevelOfDetail.Detailed,
         metadata: node
-      });
+      });*/
```
larsmoa, line 372: How would one specify a custom SectorCuller after the refac?

viewer/src/component/three/RevealManager.ts
```diff
@@ -0,0 +1,291 @@
...
+              } else {
+                throw new Error('Unknown Source');
+              }
+            },
+            (model, cadModel) => ({
+              ...model,
+              data: cadModel
+            })
+          )
+        );
```
larsmoa, line 99: Extract createCadObservable()

viewer/src/component/three/RevealManager.ts
```diff
@@ -0,0 +1,291 @@
...
+              } else {
+                throw new Error('Unknown Source');
+              }
+            },
+            (model, pointCloudModel) => ({
+              ...model,
+              data: pointCloudModel
+            })
+          )
+        );
```
larsmoa, line 117: Extract createPointCloudObservable()

viewer/src/component/three/RevealManager.ts
```diff
@@ -0,0 +1,291 @@
...
+    sectorRepository: Repository,
+    materialManager: MaterialManager,
+    dataUpdatedCallback: OnDataUpdated,
+    options?: RevealOptions
+  ) {
+    this._sectorCuller = (options && options.internal && options.internal.sectorCuller) || new ProximitySectorCuller();
+    // this._budget = (options && options.budget) || createDefaultCadBudget();
+    this._sectorRepository = sectorRepository;
+    this._materialManager = materialManager;
+    this._modelObservable = this._modelSubject.pipe(
```
larsmoa, line 80: Can pipelines be visualized with some simple ASCII-art inline in the code? Hard for me to follow what's going on

examples/src/utils/renderer-debug-widget.ts
```diff
@@ -373,18 +357,19 @@ function updateWantedSectorOverride(
...
     }
```
larsmoa, line 373: Is it possible to specify a custom SectorCuller now?

viewer/src/views/common/cad/GlobalNodeAppearance.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright 2020 Cognite AS
+ */
+
+export type GlobalColorDelegate = (
+  modelIdentifier: string,
+  treeIndex: number
+) => [number, number, number, number] | undefined;
+export type GlobalVisibilityDelegate = (modelIdentifier: string, treeIndex: number) => boolean;
+
+export interface GlobalNodeAppearance {
```
larsmoa, line 11: What is this used for? We should avoid using `Node` about `Models` as we use it about individual 3D nodes (yes, we should rename `CadNode`).

larsmoa: I don't think we need a global override method for appearance.

examples/src/utils/renderer-debug-widget.ts
j-bjorne: RevealManager has a SectorCuller in it's constructor options, this is not exposed in the SimpleRevealManager version, but if you want it to be part of the Simple API it can be added there too.

viewer/src/component/three/RevealManager.ts
```diff
@@ -0,0 +1,291 @@
...
+    let resolveCb: (cadNode: CadNode) => void | undefined;
+    let rejectCb: (message: string) => void | undefined;
+    const promise = new Promise<CadNode>((resolve, reject) => {
+      resolveCb = resolve;
+      rejectCb = reject;
+    });
+    this._modelSubject.next({
+      format: File3dFormat.RevealCadModel,
+      source: {
+        discriminator: 'cdf',
```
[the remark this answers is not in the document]
j-bjorne: Due to typescript I believe that you can't actually have a typo there. That is because the discriminator is a string type of said string and not a string type of string.

viewer/src/component/three/SimpleRevealManager.ts
```diff
@@ -0,0 +1,38 @@
...
+import { RevealManager, RevealOptions } from './RevealManager';
+import { CogniteClient } from '@cognite/sdk';
+import { CadSectorParser } from '../../data/parser/CadSectorParser';
+import { SimpleAndDetailedToSector3D } from '../../data/transformer/three/SimpleAndDetailedToSector3D';
+import { CachedRepository } from '../../repository/cad/CachedRepository';
+import { MaterialManager } from '../../views/threejs/cad/MaterialManager';
+
+// First iteration of a SimpleRevealManager. Currently tailored to examples but should be tailored to external usecase.
+// Should move to example-helpers.ts as a function without extending
+export class SimpleRevealManager extends RevealManager {
```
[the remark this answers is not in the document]
j-bjorne: Okey, I can agree with that

viewer/src/views/common/cad/GlobalNodeAppearance.ts
j-bjorne: Node appearance is logic used for say changing the colors of a node when selected. RevealManager (Base) should have a default picking handler implemented. It doesn't make sense that the user needs to for each model they have define and reuse the same appearance?

j-bjorne:
ByVisibilityGpuSectorCuller.test.ts and examples/src/gpu-sector-culler.ts are only deleted in this commit as I wanted to see if travis built the code without them. I have them untracked and ready to be added at any time. If you want I can add them so that you can check them out and modify them the way they are supposed to be.

viewer/src/data/model/ExternalSource.ts
```diff
@@ -0,0 +1,8 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+export interface ExternalSource {
+  discriminator: 'external';
+  url: string;
```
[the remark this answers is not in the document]
j-bjorne: Vet ikke om jeg forstår, hva mener du med CamelCase i dette tilfellet? CamelCase er jo camelCase?

viewer/src/component/three/RevealManager.ts
j-bjorne: as in make a private function?

viewer/src/data/model/PointCloud.ts
```diff
@@ -0,0 +1,16 @@
...
+import { PromiseCallbacks } from './PromiseCallbacks';
+import { CdfSource } from './CdfSource';
+import { ExternalSource } from './ExternalSource';
+import { PotreeGroupWrapper } from '../../views/threejs/pointcloud/PotreeGroupWrapper';
+import { PotreeNodeWrapper } from '../../views/threejs/pointcloud/PotreeNodeWrapper';
+
+export interface PointCloud {
+  format: File3dFormat.EptPointCloud;
+  source: CdfSource | ExternalSource;
+  callbacks: PromiseCallbacks<[PotreeGroupWrapper, PotreeNodeWrapper]>;
```
larsmoa, line 15: Why does this thing in `data/model` depend on something in `views/threejs`?

viewer/src/data/provider/BinaryFileProvider.ts
```diff
@@ -0,0 +1,8 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+export interface BinaryFileProvider {
+  fetchBinaryFileFromUrl(url: string): Promise<Uint8Array>;
+  fetchBinaryFileFromCdf(blobId: number, filePath: string): Promise<Uint8Array>;
```
larsmoa, line 7: To me it still looks off that we have one provider that needs to know about all possible data sources, but I've been barking up this tree before.... 

Do we really need both this interface and `BinaryBlobFileProvider?

viewer/src/views/common/cad/GlobalNodeAppearance.ts
larsmoa: Then I think that at least they should implement the same interface to avoid the need for two different types of callbacks. It's fine to always pass the model, even if it's just one model. This will simplify the code for applying colors.

However, I think we should send `CadModel` rather than the ID which doesn't have any meaning (?). Any reason not to?

j-bjorne: Due to the complexity of MaterialManager (it is being used both at the creation stage of an object, pre cadnode creation, and later at transformation from Sector and Quads to THREE.Group). we are creating a map that has a string identifier which links to a set of materials. The map's only functionality is to make sure that we are able to get the materials of a certain cadnode when we are required to modify them. I am not a fan of the implementation, and I do agree that it would be nice to change it, but I currently don't have a good solution for it. Also, the reason why we need two different sets of callbacks is because of how treeindex works. In a multimodel scenario you can't only depend on the tree index because it is the same across multiple models, hence why you need a separate identifier for the model. Say if you create a list next to the viewer and you select a node, you would need to in some way allow that node to change appearance in the 3d viewer, and that logic belongs outside of the manager itself.

viewer/src/data/model/PointCloud.ts
j-bjorne: Because of the current solution of callbacks which we want to remove at some point, but for current implementation we are stuck with the callback (this is something I really want to change as soon as we have time, I don't foresee it being a big job)

viewer/src/data/provider/BinaryFileProvider.ts
j-bjorne: No, BinaryBlobProvider is going away, I think I removed it in the latest push?
Yes, we are going back and forth on this issue. I don't think the BinaryFileProvider.ts is in use currently, so I could remove it for the current merge, last time we talked about this it seemed like you were for having a split at the repository level since it simplified the code structure, but I kinda have lost track xD

larsmoa: image
Both classes are still present

viewer/src/views/common/cad/GlobalNodeAppearance.ts
larsmoa: I understand that in one of the cases you need to identify the model, but why not just provide this to both callbacks?

PR #398, opened 2020-04-30
Add a simple debug viewer in Rust

dragly:
Just in case someone wants to have a look at how a debug viewer can be implemented in Rust with wgpu-rs (which now runs on top of WebGPU in experimental versions of your favorite browser) ;)

This makes a few changes along the lines of the interleaved buffers @christjt is introducing. You probably want to merge his changes first, if you should decide to want to include this in the repository.

It can render primitives as boxes:

image

Or as faces:

image

But it currently lacks most of the primitives. Implementing those is left as an "exercise for the reader".

It is split into a few files now, but main.rs should probably be separated into more files for event handling, rendering and data creation.

And please feel free to throw this PR away. I just wanted to upload it for reference in case you wanted to explore Rust + wgpu-rs with Reveal data at some point.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,6 +36,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "andrew"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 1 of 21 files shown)


dragly:
Run it with

    cargo run /path/to/model/reveal-v8

the reveal-v8 folder should contain a scene.json and the i3df and f3df files.

PR #443, opened 2020-05-07
Add initial support for point clouds to migration layer.

larsmoa:


viewer/src/migration.ts
```diff
@@ -2,7 +2,9 @@
  * Copyright 2020 Cognite AS
  */
 
+export { CogniteModelBase } from './migration/CogniteModelBase';
 export { Cognite3DModel } from './migration/Cognite3DModel';
+export { CognitePointCloudModel } from './migration/CognitePointCloudModel';
 export { Cognite3DViewer } from './migration/Cognite3DViewer';
 export { Intersection } from './migration/intersection';
-export { Color, Cognite3DViewerOptions, AddModelOptions } from './migration/types';
+export { Color, Cognite3DViewerOptions, AddModelOptions, WellKnownModelTypes } from './migration/types';
```
j-bjorne, line 10: I have made a class File3dFormat which I believe we can use instead of WellKnownModelTypes. Can also be renamed

viewer/src/migration/types.ts
```diff
@@ -45,3 +45,12 @@ export interface AddModelOptions {
   onProgress?: (progress: OnProgressData) => void;
   onComplete?: () => void;
 }
+
+export enum WellKnownModelTypes {
```
j-bjorne, line 49: File3dFormat, or rename the File3dFormat to this. Supported3dFormat is another name I like more than WellKnownModelTypes :)

viewer/src/migration.ts
larsmoa: I'm not sure I agree with this. I saw the existing enum, but I don't think there always will be a 1-to-1 mapping between file format and model type. There are also outputs that doesn't map directly to a model that we will use in Reveal, e.g. walkable path. Therefor I think it makes sense with two enums.

Also; migration only exposes types within that package, and Files3dFormat is defined outside the package...

viewer/src/migration/types.ts
larsmoa: I like `WellKnown[Something]` and think its a well established pattern (eg [[1]](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/enum)). However, I have no problem renaming to e.g. `SupportedModelTypes` or just `ModelTypes`

viewer/src/migration/types.ts
```diff
@@ -45,3 +45,12 @@ export interface AddModelOptions {
   onProgress?: (progress: OnProgressData) => void;
   onComplete?: () => void;
 }
+
```
j-bjorne, line 48: Sorry, not sure I was clear :) Since there is already a file with a type that covers this. I would want to avoid creating new. I would suggest renaming the File3DFormat to SupportedModelTypes or something like that :)

PR #482, opened 2020-05-11
Publish script to prevent publishing wrong packages

j-bjorne:
Added a script which makes the base package json private (unable to run npm publish from viewer)
Added npm script publishscript (please come up with a better name? autoPublish?) which takes a parameter 'version', which builds a release version, and publishes from the dist folder.


viewer/package.json
```diff
@@ -14,7 +15,9 @@
     "test": "jest",
     "coverage": "jest --coverage",
     "lint": "tslint --project .",
-    "serve": "npm install && npm link && cd examples && npm install && npm link @cognite/reveal && npm run serve"
+    "serve": "npm install && npm link && cd examples && npm install && npm link @cognite/reveal && npm run serve",
+    "prepublishscript": "npm run release",
+    "publishscript": "node script/copyPackage.js && cd dist && npm publish"
```
larsmoa, line 20: Does it make sense to have `prepublishscript` if it's just `npm run release`? I suggest we run `npm run release` before `node copyPackage` if there's no good reason not to.

viewer/script/copyPackage.js
```diff
@@ -0,0 +1,21 @@
...
+function copyPackage(version) {
+  const jsonString = fs.readFileSync('./package.json', 'utf8')
+  const json = JSON.parse(jsonString);
+
+  if(version) {
+    // anything else we should remove?
+    json.version = version;
+    delete json.private;
+    delete json.scripts.publishscript;
+    delete json.scripts.prepublishscript;
```
larsmoa, line 12: Just so I know, will the package be public if I use this script?

j-bjorne: I don't think so. We already do npm publish without the private parameter in the json file.

viewer/package.json
j-bjorne: So, the reason why we have a prepublishscript is that is will run directly before the publishscript, meaning we can't mistakingly do npm run build and then npm run publishscript.

larsmoa: This needs to be run manually right?

PR #490, opened 2020-05-12
Add postpublishscript which does the publishing

j-bjorne:
This allows the user to add the -- 'version' parameter.
Updates the viewer package with new version.
Deletes private and publish scripts regardless of version input.


larsmoa:
What's the argument for not using the version field from package.json? The version field will be useless if we do this. How do you plan to provide the correct version number when we e.g. release from CI?

viewer/package.json
```diff
@@ -16,8 +16,9 @@
     "coverage": "jest --coverage",
     "lint": "tslint --project .",
     "serve": "npm install && npm link && cd examples && npm install && npm link @cognite/reveal && npm run serve",
-    "prepublishscript": "npm run release",
-    "publishscript": "node script/copyPackage.js && cd dist && npm publish"
+    "prepublishscript": "npm run release && npm run coverage",
+    "publishscript": "node script/copyPackage.js",
+    "postpublishscript": "cd dist && npm publish"
```
maksnester, line 21: we could probably have `"files": [ "dist/**/*" ],` specified to avoid using `cd dist` here.

j-bjorne: The reason is that if we do that the package.json publish will publish the dist folder inside the dist folder :) What we do want is to publish everything in the dist folder as a package.

PR #502, opened 2020-05-14
Initial refactor of file structure to support a more module based system

christjt:


christjt:
Note: this is by no means the final structure, but simply the first step towards it.

PR #504, opened 2020-05-15
Use @ alias for imports and merge all viewer entry points to a single entry package

larsmoa:
- Use @ alias for imports to enable imports ala `'@/dataModels/cad'`
- Remove the 'migration' and 'threejs' packages and updates all references.


examples/src/clipping.ts
```diff
@@ -23,10 +23,10 @@ async function main() {
   document.body.appendChild(renderer.domElement);
 
   let modelsNeedUpdate = true;
-  const revealManager = new RevealManager(client, () => {
+  const revealManager = new reveal.RevealManager(client, () => {
```
j-bjorne, line 26: Should we stick with RevealManager, or should we go for ModelManager, or Model3DManager or something along those lines?

viewer/src/__tests__/culling/ByVisibilityGpuSectorCuller.test.ts
```diff
@@ -2,18 +2,21 @@
...
 import * as THREE from 'three';
-import { OrderSectorsByVisibilityCoverage } from '../../dataModels/cad/internal/sector/culling/OrderSectorsByVisibilityCoverage';
-import { ByVisibilityGpuSectorCuller } from '../../dataModels/cad/internal/sector/culling/ByVisibilityGpuSectorCuller';
-import { CadModel } from '../../dataModels/cad/internal';
-import { SectorMetadata } from '../../dataModels/cad/internal/sector/types';
-import { ModelDataRetriever } from '../../utilities/networking/ModelDataRetriever';
-import { SectorSceneImpl } from '../../dataModels/cad/internal/sector/SectorScene';
 import { mat4, vec3 } from 'gl-matrix';
+
+import { ModelDataRetriever } from '@/utilities/networking/ModelDataRetriever';
```
j-bjorne, line 7: Would it be possible to do the following for alias'?
@dataModels would be linked to ['src/dataModels/*]
@utilities ... ['src/utilities/*']
...

or go even further and just have @cad, @pointcloud etc.

Sure, it would make us have to add a new alias for packages, but just @ in the beginning of the import just seems weird. @root maybe, or @src?

viewer/src/dataModels/cad/internal/CadMetadataParserV8.ts
```diff
@@ -39,12 +39,12 @@ export interface CadSectorMetadataV8 {
       xz: number;
     };
     readonly recursiveCoverageFactors:
-    | {
-      xy: number;
-      yz: number;
-      xz: number;
-    }
-    | undefined;
+      | {
```
j-bjorne, line 42: We need to check our lint settings or something, it seems to save differently across machines.

viewer/src/index.ts
```diff
@@ -4,12 +4,17 @@
...
+export { CadNode } from './dataModels/cad/internal/CadNode';
+export { ModelNodeAppearance } from './dataModels/cad/internal/ModelNodeAppearance';
+
 // CAD
 export { CadModel, loadCadModelFromCdf, loadCadModelByUrl } from './dataModels/cad/internal';
 export { SectorMetadata, SectorModelTransformation } from './dataModels/cad/internal/sector/types';
 export { CadRenderHints } from './dataModels/cad/public/CadRenderHints';
 export { CadLoadingHints } from './dataModels/cad/public/CadLoadingHints';
-
+export { intersectCadNodes } from './dataModels/cad/internal/picking';
```
j-bjorne, line 17: Is there any reason why this one is not using alias?

viewer/src/utilities/networking/CogniteClient3dExtensions.ts
```diff
@@ -3,7 +3,6 @@
  */
 
 import { CogniteClient, CogniteInternalId, CogniteExternalId, IdEither } from '@cognite/sdk';
-import { HttpHeaders } from '@cognite/sdk/dist/src/utils/http/basicHttpClient';
```
j-bjorne, line 6: Why was this import removed?

viewer/tsconfig.json
```diff
@@ -21,7 +21,12 @@
     "lib": [
         "dom",
         "esnext"
-    ]
+    ],
+
+    "baseUrl": "./src/",
+    "paths": {
+      "@/*": ["*"]
```
j-bjorne, line 28: Finally the file I was looking for :) An opinion here and maybe I am wrong. @ alone seems so weird. I would prefer if it was say:
  "@datamodels/*": ["datamodels/*"],
  "@utilities/*": ["utilities/*"]
And honestly, the alias feels even more useful if used to shorten imports so:
  "@cad/*": ["datamodels/cad/*"],
  "@pointcloud/*": ["datamodels/pointcloud"]

Thoughts?

examples/src/clipping.ts
larsmoa: Not part of this PR, but we should maybe discuss the name at some time. I don't like "Manager" myself.

viewer/src/__tests__/culling/ByVisibilityGpuSectorCuller.test.ts
larsmoa: I believe using `@/...` is a convention that is quite common.

As for `@cad` etc I think we should introduce sub-packages to achieve something like this, not a ton of aliases.

viewer/tsconfig.json
larsmoa: See answer above

viewer/src/utilities/networking/CogniteClient3dExtensions.ts
larsmoa: Because it's internal to @cognite/sdk, ref `dist + we don't need it.

viewer/src/index.ts
larsmoa: Not really.

PR #505, opened 2020-05-15
Dependency updates

larsmoa:


viewer/package.json
```diff
@@ -42,11 +42,11 @@
...
     "@cognite/tslint-config-cognite-3d": "^1.0.4",
     "@types/gl-matrix": "^2.4.5",
-    "@types/jest": "^25.2.1",
-    "@types/node": "^13.13.5",
-    "@types/tween.js": "^18.5.0",
+    "@types/jest": "^25.2.3",
+    "@types/node": "^14.0.5",
+    "@types/tween.js": "^18.5.1",
     "@wasm-tool/wasm-pack-plugin": "^1.3.1",
-    "copy-webpack-plugin": "^5.1.1",
```
maksnester, line 49: `copy-webpack-plugin` I don't think we use it at all in the viewer. Only in examples. But they have their own.

PR #507, opened 2020-05-15
Extract CadManager from RevealManager

j-bjorne:
Created a CadManager
Some choices here that are open for discussion! Have a nice weekend!

IMPORTANT! Examples don't work with LOCAL URL Assets in this commit, not yet implemented multisource


viewer/src/dataModels/cad/internal/CadTransformationProvider.ts
```diff
@@ -0,0 +1,9 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+import { SectorModelTransformation } from './sector/types';
+
+export interface CadTransformationProvider {
+  getCadTransformation(): SectorModelTransformation;
```
larsmoa, line 8: Shouldn't this take `blobUrl`/some ID as parameter?

viewer/src/dataModels/cad/internal/sector/WantedSector.ts
```diff
@@ -4,12 +4,10 @@
 
 import { SectorMetadata, SectorModelTransformation } from './types';
 import { LevelOfDetail } from './LevelOfDetail';
-import { ModelDataRetriever } from '@/utilities/networking/ModelDataRetriever';
 import { SectorScene } from './SectorScene';
 
 export interface WantedSector {
-  cadModelIdentifier: string;
-  dataRetriever: ModelDataRetriever;
+  blobUrl: string;
```
larsmoa, line 10: I'm a bit unsure about `blobUrl` to identify model. We use this as a modelIdentifier in places where it doesn't make sense to know that `blobUrl` identifies a model. Also it seems that `WantedSector` should identify what to load, but `blobUrl` basically defines how to load the data...

viewer/src/public/RevealManager.ts
```diff
@@ -8,16 +8,36 @@ import { CadSectorParser } from '@/dataModels/cad/internal/sector/CadSectorParse
...
+import { CadModelFactory } from '@/dataModels/cad/internal/CadModelFactory';
+import { CadModelUpdateHandler } from '@/dataModels/cad/internal/CadModelUpdateHandler';
+import { ProximitySectorCuller } from '@/dataModels/cad/internal/sector/culling/ProximitySectorCuller';
 
 // First iteration of a RevealManager. Currently tailored to examples but should be tailored to external usecase.
 // Should move to example-helpers.ts as a function without extending
+// TODO: j-bjorne 15-05-2020: create a function that creates a default CadManager and place it in a util class?
 export class RevealManager extends RevealManagerBase {
-  constructor(client: CogniteClient, onUpdatedCallback: () => void, options?: RevealOptions) {
+  constructor(client: CogniteClient, options?: RevealOptions) {
```
larsmoa, line 24: Nice to get rid of the onUpdatedCallback! Should introduce event as alternative to polling, but not the most pressing issue.

viewer/src/public/RevealManagerBase.ts
```diff
@@ -58,110 +29,49 @@ export type OnDataUpdated = () => void;
...
+    this._cadManager = cadManager;
     this._materialManager = materialManager;
-    this._modelObservable = this._modelSubject.pipe(this.addCadOrPointcloudModel(client));
-    this._cadModelObservable = this._modelObservable.pipe(this.getAddedModels());
-    this._cameraSubject = this.createLoadSectorsPipeline(dataUpdatedCallback);
+  }
+
+  public resetRedraw(): void {
+    this._cadManager.resetRedraw();
+  }
```
larsmoa, line 45: Do we need to have a `resetRedraw`-function or can we do this using the `Mesh`-trick I showed? E.g. can we provide a new field on CadNode called lastRenderTimestamp and compare that with a timestamp of the last event that caused needsRedraw to be set?

viewer/src/utilities/networking/CogniteClient3dExtensions.ts
```diff
@@ -2,105 +2,67 @@
...
-    readonly model: CogniteModel3dIdentifier;
-    readonly outputs: Model3dOutput[];
-  }[];
-}
-
 /**
  * Provides 3D V2 specific extensions for the standard CogniteClient used by Reveal.
  */
-export class CogniteClient3dExtensions {
+export class CogniteClient3dExtensions implements ModelUrlProvider, CadSceneProvider, CadSectorProvider {
```
larsmoa, line 22: I think what you maybe are doing here is to create a specialized version for Reveal which makes sense, but I think `CogniteClient3dExtensions` should be a class that we can migrate to `@cognite/sdk` and hence closely map to the API. It's probably a good idea with something that wraps this to offer the functionality implemented here.

viewer/src/public/RevealManagerBase.ts
```diff
@@ -58,110 +29,49 @@ export type OnDataUpdated = () => void;
...
-      source: {
-        discriminator: 'external',
-        url
-      },
-      modelNodeAppearance,
-      callbacks: { success: resolveCb!, fail: rejectCb! }
-    });
-    return promise;
+  public addModelFromUrl(_modelUrl: string, _modelNodeAppearance?: ModelNodeAppearance): Promise<CadNode> {
+    throw new Error('Deprecated');
```
larsmoa, line 57: Deprecated or not implemented yet?

viewer/src/utilities/networking/ModelOutputList.ts
```diff
@@ -0,0 +1,37 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+import { IdEither } from '@cognite/sdk';
+import { BlobOutputMetadata } from './BlobOutputMetadata';
+import { File3dFormat } from '../File3dFormat';
+
+export class ModelOutputList {
```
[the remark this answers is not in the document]
j-bjorne: renamed back, it was mainly done for consistency since none of our stuff was named Model3D

viewer/src/dataModels/cad/internal/CadModelUpdateHandler.ts
```diff
@@ -0,0 +1,74 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+import * as THREE from 'three';
+import { Subject, Observable, combineLatest, animationFrameScheduler } from 'rxjs';
+import { CadLoadingHints } from '../public/CadLoadingHints';
```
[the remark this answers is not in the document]
j-bjorne: actually, @/dataModels/cad/public...

viewer/src/dataModels/cad/internal/CadTransformationProvider.ts
j-bjorne: yes, it will when it gets it from network :)

viewer/src/dataModels/cad/internal/sector/WantedSector.ts
j-bjorne: So, I have been thinking about this, and honestly, blob url makes so much more sense, since like we have said, the model being showed isn't the model, but a revision, stored in a single blob.

I have looked at this part of our code for a while, and well the most consistent way of dealing with the data loading is with blobUrl, and having a blobUrl Provider (this is also consistent with how PointClouds currently work).

What I would agree with, and this is how I would do it in the future, is to have the modelUrl cached, and then rather save the identifier which reacquires the blob url with the identifier, but I believe that to be a second step if anything.

viewer/src/public/RevealManagerBase.ts
j-bjorne: I did not like the Mesh trick due to it having to be added to scene, and I didn't just want to add it without having a proper solution for it. ie (we have 1 mesh for the entire manager, and we make sure that it follows the camera (due to well, if not can't this be outside the frustum and not rendered at some point?)

viewer/src/dataModels/cad/internal/sector/WantedSector.ts
larsmoa: Are we sure that a model will be identifies by a single blob? Could it be that a model consists of two different outputs? An example I can think of is that a model might consist of 3D geometry + 2D "minimaps". These can probably be split into two different things, but are we sure this is the best way to do it? Remember that they are tightly linked, e.g. have some common coordinate system. I'm not sure blobUrl as an ID prohibits this to work well, I'm just not sure that it _doesn't_.

And, I think the argument about choosing an ID that is as generic as possible to hold...

I won't keep arguing to much about this, but I'm still not convinced that this is the right approach.

viewer/src/public/RevealManagerBase.ts
larsmoa: `onBeforeRender` is called before frustum culling I believe, and if not it's possible to disable frustum culling for an object (https://threejs.org/docs/#api/en/core/Object3D.frustumCulled).

I agree that adding an empty mesh to trigger the callback is code smell, but I think it's worth it as `resetRedraw` is prone to user errors.

larsmoa:
@j-bjorne : notice that there is a lot of merge errors - you should try to resolve these asap before it's too late...

viewer/src/dataModels/cad/internal/CadTransformationProvider.ts
larsmoa: Not sure why we add this interface without the parameter - if we won't add support yet is there any use in adding the interface at all?

PR #514, opened 2020-05-18
Make the new sector culler the default one and remove the old one.

larsmoa:
The old default has a bug causing detailed leaf sectors not to be loaded when there is no F3DF file for the leafs.


viewer/src/public/RevealManagerBase.ts
```diff
@@ -282,16 +282,12 @@ export class RevealManagerBase {
     pipeline
       .pipe(
         auditTime(100),
-        fromThreeCameraConfig(),
+        map(camera => camera.clone()),
```
j-bjorne, line 285: Why have this here instead of inside the determine sector?

larsmoa: To ensure that implementors of `SectorCuller` (which can be customised by the end-user) doesn't forget to clone.

PR #520, opened 2020-05-19
Update Readme with Code Example and API Reference

j-bjorne:
Add a code example and api reference to readme


viewer/README.md
```diff
@@ -142,3 +142,212 @@ Y up, X to the right and Z pointing out of the screen.
 The policy in this repository is to stick with the CDF coordinate system in any code that is not
 viewer-specific.
 For viewer-specific code, the conversion should happen as early as possible.
+
+# Code Example
```
maksnester, line 146: I think we should respect header levels across the readme.

Also, it's probably a good idea to put this code sample right at the top of the readme. I think a good structure might be like this:

`# Reveal`

* Lib description - what is it? What problem it solves.
* Your nice and short example of code (however I think you should add imports to understand from where exactly you got used components)
* Installation - some dummy thing like `npm install @cognite/reveal` just to copy-paste for installation (should be above the notes about workers). Maybe some other notes if something more than npm install is required to make it work.

PR #539, opened 2020-05-21
Add support for overriding sectors to load in the side-by-side example.

larsmoa:
Add support for override loading of sectors in the side-by-side example using the debug widget.

Note that there are bugs in the RxJS loading pipeline that causes errors when providing certain overload options.


larsmoa:
I realize that saying "there are bugs" is about as helpful as saying "the software isn't working", @j-bjorne. Noted.

In this case what I was referring to was the stuff I posted on Slack during the weekend. Not sure what the bug is, but the pipeline seems to eat errors and determining cause is very hard because of bad call stacks.

PR #545, opened 2020-05-22
Extract CadManager review - change suggestions.

larsmoa:
Suggested changes for PR #507


viewer/src/utilities/networking/ModelDataRetriever.ts
```diff
@@ -2,7 +2,39 @@
...
  */
 
+import { HttpError } from '@cognite/sdk';
+
 export interface ModelDataRetriever {
   fetchJson(filename: string): Promise<any>;
   fetchData(filename: string): Promise<ArrayBuffer>;
 }
+
+export class ByUrlModelDataRetriever implements ModelDataRetriever {
```
j-bjorne, line 12: What is the point of this class? I didn't see it used anywhere. And why is this better than the LocalUrlClient & Cognite3dExtension?
This one seems to go back to the same design where we have one retriever per every model that gets added each with its own baseUrl.

viewer/src/utilities/networking/types.ts
```diff
@@ -0,0 +1,27 @@
...
+  format: File3dFormat | string;
+  version: number;
+}
+
+export interface ModelUrlProvider<TModelIdentifier> {
+  getModelUrl(params: TModelIdentifier): Promise<string>;
+}
+
+export interface ByUrlModelIdentifier {
+  discriminator: 'external';
```
j-bjorne, line 20: Why do we have discriminator here? And why does this have to be an entire url. I liked the restriction the local host imposed by locking the url to the origin of the url. It meant people only fed the path and not the entire url.

viewer/src/utilities/networking/types.ts
```diff
@@ -0,0 +1,27 @@
...
+  url: string;
+}
+
+export interface CdfModelIdentifier {
+  discriminator: 'cdf';
```
j-bjorne, line 25: Again, I think these can be without a discriminator

j-bjorne:
Also, does this actually build? I somehow doubt that the addModel expecting the new identifiers work, at least it looked like they weren't completely done there. The models had discriminators and I didn't see these created anywhere.

viewer/src/utilities/networking/ModelDataRetriever.ts
larsmoa: I didn't add it - it was in the code base you checked in. I just moved the class. If we can remove it, go ahead, @j-bjorne.

viewer/src/utilities/networking/types.ts
larsmoa: It's here because it's a type? Before it was in the same folder, just another file.

As for URL in the name it has an URL-property so if the property is used as a relative path that property must be renamed.

I don't agree that locking to localhost has any value though. A use case would be to have some automated test with datasets stored externally to avoid a ton of blobs in git.

viewer/src/utilities/networking/types.ts
larsmoa: Again, just moved stuff. I didn't add the discriminator - you did.

larsmoa:
Last snippet you posted is your code.

viewer/src/utilities/networking/ModelDataRetriever.ts
j-bjorne: Awesome xD it must have been a file that wasn't deleted since it was moved in the pr. I had originally deleted it, but the merge must have kept it

viewer/src/utilities/networking/types.ts
j-bjorne: I think the problem here is that in my commit I deleted some files, and when I merged these files had been moved, and been restored. A lot of the files deleted in this PR were files I had removed. I will go over this PR and fix it.

j-bjorne:
Yes, but my Params were set up to be { fileName: string }, and { modelRevision: IdEither, fileFormat: File3DFormat }, in which the addModel that was in my code worked. You changed the Param that was defined and changed it into CdfModelIdentifier and UrlModelIdentifier, which is what I am questioning is working now.

larsmoa:
The `CdfModelIdentifier` and `LocalModeldentifier` used in the code above is the locally defined variants, not the instances that we found out can be removed.

PR #557, opened 2020-05-26
Dependency updates

larsmoa:


viewer/package.json
```diff
@@ -76,10 +76,10 @@
   },
   "dependencies": {
     "@cognite/potree-core": "^1.1.3",
-    "@cognite/sdk": "^2.28.0",
+    "@cognite/sdk": "^2.31.0",
```
maksnester, line 79: Why we moved sdk and three from peer+dev into common deps?

maksnester: It's fine for threejs probably since we export it. But not sure about SDK. These issues about different versions between examples and e.g. viewer, but more generally between any client app that uses reveal and reveal - they happen because reveal bundles sdk into own deps, at least as I understand it. Reveal should require sdk as a peer, but shouldn't install its own version.

But at the same time, we still will struggle with inconsistent versions between examples and reveal, because we need SDK to be installed to work with it locally and to develop reveal (to have types from SDK at least). But I think if peer+devdep combination is used real client apps shouldn't be affected by that issue. By real I mean those that install reveal from NPM.

maksnester: To be discussed somewhere else

PR #559, opened 2020-05-27
Pointcloud package restructuring

j-bjorne:
Restructured package, removed public and internal


j-bjorne:
Isn't there one?
viewer\src\datamodels\pointcloud\index.ts

PR #584, opened 2020-06-02
examples: port on react almost without changes

maksnester:
I ported examples on create-react-app based setup. I tried to keep examples as they were before with little exceptions related to manipulations with DOM. So basically they're the same examples as before but all code is placed inside `useEffect` with little changes. 
~I also added prettier, hope you don't mind.~


maksnester:
Removing prettier pre-commit hook since it affects other files and that's not I wanted

examples/src/pages/Clipping.tsx
```diff
@@ -0,0 +1,255 @@
...
+CameraControls.install({ THREE });
+
+export function Clipping() {
+  const canvasRef = useRef<HTMLCanvasElement>(null);
+
+  useEffect(() => {
+    const gui = new dat.GUI();
+
+    async function main() {
+      const { project, modelUrl, modelRevision } = getParamsFromURL({
```
j-bjorne, line 26: I know that it might not be part of this PR, since you say you don't change much, but would it be possible if at least one example could be changed to fit a more React style of setup as an example on how it would look that way?

The component should be provided parameters from their parents where possible e.g. client, modelUrl / modelRevision. Thoughts?

Right now it seems like we are using react, but not really taking advantage of it. I could be mistaken but the react router makes chaging url not actually reload pages, and in that sense we should structure the examples more like an actual application?

examples/.env
```diff
@@ -0,0 +1,2 @@
+HTTPS=true
```
j-bjorne, line 1: This file should be added to the gitignore file. We should create an .env.example which people can copy and rename, so that we don't enforce environment variables.

examples/package.json
```diff
@@ -1,74 +1,63 @@
...
-    "coverage": "jest --coverage",
-    "lint": "tslint --project ."
+    "start": "react-app-rewired start",
+    "build": "react-app-rewired build",
+    "test": "react-app-rewired test",
+    "eject": "react-scripts eject"
   },
   "author": "Svenn-Arne Dragly <dragly@cognite.com>",
   "contributors": [
     "Lars Moastuen <lars.moastuen@cognite.com>",
```
j-bjorne, line 14: Just noticed, Christopher and Maksim, you guys might want to add yourself to the list

examples/package.json
```diff
@@ -1,74 +1,63 @@
...
-    "@xailabs/three-renderer-stats": "^1.0.4",
     "camera-controls": "^1.21.0",
-    "comlink": "^4.3.0",
     "dat.gui": "^0.7.7",
-    "gl-matrix": "^3.3.0",
-    "glslify": "^7.0.0",
-    "hold-event": "0.0.1",
-    "postprocessing": "^6.13.5",
-    "rxjs": "^6.5.5",
+    "hold-event": "^0.0.1",
```
j-bjorne, line 23: What has happened for the dependencies to have changed this much?

examples/src/App.tsx
```diff
@@ -0,0 +1,64 @@
...
+            background: "rgb(255, 255, 255)",
+            padding: "15px",
+            borderRadius: "4px",
+            minHeight: "95vh",
+            display: "flex",
+            flexDirection: "column",
+            overflowX: "auto",
+          }}
+        >
+          <Switch>
```
j-bjorne, line 30: Should we define the cognite client outside of the examples since I am guessing we are using react reroute and don't actually reload pages on url change?

examples/src/pages/Simple.tsx
```diff
@@ -0,0 +1,128 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+import * as THREE from "three";
+import CameraControls from "camera-controls";
+import {
```
j-bjorne, line 7: This file is probably the first one that should be "Reactified". I would like seeing more reactive code (splitting fetch model and setup renderloop etc).

Meaning for instant one has a:
```
useEffect( () => {
  // fetch new model
  // then change camera
  // request rerender?
}, [modelRevision, modelUrl])
```

Along these lines?

examples/src/routes.tsx
```diff
@@ -0,0 +1,97 @@
...
+  },
+  {
+    path: "/ssao",
+    menuTitle: "Screen space ambient occlusion shading",
+    component: <SSAO />,
+  },
+  {
+    // not really good defaults, provide something more meaningful
+    path:
+      "/two-models?project=3ddemo&model=1937792320363609&model2=5254077049582015",
```
j-bjorne, line 86: the default project is publicdata and we should not provide defaults for parameters like model

examples/src/utils/example-helpers.ts
```diff
@@ -2,37 +2,50 @@
...
+  type: "local" | "cdf",
   client: CogniteClient,
   options?: RevealOptions
 ): RevealManager | LocalHostRevealManager {
   switch (type) {
-    case 'cdf':
+    case "cdf":
       return new RevealManager(client, options);
-    case 'local':
+    case "local":
```
j-bjorne, line 21: Will there not be a lint error here? Are we all using the same linter?

examples/src/pages/Clipping.tsx
maksnester: > Right now it seems like we are using react, but not really taking advantage of it.

Indeed. I wanted to refactor at least one of the examples to be more 'reactish' but later. 
About URL params, they are still work as they worked before. You enter params, hit enter, page reloads, params parsed. 

> The component should be provided parameters from their parents where possible e.g. client, modelUrl / modelRevision. Thoughts?

Not really necessary IMO but doable. My concerns here are that different examples have different params so you can't just put URL params upper in components hierarchy and pass them to children in the same way for all examples.

examples/.env
maksnester: I don't think we need it all. I just wanted to keep the same port as prev examples have and HTTPS as far as I remember is strictly necessary because of workers. 

So because "people" you mentioned are our team, I think having preconfigured .env is just fine. But if you don't like it I can put this line right into `package.json`

```
"start": "crossenv HTTPS=true react-app-rewired start"
```

examples/package.json
maksnester: I recreated examples from scratch. So some dependencies are not needed anymore. Some was unused (rxjs), those related to build are now part of react-scripts and they are not exposed in our package.json

examples/src/App.tsx
maksnester: Yep, maybe later. We can declare it once here and then pass with context or prop.

examples/src/utils/example-helpers.ts
maksnester: examples have their one eslint installed inside react-scripts, formatting is done by prettier. I believe vscode is also should understand installed prettier and format accordingly. Please let me know if there are some issues with that.

j-bjorne:
Well, I am fine with the comments you have addressed. Only change I will want before approving is the `.env` file rename and adding a line in the todo that the user needs to copy the file and rename it `.env`

.gitignore
```diff
@@ -5,20 +5,29 @@ target
...
 coverage
 
-
 # Viewer ignores
 viewer/dist/
 viewer/public/
 viewer/pkg/
 viewer/node_modules/
 
+# Exapmles
```
j-bjorne, line 20: Nitpick :D See if you can spot it xD

examples/README.md
```diff
@@ -0,0 +1,50 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app)
+and extended with [react-app-rewired](https://github.com/timarney/react-app-rewired) to copy reveal/viewer workers and wasm files to output folder.
+
+## [Environment files](https://create-react-app.dev/docs/adding-custom-environment-variables/)
+
+See `.env.example` file to create your own `.env` file for development.
```
j-bjorne, line 6: Nitpick add:
```
create a copy of .env.example and rename it to .env
This file can be used to hold environment defaults
```
?

.gitignore
maksnester: Thanks.

examples/README.md
maksnester: Done

PR #592, opened 2020-06-03
Expose and fix issue causing RevealManagerBase.update() to trigger update() when camera was unchanged.

larsmoa:


viewer/src/public/RevealManagerBase.ts
```diff
@@ -49,7 +57,18 @@ export class RevealManagerBase<TModelIdentifier> implements RenderManager {
   }
 
   public update(camera: THREE.PerspectiveCamera) {
-    this._cadManager.updateCamera(camera);
+    const hasCameraChanged =
```
j-bjorne, line 60: This could have been handled in the update handler by doing a distinctUntilChanged on the cameraSubject and it could have been a bit prettier :)

I'll approve it for now, but I would recommend we do it that way.

PR #601, opened 2020-06-04
Enable antialiasing in Cognite3dViewer for PCs

larsmoa:
Enables antialiasing by default for Cognite3dViewer for PCs, but not mobile/tablet.


viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -709,6 +713,10 @@ export class Cognite3DViewer {
   };
 }
 
+function shouldEnableAntialiasing(): boolean {
```
maksnester, line 716: Do we need to have a function for the sake of just returning a single boolean?

Also, I don't really understand why not to have that isMobileOrTablet just as a const instead of function?

larsmoa: First the function was more complex (and I expect it to be in the future) - e.g. by checking how powerful the GPU was. I considered removing it, but I decided to keep it. Willing to ditch it if you don't like it

larsmoa: I guess `isMobileOrTablet` can be computed once and stored as a const, yes :+1:

maksnester: ok thanks for the clarification on that

PR #606, opened 2020-06-05
Cognite3DViewer: implement basic addObject3D/removeObject3D + minor examples fix

maksnester:
* add methods from old 3d viewer to add/remove Object3D on scene
* Fixed hardcoded local reveal manager in distance measurement example.

~Part of code with `additionalObjects` isn't used for now, so I kept it but commented out.~

### DIFF
--- a/viewer/src/public/migration/Cognite3DViewer.ts
+++ b/viewer/src/public/migration/Cognite3DViewer.ts
@@ -277,11 +277,20 @@ export class Cognite3DViewer {
     return SupportedModelTypes.NotSupported;
   }
 
-  addObject3D(_object: THREE.Object3D): void {
-    throw new NotSupportedInMigrationWrapperError();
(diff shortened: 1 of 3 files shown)


viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -84,6 +84,7 @@ export class Cognite3DViewer {
   private readonly materialManager: MaterialManager;
   private _slicingNeedsUpdate: boolean = false;
   private _geometryFilters: GeometryFilter[] = [];
+  // private additionalObjects: THREE.Object3D[] = [];
```
larsmoa, line 87: Get rid of commented code.

maksnester: @larsmoa done

PR #614, opened 2020-06-08
Initial functionality for selecting nodes (coloring)

christjt:


examples/src/pages/Migration.tsx
```diff
@@ -204,9 +204,10 @@ export function Migration() {
       if (modelIdStr && revisionIdStr) {
         const modelId = Number.parseInt(modelIdStr, 10);
         const revisionId = Number.parseInt(revisionIdStr, 10);
-        addModel({ modelId, revisionId });
+        await addModel({ modelId, revisionId });
       }
-      viewer.on('click', function (event) {
+
+      viewer.on('click', async function (event) {
```
larsmoa, line 210: Why has this become async? Can't find an await...

larsmoa: @christjt

christjt: Ah, sorry yeah, maybe it should not be. But ```model.setNodeColor(nodeId, 0, 255, 0); (218)``` returns a promise which i awaited intially. That is why I had it as async. But I guess this is not needed.

PR #618, opened 2020-06-08
feat(Cognite3DViewer): add loading spinner; add `getLoadingStateObserver` for RevealManager

maksnester:
Looks like this:

image

And for the RevealManager I just exposed a public method to have a loading state as in viewer.

### DIFF
--- a/examples/src/pages/Migration.tsx
+++ b/examples/src/pages/Migration.tsx
@@ -19,6 +19,7 @@ export function Migration() {
   const canvasWrapperRef = useRef<HTMLDivElement>(null);
   useEffect(() => {
     const gui = new dat.GUI();
+    let viewer: Cognite3DViewer
 
     async function main() {
       const urlParams = new URL(window.location.href).searchParams;
@@ -60,7 +61,7 @@ export function Migration() {
       await client.authenticate();
 
       // Prepare viewer
-      const viewer = new Cognite3DViewer({
+      viewer = new Cognite3DViewer({
         sdk: client,
         domElement: canvasWrapperRef.current!,
       });
@@ -226,6 +227,7 @@ export function Migration() {
 
     return () => {
       gui.destroy();
+      viewer.dispose();
     };
   });
   return <CanvasWrapper ref={canvasWrapperRef} />;
--- a/viewer/src/public/RevealManager.ts
+++ b/viewer/src/public/RevealManager.ts
@@ -22,9 +22,12 @@ import { PointCloudMetadataRepository } from '@/datamodels/pointcloud/PointCloud
 import { PointCloudFactory } from '@/datamodels/pointcloud/PointCloudFactory';
 import { PointCloudManager } from '@/datamodels/pointcloud/PointCloudManager';
 import { DefaultPointCloudTransformation } from '@/datamodels/pointcloud/DefaultPointCloudTransformation';
+import { Observable } from 'rxjs';
 
 type CdfModelIdentifier = { modelRevision: IdEither; format: File3dFormat };
 export class RevealManager extends RevealManagerBase<CdfModelIdentifier> {
+  private readonly loadingStateObserver: Observable<boolean>;
+
   constructor(client: CogniteClient, options?: RevealOptions) {
     const modelDataParser: CadSectorParser = new CadSectorParser();
     const materialManager: MaterialManager = new MaterialManager();
@@ -57,6 +60,8 @@ export class RevealManager extends RevealManagerBase<CdfModelIdentifier> {
     );
 
     super(cadManager, materialManager, pointCloudManager);
+
+    this.loadingStateObserver = sectorRepository.getLoadingStateObserver();
   }
 
   public addModel(
@@ -87,6 +92,10 @@ export class RevealManager extends RevealManagerBase<CdfModelIdentifier> {
       default:
         throw new Error(`case: ${type} not handled`);
     }
+  }
+
+  public getLoadingStateObserver(): Observable<boolean> {
+    return this.loadingStateObserver;
   }
 
   private createModelIdentifier(id: string | number): IdEither {
--- a/viewer/src/public/migration/Cognite3DViewer.ts
+++ b/viewer/src/public/migration/Cognite3DViewer.ts
@@ -39,6 +39,7 @@ import { PointCloudMetadataRepository } from '@/datamodels/pointcloud/PointCloud
 import { PointCloudFactory } from '@/datamodels/pointcloud/PointCloudFactory';
 import { DefaultPointCloudTransformation } from '@/datamodels/pointcloud/DefaultPointCloudTransformation';
 import { BoundingBoxClipper } from '@/utilities';
+import { Spinner } from '@/utilities/Spinner';
 
 export interface RelativeMouseEvent {
   offsetX: number;
@@ -85,6 +86,8 @@ export class Cognite3DViewer {
   private _slicingNeedsUpdate: boolean = false;
   private _geometryFilters: GeometryFilter[] = [];
 
+  private readonly spinner: Spinner;
+
   constructor(options: Cognite3DViewerOptions) {
     if (options.enableCache) {
       throw new NotSupportedInMigrationWrapperError('Cache is not supported');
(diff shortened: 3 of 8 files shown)


viewer/src/utilities/Spinner.ts
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2020 Cognite AS
+ */
+
+// @ts-ignore
+import css from 'raw-loader!./spinnerStyles.css';
```
larsmoa, line 6: Can this `ts-ignore` be avoided by adding a loader for CSS? How does this line affect bundling - is it embedded, and not a separate file, like e.g. wasm and workers? See https://webpack.js.org/loaders/css-loader/

viewer/src/utilities/spinnerStyles.css
```diff
@@ -0,0 +1,37 @@
+.reveal-viewer-spinner,
+.reveal-viewer-spinner:after {
+    border-radius: 50%;
+    width: 32px;
+    height: 32px;
+}
```
larsmoa, line 6: Is this 'mobile-friendly'?

viewer/src/utilities/Spinner.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (document.getElementById(Spinner.stylesId)) {
+      return;
+    }
+    const style = document.createElement('style');
+    style.id = Spinner.stylesId;
+    style.appendChild(document.createTextNode(css));
+    document.head.appendChild(style);
+  }
+
+  private el: HTMLElement;
```
larsmoa, line 22: add const

viewer/src/public/RevealManager.ts
```diff
@@ -89,6 +94,10 @@ export class RevealManager extends RevealManagerBase<CdfModelIdentifier> {
     }
   }
 
+  public getLoadingStateObserver(): Observable<boolean> {
+    return this.loadingStateObserver;
+  }
```
larsmoa, line 99: I don't think we should expose an`Observable` here as it makes RxJS 'leak out' of Reveal. We should either just have a simple `isLoading`-field that must be polled, or maybe events `loadingStarted`+ `loadingFinished` or `loadingStateChanged`.

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -148,6 +152,14 @@ export class Cognite3DViewer {
     this.revealManager = new RevealManagerBase(this.cadManager, this.materialManager, this.pointCloudManager);
     this.startPointerEventListeners();
 
+    this.sectorRepository.getLoadingStateObserver().subscribe(isLoading => {
+      if (isLoading) {
+        this.spinner.show();
+      } else {
+        this.spinner.hide();
+      }
+    });
```
larsmoa, line 161: Not a biggie, but I think we should avoid RxJS in the 'migration'-layer.

viewer/src/utilities/spinnerStyles.css
maksnester: Yep. It's still a small little spinner at the top corner.

viewer/src/utilities/Spinner.ts
maksnester: > The css-loader interprets @import and url() like import/require() and will resolve them.

About bundling, with raw loader, it should be just text. Like if everything I have in the file would be written in js string and exported.

image

But yes, I think I should just add a loader in the webpack config to avoid ts-ignore here. Too easy to break by renaming or something.

viewer/src/public/RevealManager.ts
maksnester: Yes, I was also thinking about that... Would something like that be fine?

```js
reveal.on('loadingStateChanged', (isLoading) => /* ... */ )
```

@larsmoa @j-bjorne

viewer/src/public/migration/Cognite3DViewer.ts
maksnester: Well, it's used already anyway. It's not the first usage. 

```js
    this.sectorRepository
      .getParsedData()
      .pipe(
        share(),
        filter(x => x.blobUrl === cadNode.cadModelMetadata.blobUrl)
      )
      .subscribe(parseSector => model3d.updateNodeIdMaps(parseSector));
```

TBH, I think it's not easy to avoid it because the `CachedRepository` is built on top of it. As I said before: rxjs code always tends to spread. It's just so convenient to use it for that loader case :D 

To avoid it here I will probably need to do exactly the same thing I suggested for the reveal API. Some `.on/.off` to watch for loading state changes. But it feels like way more code.

viewer/src/public/RevealManager.ts
j-bjorne: I completely disagree. Why would we have something that the user must create their custom polling on. I would much rather they add an event callback on something that gets triggered, but anything else just sounds like a step back. Also rx is not exposed in the migration. Also this doesn't expose rx as we would add the same kind of callbacks to reveal manager

viewer/src/public/migration/Cognite3DViewer.ts
j-bjorne: Yeah I agree with Maksim here. adding a single observer which you observe is not a very complex and it doesn't leak outside the migration layer.

j-bjorne: This can also be wrapped in the RevealManager as a event function like suggested earlier.

```revealManager.on('loadingStateChanged', isLoading => {
  // Do own loading logic
});```

Comment (maksnester) [reply viewer/src/public/migration/Cognite3DViewer.ts:161]: @larsmoa @j-bjorne so just to summarize both threads. Are we agreed on:

1. I should change the RevealManager public API like that

```diff
const loadingHandler = (isLoading) => /* custom loading handling */
+ revealManager.on('loadingStateChanged', loadingHandler);
- revealManager.getLoadingStateObserver().subscribe(loadingHandler );
```

2. Cognite3DViewer is ok now. Since rxjs is internal and is not exposed to public API.

?

j-bjorne: I agree with both of those

viewer/src/public/RevealManager.ts
larsmoa: `RevealManager` and friends must be considered public, and they therefore should not expose RxJS. 

I agree that polling probably is a bad idea.

viewer/src/public/migration/Cognite3DViewer.ts
larsmoa: Agreed, @maksnester

viewer/src/public/RevealManager.ts
j-bjorne: Yes, they aren't at this point. Rxjs is not being exposed to the public with this solution :)

examples/src/pages/Migration.tsx
```diff
@@ -19,6 +19,7 @@ export function Migration() {
   const canvasWrapperRef = useRef<HTMLDivElement>(null);
   useEffect(() => {
     const gui = new dat.GUI();
+    let viewer: Cognite3DViewer;
```
j-bjorne, line 22: Probably part of another pr, but thought I would ask since you are making the viewer a let variable.
Should we not rather do something like
```
cons [setClient, client] = useState();

useEffect(() => {
  const viewer = new Cognite3DViewer();
  return () => {
    viewer.dispose();
  }
}, [client])
```

Or something along those lines?

viewer/src/utilities/Spinner.ts
```diff
@@ -0,0 +1,52 @@
+/*
+ * Copyright 2020 Cognite AS
+ */
+
+import css from './spinnerStyles.css';
```
j-bjorne, line 5: We should make sure that the user can override the style if they want to :)

examples/src/pages/Migration.tsx
maksnester: @j-bjorne We probably should. At least for such things as the viewer (where it makes sense). 

Here it makes sense to have `let` because we should clean up things when component is unmounted. But I don't see the need to declare it as a state, because nothing else needs it. 

We also need to clean up after the reveal manager in a similar way. That's why I asked about `dispose()` for the reveal manager yesterday. 

About putting `client` as a state thing, it doesn't really make sense to me since it's declared and used inside the same effect and doesn't change. But it would be exactly as you wrote if to lift client declaration higher (as a context or prop, for example). 

I promised to make at least some examples to be more in "react way" and hopefully will deal with it once I have some spare time in that sprint.

viewer/src/utilities/Spinner.ts
maksnester: I think there is nothing impossible right now. Just specifying a more specific CSS selector should do the trick.

PR #642, opened 2020-06-12
feat(CognitePointCloudModel): add pointColorType get/set, requestRedraw in setters

maksnester:
Need that to make it possible to change colorType for pointcloud models with Cognite3DViewer in use (in the console).

### DIFF
--- a/viewer/src/datamodels/pointcloud/PotreeGroupWrapper.ts
+++ b/viewer/src/datamodels/pointcloud/PotreeGroupWrapper.ts
@@ -8,13 +8,18 @@ import * as THREE from 'three';
 
 import { PotreeNodeWrapper } from './PotreeNodeWrapper';
 
+export interface Redrawable {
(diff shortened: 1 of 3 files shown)


viewer/src/datamodels/pointcloud/PotreeGroupWrapper.ts
```diff
@@ -46,7 +52,12 @@ export class PotreeGroupWrapper extends THREE.Object3D {
     }
   }
 
+  requestRedraw() {
+    this.redrawRequested = true;
+  }
```
larsmoa, line 57: Can we omit `PotreeGroupWrapper` implementing `Redrawable` and rather handle this internally? I'm thinking we can e.g. have a `needsRedraw` in `PotreeNodeWrapper` which is set to true when its properties change and reset that field when `PotreeGroupWrapper.resetNeedsRedraw()` is invoked.

PR #647, opened 2020-06-15
Test for high detailed queue (from proximity) being cut

christjt:


viewer/src/datamodels/cad/CadManager.ts
```diff
@@ -30,6 +30,9 @@ export class CadManager<TModelIdentifier> {
     this._cadModelUpdateHandler.observable().subscribe(
       sector => {
         const cadModel = this._cadModelMap.get(sector.blobUrl);
+        if (cadModel!.renderHints.showSectorBoundingBoxes) {
```
larsmoa, line 33: I think we should add a `if (cadModel) throw` test instead of using (`!`) which will cause 'undefined reference' if it fails runtime.

PR #650, opened 2020-06-15
Recover pipeline on network errors

j-bjorne:
I threw up this PR, no pun intended.


viewer/src/datamodels/cad/sector/CachedRepository.ts
```diff
@@ -54,200 +68,183 @@ export class CachedRepository implements Repository {
...
+  getParsedData(): Observable<{ blobUrl: string; lod: string; data: SectorGeometry | SectorQuads }> {
+    return this._parsedDataSubject.pipe(distinct(keySelector => '' + keySelector.blobUrl + '.' + keySelector.sectorId)); // TODO: Should we do replay subject here instead of variable type?
+  }
   public dispose(): void {
     if (this.isDisposed) {
       return;
     }
     this.isDisposed = true;
     this._isLoadingSubject.complete();
-    this.clearSemaphore();
```
maksnester, line 79: @j-bjorne Before that thing stopped queued requests from being run after disposal. How is it handled now?

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
```diff
@@ -37,18 +38,19 @@ export class CadModelUpdateHandler {
...
       this._modelSubject.pipe(
+        share(),
         scan((array, next) => {
           array.push(next);
           return array;
         }, modelsArray)
       )
     ).pipe(
-      auditTime(700),
+      auditTime(250),
```
larsmoa, line 53: Why?

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
```diff
@@ -37,18 +38,19 @@ export class CadModelUpdateHandler {
   constructor(sectorRepository: CachedRepository, sectorCuller: SectorCuller) {
     const modelsArray: CadNode[] = [];
     this._updateObservable = combineLatest(
-      this._cameraSubject.pipe(),
+      this._cameraSubject.pipe(auditTime(1000)),
```
larsmoa, line 41: Why?

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
```diff
@@ -71,14 +73,43 @@ export class CadModelUpdateHandler {
...
+          flatMap(wantedSectorArray => {
+            return from(wantedSectorArray).pipe(
+              filter(wantedSector => {
+                try {
+                  const sectorStates = modelSectorStates[wantedSector.blobUrl];
+                  const sectorState = sectorStates[wantedSector.metadata.id];
+                  return !(sectorState && sectorState === wantedSector.levelOfDetail);
+                } catch (error) {
+                  return true;
+                }
```
larsmoa, line 88: Why do we need to hold state? And when will this fail (i.e. error is catched and true is returned)?

viewer/src/datamodels/cad/sector/sectorUtilities.ts
```diff
@@ -133,8 +133,13 @@ export function filterCurrentWantedSectors(
...
       for (const wantedSector of wanted) {
-        if (loaded.metadata.id === wantedSector.metadata.id && loaded.levelOfDetail === wantedSector.levelOfDetail) {
-          return of(loaded);
+        try {
+          if (loaded.metadata.id === wantedSector.metadata.id && loaded.levelOfDetail === wantedSector.levelOfDetail) {
+            return of(loaded);
+          }
+        } catch (error) {
+          // tslint:disable-next-line: no-console
+          console.log(error);
```
larsmoa, line 142: `consol.error`. Any why are we eating the error?

viewer/src/utilities/cache/RequestCache.ts
```diff
@@ -6,5 +6,7 @@ export interface RequestCache<Key, Data> {
   has(key: Key): boolean;
   get(key: Key): Data;
   add(key: Key, data: Data): void;
+  remove(key: Key): void;
+  isFull(): boolean;
```
larsmoa, line 10: There are not in use anywhere (?)

larsmoa: nm, diff for CachedRepository was not loaded

viewer/src/datamodels/cad/sector/CachedRepository.ts
```diff
@@ -54,200 +68,183 @@ export class CachedRepository implements Repository {
...
+        }),
+        map(group => ({ ...wantedSector, group })),
+        shareReplay(1),
+        take(1)
+      )
+    );
+    if (this._modelDataCache.isFull()) {
+      this._modelDataCache.cleanCache(10);
+    }
+    this._modelDataCache.add(this.cacheKey(wantedSector), networkObservable);
```
larsmoa, line 198: I think `add` should free space itself and don't bother the caller to do it.

viewer/src/datamodels/cad/sector/CachedRepository.ts
j-bjorne: Added some more files to handle unsubscribing. I am using the rx Subscription instead, which is made for clean up.

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
j-bjorne: Because you don't want say almost a .7s long delay on things like loading hints / clipping planes etc?

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
j-bjorne: Camera is the heavy hitter on updates, so we can change that to 700ms as it was. I just thought the others would be nice to have more responsiveness to them.

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
j-bjorne: true means that it will be requested. In this scenario if say the model doesn't exist and is undefined it will ask for it. Further down it does the creation of the "model" in the state if it is undefined.

viewer/src/datamodels/cad/sector/sectorUtilities.ts
j-bjorne: because there shouldn't be an error there, and if there is we don't want the entire pipeline to go down.. This was only due to some errors not being handled and they caused the id undefined error.

viewer/src/datamodels/cad/sector/CachedRepository.ts
j-bjorne: I guess we could do that, but should it be part of this PR?

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
larsmoa: I bumped to 0.7s because it's causing lags (probably not on your workstation...). I think we need to have it high until we fix the underlying issue.

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
larsmoa: Ok, 700 ms for camera sounds "good"

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
larsmoa: But if this doesn't affect the camera it might be ok like this as the user doesn't notice the lag like they will with the camera.

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
larsmoa: I think we should avoid catching error like this, and rather check for undefined and return true in that case.

viewer/src/datamodels/cad/sector/CachedRepository.ts
larsmoa: As it's being changed now, I think so yes.

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
j-bjorne: Changed

viewer/src/datamodels/cad/CadModelUpdateHandler.ts
j-bjorne: Sorted

viewer/src/datamodels/cad/sector/CachedRepository.ts
j-bjorne: I made a compromise

larsmoa: What compromise was that? To me it looks identical still?

viewer/src/datamodels/cad/sector/CachedRepository.ts
```diff
@@ -437,11 +411,36 @@ export class CachedRepository implements Repository {
     return meshesGroupedByFile;
   }
 
-  private cacheKey(wantedSector: WantedSector) {
+  private wantedSectorCacheKey(wantedSector: WantedSector) {
     return '' + wantedSector.blobUrl + '.' + wantedSector.metadata.id + '.' + wantedSector.levelOfDetail;
   }
 
-  private ctmKey(request: { blobUrl: string; fileName: string }) {
+  private updateConsumedSectorCache(wantedSector: WantedSector, networkObservable?: Observable<ConsumedSector>) {
```
j-bjorne, line 418: Added two functions instead of internalizing cleanup in memorycache, for both ctm and consumedsector caches.

viewer/src/datamodels/cad/sector/CachedRepository.ts
j-bjorne: I added two functions inside cached repository called `updateConsumedSectorCache` and `updateCtmFilesCache`. I prefer that to internalizing clean up, and even our test is based on the cache throw failing logic. If we are changing it to internally clean up then I want more than your and mine opinion on it, because I think that it sets a precedence on our coding guidelines

PR #651, opened 2020-06-16
Implement functionality missing in the new viewer

larsmoa:
- Add `Cognite3dModel.mapFromCdfToModelCoordinates()` and `mapPositionFromModelToCdfCoordinates()` for mapping between "ThreeJS coordinates" and "CDF coordinates"
- Add `Cognite3dModel. selectNodeByTreeIndex` and `deselectNodeByTreeIndex`
- Fixed signature of `PointerEventDelegate`
- Renamed event `cameraChanged` to `cameraChange` for consistency with old viewer
- Add `Cognite3dViewer. setBackgroundColor`
- Export THREE through @cognitedata/reveal


viewer/src/index.ts
```diff
@@ -10,3 +10,7 @@ export { Intersection } from './public/migration/intersection';
 export { Color, Cognite3DViewerOptions, AddModelOptions, SupportedModelTypes } from './public/migration/types';
 export { CadLoadingHints } from './datamodels/cad/CadLoadingHints';
 export { BoundingBoxClipper } from './utilities';
+
+// Export ThreeJS to enable easy import for our users
```
maksnester, line 14: But why to do so if it's a peer-dependency?

larsmoa: Not sure if it's the right thing to do, but it's what the old viewer did... Seems odd to me as well

PR #654, opened 2020-06-16
Code Guidelines

j-bjorne:
Initial commit of coding guidelines


coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+}
+
+const isOpen = (hour24: number) => {
+  return hour24 > 6 && hour24 < 20;
+}
+
+const butcherShop = {
+  name: 'Meaty',
+  isOpen // AVOID!
+}
```
larsmoa, line 83: Why is this bad?

coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+    this._owner = owner;
+  }
+
+  get name(): string {
+    return this._name;
+  }
+
+  get owner(): Owner {
+    return this._owner;
+  }
```
larsmoa, line 151: I think this is perfectly valid. This is how e.g. C# works too and should be familiar. If we want to be explicit about this sort of stuff, we should return `[Readonly<Owner>](https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlyt)` instead.

coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+
+owner.name = 'Billy';
+
+console.log(candyShop.name); // CandyPlusPlus
+console.log(candyShop.owner.name); // Billy
+```
+> In this example it might make more sense if the owner name was `readonly` and the owner had both a setter and a getter.
+
+### [Private Variables](#private-variables)
+Prefix private variables with `_`, e.g. `_name`, `_age`, `_height`.
```
larsmoa, line 169: 👍

coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+
+### [Naming Guidelines](#naming-guidelines)
```
larsmoa, line 171: I don't think we can avoid bad code using a coding style tbh...

coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+...
+```
+Now in addition to describing what happens inside the function it also matches with the return type of the function.
+
+# Arrays
+
+### [Array Instantiation](#array-instantiation)
+For array instantiation you should use the literal syntax instead of `new` constructor.
+```typescript
+// const items: number[] = new Array(); Don't do this.
```
larsmoa, line 246: Makes perfect sense to use `new Array(...)`, e.g. when you want to specify size up front

coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+### [Expression Vs Declaration](#expression-vs-declaration)
+When defining a function we prefer function declaration over function expression, and when dealing with assigned functions we prefer using using an anonymous [arrow function](#arrow-functions) expression.
+```typescript
+function helloWorld(): void {} // Yes
+const helloWorld = function(): Void => {}; // No
+const helloWorld = (): void => {}; // No
+```
+> **Note:** Performance based on numbers from: https://jsperf.com/function-vs-arrow-function/14.
+
+> **Warning:** Declaring a function in a non-function block (if, while etc) has had a history of having different behavior across browsers. It is recommended to assign the function to a variable instead.
```
larsmoa, line 269: Doesn't webpack take care of this?

coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+
+### [Default Parameters](#default-parameters)
+Instead of having mutating function arguments, consider using default parameters.
+```typescript
+// Worst implementation
+function handleThings(opts): void {
+  // No! We shouldn't mutate function arguments.
+  // Double bad: if opts is falsy it'll be set to an object which may
+  // be what you want but it can introduce subtle bugs.
+  opts = opts || {};
```
larsmoa, line 282: This is a pattern we use quite extensively... What do you propose we do with the usage we have? I think it's a quite nice pattern to use with options-objects, but I agree that falsy values is a bit risky. Maybe we should be explicit about matching undefined?

coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+      this._targetTemperature = this._maximumTemperatureLimit;
+    }
+    else if(this._targetTemperature < this._minimumTemperaturLimit) {
+      this._targetTemperature = this._minimumTemperatureLimit;
+    }
+    return this._targetTemperature;
+  }
+
+  // Good function
+  public getTargetTemperatureAfterUpdatingTemperatureBy(degrees: number): number {
```
larsmoa, line 363: I disagree. What does 'update temperature by' mean? Increase? I think `increaseTemperatur(deltaDegrees: number): number` would be good...

However, I'm not sure if I agree that we should be this granular in our code style definitions.

coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+}
+
+// Getting there!
+function authenticate(credentials: Credentials): boolean {
+  const { username, password } = credentials;
+  return this._apiService(username, password);
+}
+
+// And the reward goes to!
+function authenticate({ username, password }: Credentials): boolean {
```
larsmoa, line 406: Don't agree with this... The function declarations get awfully messy if they must be destructed in the argument list. I'd rather argument that if only a couple of fields of `Credentials` is used it's a sign of code smell and `username` and `password` should be passed as separate arguments.

coding_guidelines.md
```diff
@@ -0,0 +1,500 @@
...
+# Functions
+
+### [Expression Vs Declaration](#expression-vs-declaration)
+When defining a function we prefer function declaration over function expression, and when dealing with assigned functions we prefer using using an anonymous [arrow function](#arrow-functions) expression.
+```typescript
+function helloWorld(): void {} // Yes
+const helloWorld = function(): Void => {}; // No
+const helloWorld = (): void => {}; // No
+```
+> **Note:** Performance based on numbers from: https://jsperf.com/function-vs-arrow-function/14.
```
larsmoa, line 267: I don't think that test is conclusive - my results vary with the choice of browser

coding_guidelines.md
j-bjorne: This goes against the rule we set up that we want classes. Literal syntax is what we would use if we went for a more functional programming approach.

coding_guidelines.md
j-bjorne: No, actually that is bad. Saying new Array(5); actually doesn't create an array with the size 5, it just creates an array with length 5. This can actually trick you.

First of all you can no longer use length to see if an array is empty.
I would completely agree if there was some merit into doing this, but the way Arrays are implemented in js seems to not strengthen this.

https://stackoverflow.com/questions/18947892/creating-range-in-javascript-strange-syntax

Basically, if you do it this way, you can't find out if an array is empty, since lenght != 0 this way.

coding_guidelines.md
j-bjorne: Well, as proposed further down is to rather add the default to the function variable, and yes, that means if we agree on this we should change our old code to match the guideline.

coding_guidelines.md
j-bjorne: I am saying, either a function is:
A) Super Granular
B) Documented with /** **/

coding_guidelines.md
j-bjorne: At some point you need to destructure, whether you do it here or outside the function the fact of the matter is you either need:
```
const { username, password } = credentials
authenticate(username, password);

function authenticate(username: string, password: string) {}
```
or just
```
authenticate(credentials);

function authenticate({ username, password } : Credentials) {}
```

coding_guidelines.md
j-bjorne: Sure, I just mean whatever codestyle we go for I would prefer we have 1, and not 2.

coding_guidelines.md
larsmoa: or
```
authenticate(credentials);
function authenticate(credentials: Credentials) {
const { username, password } = credentials;
}
```
or just use `credentials.username` or `credentials.password` within the function. To me this isn't something we need to set in stone, but I dont think we should use destruction in function declarations.

j-bjorne: Destructuring and not destruction :D

Your example is covered as a do not, but we can change it to do. The first one:

```
function authenticate(credentials: Credentials): boolean {
  const username: string = credentials.username;
  const password: string = credentials.password;
  return this._apiService(username, password);
}
```

We can try and avoid?

j-bjorne: Or are we saying, don't use destructuring at all?

christjt:
What are we going to do with this PR?

maksnester:
> What are we going to do with this PR?

I think that it's great that we discussed all that stuff but I don't really see any practical use from it, to be honest. No one will look into that guide while coding. I think any code issues should be resolved either with linter rules or during code reviews. Having different opinions is fine, if both opinions very strong we can discuss altogether. That's my opinion on that. I don't mind if we merge that file, I just don't expect everybody will follow it (we don't even have consensus on its content – it's big enough).

PR #656, opened 2020-06-17
Highlighted objects are now rendered in front of other objects in the scene

christjt:


viewer/src/datamodels/cad/rendering/postRenderEffects.ts
```diff
@@ -0,0 +1,22 @@
...
+  renderer: THREE.WebGLRenderer,
+  camera: THREE.PerspectiveCamera,
+  scene: THREE.Scene
+) {
+  const currentRenderModes = cadNode.map(x => x.renderMode);
+  cadNode.forEach(x => (x.renderMode = RenderMode.Effects));
+  renderer.clearDepth();
+  renderer.autoClearColor = false;
+  renderer.render(scene, camera);
+  renderer.autoClearColor = true;
```
larsmoa, line 20: Maybe we should restore the previous value rather than setting to true?

viewer/src/datamodels/cad/ModelRenderAppearance.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type NodeBorderMetadata = {
+  color: [number, number, number, number];
+  width: number;
+};
+
+export interface ModelRenderAppearance {
+  renderInFront?: NodeFrontRenderDelegate;
+  renderBorder?: NodeBorderDelegate;
+}
```
larsmoa, line 16: I think this should be merged with `ModelNodeApperance`. Also (but not related to this PR), `GlobalNodeApperance` needs to die.

viewer/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -15,7 +16,7 @@ vec3 packNormalToRgb( const in vec3 normal ) {
 }
 
 void updateFragmentColor(int renderMode, vec3 color, float treeIndex, vec3 normal, float depth, sampler2D matCapTexture) {
-    if (renderMode == RenderTypeColor) {
+    if (renderMode == RenderTypeColor || renderMode == RenderTypeEffects) {
```
larsmoa, line 19: What am I missing here - why do we need to have this as a separate render mode? Seems like it's the same as `RenderTypeColor`?

christjt: Usage is in update visibility shader

PR #657, opened 2020-06-17
Remove intermediary interface ParseQuadsResult

aslettemark:
This also has the effect of eliminating the finalizeSimple step, skipping straight to SectorQuads for rendering.


aslettemark:
I decided to eliminate ParseQuadsResult, not sure if anyone has any thoughts on this. It arguably served the non-functional purpose of being part of the interface to the parsing part of the program, which might have some value. Since we already have a different interface to Rust, it seemed redundant, and we're rethinking the boundary between Rust and TS for I3D+OCTM anyways.

PR #660, opened 2020-06-18
Replace ModelNodeAppearance and ModelRenderAppearance with NodeAppearanceProvider

larsmoa:


viewer/src/datamodels/cad/MaterialManager.ts
```diff
@@ -141,61 +68,55 @@ export class MaterialManager {
...
+        materials.overrideColorPerTreeIndex.needsUpdate = true;
+      }
+
+      // Hide node?
+      if (style && style.visible !== undefined) {
+        materials.overrideVisibilityPerTreeIndex.image.data[4 * treeIndex] = style.visible ? 255 : 0;
+        materials.overrideVisibilityPerTreeIndex.needsUpdate = true;
+      }
+
+      // Render in front of everything (i.e. skip depth testing)?
```
christjt, line 111: We don't skip depth testing, we clear the depth buffer between passes; but depth test in the highlighting pass (else the ordering between highlighted objects would be wack)

larsmoa: good point

viewer/src/datamodels/cad/rendering/materials.ts
```diff
@@ -41,7 +41,9 @@ export function createMaterials(
 
   const colors = new Uint8Array(4 * textureElementCount);
   const visibility = new Uint8Array(4 * textureElementCount);
-  visibility.fill(255);
+  for (let i = 0; i < textureElementCount; i++) {
+    visibility[4 * i] = 255;
+  }
```
christjt, line 46: Concider:

```js
visibility.map((p, n) => (n % 4 == 0) ? 255 : 0);
```

PR #670, opened 2020-06-18
fix(Spinner): specify top-left position explicitly

maksnester:
If parent doesn't have `display: flex` it doesn't work the same way with defaults.


maksnester:
Will merge it with the spinner fix for pointcloud

PR #685, opened 2020-06-22
Automatically adjust camera near/far planes based on models loaded and objects added to the scene.

larsmoa:


viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -51,6 +51,23 @@ type RequestParams = { modelRevision: IdEither; format: File3dFormat };
 type PointerEventDelegate = (event: { offsetX: number; offsetY: number }) => void;
 type CameraChangeDelegate = (position: THREE.Vector3, target: THREE.Vector3) => void;
 
+// Reusable buffers used by functions in Cognite3dViewer to avoid allocations
+const UpdateNearAndFarPlaneBuffers = {
```
maksnester, line 55: It's rare when we have multiple viewers on the same page, isn't it? I mean, not sure if we should keep it outside, because it won't be garbage-collected ever. I think it should be a part of the instance, just to be disposed when the viewer is disposed.

larsmoa: Agree, not really an issue but lets avoid it.

PR #687, opened 2020-06-23
chore: use yarn, simplify publish scripts

maksnester:
Not much here. I just created yarn.lock from package-lock and replaced commands to emphasize that we're using yarn here. 

Let's remove package-lock a bit latter once all of us start using yarn for installing stuff.

**UPD:** accidentally merged here [that PR](https://github.com/cognitedata/reveal/pull/688), sorry. But don't really want to split it back again. If it's really an issue - let me know.


aslettemark:
Should the Github workflow be changed as well?

maksnester:
@aslettemark sure, thanks for reminding me that.

maksnester:
Not sure what the heck is going on with the examples build, will check it later.

larsmoa:
@maksnester : yeah, me neither. Tried reproducing the failure locally without success (or with success, depending on how you look at it ;) )

maksnester:
@larsmoa it feels like before (when examples were built with npm) missing dependencies somehow being resolved from viewer/node_modules. Maybe because reveal is set as local dependency `"@cognite/reveal": "../viewer/dist",`.

PR #694, opened 2020-06-23
fix(bug): Fix bug causing styles not to be reset to default, causing deselectAllNodes() to not have any effect

larsmoa:


viewer/src/datamodels/cad/MaterialManager.ts
```diff
@@ -84,40 +84,57 @@ export class MaterialManager {
...
 
+    const colorData = materials.overrideColorPerTreeIndex.image.data;
+    const visibilityData = materials.overrideVisibilityPerTreeIndex.image.data;
+
     const count = treeIndices.length;
     for (let i = 0; i < count; ++i) {
       const treeIndex = treeIndices[i];
+
+      // Reset to original style (matches defaults from materials.ts)
+      colorData[4 * treeIndex + 0] = 0;
```
j-bjorne, line 99: Suggest we wait on merging this until @christjt has finished the remaining work on. I personally do not like having a default hardcoded inside a function here, and would rather see the `DefaultNodeAppearance.NoOverrides` contain the default color and visibility data.

larsmoa: We can wait with merge. 

Note that the values that are defaulted are not values used, but special values meaning "no override". But I agree that having them coded in NoOverrides might be better to avoid two writes and multiple definitions of values rather than one source of truth 👍

larsmoa:
Also fixed in https://github.com/cognitedata/reveal/pull/706/files, closing.

PR #704, opened 2020-06-25
Switch to use V1 endpoints instead of playgrounds endpoints.

larsmoa:
This introduces the use of modelId and revisionId to identify a model in CDF. Note that the "model"-query parameter used in the examples now has been replaced by "modelId" and "revisionId".


yarn-error.log
```diff
@@ -0,0 +1,36 @@
...
+Yarn version: 
+  1.19.1
+
+Node version: 
+  10.16.3
+
+Platform: 
+  darwin x64
+
+Trace:
```
maksnester, line 16: @larsmoa I added some .gitignore for yarn stuff in your branch, since your changes reminded me about it, but I added it for viewer/. This file should be deleted I guess. 

I see you wanted to add some eslint stuff here, but doesn't look like you finished on that.

viewer/src/utilities/networking/CogniteClient3dExtensions.ts
```diff
@@ -58,37 +55,40 @@ export class CogniteClient3dExtensions
...
-    const response = await this.client.post<ItemsResponse<{ model: IdEither; outputs: BlobOutputMetadata[] }>>(url, {
-      data: request
-    });
+  public async getOutputs(modelIdentifier: {
+    modelId: number;
+    revisionId: number;
+    format?: File3dFormat | string;
+  }): Promise<Model3DOutputList> {
+    const { modelId, revisionId, format } = modelIdentifier;
+    const url = `/api/v1/projects/${this.client.project}/3d/models/${modelId}/revisions/${revisionId}/outputs`;
```
maksnester, line 81: @larsmoa why don't we have it in public API? I mean, I don't see anything like [outputs](https://docs.cognite.com/api/v1/#tag/3D-Models) in the API reference.

yarn-error.log
larsmoa: Experiment that I didn't follow up on - oops

viewer/src/utilities/networking/CogniteClient3dExtensions.ts
larsmoa: It's actually on its way, just hasn't made it to the doc yet. We should move the stuff we implement here to @cognite/sdk

PR #707, opened 2020-06-26
Remove package-lock.json

aslettemark:
package-lock.json is not needed now that we're running yarn


aslettemark:
@maksnester I think you're the best reviewer for this.

PR #708, opened 2020-06-26
Upload image diff as workflow artifact when failing visual tests

aslettemark:


aslettemark:
Staying as a draft while I mess with CI

PR #712, opened 2020-06-26
mn/use eslint instead of tslint

maksnester:
Aligned as much as I could with our current configs. So now basically we should have everything we had but now it really works (some rules were declared but I've never seen any warnings or errors related, e.g. unbound methods).

I haven't tested it in VSCode, but I think it should work best for you if you will run eslint fix on save. Something like this: (?)
```
{
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  },
}
```
however, it shouldn't really matter because I added pre-commit hook with linting on staged files, so if you in rage and want to commit whatever you have, you can ignore git-hooks (e.g. `git commit -m 'foo' --no-verify`).

It's not anyhow related to Joachim's coding guides yet, it's more about replacing tslint with eslint. 

---

Regarding optional chaining, well, now linting plays well with it, but compiler doesn't. It's easy to enable it, but after some consideration, I think we shouldn't use it. This is the example of how result js code will look after ts transpilation:

```diff
- const sectorCuller =
-     (options && options.internal && options.internal.sectorCuller) || new ByVisibilityGpuSectorCuller();
+ const sectorCuller = options?.internal?.sectorCuller || new ByVisibilityGpuSectorCuller();
```

```js
var _a;
const sectorCuller = ((_a = options === null || options === void 0 ? void 0 : options.internal) === null || _a === void 0 ? void 0 : _a.sectorCuller) || new _internal__WEBPACK_IMPORTED_MODULE_10__["ByVisibilityGpuSectorCuller"]();
```

You can play with it [in babel playground](https://babeljs.io/en/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAZiEMC8MDeMCGAuGBGABhgF8AoUSEAGwFMA6KkAcwAoEQB-OzLiEAWxpQAFgEswTLvxAAnGgEoA3EA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Ces2015%2Cenv&prettier=false&targets=&version=7.10.3&externalPlugins=) output is the same.

Also, some evil tongues say [there might be performance impact](https://dev.to/smeijer/the-costs-of-optional-chaining-5bno) if optchain is overused. So I don't think it's a good idea to enable it. At least in a library like reveal that should be highly performant.


larsmoa:
Thx for the thorough check wrt optional chaining 👍 

I get problems with `prettier`, which wants to add double newlines everywhere:
```
  118:5   error  Insert `␍`  prettier/prettier
  119:1   error  Insert `␍`  prettier/prettier
  120:17  error  Insert `␍`  prettier/prettier
  121:3   error  Insert `␍`  prettier/prettier

✖ 11107 problems (11107 errors, 0 warnings)
  11104 errors and 0 warnings potentially fixable with the `--fix` option.
```
Any idea?

maksnester:
>  get problems with prettier, which wants to add double newlines everywhere:

@larsmoa Yeah, I had exactly the same but decided that we mostly use `crlf`, but probably we just use a mix of them everywhere. I'll set `lf` because it is a default setting in prettier. If any of you will have that issue, just run `yarn run lint:fix` and it will be fixed. 

I've pushed an update. Let me know if anything.

PR #722, opened 2020-07-01
Rename ParserWorker to RevealParserWorker

aslettemark:
More importantly renames its corresponding .worker.js file from `parser.worker.js` to `reveal.parser.worker.js` and the web worker from `parser`to `reveal.parser`. This avoids any potential conflicts with other packages.


examples/config-overrides.js
```diff
@@ -14,6 +14,11 @@ module.exports = function override(config) {
   config.plugins.push(
     new CopyWebpackPlugin({
       patterns: [
+        {
+          from: `${revealSource}/**/reveal.*.worker.js`,
```
maksnester, line 18: Isn't it handled already by next pattern `*.worker.js` ?

aslettemark: You're right, this was the result of some back and forth changes

PR #724, opened 2020-07-02
feat(metrics): init, addModel, error - tracked in mixpanel

maksnester:
Looks like that in the mixpanel (reports are configurable):

**Add model overview**
image

**Errors**
image

**Errors with messages** (prod stack doesn't look very useful though)
image


viewer/src/utilities/metrics.ts
```diff
@@ -0,0 +1,70 @@
...
+    return;
+  }
+  mixpanel.init(MIXPANEL_TOKEN, { persistence: 'localStorage' });
+  globalProps.VERSION = VERSION;
+  if (project) {
+    globalProps.project = project;
+  }
+
+  trackEvent('init', eventProps);
+}
```
larsmoa, line 46: Just to make sure, does this handle when Mixpanel is blocked, e.g. by an exentsion? Just making sure that Adblock or Disconnect extensions etc doesn't cause Reveal to fail.

maksnester: @larsmoa good point. I double-checked that. And no, mixpanel isn't so poorly designed, so we don't need to cover it with try/catch every time we use it. It won't throw in case there is a network error, or even a wrong token passed.

PR #727, opened 2020-07-03
Implement screenshot test presets

aslettemark:


aslettemark:
Didn't migrate a lot from the old viewer, as this setup is based on RevealManager, not Cognite3DViewer. The latter does not support serving local models.

examples/src/visual_tests/basic.test.ts
```diff
@@ -1,15 +1,18 @@
-import { gotoAndWaitForRender } from "./VisualTestUtils";
+import { screenShotTest } from "./VisualTestUtils";
+
+const test_presets = [
+    "clipping",
+    "default_camera",
+]
 
 describe('Testable', () => {
```
maksnester, line 8: Instead of having that Testable thing + test_presets I suggest having different test files, like `clipping.test.ts` etc. Yes, they all will look the same like 

```ts
import { screenShotTest } from "./VisualTestUtils";

describe('Clipping', () => {
    it('renders primitives model', async () => {
        await screenShotTest('http://localhost:3000/testable');
    });

   it('matches the screenshot', async () => {
        await screenShotTest('http://localhost:3000/testable?test=' + test);
    });
});
```

But I think it's absolutely fine, no need to be too smart on that in tests. As a bonus, artifact names will be a bit shorter, like 

```diff
-basic-test-ts-testable-matches-the-screenshot-for-test-preset-default-camera-1-snap.png
+basic-test-ts-clipping-matches-the-screenshot-1-snap.png
```

The same thing I would say about `examples/src/pages/Testable.tsx` - now it's all tests in a single file. So it'll be worse when more tests added. Though it's relatively simple for now, so I think it can wait for it's time to be refactored.

aslettemark: I think we should keep the preset list as long as we still have all the preset tests in the same tsx page.

maksnester: 👌

PR #728, opened 2020-07-03
Move finalize to Rust

Mattman22:


viewer/src/datamodels/cad/sector/CachedRepository.ts
```diff
@@ -204,25 +200,9 @@ export class CachedRepository implements Repository {
   }
 
   private loadDetailedSectorFromNetwork(wantedSector: WantedSector): Observable<ConsumedSector> {
```
j-bjorne, line 202:
```
private loadDetailedSectorFromNetwork(wantedSector: WantedSector): Observable<ConsumedSector> {
    const detailedSectorObservable = from(
      /** Needs to be renamed so that it is clear this also takes over network requests etc.
       * Say we rename it to loadDetailedFromNetwork() and we send in a bit more data.
       * {
       *   blobUrl: wantedSector.blobUrl,
       *   headers: this._modelSectorProvider.headers,
       *   i3dFile: wantedSector.metadata.indexFile.fileName,
       *   ctmFiles: wantedSector.metadata.indexFile.peripheralFiles,
       * }
       *
       * That way you don't have to first request the i3d file, then parse to see what ctm files are needed to download.
       */
      this._modelDataParser.parseAndFinalizeDetailed(wantedSector.metadata.indexFile.fileName, {
        blobUrl: wantedSector.blobUrl,
        headers: this._modelSectorProvider.headers
      })
    ).pipe(
      catchError(error => {
        trackError(error, {
          moduleName: 'CachedRepository',
          methodName: 'loadDetailedSectorFromNetwork'
        });
        this._consumedSectorCache.remove(this.wantedSectorCacheKey(wantedSector));
        throw error;
      }),
      map(finalizedDetailed => {
        const sector = this.unflattenSector(finalizedDetailed);
        this._parsedDataSubject.next({
          blobUrl: wantedSector.blobUrl,
          sectorId: wantedSector.metadata.id,
          lod: 'detailed',
          data: sector
        }); // TODO: Remove when migration is gone.
        return { ...wantedSector, data: sector };
      }),
      this._modelDataTransformer.transform(),
      map(group => ({ ...wantedSector, group })),
      shareReplay(1),
      take(1)
    );
    this._consumedSectorCache.forceInsert(this.wantedSectorCacheKey(wantedSector), detailedSectorObservable);
    return detailedSectorObservable;
  }
```

This entire function can be simplified

Mattman22:
### Status of this PR

* Performance is comparable to `master`, but probably slightly worse
* Network requests take significatly longer (sometimes over twice as long)
* Multiple requests for the same `ctm` file are sent for different sectors, because they are only added to the cache after they are completely loaded (adding a Rust `Future` to a cache is much harder than adding a RxJS `Observable`)
* The changes made are not very readable, primarily due to using async Rust for the networking

### Steps Forward
Based on the results of this PR, possible future tasks (for `master`) could be:
* Move `finalizeDetailed` to the workers
  * This would reduce the work for the main thread and is not a huge change (see the first commit of this PR)
* Use `Comlink.transfer` from the workers where possible
  * For example, if returning a result containing a buffer from the worker, 
  ```js
  return result;
  ```
  becomes
  ```js
  return Comlink.transfer(result, [result.buffer]);
  ```
* If the wasm [interface types proposal](https://github.com/WebAssembly/interface-types/blob/master/proposals/interface-types/Explainer.md) is implemented, then we could hopefully avoid flattening our data to send it between Rust and JS. This could help us eliminate the de/serialization, flattening, and conversion we do at the Rust-JS interface, so it is something to keep an eye on.

### Takeaways
* Sending buffers between Rust and JS is fast.
  * Sending multiple MB flattened buffers from Rust happened in at most a couple milliseconds.
* De/serializing data is the real bottleneck for the interface between Rust and JS
  * Using `serde` to de/serialize to/from JSON took hundreds of milliseconds (converting the `ctm_map` mapping a single filename to a parsed CTM file took ~600ms).
  * Using `flatbuffers` to serialize took ~200ms in the worst case I saw (about 5MB of data) and deserializes in just a couple milliseconds without copies
  * Interface types may improve this in the future
* Networking is the main bottleneck for this PR.
* Caching is harder from the workers (because of duplicates) and from Rust (because it's harder to store a `Future` than an `Observable`).

### Future Improvements for this PR
If this PR were to be salvaged, performance would need to be improved. Some ways forward, in order of impact:
1. Network requests currently take longer from the workers than the main thread; figure out why and fix it.
   * Is this because too many concurrent requests are being sent at once? 
   * Is this somehow a limitation of workers?
1. The current implementation uses an LRU cache that is only added to after the file is completely downloaded and parsed, but this can be improved.
   * Figure out a way to store a `Future` in the cache, so don't send another network request if the file is already being fetched.
   * Use a shared cache for all workers, probably with a shared instance of a JS class that can do network requests and has an internal cache (which would also avoid the need to send the `blobUrl` and `headers` to Rust for every sector)
1. After networking, the most time is spent serializing the data to send to JS.
   * Use [interface types](https://github.com/WebAssembly/interface-types/blob/master/proposals/interface-types/Explainer.md) if available
   * Try other libraries than `flatbuffers`
   * Allocate and reuse a single buffer (current flattening times can be reduced by around half in many cases by preallocating a large buffer and reusing it, because allocating such large buffers by doubling when out of space is slow)

Improvements 1. and 2. were not made because they are likely to be high effort and its unclear how effective they would be.  
Improvement 3. would probably be most effective with interface types, which are not currently available and `flatbuffers` is already very fast and reusing the same buffer did not have a noticable performace impact in the face of the other issues, so it should only be considered if significant improvements are made elsewhere first.

PR #738, opened 2020-07-03
Add Cognite3dViewer.getVersion() and VERSION export

larsmoa:


viewer/webpack.config.js
```diff
@@ -110,7 +110,7 @@ module.exports = env => {
     },
     plugins: [
       new webpack.DefinePlugin({
-        VERSION: JSON.stringify(packageJSON.version),
+        VERSION: JSON.stringify(packageJSON.version) + JSON.stringify(development ? '-dev' : ''),
```
maksnester, line 113: For what reason do we need that `-dev` suffix? It will appear only when you run build:watch I believe. Or build manually with `yarn run build --env.development`.

Also, it could be just `JSON.stringify(packageJSON.version + development ? '-dev' : '')`

Also, I don't really like how it looks as it spreads. I suggest the following changes:
* use process.env with webpack.DefinePlugin
* somewhere create environment.d.ts file with a declaration of what process.env has, it would be like 

```ts
declare global {
  namespace NodeJS {
    interface ProcessEnv {
      VERSION: string;
      MIXPANEL_TOKEN: string
      // maybe something else later
    }
  }
}

// If this file has no import/export statements (i.e. is a script)
// convert it into a module by adding an empty export statement.
export {}
```

So we won't need then to declare that version in every file we want it. And no need for additional globalProps because `process.env` is a quite common thing for that case.

viewer/src/index.ts
```diff
@@ -1,6 +1,10 @@
 /*!
  * Copyright 2020 Cognite AS
  */
+declare let VERSION: string;
+const globalProps = {
```
maksnester, line 5: isn't it unused here? Not sure why linter allowed to commit it...

maksnester: Created a PR specifically for such cases https://github.com/cognitedata/reveal/pull/739

PR #743, opened 2020-07-05
refac: Get rid of duplicated code in RevealManagerBase-subclasses.

larsmoa:
- Also moved some logic from Cognite3dViewer to new class CdfModelManager.
- Laying the ground work for adding support for local models to Cognite3dViewer


viewer/src/public/LocalHostRevealManager.ts
```diff
@@ -4,89 +4,24 @@
...
-import { PointCloudMetadataRepository } from '@/datamodels/pointcloud/PointCloudMetadataRepository';
-import { PointCloudFactory } from '@/datamodels/pointcloud/PointCloudFactory';
-import { PointCloudManager } from '@/datamodels/pointcloud/PointCloudManager';
-import { DefaultPointCloudTransformation } from '@/datamodels/pointcloud/DefaultPointCloudTransformation';
 import { RevealOptions } from './types';
-import { Subscription, combineLatest } from 'rxjs';
-import { distinctUntilChanged, map } from 'rxjs/operators';
-import { trackError } from '@/utilities/metrics';
+import { initMetrics } from '@/utilities/metrics';
+import { omit } from 'lodash';
```
maksnester, line 12: As far as I remember lodash always worked poorly with tree-shaking. Need to check if it enlarges our bundle on the whole lodash, or it's tree-shaken. But if it adds the whole lodash it's better to use `import omit from 'lodash/omit'`

viewer/src/datamodels/pointcloud/PointCloudMetadataRepository.ts
```diff
@@ -9,11 +9,14 @@ import { PointCloudMetadata } from '@/datamodels/pointcloud/PointCloudMetadata';
 // TODO 25-05-2020 j-bjorne: Move into common type folder
 import { DataRepository } from '@/datamodels/cad/DataRepository';
 import { CadTransformationProvider } from '@/datamodels/cad/CadTransformationProvider';
+import { File3dFormat } from '@/utilities';
 
+type ModelIdentifierWithFormat<T> = T & { format: File3dFormat };
```
maksnester, line 14: It's copypasted 4 times... I'm not sure why do you need it at all, but If the `format` is something always needed then it's better to declare it somewhere like 

```ts
export class PointCloudMetadataRepository<TModelIdentifier extends { format: File3dFormat }>
```

viewer/src/public/LocalHostRevealManager.ts
maksnester: I tried to find a difference in result bundle for examples and there is no difference. Looks like cognite SDK already brings the whole lodash. But still, I think it's better not to import it like that and on lodash page it also said 

> Cherry-pick methods for smaller browserify/rollup/webpack bundles.

https://www.npmjs.com/package/lodash

viewer/src/datamodels/pointcloud/PointCloudMetadataRepository.ts
larsmoa: Extends won't do the trick here as we add the format field in the implementation of `loadData`. I don't really like the need for a format like this, but I didn't want to go down that path now...

maksnester: Ok, it's fine. I just pointed on what I noticed and didn't really understand.

viewer/src/public/LocalHostRevealManager.ts
maksnester: I sent a PR to our js sdk as well.

viewer/src/public/RevealManagerBase.ts
```diff
@@ -72,21 +114,191 @@ export class RevealManagerBase<TModelIdentifier> implements RenderManager {
...
+    modelIdentifier: TModelIdentifier
+  ): Promise<[PotreeGroupWrapper, PotreeNodeWrapper]>;
+  protected async loadModel(
+    type: 'cad' | 'pointcloud',
+    modelIdentifier: TModelIdentifier,
+    nodeApperanceProvider?: NodeAppearanceProvider
+  ): Promise<[PotreeGroupWrapper, PotreeNodeWrapper] | CadNode> {
+    trackAddModel(
+      {
+        moduleName: 'RevealManager',
```
maksnester, line 197: oh, that's lying now. I suggest to use RevealMangerBase and `loadModel` as method name.

viewer/src/public/RevealManagerBase.ts
```diff
@@ -72,21 +114,191 @@ export class RevealManagerBase<TModelIdentifier> implements RenderManager {
...
+        methodName: 'addModel',
+        options: { nodeApperanceProvider }
+      },
+      modelIdentifier
```
maksnester, line 201: Is it compiled? `modelId: number; revisionId: number` are required for trackAddModel

maksnester: Oh, I found changes, sry.

maksnester: But still not sure why does it compile, you don't pass a type here for trackAddModel<MyType>. But if you declared a generic type without a default value, you must then pass it everywhere.

maksnester: ok, I understood it... That `trackAddModel<TModelIdentifier>` looks confusing to me, in fact `modelIdentifier` is literally `any`. I suggest putting `any` here to make it clear.

```ts
export function trackAddModel(eventProps: EventProps, modelIdentifier: any)
```

viewer/src/public/LocalHostRevealManager.ts
larsmoa: Nice, great!

viewer/src/public/RevealManagerBase.ts
j-bjorne: Yep, change to `moduleName: 'RevealManagerBase'` and `methodName: 'loadModel'` not sure what to do with the `options` parameter here, prbably leave it as is?

viewer/src/public/RevealManagerBase.ts
```diff
@@ -72,21 +114,191 @@ export class RevealManagerBase<TModelIdentifier> implements RenderManager {
     }
   }
 
+  public get renderMode(): RenderMode {
```
j-bjorne, line 117: I will accept this for now, I do wish to have a talk about the consequences of having a single point of entry.

viewer/src/public/RevealManagerBase.ts
larsmoa: I think maybe we should detach the events and modules from code, such that the log is `CAD Model::load` instead of `RevealManagerBase::loadModel` to allow code to change without worrying about being able to compare data in MixPanel historically.

larsmoa: (Fixed as commented now)

viewer/src/public/RevealManagerBase.ts
larsmoa: Node that I did not add `implements RenderManager` to `RevealManagerBase`. As long as `RevealManagerBase` actually implements that I can't see why `RenderMode` doesn't belong there...

PR #775, opened 2020-07-08
RevealManager cleanup

j-bjorne:
Some early interface cleaning turned into a complete re-haul av RevealManager constructor.

```
  initMetrics(revealOptions.logMetrics !== false, 'local', {
    moduleName: 'RevealManager',
    methodName: 'constructor',
    constructorOptions: omit(revealOptions, ['internal'])
  });
```
Was moved to outside of RevealManager, I don't really like it being there and would rather move it inside it. We might not have it on RevealManager, but rather place somewhere else, say inside the Cad- and PointCloudManager, or inside RevealManager, but with different info as neither the client with the project or the revealOptions are part of the constructor anymore.

Also removed some classes and simplified some things.


larsmoa:
Why don't you like `initMetrics` in the constructor?

j-bjorne:
I do like it in the constructor, and that is where I want it. Current reveal manager constructor only takes in Cad- and PointCloudManager, and none of the other parts which current init metrics did so I didn't know what to change with it as I was not aware of what it needed.

j-bjorne:
Also they won't need the unknown as most cases they don't want to do it the way it is setup now. They would do say const revealManager: RevealManager<CdfModelIdentifier> = createCdfRevealmanager(..);

It is only due to the way we have setup the react examples that we have to use the unknown.

j-bjorne:
Also your suggestion sounds like my old union approach which you did not like, where I said input was CdfModelIdentifier | LocalUrlModelIdentifier and you disliked the handling of if one send to a data source, if another send to another datasource

viewer/src/public/createRevealManager.ts
```diff
@@ -0,0 +1,45 @@
...
+export function createLocalRevealManager(revealOptions: RevealOptions = {}): RevealManager<LocalModelIdentifier> {
+  const modelDataClient = new LocalModelDataClient();
+  initMetrics(revealOptions.logMetrics !== false, 'local', {
+    moduleName: 'RevealManager',
+    methodName: 'constructor',
+    constructorOptions: omit(revealOptions, ['internal'])
+  });
+  return createRevealManager(modelDataClient, revealOptions);
+}
+export function createCdfRevealManager(
```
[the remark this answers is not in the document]
maksnester: Might be handled by linter 

https://eslint.org/docs/rules/padding-line-between-statements

examples/src/pages/Clipping.tsx
```diff
@@ -22,6 +19,7 @@ export function Clipping() {
 
   useEffect(() => {
     const gui = new dat.GUI();
+    let revealManager: reveal.RevealManager<unknown>;
```
maksnester, line 22: It's not really unknown, we either use modelId + revisionId or modelUrl, don't we?

I would personally like to have some sane defaults for that template identifier. For most cases, it probably would be modelId + revisionId.

viewer/src/public/createRevealManager.ts
```diff
@@ -0,0 +1,45 @@
...
+  return createRevealManager(modelDataClient, revealOptions);
+}
+
+export function createRevealManager<T>(
+  client: ModelDataClient<T>,
+  revealOptions: RevealOptions = {}
+): RevealManager<T> {
+  const cadManager = createCadManager(client, revealOptions);
+  const pointCloudManager = createPointCloudManager(client);
+  return new RevealManager(cadManager, pointCloudManager);
```
maksnester, line 44: so if eventually constructor is called, why not to have initMetrics in that function or constructor directly?

viewer/src/public/RevealManager.ts
```diff
@@ -2,50 +2,225 @@
...
+  };
+
+  private _isDisposed = false;
+  private readonly _subscriptions = new Subscription();
+  private readonly eventListeners = {
+    sectorNodeIdToTreeIndexMapLoaded: new Array<SectorNodeIdToTreeIndexMapLoadedListener>(),
+    loadingStateChanged: new Array<LoadingStateChangeListener>()
+  };
+
+  constructor(cadManager: CadManager<TModelIdentifier>, pointCloudManager: PointCloudManager<TModelIdentifier>) {
```
maksnester, line 39: Probably should be marked with `/** @interhal */` if you don't expect any outside usage of it

viewer/src/public/createRevealManager.ts
maksnester: Yep, sorry, I noticed that there is no client in RevealManager itself.

j-bjorne: Since I didn't know the intention of mixpanel, I didn't want to change it's behavior :)

j-bjorne: The reason why it isn't in the constructur is that none of the info added to initmetric exists inside the current constructor.

maksnester: Well, I just would put these metrics into that createRevealManager to not copy it twice 🤷‍♂️ 

But with honest module/method names at least.

```
 initMetrics(revealOptions.logMetrics !== false, client.project || 'local', {
    moduleName: 'createRevealManager',
    methodName: 'createRevealManager',
    constructorOptions: omit(revealOptions, ['internal'])
  });
```

j-bjorne:
Awesome Maksim, now it just needs an approval :D

examples/src/pages/Clipping.tsx
j-bjorne: This is more complicated than just changing it. I will setup a proposal for our examples later which will change this behavior.

viewer/src/public/RevealManager.ts
j-bjorne: marked it with internal for now :)

PR #802, opened 2020-07-13
[WIP] Mock suggestion for architecture for a new CAD loading pipline.

larsmoa:
Still very much a draft. Making a PR to make sure this is not lost as I enter vacation mode.


j-bjorne:
Does each CadModel/Node have its own cad loader?

PR #808, opened 2020-07-15
Don't schedule new sectors while camera is moving.

larsmoa:
Temporary disabling scheduling while on the move to avoid some choppyness. Should be enabled when we improve performance in culler.


larsmoa:
Await with merging this until we decide if necessary

PR #822, opened 2020-07-16
Avoid memory copies in OpenCTM parser

aslettemark:


viewer/rust/lib.rs
```diff
@@ -17,27 +17,11 @@ impl CtmResult {
...
-        };
-        let data_as_f32 = unsafe {
-            std::slice::from_raw_parts(
-                data_as_vector3.as_ptr() as *const f32,
-                data_as_vector3.len() * 3,
-            )
-        };
-        Some(Float32Array::from(data_as_f32))
+        let components = self.file.normal_components.as_ref()?;
+        Some(Float32Array::from(components.as_slice()))
```
Mattman22, line 24: This could also be done with a map
`self.file.normal_components.map(|x| Float32Array::from(x.as_slice()))`
but idk if that's more clear or not

PR #835, opened 2020-07-17
chore(examples): disable dependabot for linked deps

maksnester:
https://github.com/cognitedata/reveal/issues/834

Should fix these messages


maksnester:
@aslettemark hopefully 🤞

PR #846, opened 2020-07-21
Dependency updates

aslettemark:


larsmoa:
> Well if it builds, then it should be fine.

Haha, famous last words @maksnester

PR #847, opened 2020-07-21
add documentation site

maksnester:
I guess I could submit it as a PR and add jsdoc or whatever for generated API reference later. 

At the moment the site is already deployed [https://cognitedata.github.io/reveal/](https://cognitedata.github.io/reveal/). There is not much documentation yet, but at least it's ready to add examples/guides/whatever.


aslettemark:
Should we have some branch protection rules for the gh-pages branch?

maksnester:
> Should we have some branch protection rules for the gh-pages branch?

maybe 🤷‍♂️

maksnester:
@aslettemark also, I would like to not have deploy commit associated with my account 😅. But for now, I just set up deployment with my personal access token, which is not something we want to have for company projects.

larsmoa:
Love it! Haven't read through all docs yet but I really like the structure.

maksnester:
@larsmoa let's merge it? Only API reference will be different from what is deployed already.

PR #849, opened 2020-07-22
add api-reference generation from typedoc; add docs for Cognite3DViewer

maksnester:


viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -312,6 +400,10 @@ export class Cognite3DViewer {
     if (options.orthographicCamera) {
       throw new NotSupportedInMigrationWrapperError();
     }
+    if (options.onComplete) {
```
maksnester, line 403: @j-bjorne is it ok to throw on it? I didn't see any usage, so it would be implicitly ignored without throwing.

viewer/src/public/migration/types.ts
```diff
@@ -14,23 +14,32 @@ export interface Color {
...
   sdk: CogniteClient;
+
+  /** An existing DOM element that we will render canvas into. */
   domElement?: HTMLElement;
-  noBackground?: boolean;
+
+  /** Send anonymous usage statistics. */
   logMetrics?: boolean;
+
+  /** @deprecated and ignored */
```
maksnester, line 24: @j-bjorne I just mentioned how it works right now. Some params are ignored, some will throw. 

Is it ok to deprecate these params from now?

viewer/src/public/types.ts
```diff
@@ -7,31 +7,31 @@ import { SectorGeometry } from '@/datamodels/cad/sector/types';
 import { SectorQuads } from '@/datamodels/cad/rendering/types';
 import { SectorCuller } from '@/internal';
 
-export type RevealOptions = {
+export interface RevealOptions {
```
maksnester, line 10: There is some difference between interface and type alias for typedoc and how to document it. Type declaration is more compact, but you can't document props without jsdoc. That how it looks:

image

But with interface keyword and prop-based docs it looks like that:

image

It's much more verbose for interface declaration. But you don't have to write a separate jsdoc to document props. Not really sure which approach should be preferred. I think output for type alias is much easier to read, but interfaces are easier to document.

maksnester: I decided to stick with type + jsdoc here.

viewer/src/public/migration/types.ts
larsmoa: What is the effect from a client perspective on marking as deprecated? Will there be a warning on compile or is it strictly a documentation thingy?

maksnester: Documentation + clarification in code like that:

image

image

Doesn't affect compilation.

PR #864, opened 2020-07-28
Simplifying Pipeline

j-bjorne:
Making some changes to the pipeline to simplify some of its parts. Built upon the rx pipeline fix so this can be merged instead.


viewer/src/datamodels/cad/sector/CachedRepository.ts
```diff
@@ -100,100 +98,95 @@ export class CachedRepository implements Repository {
...
+          return this.loadDetailedSectorFromNetwork(sector);
+        } else if (sector.levelOfDetail == LevelOfDetail.Simple) {
+          return this.loadSimpleSectorFromNetwork(sector);
+        } else if (sector.levelOfDetail == LevelOfDetail.Discarded) {
+          return of(sector).pipe(map(wantedSector => ({ ...wantedSector, group: undefined } as ConsumedSector)));
+        } else {
+          throw new Error('Unhandled case');
+        }
+      }, this._concurrentNetworkOperations),
+      this._loadingCounter.remove(),
```
larsmoa, line 126: What happens if the `flatMap`-thing throws? Are we forever in loading-state because the counter is never decremented?

viewer/src/datamodels/cad/sector/CachedRepository.ts
```diff
@@ -100,100 +98,95 @@ export class CachedRepository implements Repository {
...
+      this._loadingCounter.clear()
```
larsmoa, line 127: Won't this always reset, even if we still are loading?

viewer/src/datamodels/cad/sector/CachedRepository.ts
```diff
@@ -64,7 +58,8 @@ export class CachedRepository implements Repository {
   private readonly _modelSectorProvider: BinaryFileProvider;
   private readonly _modelDataParser: CadSectorParser;
   private readonly _modelDataTransformer: SimpleAndDetailedToSector3D;
-  private readonly _isLoadingSubject: Subject<boolean> = new BehaviorSubject<boolean>(false);
+  private readonly _loadingCounter: RxCounter = new RxCounter();
+  // private readonly _isLoadingSubject: Subject<boolean> = new BehaviorSubject<boolean>(false);
```
larsmoa, line 62: Remvoe

viewer/src/datamodels/cad/sector/CachedRepository.ts
larsmoa, line 127: What happens if the flatMap above throws? Is the decrement not done and hence the isLoading status invalid? And should clear be here?

viewer/src/utilities/RxCounter.ts
```diff
@@ -0,0 +1,52 @@
...
+import { Subject, MonoTypeOperatorFunction } from 'rxjs';
+import { scan, tap, finalize, shareReplay } from 'rxjs/operators';
+
+enum Action {
+  Add,
+  Remove,
+  Clear
+}
+
+export class RxCounter {
```
larsmoa, line 14: Add some doc and maybe a unit test?

viewer/src/utilities/RxCounter.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+  }
+
+  public clear<T>(): MonoTypeOperatorFunction<T> {
+    return finalize<T>(() => {
+      this._actionSubject.next(Action.Clear);
+    });
+  }
+
+  public observeCount() {
```
larsmoa, line 35: Declare return type

viewer/src/utilities/RxCounter.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+  }
+
+  public remove<T>(): MonoTypeOperatorFunction<T> {
+    return tap<T>({
+      next: () => this._actionSubject.next(Action.Remove)
+    });
+  }
+
+  public clear<T>(): MonoTypeOperatorFunction<T> {
```
larsmoa, line 29: Nitpick, but I'd call the functions `increment`, `decrement` and `reset` instead as add/remove/clear sound like a collection.

viewer/src/utilities/RxCounter.ts
```diff
@@ -0,0 +1,52 @@
...
+      scan((count, action) => {
+        switch (action) {
+          case Action.Add:
+            return count + 1;
+          case Action.Remove:
+            return count - 1;
+          case Action.Clear:
+            return 0;
+          default:
+            throw new Error('Unhandled switch case');
```
larsmoa, line 46: Switch to 'Unsupported action ${action)` or something.

viewer/src/datamodels/cad/sector/CachedRepository.ts
j-bjorne: Finalize runs if the observable completes, and that should reset the counter to 0 and loading. The reason why I introduced this is due to loading not being very helpful and I want to change it to Loading X sectors or something, so that there is a bit more indication that something is happening :)

viewer/src/datamodels/cad/sector/CachedRepository.ts
j-bjorne: Clear is using finalize, while remove is using tap

j-bjorne: So the reason we have clear is due to the switchmap effect. When the observable is removed from the switchmap the observable gets cancelled, hence it completes, and resets the counter.

viewer/src/utilities/RxCounter.ts
j-bjorne: Yes!

viewer/src/utilities/RxCounter.ts
j-bjorne: I agree, I will do that

viewer/src/utilities/RxCounter.ts
j-bjorne: Done

viewer/src/utilities/RxCounter.ts
j-bjorne: Did not add a doc, but made a test

viewer/src/datamodels/cad/sector/CachedRepository.ts
j-bjorne: Gone

viewer/src/utilities/RxCounter.ts
```diff
@@ -0,0 +1,52 @@
...
+    return tap<T>({
+      next: () => this._actionSubject.next(Action.Remove)
+    });
+  }
+
+  public resetOnComplete<T>(): MonoTypeOperatorFunction<T> {
+    return finalize<T>(() => {
+      this._actionSubject.next(Action.Clear);
+    });
+  }
```
larsmoa, line 33: Doc what is meant by `onNext` and `onComplete` (because I don't understand)

viewer/src/utilities/RxCounter.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  public observeCount(): Observable<number> {
+    return this._actionSubject.pipe(
+      scan((count, action) => {
+        switch (action) {
+          case Action.Add:
```
larsmoa, line 39: Change enum to `Increment`, `Decrement` and `Reset` too

viewer/src/utilities/RxCounter.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+  });
+
+  test('decrement on next', done => {
+    // const incrementCount = 3;
+    // const decrementCount = 2;
+    const counter = new RxCounter();
+    const operationCount = 4;
+    const wait = flatMap(value => new Promise(resolve => setTimeout(resolve, 100, value)));
+    expect.assertions(operationCount * 2);
```
larsmoa, line 35: What is this assertion?

viewer/src/utilities/RxCounter.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+import { Subject, MonoTypeOperatorFunction, Observable } from 'rxjs';
+import { scan, tap, finalize, shareReplay } from 'rxjs/operators';
+
+enum Action {
+  Add,
+  Remove,
+  Clear
+}
```
larsmoa, line 12: Change enum to `Increment`, `Decrement` and `Reset` too

viewer/src/utilities/RxCounter.ts
j-bjorne: onNext and onComplete are rx terms. This is an rx operation counter, say if you have a pipeline with events:
-----a-----b-----c-----d-----e
if you add the incrementOnNext then on each submission you will increment.
onComplete will be called once the observable completes.

viewer/src/utilities/RxCounter.test.ts
j-bjorne: Assertion is a way to make sure the test has x numbers of assertions. It is a way to make sure that your loops etc and the number of tests that you assert are correct. In this case since we expect the increment to happen 4 times and in the subscriber we expect 4 values, assertions basically counts the number of expects and if it for some reason is more than 4 it fails it.

viewer/src/utilities/RxCounter.ts
j-bjorne: So the way this work, imagine you have something like this.

```
const rxCounter = new RxCounter();
...

const sector$: Observable<Sector> = this.getSectorObservable();
return sector$.pipe(
  rxCounter.incrementOnNext(),
  this.getSectorData(),
  rxCounter.decrementOnNext(),
  rxCounter.resetOnComplete()
);

each sector will increase the counter by 1, each sector with data returned will decrease the counter by 1, if the observable completes the counter is reset
```

PR #867, opened 2020-07-30
fix: optimize hasIntersection() by using iterator instead of forEach().

larsmoa:
See https://jsben.ch/zRusx for performance comparison. This code was a hotspot in the code for rendering highlighted objects and hence worth the effort to optimize.


larsmoa:
See https://jsben.ch/zRusx for performance comparison. This code was a hotspot in the code for rendering highlighted objects and hence worth the effort to optimize.

PR #895, opened 2020-08-04
Add feature to include coloring children in setNodeColorByTreeIndex endpoint

christjt:
Added do-not-merge until we figure out what to do with missing NodeId for a given treeIndex.


viewer/src/public/migration/Cognite3DModel.ts
```diff
@@ -209,48 +221,72 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
 
   async setNodeColor(nodeId: number, r: number, g: number, b: number): Promise<void> {
```
j-bjorne, line 222: I still think this sounds too much like the old api's set color, where you changed the color of the node. If I am not mistaken this sets the color of the node if it isn't highlighted currently.

Last thing I want is people asking us why the color doesn't change when you try and set the color, and we figure out that they already have the node selected or filtered or who knows what effect we have on it which overrides the color.
Suggestions:
`setBaseColor`
`setDefaultColor` <-- don't really like this one.
`setStandardColor`

viewer/src/public/migration/Cognite3DModel.ts
```diff
@@ -209,48 +221,72 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
...
   async resetNodeColor(nodeId: number): Promise<void> {
     const treeIndex = await this.nodeIdAndTreeIndexMaps.getTreeIndex(nodeId);
-    this.resetNodeColorByTreeIndex(treeIndex);
+    await this.resetNodeColorByTreeIndex(treeIndex);
   }
 
-  resetNodeColorByTreeIndex(treeIndex: number) {
-    this.nodeColors.delete(treeIndex);
-    this.cadNode.requestNodeUpdate([treeIndex]);
+  async resetNodeColorByTreeIndex(treeIndex: number, applyToChildren = false): Promise<number> {
```
j-bjorne, line 255: Same comment applies. I feel like people will do something like.
```
let selected: [] = [];

const select = (model: Cognite3DModel, treeIndices: []) => {
  if (selected.length > 0 ) {
    await model.resetNodeColorByTreeIndex(selected);
  }
  selected = treeIndices;
  await model.setNodeColorByTreeIndex(selected, selectedColor);
};
```

And wonder why it doesn't work.

viewer/src/utilities/NumericRange.ts
```diff
@@ -0,0 +1,38 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+export class NumericRange {
+  readonly from: number;
+  readonly count: number;
```
j-bjorne, line 6: Nice

viewer/src/public/migration/Cognite3DModel.ts
larsmoa: I agree, we should aim for people providing `NodeStyle` instead to define all styling aspects. For now however I think keeping it close to the old API makes sense.

viewer/src/public/migration/Cognite3DModel.ts
larsmoa: Hm, why won't that work?

j-bjorne: sorry, I was missing a line there..

```
let selected: [] = [];

const select = (model: Cognite3DModel, treeIndices: []) => {
  if (selected.length > 0 ) {
    model.deselectAllNodes();
    await model.resetNodeColorByTreeIndex(selected);
  }
  selected = treeIndices;
  await model.setNodeColorByTreeIndex(selected, selectedColor);
  treeIndices.forEach(treeIndex => model.selectNodeByTreeIndex(treeIndex));
};
```

If I am not mistaken the setColor sets what is returned by the model appearance when it isn't selected?

larsmoa: The code you provide will cause the `selectedColor` to not have any effect untill `deselectNode` is called, yes.

PR #900, opened 2020-08-05
feat(viewer): use cdn for releases

maksnester:
So... the trick here is just to set correct publicPath for our build and overcome CORS worker import with importScript. 

When we will release it, our next release will be available automatically at CDN. Example with my test library is here: https://cdn.jsdelivr.net/npm/mn-reveal-worker-test@0.0.12/reveal.parser.worker.js. You might notice the part like `i.p="https://cdn.jsdelivr.net/npm/mn-reveal-worker-test@0.0.12/",` it is webpack publicPath being set and used to point on other files from cdn (worker chunk, wasm files).

When we use it locally with examples, we should keep the old way with copying. Because our local changes obviously aren't published anywhere, so just keep copying our files and host them in examples manually (nothing changed for devs).

I haven't updated the docs yet, but after that being released all that stuff documented here: https://cognitedata.github.io/reveal/docs/installation should be removed (except normal installation experience with `npm i @cognite/reveal`


larsmoa:
Trying to wrap my head on how this will work:
1. Is the CDN based on the npm packages?
2. Is there any way to support embedding everything within the same package currently, or is the CDN required? (not necessarily a blocker, but something we maybe should add support for in the future)
3. Is it possible to override the path to the worker JS somehow?

maksnester:
@larsmoa 

> Is the CDN based on the npm packages?

Yes. You publish it and it just becomes available at jsdelivr by link with package name and version. Like 30 seconds after publishing or something like that.

> Is there any way to support embedding everything within the same package currently, or is the CDN required? (not necessarily a blocker, but something we maybe should add support for in the future)

If you mean inlining everything into js bundle - I think no, it's not possible at the moment or I haven't found a way to do so. The main problem I see here is that we use wasm bindgen that produces a js output that we can't control. 

> Is it possible to override the path to the worker JS somehow?

The short answer is no, but for what use-case we might want it? The only way to change paths is by running the viewer build. 

It would be a bit inconvenient for us to debug the latest dev versions on projects, e.g console, but it's doable. The simplest workaround is to run the viewer build with the publicPath pointing on our running examples. I understand it's not convenient but if to choose between inconveniences for users and us, I think we have to choose us. And after all our current workaround with copying and HTML `<base>` tag is quite bad and won't work for every project. For example, I had to write some really [hacky stuff](https://github.com/cognitedata/reveal/blob/master/documentation/src/hooks/useBaseTag.tsx#L10) to make it work in our docusaurus based docs site.

larsmoa:
> @larsmoa
> 
> > Is the CDN based on the npm packages?
> 
> Yes. You publish it and it just becomes available at jsdelivr by link with package name and version. Like 30 seconds after publishing or something like that.

Great!

> 
> > Is there any way to support embedding everything within the same package currently, or is the CDN required? (not necessarily a blocker, but something we maybe should add support for in the future)
> 
> If you mean inlining everything into js bundle - I think no, it's not possible at the moment or I haven't found a way to do so. The main problem I see here is that we use wasm bindgen that produces a js output that we can't control.

Not necessarily inline, but that it's hosted somewhere the user chooses (e.g. for a native Electron app it's nice if all code is hosted locally).

> 
> > Is it possible to override the path to the worker JS somehow?
> 
> The short answer is no, but for what use-case we might want it? The only way to change paths is by running the viewer build.
> 
> It would be a bit inconvenient for us to debug the latest dev versions on projects, e.g console, but it's doable. The simplest workaround is to run the viewer build with the publicPath pointing on our running examples. I understand it's not convenient but if to choose between inconveniences for users and us, I think we have to choose us. And after all our current workaround with copying and HTML `<base>` tag is quite bad and won't work for every project. For example, I had to write some really [hacky stuff](https://github.com/cognitedata/reveal/blob/master/documentation/src/hooks/useBaseTag.tsx#L10) to make it work in our docusaurus based docs site.

Should we maybe update the `README.md` or something with information on how to do this local debugging?

maksnester:
> Not necessarily inline, but that it's hosted somewhere the user chooses (e.g. for a native Electron app it's nice if all code is hosted locally).

@larsmoa No, there is no such option, unfortunately. I tried and found a workaround to replace the publicPath in main thread, but it won't work for worker. And as you said in slack, an electron app will need to access network to fetch models anyway, so it's not something that should be a blocker for an electron app. 

>  update the README.md
sure, will add some info on that.

larsmoa:
@maksnester : last question, is adding an alternative package with everything embedded, but that requires the copy-step as an alternative an option for later? I.e. that we provide two different packages depending on the requirements from the user.

maksnester:
> alternative package with everything embedded

@larsmoa 
A quick note on embedding just to clarify it:
In our current setup and how I left it for dev examples - we don' embed anything. Our bundle just has the publicPath set to `''`. This means that resources are requested relatively to current webpage's URL. Our bundle requests worker/wasm like `publicPath + workerName'. To make it work we started to use `<base>` tag which changes behavior of relative links (that why it breaks anchor links when used). So instead of fetching our workers from current page url, with the base tag it's fetched from the path specified in `<base>` tag. 

Back to your question, I think it should be possible to provide an alternative bundle that requires copying. But as I imagine it now, it would be some different entry point, so user will have to do something like
``` 
import { Cognite3DViewer } from '@cognite/reveal/local-entry' // or whatever name
```

PR #905, opened 2020-08-07
refactor(SupportedModelTypes): use type union, remove NotSupported

maksnester:
Replaced `SupportedModelTypes` enum with type union to avoid strings that mimic enum values in revealManager.addModel. Also removed NotSupported `''` from that type because it doesn't really make sense to have it in type that declares supported stuff.

### DIFF
--- a/viewer/src/public/RevealManager.ts
+++ b/viewer/src/public/RevealManager.ts
@@ -19,6 +19,7 @@ import { PotreeGroupWrapper } from '@/datamodels/pointcloud/PotreeGroupWrapper';
 import { PotreeNodeWrapper } from '@/datamodels/pointcloud/PotreeNodeWrapper';
 import { RenderMode } from '@/datamodels/cad/rendering/RenderMode';
 import { EffectRenderManager } from '@/datamodels/cad/rendering/EffectRenderManager';
+import { SupportedModelTypes } from '@/public/migration/types';
 
 export class RevealManager<TModelIdentifier> {
   private readonly _cadManager: CadManager<TModelIdentifier>;
@@ -161,7 +162,7 @@ export class RevealManager<TModelIdentifier> {
     modelIdentifier: TModelIdentifier
   ): Promise<[PotreeGroupWrapper, PotreeNodeWrapper]>;
   public async addModel(
-    type: 'cad' | 'pointcloud',
+    type: SupportedModelTypes,
     modelIdentifier: TModelIdentifier,
     nodeApperanceProvider?: NodeAppearanceProvider
   ): Promise<[PotreeGroupWrapper, PotreeNodeWrapper] | CadNode> {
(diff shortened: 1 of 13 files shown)


viewer/src/public/RevealManager.ts
```diff
@@ -19,6 +19,7 @@ import { PotreeGroupWrapper } from '@/datamodels/pointcloud/PotreeGroupWrapper';
 import { PotreeNodeWrapper } from '@/datamodels/pointcloud/PotreeNodeWrapper';
 import { RenderMode } from '@/datamodels/cad/rendering/RenderMode';
 import { EffectRenderManager } from '@/datamodels/cad/rendering/EffectRenderManager';
+import { SupportedModelTypes } from '@/public/migration/types';
```
j-bjorne, line 22: It would be better if you made the types part of utilities or datamodels. I believe we try and limit the import dependency paths. (I think the rules were:
datamodels can import from utilities but not the other way around.
public can import from utilities and datamodels but not the other way around.

maksnester: @j-bjorne Do you mean just move all the types under utilities? I don't think it's actually a good idea because we will lose that type in the documentation then. Now it's [documented here](https://cognitedata.github.io/reveal/docs/API%20Reference#enums_migration_types_supportedmodeltypesmd)

Utilities folder is ignored from docs. Because otherwise there would be tons of stuff from utils that also will pop up into the API reference.

eXigiouS: Then add it under datamodel/common or something, or datamodel/core some folder to indicate its shared state between datamodels I guess. I just know we made a choice a few weeks back and determined that we would have import restrictions, and I believe this import is breaking that restriction.

`src/datamodels/{common/core/shared}/types.ts` or something could contain this information instead.

j-bjorne: Then add it under datamodel/common or something, or datamodel/core some folder to indicate its shared state between datamodels I guess. I just know we made a choice a few weeks back and determined that we would have import restrictions, and I believe this import is breaking that restriction.

src/datamodels/{common/core/shared}/types.ts or something could contain this information instead.

maksnester: @j-bjorne OK, I moved `SupportedModelTypes` under datamodels, but had to reexport them in public/types to keep in the API reference.

viewer/src/public/types.ts
```diff
@@ -7,6 +7,12 @@ import { SectorGeometry } from '@/datamodels/cad/sector/types';
 import { SectorQuads } from '@/datamodels/cad/rendering/types';
 import { SectorCuller } from '@/internal';
 
+// we use these types in public API so they should be reexported here
+// to appear in the api reference docs
+export { CadRenderHints } from '../datamodels/cad/rendering/CadRenderHints';
```
j-bjorne, line 12: minor nitpick.. `@/datamodels`, but seeing as they are exported to experimentals it could be `@/experimental`. Anyways, that can be apart from this pr.

maksnester: Hm... Looks like typedoc can't resolve `@` alias. Will try to figure out on it a bit later then.

PR #913, opened 2020-08-10
docs: add tsdoc for Cognite3DModel and PointCloudModel

maksnester:
While reading through the code, I noticed a few things that probably should be fixed.
Please comment if you agree on that, and I'll submit the changes later:

Methods from the Cognite3DModel:

* `getBoundingBox` === `getModelBoundingBox` -> I suggest to deprecate the first one. And have the implementation under the second one to avoid usage of the deprecated method.
* `getNodeColor` - throws `NotSupportedInMigrationWrapperError` exception that user will never see. Should not throw it at all, or avoid of catching it locally and instead really throw it outside. But I guess that not throwing it and returning the fallback value is fine
* `updateNodeIdMaps` - unused? Why do we need this?


viewer/src/public/migration/CognitePointCloudModel.ts
```diff
@@ -11,12 +11,18 @@ import { PotreeGroupWrapper } from '@/datamodels/pointcloud/PotreeGroupWrapper';
 import { PotreePointColorType } from '@/datamodels/pointcloud/types';
 import { Matrix4 } from 'three';
 
+/**
+ * Documentation for the Cognite3DModel class
```
maksnester, line 15: @j-bjorne @larsmoa @christjt I'm not really sure what to write as a general overview for this class and for the cad model class as well. Would be happy to get some suggestions on it.

viewer/src/public/migration/Cognite3DModel.ts
```diff
@@ -133,7 +133,7 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
   }
 
   /**
-   * Used to clean up memory.
+   * Cleans up used resources.
```
maksnester, line 143: BTW I haven't found any usage of it in reveal manager only in migration wrapper

cc @larsmoa

viewer/src/public/migration/Cognite3DModel.ts
```diff
@@ -361,7 +361,6 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
    * Highlight node by node ID.
    * This method is async because node ID might be not loaded yet.
    * @deprecated {@link Use Cognite3DModel.selectNodeByTreeIndex}
-   * @see {@link Cognite3DModel.selectNodeByTreeIndex}
```
maksnester: Why remove it? @larsmoa

maksnester: Oh... I see

PR #914, opened 2020-08-10
Revert three to 117 as there is a bug that causes missing geometry

christjt:


viewer/package.json
```diff
@@ -71,7 +71,7 @@
     "raw-loader": "^4.0.1",
     "rimraf": "^3.0.2",
     "run-script-os": "^1.1.1",
-    "three": "^0.119.1",
+    "three": "^0.117.1",
```
maksnester, line 74: I guess it should be fixed to 0.117.1 i.e. without `^`

PR #922, opened 2020-08-11
chore: bump @cognite/sdk

larsmoa:


viewer/src/public/migration/Cognite3DModel.ts
```diff
@@ -224,6 +224,14 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
       throw new Error('NodeId not found');
     }
     const boundingBox3D = response[0].boundingBox;
+    if (boundingBox3D === undefined) {
+      trackError(new Error(`Node ${nodeId} doesn't have a defined bounding box, returning model bounding box`), {
+        moduleName: 'Cognite3DModel',
+        methodName: Cognite3DModel.prototype.getBoundingBoxFromCdf.name
```
maksnester, line 230: I would avoid using the name here. The resulting code might be minified/uglified. String feels safer. And shorter :)

viewer/src/utilities/networking/HttpHeadersProvider.ts
```diff
@@ -2,7 +2,7 @@
  * Copyright 2020 Cognite AS
  */
 
-import { HttpHeaders } from '@cognite/sdk/dist/src/utils/http/basicHttpClient';
+export type HttpHeaders = { [key: string]: string };
```
maksnester, line 5: https://github.com/cognitedata/cognite-sdk-js/blob/a569720c188d3413f46b997f1d17b6efaf7ff65f/packages/core/src/index.ts

Should we just add sdk-core into dev deps and import HttpHeaders from it?

viewer/src/public/migration/Cognite3DModel.ts
larsmoa: Good point 👍

viewer/src/utilities/networking/HttpHeadersProvider.ts
larsmoa: I'm not sure - the reason I switched away from importing was that I thought that it might break compatibility rules, i.e. that internal files might be moved around. Or is it something that is public and that we can consider stable?

maksnester: I guess it's public and should be just available directly from sdk-core as `import { HttprHeaders } from '@cognite/sdk-core'`

larsmoa: Oh, is sdk-core something new? Will update

PR #923, opened 2020-08-11
fix(examples): use model ids from .env file

maksnester:


larsmoa:
Great if you provide a short howto on Slack so everyone is aware on how to use this :)

PR #932, opened 2020-08-12
Migration Progress

j-bjorne:
Added progress to migration wrapper with current placeholder.


maksnester:
@j-bjorne Please add a simple note to the migration guide - it has some hidden notes. Like: 
`* use <callbackname> instead of onProgress/onComplete.`

PR #956, opened 2020-08-18
doc: add live examples for coloring CAD models

larsmoa:
TODO: 
- Remove need for boilerplate for each snippet
- Replace "Result" panel under each snippet with a Run-button
- Split layout into two panels such that the running example is visible regardless of which snippet is focused


documentation/src/components/CodeBlock.js
```diff
@@ -0,0 +1,49 @@
+/* eslint react/jsx-key: 0 */
+import React from 'react'
+import Highlight, {defaultProps} from 'prism-react-renderer'
+import {LiveProvider, LiveEditor, LiveError, LivePreview} from 'react-live'
+import {mdx} from '@mdx-js/react'
+
+export default ({children, className, live, render}) => {
```
maksnester, line 7: What were your changes in that swizzled file? @larsmoa

documentation/src/components/Layout.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import {MDXProvider} from '@mdx-js/react'
+import {Container, baseStyles} from 'unified-ui'
```
maksnester, line 3: Are these your changes for that component? What's the unified-ui?

documentation/src/components/CodeBlock.js
larsmoa: Oops, unnecessary file. Removed.

larsmoa:
> > Remove need for boilerplate for each snippet
> > Replace “Result” panel under each snippet with a Run-button
> 
> These 2 will be connected I guess. Can’t say whether it would be hard or not, it depends on implementation. If we'll keep using `window.viewer` it shouldn’t be very hard I guess.

I was kinda hoping it would be easy to extract the `sdk`, `viewer` and `model` from the component state - not so? I guess an alternative is to inject them into `DemoWrapper` instead, to avoid the need to store them on `viewer`?

maksnester:
> I was kinda hoping it would be easy to extract the sdk, viewer and model from the component state - not so? I guess an alternative is to inject them into DemoWrapper instead, to avoid the need to store them on viewer?

Well, yes, putting state upper shouldn't be a problem. Though I don't see yet what it supposed to solve. 🤔

larsmoa:
> > I was kinda hoping it would be easy to extract the sdk, viewer and model from the component state - not so? I guess an alternative is to inject them into DemoWrapper instead, to avoid the need to store them on viewer?
> 
> Well, yes, putting state upper shouldn't be a problem. Though I don't see yet what it supposed to solve. 🤔

Just felt wrong putting the model and viewer in the global viewer state, but I don't have a big problem with it.

christjt (approved):

documentation/docs/examples/cad-colors.mdx
```diff
@@ -0,0 +1,54 @@
...
+<LiveCodeSnippet>
+{`
+model.resetAllNodeColors();
+`}
+</LiveCodeSnippet>
+
+### Update the colors of a subtree
+The 3D nodes are organized in a tree structure, where e.g. equipment will have many sub-components. In 
+order to update the colors of equipment it's typically necessary to apply the same color to the entire
+subtree. Reveal supports this through the use of `Cognite3DModel.setNodesColorsByTreeIndex` with `applyToChildren`
```
maksnester, line 48: TBH I can't notice how it works. The model itself has some colors already, and when I click run, I have no idea where my subtree is and what changed. Perhaps we should just add here this line: 
```js
model.iterateNodesByTreeIndex(treeIndex => {
  model.setNodeColorByTreeIndex(treeIndex, 127, 127, 127);
});
```
 to make setNodeColorByTreeIndex visible

I tried to do that but still can't see the subtree that changed 🤷‍♂️  Maybe we can to point camera on that subtree somehow?

maksnester:
We can use beta-1 in docs. Otherwise, these guys won't work because of the <base> tag

Screenshot 2020-08-19 at 13 08 43

documentation/docs/concepts.md
```diff
@@ -0,0 +1,18 @@
+---
+id: concepts
+title: Concepts
+---
+
+This page explains some basic concepts implemented by Reveal.
+
+# CAD model structure
```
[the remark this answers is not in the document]
larsmoa: I now use Grammarly :)

Noted about # being used for title, I thought it was something "above" #

documentation/docs/examples/Cognite3DViewerDemo.tsx
```diff
@@ -23,13 +24,15 @@ export default function Cognite3DViewerDemo({ client }: DemoProps) {
       sdk: client,
       domElement: canvasWrapperRef.current,
     });
-    addModel({ modelId: 5641986602571236, revisionId: 5254077049582015 });
+    addModel({ modelId, revisionId });
 
     async function addModel(options: AddModelOptions) {
       const model = await viewer.addModel(options);
       viewer.fitCameraToModel(model);
+      (window as any).model = model;
```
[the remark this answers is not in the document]
larsmoa: Aha, thx

PR #960, opened 2020-08-20
Added functional LoadingState to update pipeline

j-bjorne:
Now provides loading state instead of outputing 0, 1 when loading items from network


larsmoa:
Doesn't seem to work. I get an infinite loading indicator that never completes. See my comment on `RxCounter`

viewer/src/utilities/RxCounter.ts
```diff
@@ -48,20 +53,29 @@ export class RxCounter {
...
+              loadingState.itemsRequested = 0;
+              break;
+            default:
+              throw new Error(`Unsupported action ${action}`);
+          }
+          return loadingState;
+        },
+        { itemsLoaded: 0, itemsRequested: 0 } as LoadingState
+      ),
+      throttleTime(200, asyncScheduler, { trailing: true }),
```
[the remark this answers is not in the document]
j-bjorne: I changed it in a commit I did from home, but since I was running for the bus it seems it failed and I didn't catch it. Wasn't a lot of change, and I remember what I did so I'll sort it out quickly. The throttle is not needed there, but it is nice to have a throttle, since say the count goes up super fast, you don't need all the emissions, what you are really interested in is the last emission, and not necessarily all the emissions.

viewer/src/utilities/RxTaskTracker.test.ts
```diff
@@ -0,0 +1,100 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+import { RxTaskTracker, TaskTracker } from './RxTaskTracker';
+import { interval, queueScheduler } from 'rxjs';
+import { take, flatMap, observeOn } from 'rxjs/operators';
+
+describe('RxCounter', () => {
```
larsmoa, line 9: Does not match class name

j-bjorne: It haz been renamed

larsmoa:
Looks much better, but I notice a couple of things. When reporting progress, lots of events are triggered for progress, e.g. `0/100`, `14/100`, `35/100`, `80/100`. However, the last progress event fired will always be `0/0` which is very confusing and is prone to bugs on the user end (e.g. division by zero). Can we ensure that `100/100` is sent instead of `0/0`?

Also, I'm noticing that several sectors are loaded twice:
image
(note! There should be two requests per sector, one for `OPTIONS` and one for `GET`).
Not sure if that is caused by this fix, but something that needs to be addressed at least.

j-bjorne:
Well progress is not a %. It is 50 out of 135 files etc. The sectors loaded seems weird, and I should look into it. Can you verify that this is not also in master. Also it seems your list is sorted by name. It will fetch sectors multiple time if they are out of the cache, but I am making assumptions here.

larsmoa:
The issue with sectors being loaded multiple times (after moving the camera _once_) occurs in master too, so we just need to create (and fix) a separate issue for that.

As for the `loadingStateChanged` going from `99/100` to `0/0` I still think it would be a lot better if it wasn't reset to 0 when it completed.

j-bjorne:
@larsmoa added a filter to exclude the 0:0 event while still maintaining the reset logic of the counter.

larsmoa (commented):
> @larsmoa added a filter to exclude the 0:0 event while still maintaining the reset logic of the counter.

@j-bjorne: But how will this affect `OnLoadingCallback`? Will there ever be an event for `100/100`, i.e. when loading has completed?

viewer/src/utilities/RxTaskTracker.ts
```diff
@@ -0,0 +1,87 @@
...
+              break;
+            default:
+              throw new Error(`Unsupported action ${action}`);
+          }
+          return taskTracker;
+        },
+        { taskCount: 0, taskCompleted: 0 } as TaskTracker
+      ),
+      // TODO: 24-08-2020 j-bjorne: figure out a way to get last emission in a time window
+      // throttleTime(200, asyncScheduler, { trailing: true }),
```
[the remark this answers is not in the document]
j-bjorne: Sure, I can remove it

j-bjorne:
1/1 is loading complete currently. Commit 630260c is the one that maps 0,0 to 1,1 instead. So we no longer filter it away but remap it. For interrnal counter logic it remains 0/0

larsmoa:
> 1/1 is loading complete currently. Commit [630260c](https://github.com/cognitedata/reveal/commit/630260c17b12900605ddd28ece64b550916c5b75) is the one that maps 0,0 to 1,1 instead. So we no longer filter it away but remap it. For interrnal counter logic it remains 0/0

and fixing so it reports 100/100 instead (or whatever the totalt count), is hard?

j-bjorne:
no, but 100/100 doesn't make sense. We are reporting requests / total. 100/100 seems like a progress thing.

j-bjorne:
If you want we can have it just report progress which is a number from 0 to 100 :)

larsmoa:
I mean completed/total, so the last progress report will be total/total

larsmoa:
@j-bjorne: if easy, change such that last reported progress is `total/total`. If not, just change progress to a number between 0 and 1.

j-bjorne:
The problem is that changing it to total/total is wrong, because when camera moves that is when it resets. And it resets mid load. So say you are at 102/184 in progress, when you move the camera and stop loading the next ones, what is the right value to emit as reset. This is why we used 0 / 0. I mean, I get the 0 division error, but one can also do something as simple as have a helper function like this:

```
function getProgress(completed, total) {
  return total > 0 ? total : completed / total;
}
```

larsmoa:
I realize that its not really progress, but that the counter goes from 88/89 to 1/1, or worse, 0/0 is indeed a bit confusing. Imagine if there is some UI component that is based of the progress that uses the actual count to e.g. determine something about how long it will take to load the sectors. This will probably break down if we report 1/1 or 0/0. That's why I think we should either report `total/total` or switch to a single number between 0 and 1.

examples/src/pages/SimplePointcloud.tsx
```diff
@@ -182,7 +182,7 @@ export function SimplePointcloud() {
 
   return (
     <CanvasWrapper>
-      <Loader isLoading={isLoading} style={{ position: 'absolute' }}>
+      <Loader isLoading={loadingState.itemsLoaded !== loadingState.itemsRequested} style={{ position: 'absolute' }}>
```
maksnester, line 185: From my perspective, it's become harder to use, to be honest. 

Maybe we should have some simple loading state as we had which just indicates if something is loading at the moment? And in addition, have that one with progress? 

I just think that it's a very typical use-case to know is something loading or not and it should be solved as simple as possible. But drawing some things about progress is more advanced usage which shouldn't replace or make basic use-case more difficult. 

So I would suggest having another event for progress like `on('loadingProgressChanged', progressObject)` in addition to `loadingStateChanged` which returns boolean. Also, in that way we can keep API similar to the Cognite3DViewer which is not a bad thing.

maksnester: > counter goes from 88/89 to 1/1, or worse, 0/0

I think it's fine, to be honest. We don't provide entire progress functional here, right? Also, that information might be used in UI if you want to display it. Or if you don't you can implement that percentage stuff manually. Or we might extend loading state object to have that. E.g. every event will have:

```ts
export type LoadingState = {
  itemsLoaded: number;
  itemsRequested: number;
  progress: number; // 0 - 100 based on `itemsLoaded / itemsRequested`. It's a dummy but a convenient thing. 
};
```

maksnester: @larsmoa @j-bjorne gently nudging you to agree or disagree with that suggestion.

j-bjorne: hehe, just reloaded the page and saw your comments. I think we went for
```
export interface LoadingState {
  isLoading: boolean;
  itemsLoaded: number;
  itemsRequested: number;
}
```

j-bjorne: That way you can use only the isLoading if you want to, and then if you want the details you use itemsLoaded and itemsRequested

larsmoa (approved):

PR #969, opened 2020-08-24
doc: add example for adding HTML overlays

larsmoa:


Mattman22 (approved):
Looks good, just need to remember to add the custom 3D example, but there's already a Miro task for that, so should be fine.

PR #975, opened 2020-08-24
doc: reduce the number of documented modules in the API reference

larsmoa:


maksnester (approved):
LGTM. I had some issues with config file for docs that why I put that very long line into config. IDK what was the reason but output wasn't correct and felt like some options are just ignored. Maybe some Windows related issue or I just screwed up somewhere at that time.

PR #980, opened 2020-08-25
fix: issue causing worldToScreen() returning wrong values when camera is being moved

larsmoa:
This fixes an issue causing 2D HTML overlays to be appearing to "lag behind" when they their position was updated when Cognite3DViewer triggered 'cameraChanged'.

### DIFF
--- a/documentation/docs/examples/cad-colors.mdx
+++ b/documentation/docs/examples/cad-colors.mdx
@@ -51,6 +51,6 @@ subtree. Reveal supports this through the use of `Cognite3DModel.setNodesColorsB
 set to `true`:
 <LiveCodeSnippet>
 {`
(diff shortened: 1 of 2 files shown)


documentation/docs/examples/cad-colors.mdx
```diff
@@ -51,6 +51,6 @@ subtree. Reveal supports this through the use of `Cognite3DModel.setNodesColorsB
 set to `true`:
 <LiveCodeSnippet>
 {`
-model.setNodeColorByTreeIndex(122, 240, 100, 60, true /* apply to children */);
+model.setNodeColorByTreeIndex(122, 240, 100, 60, applyToChildren=true);
```
larsmoa, line 54: Unrelated change which I accidentally added to the commit, but I want to change this so I'll leave it.

Mattman22 (approved):

PR #981, opened 2020-08-25
chore: extract parser-worker as separated package

maksnester:
* worker and rust now a dependency for viewer and will live in separate package called `@cognite/reveal-parser-worker`. It's inside the `parser-worker` folder.
* worker's build actually produces the same output twice, one for CDN with correct CDN publicPath, and another for local needs like copying in environments under CSP.
* When CDN is unavailable or you need local files in examples - you can just specify the path that should be used like:

```js
// if you want to test your latest changes in workers or rust files
// copy your worker files to some folder in /public and specify the path below
import { revealEnv } from '@cognite/reveal';
revealEnv.publicPath = `https://localhost:8080${process.env.PUBLIC_URL}/worker-custom-folder/`;
```

but that will work only in case if a project allows hosting of static files on the same domain. If it's not, then it's only possible to build parser-worker from sources and put these built files to someplace that used for static files (that CDF case, unfortunately).

I will add more README stuff about local usage tomorrow. It's a bit cumbersome, but not impossible to use. Installation docs also will be here.


maksnester:
Regarding failed build: 

> @cognite/reveal-parser-worker

is not published, so once it will - build errors should go away.

viewer/package.json
```diff
@@ -81,11 +80,11 @@
     "webpack-dev-server": "^3.11.0",
     "webpack-log": "^3.0.1",
     "webpack-node-externals": "^2.5.1",
-    "whatwg-fetch": "^3.4.0",
-    "worker-plugin": "^5.0.0"
+    "whatwg-fetch": "^3.4.0"
   },
   "dependencies": {
     "@cognite/potree-core": "^1.1.3",
+    "@cognite/reveal-parser-worker": "link:../parser-worker",
```
maksnester, line 87: Before merging this PR I'm going to publish parser-worker as 1.0.0-beta-1 and change this line. Then merge this branch to master.

maksnester:
@larsmoa merge?

larsmoa:
@maksnester: Sorry this got hanging around

PR #986, opened 2020-08-26
feat: getNodeColor is now deprecated and will throw error

larsmoa:


viewer/src/public/migration/Cognite3DModel.ts
```diff
@@ -290,40 +290,12 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
...
-        moduleName: 'Cognite3DModel',
-        methodName: 'getNodeColor'
-      });
-      return {
-        r: 255,
-        g: 255,
-        b: 255
-      };
-    }
+  getNodeColor(_nodeId: number): Promise<Color> {
```
maksnester, line 297: But there is still setNodeColor is it ok?

larsmoa: Yes, I think so. We _might_ want to rename it to `setNodeOverrideColor` (and could have a getter for it)

larsmoa: The reason for removing `getNodeColor` is that it gives the impression that we know what color a node has - we dont, since we potentially haven't loaded geometry for the node yet.

maksnester (approved):
I just concerned that we have unpaired methods. Feels weird to me that I can set a color by nodeid, but can't get the color by nodeid 🤷‍♂️   

But if you think it's fine then ok. I don't really have any strong arguments against that.

maksnester:
@larsmoa BTW, in docs you mentioned:

> - `Cognite3DModel.setNodeColor(nodeId, r, g, b)` has been replaced by `setNodeColorByTreeIndex(treeIndex, r, g, b, applyToChildren?)`

So perhaps we should deprecate it too or rephrase docs.

PR #1010, opened 2020-08-31
doc: add example for adding custom 3D content to the scene

larsmoa:


documentation/docs/examples/cad-3dobjects.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+:::tip
+Reveal uses an "implicit lighting model", i.e. no lights are added to the scene. If you need lights this needs to be added to the viewer scene. 
+:::
+
+<DemoWrapper modelId={4715379429968321} revisionId={5688854005909501} name="Cognite3DViewerDemo" />
+
+## Adding markers to the scene on clicks
+
+The following examples reacts to clicks in the scene and adds markers to the positions clicked.
```
maksnester, line 23: looks like a typo here. should either be "example reacts adds" or "examples react add". Consider using google docs to catch some typos like "lets/let's".

documentation/docs/examples/cad-3dobjects.mdx
```diff
@@ -0,0 +1,120 @@
...
+`}
+</LiveCodeSnippet>
+
+## Add environment
+
+To make a model feel more natural it might be useful to add an environment. The following examples adds sky and a sea effect for the example oil rig. 
+
+First, lets encapsulate the model in a large sky-textured sphere to create an illusion of skies. The sphere map is available from [Free HDR - EXR SKIES-blog](http://freepanorama.blogspot.com/).
+
+<LiveCodeSnippet scope={{skyUrl: useBaseUrl('/img/sky007.jpg')}}>
```
maksnester, line 46: That's just an opinion and you can ignore it if you want, but I think not every example must be interactive. I think it's even better if you have a complete demo in the beginning or at the end of the page with some static code snippets that show how to achieve that state. It's just might not be obvious (or convenient) to a user to click "run" on every code snippet to get the cool final state with the sky and sea.  

This article looks like a tutorial, so perhaps it's better to just illustrate each step with a picture rather than expect that a user will click "run" and play with the model after each step.

documentation/docs/examples/cad-3dobjects.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+This looks a bit better, but the oil rig appears to be floating mid-air at this point. Lets add an ocean. A plane is rendered at the sea level and an ocean effect added to the plane. The example builds upon [the ocean example from ThreeJS](https://threejs.org/examples/?q=ocea#webgl_shaders_ocean).
+
+Also note that since the ocean has moving waves we want to render the model continuously and not only when camera changes or data is loaded. We do this by repeatedly calling `Cognite3DViewer.forceRerender()`.
+
+<LiveCodeSnippet scope={{Water}}>
+{`
+// The following has been imported outside this scope:
+// import { Water } from 'three/examples/jsm/objects/Water';
+\
```
maksnester, line 75: notice that in that way only in source code there will be a space between lines. You probably want it in examples too. For that just add `\n` to every line before `\`. So should be like

```diff
- // import { Water } from 'three/examples/jsm/objects/Water';
+ // import { Water } from 'three/examples/jsm/objects/Water';\n
\
```

documentation/docs/examples/cad-3dobjects.mdx
```diff
@@ -0,0 +1,120 @@
...
+        }
+      ),
+    waterColor: 0x442e1f,
+  }
+);
+water.rotation.x = - Math.PI / 2;
+viewer.addObject3D(water);
+\
+function updateOcean() {
+  var time = performance.now() * 0.001;
```
maksnester, line 100: Should it be here?

documentation/docs/examples/cad-3dobjects.mdx
```diff
@@ -0,0 +1,120 @@
...
+    waterNormals: new THREE.TextureLoader()
+      .load('https://threejs.org/examples/textures/waternormals.jpg', 
+        texture => {
+          texture.wrapS = texture.wrapT = THREE.RepeatWrapping;
+        }
+      ),
+    waterColor: 0x442e1f,
+  }
+);
+water.rotation.x = - Math.PI / 2;
```
maksnester, line 95: Might not be obvious why it should be rotated here. Perhaps we should have a note/comment on that here.

documentation/docs/examples/cad-3dobjects.mdx
```diff
@@ -0,0 +1,120 @@
...
+{`
+const boundingSphere = new THREE.Sphere(new THREE.Vector3(0,0,0), 2000);
+viewer.on('cameraChange', (position, target) => {
+  // Keep camera above sea level
+  if (position.y < 20) {
+    position.y = 21;
+    viewer.setCameraPosition(position);
+  }
+\
+  // Clamp point to be within bounding sphere
```
maksnester, line 124: Well, that might be just me, but I don't understand that part about clamp point. What are the clamp point and the bounding sphere? Some explaining link on 3d basics might be good to have nearby.

documentation/docs/examples/cad-3dobjects.mdx
maksnester: typo is still here 

> should either be "example reacts adds" or "examples react add".

documentation/docs/examples/cad-3dobjects.mdx
maksnester: ⏫

maksnester (approved):
LGTM

documentation/docs/examples/cad-3dobjects.mdx
larsmoa: I've split into static + interactive example

documentation/docs/examples/cad-3dobjects.mdx
larsmoa: Thx, fixed @maksnester

PR #1011, opened 2020-08-31
doc: fix unused variable in code example

larsmoa:


documentation/docs/examples/cad-3dobjects.mdx
```diff
@@ -96,8 +96,8 @@ water.position.copy(modelCenterAtSeaLevel);
 viewer.addObject3D(water);
 
 function updateOcean() {
-  var time = performance.now() * 0.001;
-  water.material.uniforms[ 'time' ].value += 1.0 / 60.0;
+  const time = performance.now() * 0.001;
+  water.material.uniforms[ 'time' ].value = time;
```
maksnester, line 100: That doesn't look obvious to me too. I have no idea what is uniforms and why it has time and why it's ok to take performance.now() * 0.001 as value for that. So maybe worth to comment on it as well.

maksnester (approved):

PR #1017, opened 2020-08-31
update release 1.0 branch with current beta-3 state

maksnester:


maksnester:
I have no idea at what point branches became diverged and I don't like the process of resolving these conflicts. 

Wouldn't it be enough to just tag releases in master? I don't really understand for what purpose we need that release branch now.

PR #1023, opened 2020-09-01
feat: add Cognite3DViewer.loadCameraFromModel

larsmoa:
Note that example cad-basic is broken untill new version will be released as it uses new functionality.


viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -647,6 +647,31 @@ export class Cognite3DViewer {
...
+   * provided model. See {@link https://docs.cognite.com/api/v1/#operation/get3DRevision}
+   * and {@link https://docs.cognite.com/api/v1/#operation/update3DRevisions} for
+   * information on how this settings is retrieved and stored. This setting can
+   * also be changed through the 3D models management interface in Cognite Fustion.
+   * @param model The model to load camera settings from.
+   * @returns True if a camera was stored for the provided model and applied to the
+   * current camera settings.
+   * @example
+   * ```js
+   * if (!viewer.loadCameraFromModel(model)) {
```
maksnester, line 661: Can you imagine a use case when a user wants to handle that in some different way? I'm just thinking maybe we should use that fallback internally by default, without forcing users to write all the same code every time. Like 

```js
  loadCameraFromModel(model: CogniteModelBase): void {
    const config = model.getCameraConfiguration();
    if (config) {
      this.controls.setState(config.position, config.target);
    } else {
      this.fitCameraToModel(model)
   }
  }
```

viewer/src/public/migration/CogniteModelBase.ts
```diff
@@ -12,6 +13,13 @@ export interface CogniteModelBase {
   readonly type: SupportedModelTypes;
   dispose(): void;
   getModelBoundingBox(outBbox?: THREE.Box3): THREE.Box3;
+  /**
+   * Retrieves the camera position and target stored for the model. Typically this
+   * is used to store a good starting position for a model. Returns `undefined` if there
+   * isn't any stored camera configuration for the model.
+   */
+  getCameraConfiguration(): CameraConfiguration | undefined;
```
maksnester, line 21: I'm a bit unsure if we should have any documentation in the base model at all (or contrary in implementations). I just don't like that you have to copy-paste things from an interface to its implementations N times. Maybe we just should document things only once here in the interface.

viewer/src/utilities/networking/CdfModelDataClient.ts
```diff
@@ -90,6 +90,23 @@ export class CdfModelDataClient
...
   }
 
+  public async getModelCamera(modelIdentifier: {
+    modelId: number;
+    revisionId: number;
+    format: File3dFormat | string;
+  }): Promise<CameraConfiguration | undefined> {
+    const { modelId, revisionId } = modelIdentifier;
+    const model = await this.client.revisions3D.retrieve(modelId, revisionId);
+    if (model.camera !== undefined && model.camera.position !== undefined && model.camera.target !== undefined) {
```
maksnester, line 100: I understand that probably we have a stable API which either sets a value or undefined, but still if it doesn't really matter, I suggest checking for truthness here, not undefined. Because if API will send null instead of undefined that condition will be incorrectly evaluated as true because `null !== undefined` is true.

```suggestion
    if (model.camera && model.camera.position && model.camera.target) {
```

also it's just more readable.

examples/src/pages/Migration.tsx
```diff
@@ -71,7 +71,13 @@ export function Migration() {
       async function addModel(options: AddModelOptions) {
         try {
           const model = await viewer.addModel(options);
-          viewer.fitCameraToModel(model);
+          const cameraConfig = model.getCameraConfiguration();
+          if (cameraConfig) {
```
maksnester, line 75: You don't use your new method here by the way.

viewer/src/public/migration/Cognite3DViewer.ts
larsmoa: I was thinking about that and I was a bit unsure if that would be better. I chose the current approach as it's more flexible and that it allows the user to handle this themselves, but I dont have any really good arguments against doing what you propose... Not really sure. Any input @christjt @j-bjorne ?

viewer/src/public/migration/CogniteModelBase.ts
larsmoa: Removing for now.

examples/src/pages/Migration.tsx
larsmoa: Thx, forgot to update.

viewer/src/public/migration/Cognite3DViewer.ts
maksnester: Just to add an argument here, I think for rare cases when a user actually wants different behavior it's still can be achieved with model.getCameraConfiguration(); (as you did in examples before).

larsmoa: I'm convinced, I'll make the change.

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -647,6 +647,30 @@ export class Cognite3DViewer {
...
+   * information on how this settings is retrieved and stored. This setting can
+   * also be changed through the 3D models management interface in Cognite Fustion.
+   * If no camera configuration is stored in CDF, {@link Cognite3DViewer.fitCameraToModel}
+   * is used as a fallback
+   * @param model The model to load camera settings from.
+   * @returns True if a camera was stored for the provided model and applied to the
+   * current camera settings. False if no camera configuration was stored in CDF and
+   * {@link Cognite3DViewer.fitCameraToModel} is used instead.
+   */
+  loadCameraFromModel(model: CogniteModelBase): boolean {
```
maksnester, line 663: Cool that we have a fallback now but now it feels like fn does 2 things at once. I would simplify to just:

```suggestion
  loadCameraFromModel(model: CogniteModelBase): void {
    const config = model.getCameraConfiguration();
    if (config) {
      this.controls.setState(config.position, config.target);
    } else {
      this.fitCameraToModel(model, 0);
    }
  }
```

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -647,6 +647,47 @@ export class Cognite3DViewer {
...
+  }
+    const config = model.getCameraConfiguration();
+    if (config) {
+      this.controls.setState(config.position, config.target);
+      return true;
+    } else {
+      this.fitCameraToModel(model, 0);
+      return false;
+    }
+  }
```
maksnester, line 689:
```suggestion
  /**
   * Attempts to load the camera settings from the settings stored for the
   * provided model. See {@link https://docs.cognite.com/api/v1/#operation/get3DRevision}
   * and {@link https://docs.cognite.com/api/v1/#operation/update3DRevisions} for
   * information on how this setting is retrieved and stored. This setting can
   * also be changed through the 3D models management interface in Cognite Fusion.
   * If no camera configuration is stored in CDF, {@link Cognite3DViewer.fitCameraToModel}
   * is used as a fallback
   * @param model The model to load camera settings from.
   */
  loadCameraFromModel(model: CogniteModelBase): void {
    const config = model.getCameraConfiguration();
    if (config) {
      this.controls.setState(config.position, config.target);
    } else {
      this.fitCameraToModel(model, 0);
    }
  }
```

maksnester (approved):

PR #1027, opened 2020-09-02
chore: organize tests in correct folders

larsmoa:
Note that I think we should evaluate how we organize the tests, but that we should keep to the current scheme until we make a firm decision to transition to 'inline' organized tests.


larsmoa:
Note that I think we should evaluate how we organize the tests, but that we should keep to the current scheme until we make a firm decision to transition to 'inline' organized tests.

maksnester:
@larsmoa Please remind me why we don't want to put test files right near the source files? It's so inconvenient to have a duplicated folder structure. The last time when I needed to add a test (I added a unit-tests for isTheSameDomain function) I dropped the test file near the source file and it just works. And it's obvious when a module has some tests or not. And it's easier to add new tests or find existing tests. And it would unlikely require changes as you have here in the future.

maksnester (approved):

larsmoa:
Totally agreed @maksnester, I think we should move away from separate folders ASAP (read: the next sprint). I just want to be sure this won't cause our binary to grow because tests are included.

maksnester:
> I just want to be sure this won't cause our binary to grow because tests are included.

Shouldn't be the case unless we import test files in our source code files (which doesn't make any sense to me and we probably don't)

PR #1028, opened 2020-09-02
fix(Cognite3DModel): async iteration over nodes

maksnester:
Motivation for that was [that example](https://cognitedata.github.io/reveal/docs/examples/cad-colors) which currently just freezes my browser if I click "run" in the first example.

Would it be better to do that iteration without freezing a browser? We can make it to be async and it will call an action ~10k times per event loop cycle. It still will decrease FPS but won’t completely freeze a browser.

### DIFF
--- a/viewer/src/__tests__/utilities/NumericRange.test.ts
+++ b/viewer/src/__tests__/utilities/NumericRange.test.ts
@@ -39,10 +39,10 @@ describe('NumericRange', () => {
     expect(range.contains(10)).toBeFalse();
   });
 
-  test('forEach applies callback to each value', () => {
+  test('forEach applies callback to each value', async () => {
     const applied: number[] = [];
     const range = new NumericRange(1, 4);
-    range.forEach(v => {
+    await range.forEach(v => {
       applied.push(v);
     });
     expect(applied).toEqual([1, 2, 3, 4]);
--- a/viewer/src/__tests__/utilities/callActionWithIndicesAsync.test.ts
+++ b/viewer/src/__tests__/utilities/callActionWithIndicesAsync.test.ts
@@ -1,0 +1,18 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+import { callActionWithIndicesAsync } from '@/utilities/callActionWithIndicesAsync';
+
+describe('test callActionWithIndicesAsync', () => {
+  it('calls an action with specified indices range', async () => {
+    const start = 10;
+    const end = 100000;
+    const timesToCall = end - start + 1;
+    const action = jest.fn();
+    await callActionWithIndicesAsync(start, end, action);
+    expect(action).toBeCalledTimes(timesToCall);
+    expect(action).toHaveBeenNthCalledWith(1, start);
+    expect(action).toHaveBeenNthCalledWith(1 + 1234, start + 1234);
+    expect(action).toHaveBeenNthCalledWith(timesToCall, end);
+  });
+});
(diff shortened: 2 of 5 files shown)


larsmoa (commented):
Smart. My browser sometimes freezes, other times not. It also makes more sense to have the function async since `iterateSubtreeByTreeIndex` is async. 
A couple of comments:
- Document that `iterateNodesByTreeIndex` now is async in migration guide
- Make `iterateSubtreeByTreeIndex` work similarly to `iterateNodesByTreeIndex`
- Document in the functions that the operation is incremental and that the change may occur gradually and not in one go (e.g. if changing the color of all nodes).

I think we might need a `setAllNodeColors` or some similar functionality to override all nodes as its a quite common operation. Let's not introduce that now, though

maksnester:
> Make iterateSubtreeByTreeIndex work similarly to iterateNodesByTreeIndex

@larsmoa could you elaborate on that? I thought I already did it (it's async under the hood).

But I actually not really understood how `iterateSubtreeByTreeIndex` works. Now and before my changes, it always gives me a single treeIndex when I tested with the code below and clicked the  mode from [demo](https://cognitedata.github.io/reveal/docs/examples/cad-colors#update-all-node-colors):

```
// Handle mouse clicks
viewer.on('click', event => {
  // Find 3D coordinate of what the user clicks
  const intersection = viewer.getIntersectionFromPixel(event.offsetX, event.offsetY);
  if (intersection) {
    model.resetAllNodeColors();
    model.iterateSubtreeByTreeIndex(intersection.treeIndex, (i) => {
      console.log(i);
      model.setNodeColorByTreeIndex(i, 123, 123, 123);
    })
  }
});
```

It colors (and prints) one specific treeIndex that I click (maybe it's expected behavior) I just thought that subtree should have more than one treeindex.

maksnester:
@larsmoa actually I just noticed that chrome freezes only with opened dev tools. 

> [Violation] 'click' handler took 37516ms

When devtools are closed iteration is finished quite fast. Checked firefox without devtools - some freeze is here but I think iteration takes ~5s, and when dev tools are opened takes longer ~10s but still not so bad as with chrome dev tools. 

Rechecked async iteration - also goes much faster with closed devtools in Chrome and still relatively slow for Firefox. 

So, anyway, it's potentially freezing operation and I think these changes make sense. 

And one more thing - you recently showed me a black screen on iOS. So I see the same in desktop safari, so it's probably not mobile issue, but Safari issue. We lost the support of it at some point :(

larsmoa:
> @larsmoa could you elaborate on that? I thought I already did it (it's async under the hood).
> 
> But I actually not really understood how `iterateSubtreeByTreeIndex` works.

`iterateSubtreeByTreeIndex` is async because it maps treeIndex -> nodeId in order to determine subtreeSize, but once the subtreeSize has been determine it processes all indices in one go, right? If you do `iterateSubtreeByTreeIndex` on the root node, it will behave equally to how `iterateNodesByTreeIndex` did before I belive.

larsmoa:
> And one more thing - you recently showed me a black screen on iOS. So I see the same in desktop safari, so it's probably not mobile issue, but Safari issue. We lost the support of it at some point :(

Good catch. I don't believe its a very serious issue but something we definitely need to look into. `viewer.getScreenshot()` works and produces an image, but it's very wide so I actually think its a page layout issue and not a Reveal issue. Disclaimer: I have not tried the standard examples, only docs.

maksnester:
@larsmoa 

> page layout issue and not a Reveal issue

Doesn't mean that it's not a Reveal issue though, because it's a totally fine responsive layout. 

Did some more tests and turned out it's yet another pixel ratio issue. Will pin some sticker in Miro on that.

maksnester:
> iterateSubtreeByTreeIndex is async because it maps treeIndex -> nodeId in order to determine subtreeSize, but once the subtreeSize has been determine it processes all indices in one go, right? If you do iterateSubtreeByTreeIndex on the root node, it will behave equally to how iterateNodesByTreeIndex did before I belive.

@larsmoa I still don't understand what changes you're suggesting regarding that... 

```js
    model.iterateSubtreeByTreeIndex(0, (i) => {
      console.log(i);
    })
```
currently prints only 0 and that's it. Both in that branch and in published docs. Does 0 treeIndex represent the rootNode?

I didn't really change any logic behind the iteration, just made that forEach thing to be async for the subtree, but all methods already were async.

larsmoa (approved):

PR #1037, opened 2020-09-03
chore: make jest config more forgiving to avoid lint errors during development

larsmoa:
- errors will still be catched by lint step

### DIFF
--- a/viewer/tsconfig.test.json
+++ b/viewer/tsconfig.test.json
@@ -1,35 +1,7 @@
 {
+  "extends": "./tsconfig.json",
   "compilerOptions": {
-    "incremental": true,
-    "target": "ES2019",
-    "module": "esnext",
-    "declaration": true,
-    "declarationDir": "dist",
-    "outDir": "dist-ts",
-    "rootDir": "src",
-    "esModuleInterop": true,
-    "downlevelIteration": true,
-    "sourceMap": true,
-    "moduleResolution": "node",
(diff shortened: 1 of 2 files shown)


viewer/tsconfig.test.json
```diff
@@ -0,0 +1,35 @@
...
+        "dom",
+        "esnext"
+    ],
+
+    "baseUrl": "./src/",
+    "paths": {
+      "@/*": ["*"]
+    }
+  },
+  // include is also used for eslint to understand which files should be linted
```
maksnester, line 32: that's not valid json file

maksnester (changes requested):

viewer/tsconfig.test.json
```diff
@@ -0,0 +1,35 @@
...
+  "include": ["*.js", "*.ts", "src/**/*", ".eslintrc.js", "script", "src/@types"],
+  "exclude": ["node_modules/**/*", "dist/**/*"]
+}
```
maksnester, line 35: Let's use [extend](https://www.typescriptlang.org/tsconfig#extends) here.

```suggestion
{
  "extends": "./tsconfig.json",
  "compilerOptions": {
    "noUnusedLocals": false,
    "noUnusedParameters": false
  }
}
```

maksnester (changes requested):

maksnester (approved):

PR #1038, opened 2020-09-03
feat: use new endpoints for mapping between node ID and tree index

larsmoa:


documentation/docs/migration-guide.md
```diff
@@ -55,23 +55,31 @@ Reveal has functions for iterating over nodes, either all nodes in a 3D model or
...
+Note that the operation requires communication with CDF servers and should be kept at a minimum. It's recommended to use the functions that accepts a list of nodes to map whenever possible. Batching requests is recommended to achieve the best performance.
 
 Access to node information such as name, bounding box and properties/attributes is done through the [3D revision API](https://cognitedata.github.io/cognite-sdk-js/classes/revisions3dapi.html), specifically:
 
 - [`list3DNodes`](https://cognitedata.github.io/cognite-sdk-js/classes/revisions3dapi.html#list3dnodes) for retrieving a set of nodes using a filter
 - [`list3DNodeAncestors`](https://cognitedata.github.io/cognite-sdk-js/classes/revisions3dapi.html#list3dnodeancestors) for retrieving ancestors of a specific node
 - [`retrieve3DNodes`](https://cognitedata.github.io/cognite-sdk-js/classes/revisions3dapi.html#retrieve3dnodes) for retrieving a set of nodes identified by their node ids
 
-This is not changed from `@cognite/3d-viewer`.
+The API for accessing node information has not changed `@cognite/3d-viewer`, but since Reveal now uses tree indices explicit mapping tree index to node ID is necessary.
```
maksnester, line 72: > The API for accessing node information has not changed `@cognite/3d-viewer`

Is everything OK with that sentence?

viewer/package.json
```diff
@@ -52,6 +52,7 @@
     "@types/tween.js": "^18.5.1",
     "@typescript-eslint/eslint-plugin": "^3.9.0",
     "@typescript-eslint/parser": "^3.9.0",
+    "core-js": "^3.6.5",
```
maksnester, line 55: could you share what problems did you have with the tests to add that?

viewer/tsconfig.test.json
```diff
@@ -0,0 +1,35 @@
...
+        "dom",
+        "esnext"
+    ],
+
+    "baseUrl": "./src/",
+    "paths": {
+      "@/*": ["*"]
+    }
+  },
+  // include is also used for eslint to understand which files should be linted
```
maksnester, line 32: That shouldn't work too.

maksnester: Sorry, I shouldn't review anything until morning coffee. Yep, json doesn't allow comments, but tsconfig.json does.

viewer/package.json
larsmoa: `Array.flat()` doesn't work in jest (node) without it

maksnester: Mmmmm.... which node version are you using?

```
Welcome to Node.js v12.18.3.
Type ".help" for more information.
> Array.prototype.flat
[Function: flat]
> [].flat
[Function: flat]
```

Should be fine after node > 11

larsmoa: My jest is node 10 apparently (I use 12 for the rest of the toolchain though)

larsmoa: Since it's just a dev dep I think it's ok to keep though

j-bjorne (approved):
There are a couple of things I wish were different, but nothing that can't be handled in a separate commit, so if you want it in before the weekend you have my go ahead, but if not I would consider not sending in the client and rather adding these new funcitons to the client with an interface

PR #1041, opened 2020-09-04
Fix: Duplicate network requests

j-bjorne:
Added publish to pipeable operations (removing had unforseen consequences with defer).
Replaced partition with just pipeable operators, to prevent long running tasks once publish was reintroduced.


larsmoa (approved):
Reduces network request for a test model from 708 to 445 request - nice! Doesn't reduce draw calls/triangle count.

PR #1050, opened 2020-09-06
chore: deploy preview documentation on PR build

larsmoa:
Deploy documentation to a Github pages site and post a comment about the URL. Uses repository https://github.com/cognitedata/reveal-docs-preview to store the preview docs.

### DIFF
--- a/.github/workflows/preview-documentation.yml
+++ b/.github/workflows/preview-documentation.yml
@@ -40,5 +40,6 @@ jobs:
         uses: JamesIves/github-pages-deploy-action@releases/v3
         with:
           ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
-          BRANCH: gh-pages # The branch the action should deploy to.
-          FOLDER: documentation/docs-preview # The folder the action should deploy.
+          BRANCH: master # The branch the action should deploy to.
+          FOLDER: documentation/docs-preview/build # The folder the action should deploy.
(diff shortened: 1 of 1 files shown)


.github/workflows/preview-documentation.yml
```diff
@@ -0,0 +1,44 @@
...
+          export PR_NUMBER=$(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
+          mkdir -p "docs-preview/$PR_NUMBER"
+          ls "docs-preview/"
+          cp -R build/ "docs-preview/$PR_NUMBER/"
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@releases/v3
+        with:
+          ACCESS_TOKEN: ${{ secrets.ACCESS_TOKEN }}
+          BRANCH: gh-pages # The branch the action should deploy to.
```
maksnester, line 43: I'm a bit concerned that it's the same branch we use for published docs. Also, as I understand, that action will just create folders for each PR without cleaning them up. 

I think we should look for something else.

larsmoa: I agree. Branch to test concept. I think we should move doc to a separate repo too to avoid repo growing in size

larsmoa: I think also we should clean up but cleanup in git isn't really saving any space since history is kept anyways

PR #1053, opened 2020-09-07
doc: add example of how to filter based on node properties

larsmoa:
Also fixes an issue causing Cognite3dModel.showNodeByTreeIndex() not to work when applyToChildren was true.

### DIFF
--- a/documentation/docs/examples/cad-highlighting.mdx
+++ b/documentation/docs/examples/cad-highlighting.mdx
@@ -106,7 +106,7 @@ viewer.on('click', event => {
         if (taggedEquipment) {
           // Highlight tagged equipment and all children
           model.selectNodeByTreeIndex(taggedEquipment.treeIndex, true);
-        }
+        } 
       })
     })
   }
--- a/documentation/docs/examples/cad-highlighting.mdx
+++ b/documentation/docs/examples/cad-highlighting.mdx
@@ -91,12 +91,13 @@ viewer.on('click', event => {
   const intersection = viewer.getIntersectionFromPixel(
     event.offsetX, event.offsetY
   );
+  model.deselectAllNodes();
   if (intersection) {
-    const { modelId, revisionId } = intersection.model;
-    const nodeId = intersection.nodeId;
-    // Retrieve all ancestors of the clicked node
-    sdk.revisions3D.list3DNodeAncestors(modelId, revisionId, nodeId)
-      .then(ancestors => {
+    const model = intersection.model;
+    const { modelId, revisionId } = model;
+    // Map treeIndex to nodeId to be able to lookup node in CDF
+    model.mapTreeIndexToNodeId(intersection.treeIndex).then(nodeId => {
+      sdk.revisions3D.list3DNodeAncestors(modelId, revisionId, nodeId).then(ancestors => {
         // Find first ancestor whose name starts with "/" - we assume this
         // is the a tagged equipment
         const taggedEquipment = ancestors.items.find(
@@ -106,7 +107,8 @@ viewer.on('click', event => {
           // Highlight tagged equipment and all children
           model.selectNodeByTreeIndex(taggedEquipment.treeIndex, true);
         }
-      });
+      })
+    })
   }
 });
 `}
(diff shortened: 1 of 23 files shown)


documentation/docs/examples/cad-nodefiltering.mdx
```diff
@@ -0,0 +1,61 @@
...
+<DemoWrapper
+  modelId={4715379429968321}
+  revisionId={5688854005909501}
+  name="Cognite3DViewerDemo"
+/>
+
+## Only show nodes that has a given node attribute value
+
+The available node properties varies a lot from model to model, and to use this information you will need to investigate
+the structure of the underlying models. For the purpose of this example, we the attributes of the nodes typically looks something like:
```
maksnester, line 24: >  For the purpose of this example, we the attributes of the nodes typically looks something like

Could you rephrase that part? I'm not sure what to suggest.

documentation/docs/examples/cad-nodefiltering.mdx
```diff
@@ -0,0 +1,61 @@
...
+```json
+"PDMS": {
```
maksnester, line 26: That thing looks scary to me. Maybe to simplify it?

documentation/docs/examples/cad-nodefiltering.mdx
```diff
@@ -0,0 +1,61 @@
...
+from the [Cognite SDK](https://www.npmjs.com/package/@cognite/sdk) which maps to the [List 3D Nodes API endpoint](https://docs.cognite.com/api/v1/#operation/get3DNodes).
+For this example we query for nodes that are within the 'EP' function and hide all other nodes.
+
+<LiveCodeSnippet>
+{`
+model.hideAllNodes();
+sdk.revisions3D.list3DNodes(model.modelId, model.revisionId, 
+  { 
+    limit: 1000, 
+    properties: { PDMS: { Function: 'EA' } }
```
maksnester, line 50: I suggest using something more readable to demonstrate the point that we can filter properties. Function "EA" doesn't make any sense to me personally, so it's just harder to read than some filtering by name/type/date or whatever more meaningful.

examples/src/pages/Migration.tsx
```diff
@@ -68,6 +68,7 @@ export function Migration() {
       });
       (window as any).viewer = viewer;
 
+
```
maksnester, line 71:
```suggestion
```

viewer/src/public/migration/Cognite3DModel.ts
```diff
@@ -487,8 +487,8 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
    */
   async showNodeByTreeIndex(treeIndex: number, applyToChildren = false): Promise<number> {
     const treeIndices = await this.determineTreeIndices(treeIndex, applyToChildren);
-    this.hiddenNodes.delete(treeIndex);
-    this.cadNode.requestNodeUpdate([treeIndex]);
+    treeIndices.forEach(idx => this.hiddenNodes.delete(idx));
+    this.cadNode.requestNodeUpdate(treeIndices.asArray());
```
maksnester, line 491: That's not related to that PR at all, but I think that conversion of treeIndices to an array should be called `toArray` I guess that name is more common to what we have in js with its toString toNumber toLowercase and other similarly named functions.

maksnester (changes requested):

documentation/docs/examples/cad-nodefiltering.mdx
larsmoa: I agree that this doesn't make much sense, but I haven't really been able to find good examples of filters I think make sense. The input model is not great when it comes to node metadata so finding good examples is hard.

viewer/src/public/migration/Cognite3DModel.ts
larsmoa: Agreed, fixed.

documentation/docs/examples/cad-highlighting.mdx
```diff
@@ -106,7 +106,7 @@ viewer.on('click', event => {
         if (taggedEquipment) {
           // Highlight tagged equipment and all children
           model.selectNodeByTreeIndex(taggedEquipment.treeIndex, true);
-        }
+        }
```
maksnester, line 109:
```suggestion
        }
```

Will create a task on miro and try to handle linter for all packages, not only the viewer...

maksnester (approved):
LGTM. 

I removed auto-merge in case if you're on something still.

PR #1060, opened 2020-09-07
refactor(docs): add `runnable` meta for examples

maksnester:
What it does basically is this:

```diff
-<LiveCodeSnippet>
-{`
+```jsx runnable
model.iterateNodesByTreeIndex(treeIndex => {
  model.setNodeColorByTreeIndex(treeIndex, 127, 127, 127);
});
+```
-`}
-</LiveCodeSnippet>
```

Also, backticks and newlines are handled under the hood so you shouldn't think about them, just write code normally. Also, IDE should give you some nice code highlighting for these blocks now (at least I have a plugin for that in webstorm, so I believe vscode should also have something for mdx files). 

The main downside here is that I had to get rid of the `scope` prop and did some ugly things inside LiveCodeSnippet with the scope and useBaseUrl hook-function, but I hope it's fine, we shouldn't have too much custom stuff. The custom scope now is shared between all examples. But that's something I can try to fix later.


documentation/src/components/customScope.ts
```diff
@@ -0,0 +1,11 @@
...
+import { resetViewerEventHandlers } from "../viewerUtilities";
+import * as THREE from 'three';
+import useBaseUrl from "@docusaurus/useBaseUrl";
+
+export const customScope = {
+  resetViewerEventHandlers,
+  THREE,
+  Water,
+  skyUrl: useBaseUrl('/img/sky007.jpg')
+}
```
larsmoa, line 11: Don't think we need to do it now, but could we in theory make this a function and have another function `addToScope` that allowed us to extend the scope that was provided to the examples?

larsmoa (commented):
Very cool indeed! Awaiting full review untill it builds

documentation/src/components/customScope.ts
maksnester: I don't know really. Right now the easiest alternative option I see is having some component that will pollute global scope with custom scope. Something like that:

```mdx
<MyScope scope={{ foo: 'bar' }} />

\```jsx runnable
useFoo(foo);
\```
```

And `MyScope` will basically do something dirty like `Object.assign(window.scope, props.scope)`. Not sure if it's any better than it's now.

larsmoa: Works for me, but don't need to do it now

larsmoa (approved):

PR #1063, opened 2020-09-08
feat(docs): use .env to override model/project

maksnester:
Since we might have troubles now with logging in to publicdata from localhost these changes should be useful for local development. 

Just copy .env.sample as .env and uncomment lines. Changes in .env require restart of dev server.


maksnester:
@larsmoa I rolled back your changes with passing ids directly from .mdx files and my original idea was that these things will live inside demo components. You see Cognite3DViewerDemo.tsx is a demo component, we specify its name for DemoWrapper. 

```jsx
<DemoWrapper name="Cognite3DViewerDemo" />
```

We might have other demos that might be more complicated than a simple 3d viewer. For example, in your demo with overlay, It could be done with a different demo. Instead of having live example with an overlay, you could create that overlay in demo instead and just put not-editable code snippet. But since you are still reusing simple 3d demo you could use it too and new component would look like that

```jsx
// HtmlOverlayDemo
export function HtmlOverlayDemo({ client }: DemoProps) {
  useEffect(() => { /* create a hiddent overlay and add it to the page */ }, [])
  return <Cognite3DViewerDemo client={client} />
}
```

Then you can register a demo by name at DemoWrapper.tsx

```diff
const components: Record<string, ComponentType<DemoProps>> = {
  Cognite3DViewerDemo: React.lazy(() => import('./demos/Cognite3DViewerDemo')),
+ HtmlOverlayDemo: React.lazy(() => import('./demos/HtmlOverlayDemo')),
};
```

And finally use it like in .mdx as 

```jsx
<DemoWrapper name="HtmlOverlayDemo" />
```

I don't say that it's better for HTML overlay, just trying to show an alternative if you need some very different demo, but don't want to relay on a user that will run live code snippets.

larsmoa (approved):

maksnester:
@larsmoa in addition to what I wrote above, please check this commit https://github.com/cognitedata/reveal/pull/1063/commits/bbe8332128f1da8b6fc04f71ba7fe8d498e2e031

I just made Matthew's demo to be yet another demo like I described.

PR #1065, opened 2020-09-08
docs: login fixes, add .env.example, bigger run button, mobile nav

maksnester:


maksnester:
@larsmoa rolled back parsing of function source code since it doesn't work with minified code. It wasn't a good idea. It's ready to review I think.

maksnester:
@larsmoa pushed login fixes here too. Now should be possible to log in and not click that sign-in button again.

larsmoa (approved):

PR #1066, opened 2020-09-08
fix: don't trigger click event for long clicks (>250ms)

larsmoa:


maksnester (approved):
Yeah, that was an annoying thing :D I'm glad that it's gone.

PR #1073, opened 2020-09-09
fix(docs): duplicate sidebar content to avoid having almost empty navbar on mobile

maksnester:
OK, this is quite lame but it lets us have some content in that left sidebar on mobile. 

I simply copied HTML content from the right-bottom sidebar into the top-left sidebar. 

Why it's bad - see comments:

```
// Since it's dummy HTML copying – links are common <a> tags and won't be handled by the app.
// It means the page will be reloaded every time one of these links is clicked (like in multipage or common websites, i.e. not SPA).
// Also, 'active' state classes are removed from here because they won't be updated that way
``` 

Also, it's just not recommended to swizzle sidebar, because it is an internal component. 

Anyway, I just hope that we will have some more links for the navbar later (e.g. versions). And that won't look so empty. And then I'll remove these hacks.

### DIFF
--- a/documentation/src/theme/Navbar/index.tsx
+++ b/documentation/src/theme/Navbar/index.tsx
@@ -1,0 +1,225 @@
+import React, {useCallback, useState, useEffect} from 'react';
+import clsx from 'clsx';
+import Link from '@docusaurus/Link';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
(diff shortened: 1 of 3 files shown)


maksnester:
@larsmoa I'll merge it to the docs-improvements branch. But at least you can review it easier with the shorter diff here.

PR #1089, opened 2020-09-14
Update migration guide based on migrating Maintain

danlevings:
[WIP] https://github.com/cognitedata/maintenance-planner/pull/628

Notes:
- for onClick events, the `PointerEventDelegate` type currently isn't exported. Would be nice to have that
- the `.getCamera()` function has the type 'any' - could be nice to change that to something more relevant


documentation/docs/migration-guide.md
```diff
@@ -82,12 +82,15 @@ The API for accessing node information has not changed `@cognite/3d-viewer`, but
...
 There are a few other noticeable changes from `@cognite/3d-viewer` and `@cognite/reveal`:
 
 - `@cognite/3d-viewer` supports local caching to reduce the time to load previously opened 3D models. Currently, this is not supported by `@cognite/reveal`, but the need for such functionality is reduced by adding streaming capabilities.
 - In `@cognite/3d-viewer` `Cognite3DViewer.addModel(...)` will always return a `Cognite3DModel`. In `@cognite/reveal` this function might also return a `CognitePointCloudModel`. To explicitly add a CAD model or point cloud model use `Cognite3DViewer.addCadModel(...)` or `Cognite3DViewer.addPointCloudModel(...)`
-- `Cognite3DViewer.loadCameraFromModel(...)`] has been added for loading camera settings from CDF when available.
\ No newline at end of file
+- `Cognite3DViewer.loadCameraFromModel(...)`] has been added for loading camera settings from CDF when available.
+
+- Due to the way `@cognite/reveal` streams data, the `OnProgressData` is no longer exported, and the `addModel` function no longer accepts a `onProgress` parameter
+- If you were accessing the private property `_camera` in the past, you can do so now by using `viewer.getCamera()`
```
danlevings, line 96:
```suggestion
- If you were accessing the private property `_camera` in the past, you can do so now by using `viewer.getCamera()`
- You need to do `yarn add three`.
```

maksnester: @larsmoa perhaps since we still export THREE from the Reveal we should keep it inside dependencies. Yeah, we will have some pain with it when used locally with linking, but it should be easier for people, because, I think most projects don't have a direct dependency on threejs so there is no point to ask about it as a peer dependency.

danlevings, line 96:
```suggestion
- If you were accessing the private property `_camera` in the past, you can do so now by using `viewer.getCamera()`
- You can no longer pass `onComplete` to the `addModel` functions - it will throw (without a reasonable error)
```

larsmoa: Not sure on the best approach tbh, @maksnester. I thought we mimiced how 3d-viewer does it, but not sure.

larsmoa (approved):

PR #1092, opened 2020-09-14
add `bump` and `release` scripts to viewer and parser-worker, update readme on publishing

maksnester:
Basically it's useful to simplify the tagging. Both packages now have `bump` and `release` scripts. 

* `yarn bump` – updates the version number (asks you to enter new version) and also creates a tag with the clear name, like `@cognite/reveal@1.0.1` or `@cognite/reveal-parser-worker@1.0.1`, and pushes everything to the remote branch. 
  And yes, we should name our tags, since we have more than one package here. Our colleagues who work on sdk do it in [the same way](https://github.com/cognitedata/cognite-sdk-js/tags). 
* `yarn release` – creates a build and asks for OTP as `publish` normally does. 

And that's it. It's not much of automatization here or fancy GitHub actions. 

---

Here is also my note about the spike I had where I supposed to try https://github.com/marketplace/actions/create-a-release

https://github.com/marketplace/actions/create-a-release is pretty much useless by itself only. It just automates the releasing part, which normally is done by clicking some UI in the GitHub releases section.

Also, I discovered that it will be impossible to publish new versions to npm automatically, because of 2FA on npm packages. I guess we shouldn’t turn it off for cognite packages (however, our colleagues from SDK just disabled 2FA for publishing 🤷‍♂️ ). The build will fail because it will want you to enter one-time password for publishing - that’s not the case for a GitHub action. 

---

Some thoughts on better releases - would be cool to automate release notes (BTW we still don't have release notes for our 1.0 release ⚠️).

Ideally, I would like us to mark PRs with some tags, e.g. `viewer:feature`, or `viewer:fix`. And if PR is merged and marked with that tag it’s title should appear in the changelog for a corresponding package release. We basically need to generate a changelog somehow and publish a new GitHub release with that generated changelog. A published tag could be the trigger for that action. But right now it's not really clear to me how to implement all that stuff. Need more time to play with it.

BTW, Docusaurus have a similar approach with tagging PRs, but they publish GitHub-releases manually with the UI though it could be done with that create-a-release action mentioned above.


larsmoa (commented):
Very much an upgrade and makes the release process a lot clearer. A couple of things:
- Would be great if `yarn release` also would tag the release and push the tag
- Is this basically emulating lerna?

maksnester:
> Would be great if yarn release also would tag the release and push the tag
That was my initial idea too (to make it to be just a single command), but I decided that it shouldn't be. Because when there are 2 steps, it leaves you a chance to wait until CI checks are done, maybe to play with docs preview, etc, and publish to NPM only after that. So `bump` does tagging and version bumping, and `release` (which you supposed to run once you 100% sure) does publishing. 

> Is this basically emulating lerna?
Not sure, not intentionally at least. But there is definitely no version lock between packages (and shouldn't be in my opinion). But we do some stuff like local linking which is also handled by lerna as far as I know.

larsmoa:
> > Would be great if yarn release also would tag the release and push the tag
> 
> That was my initial idea too (to make it to be just a single command), but I decided that it shouldn't be. Because when there are 2 steps, it leaves you a chance to wait until CI checks are done, maybe to play with docs preview, etc, and publish to NPM only after that. So `bump` does tagging and version bumping, and `release` (which you supposed to run once you 100% sure) does publishing.

Ah, didn't notice that `bump` tagged. I think we should add an explicit step to push tags, i.e. `git push --tags` as they are not published on `git push`.

maksnester:
> I think we should add an explicit step to push tags

`bump` already handles tagging. You don't have to tag anything manually. 

```json
    "prebump": "yarn version --no-git-tag-version && yarn",
    "bump": " yarn bump:commit && yarn bump:tag && yarn bump:push",
    "bump:commit": "cross-var git commit -am \"chore(parser-worker): bump version $npm_package_version\"",
    "bump:tag": "cross-var git tag $npm_package_name@$npm_package_version",
    "bump:push": "git push -u origin head && git push --tags",
```

larsmoa (approved):

PR #1107, opened 2020-09-18
add eslint-plugin-jsdoc

maksnester:
I personally already hate it. But hopefully, you will find it useful 😅

### DIFF
--- a/viewer/.eslintrc.js
+++ b/viewer/.eslintrc.js
@@ -14,7 +14,7 @@ module.exports = {
     ecmaVersion: 2020, // Allows for the parsing of modern ECMAScript features
     sourceType: 'module' // Allows for the use of imports
   },
-  plugins: ['header', '@typescript-eslint'],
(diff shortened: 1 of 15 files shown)


larsmoa (commented):
Should we always enforce `@param` and `@return`?

maksnester:
> Should we always enforce @param and @return?

Param - yes, it's enforced. Return is optional. Because you don't have to specify type for return and description is often redundant becaus it's obvious from return type and function name. That's why I don't want to enforce it. But I can, if you think it makes sense.

larsmoa:
Regarding param I'm fine with enforcing it, but sometimes I feel doc is a bit redundant, e.g. doc for treeIndex

larsmoa (approved):

maksnester:
> Regarding param I'm fine with enforcing it, but sometimes I feel doc is a bit redundant, e.g. doc for treeIndex

I agree. It's really hard to find some balance here in requiring things and keeping configuration not stupidly annoying.

PR #1114, opened 2020-09-19
feat: reduce framerate while loading

larsmoa:
Gives more time for loading operations when camera isn't moving


larsmoa:
I realize now that this might be a bad idea in some cases, e.g. when animating the 3D view. Leaving to be considered.

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1041,16 +1047,29 @@ export class Cognite3DViewer {
...
+      if (
+        (renderController.needsRedraw || this._revealManager.needsRedraw || this._slicingNeedsUpdate) &&
+        !this.shouldSuspendRendering()
+      ) {
         this.triggerUpdateCameraNearAndFar();
         this._revealManager.render(this.renderer, this.camera, this.scene);
         renderController.clearNeedsRedraw();
         this._revealManager.resetRedraw();
         this._slicingNeedsUpdate = false;
+        this._lastFrameTimestamp = performance.now();
```
maksnester, line 1059: @larsmoa  You probably don't need performance.now in render loop. It has faster alternatives (but less precise though). 

Check it out
* https://jsbench.me/k3kh4nni14/1
* https://stackoverflow.com/questions/30795525/performance-now-vs-date-now

maksnester (changes requested):

viewer/src/public/migration/Cognite3DViewer.ts
larsmoa: Oh, didnt know - thx. Fixed.

PR #1125, opened 2020-09-22
refac: naming and comments

larsmoa:


maksnester (approved):
See nothing criminal here 🙂

PR #1132, opened 2020-09-23
feat: add Cognite3DModel.setAllNodeColors

larsmoa:


viewer/src/public/migration/Cognite3DModel.ts
```diff
@@ -451,6 +451,21 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
     return treeIndices.count;
   }
 
+  /**
+   * Overrrides color for all nodes in the scene.
+   * @param r Red component between 0 and 255.
+   * @param g Green component between 0 and 255.
+   * @param b Blue component between 0 and 255.
+   * @version New in 1.1
```
maksnester, line 459:
```suggestion
   * @version new in 1.1.0
```

Boomer suggestion here. Since it's not a sentence that is finished with `.` I think it shouldn't start with a capital letter. And let's specify minor `.0` as we do for documentation.

documentation/docs/examples/cad-colors.mdx
```diff
@@ -15,17 +15,25 @@ Reveal supports updating the color of individual objects by providing [tree indi
...
+to be applied to all nodes is to use `Cognite3DModel.setAllNodeColors`:
+
+```jsx runnable
+model.setAllNodeColors(127, 127, 127);
+```
+
 ## Reset node colors
 
-Color overrides can be overridden using `Cognite3DModel.resetNodeColorByTreeIndex`, e.g.
+Color overrides can be reset using `Cognite3DModel.resetNodeColorByTreeIndex`, e.g.
```
maksnester, line 36: But can't we have some reset function as well? A bit weird that we have setAll but not clearAll.

larsmoa: We have `resetAllNodeColors`?

maksnester: Sorry, my bad. There is resetAllNodeColors below already.

maksnester (approved):

PR #1133, opened 2020-09-23
Docs: split 1.x and next; Viewer: fix exports of types from Cognite3DViewer

maksnester:
That's will be a bit hard to review since there are a lot of files copied from /docs under /versioned_docs, but I guess it should be easier to review per commit. 

Regarding docs deployment let me just paste a quote from [documentation/Readme.md](https://github.com/cognitedata/reveal/blob/mn/docs-split-next-from-published/documentation/README.md). 

We deploy docs automatically for every merge into master. We have documentation for 2 versions at the moment:
* `1.x` version is default. 
    There should be only documentation for the published versions of `@cognite/reveal`.
    Files for `1.x` are in `/versioned_docs/version-1.x`. **You shouldn't edit these files** unless you want to publish some fix ASAP.
    In that case you need to apply the fix for the both `next` and `1.x` versions.
* `next` contains docs for the latest unpublished `@cognite/reveal` (master branch). 
    You should always edit files for the next version only (they are stored in `/docs` folder).
    Files for `1.x` version are replaced with files from `/docs` for every release. It's handled inside viewer `bump` script.
    
When you write documentation for new features please specify from which version feature is available.
Use the following comment:

```
> **New in 1.1.0**
```

> **New in 1.1.0**

---

Other fixes:
* export all types from Cognite3DViewer, see
  ```diff
  - import { Cognite3DViewer } from '@cognite/reveal';
  - import { CameraChangeDelegate, PointerEventDelegate } from '@cognite/reveal/public/migration/Cognite3DViewer';  
  + import {
  +    CameraChangeDelegate,
  +    PointerEventDelegate,
  +    Cognite3DViewer,
  +  } from '@cognite/reveal';
  ```
* minor fix for auth in documentation for case when you was authorized in one project and then do changes in .env file, reload the server and can't see anything because of stale access token.


larsmoa:
@maksnester : should we also mark API reference docs with version somehow?

maksnester:
@larsmoa not really. Next always have actual api ref. When you perform release that actual API ref will be copied under 1.x and won't change until the next release. So it's already "versioned" as all the other docs.

larsmoa (approved):
Approved - but should we also require some way of marking new API changes with version, and how should we do this? Also, should release procedure doc be updated?

larsmoa:
> @larsmoa not really. Next always have actual api ref. When you perform release that actual API ref will be copied under 1.x and won't change until the next release. So it's already "versioned" as all the other docs.

I'm thinking since doc for 1.0 and 1.1 will be the same and that we add functions in 1.1, it would make sense to mark those functions with 1.1 somehow?

maksnester:
> but should we also require some way of marking new API changes with version, and how should we do this?

Absolutely. I proposed the way of doing that in examples and hand-written docs, but not in the tsdoc. But yes, we should probably mark new features somehow. Will take a look and add a note on that.

PR #1140, opened 2020-09-24
chore: add release-template.md

maksnester:
There is no fancy automatization of anything, I just thought that it might be handy to have for creating release notes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,4 +166,18 @@ Also, you can check what's committed from the previous tag with that command:
     ```bash
     git log --pretty=format:"%s <%an> – %h" v1.0.0...HEAD
     ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -166,4 +166,18 @@ Also, you can check what's committed from the previous tag with that command:
...
+       * commit message
+       
+       ### 🐞 Bug fixes and enhancements
+       
+       * commit message
+       
+       ### 📖 Documentation
+       
+       * commit message
+   ```
```
larsmoa, line 182: I think we should reference release-template.md to avoid them going out of sync

larsmoa (approved):

PR #1141, opened 2020-09-24
fix: bug causing custom 3d objects always to be rendered in front of ghosted objects

larsmoa:
Can be tested with the following snippet in docs:
```
model.ghostAllNodes();
model.unghostNodeByTreeIndex(409228, applyToChildren=true);
const box = new THREE.BoxGeometry(10, 10, 10);
const material = new THREE.MeshBasicMaterial({color: 'white'});
const mesh = new THREE.Mesh(box, material);
mesh.position.set(121.84861758965458, 58.620189040381064, -89.28303211581611);
viewer.addObject3D(mesh);
```
Before this would render the custom object in front of ghosted objects:

image

Now its blended correctly:

image

### DIFF
--- a/viewer/src/glsl/post-processing/edge-detect-combine.frag
+++ b/viewer/src/glsl/post-processing/edge-detect-combine.frag
@@ -74,36 +74,44 @@ void main() {
     return;
   }
 
-  if(customDepth < backDepth){
-    gl_FragColor = customAlbedo;
-    return;
-  }
+  // if(customDepth < backDepth){
+  //  gl_FragColor = customAlbedo;
+  //  return;
+  // }
 
-  float backOutlineIndex = floatBitsSubset(floor((backAlbedo.a * 255.0) + 0.5), 3, 6);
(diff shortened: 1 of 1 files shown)


viewer/src/glsl/post-processing/edge-detect-combine.frag
```diff
@@ -74,36 +74,44 @@ void main() {
...
-    float averageNeighbourFragmentDepth = (d1 + d2 + d3 + d4) / 4.0;
-
-    if(d0 < averageNeighbourFragmentDepth){
-      float borderColorIndex = max(max(backOutlineIndex0, backOutlineIndex1), max(backOutlineIndex2, backOutlineIndex3));
-      gl_FragColor = texture2D(tOutlineColors, vec2(0.125 * borderColorIndex + (0.125 / 2.0), 0.5));
-      return;
+  // if(customDepth < backDepth){
+  //  gl_FragColor = customAlbedo;
+  //  return;
+  // }
```
christjt, line 80:
```suggestion

```

christjt (approved):

PR #1145, opened 2020-09-25
chore: improve error message when model is incompatible

larsmoa:


viewer/src/utilities/networking/CdfModelDataClient.ts
```diff
@@ -52,7 +52,7 @@ export class CdfModelDataClient
     const mostRecentOutput = outputs.findMostRecentOutput(format);
     if (!mostRecentOutput) {
       throw new Error(
-        `Model '${modelId}/${revisionId} (${format})' is not compatible with this version of Reveal. If this model works with a previous version of Reveal it must be reconverted to support this version.`
+        `Model '${modelId}/${revisionId}' is not compatible with this version of Reveal, because no outputs for format '(${format})' was found. If this model works with a previous version of Reveal it must be reconverted to support this version.`
```
maksnester, line 55:
```suggestion
        `Model '${modelId}/${revisionId}' is not compatible with this version of Reveal, because no outputs for format '(${format})' was found. If this model works with a previous version of Reveal it must be reprocessed to support this version.`
```

maksnester (approved):

PR #1147, opened 2020-09-25
doc: minor improvement of API doc of Cognite3DModel

larsmoa:


maksnester (approved):
Just a reminder from me that if you want to update the published version of the API reference you need to manually edit the published version docs under `versioned_docs/1x/API reference.md`. They won't be generated after changes, only the next version is updated.

PR #1148, opened 2020-09-25
doc: add section about common issues in migration

larsmoa:


documentation/docs/migration-guide.md
```diff
@@ -144,3 +144,15 @@ where `PROJECT`, `MODELID`, `REVISIONID` and `APIKEY` must be set in the environ
...
+## Common issues
+
+**`TypeError: e.applyMatrix4 is not a function`**
+
+If you experience  `TypeError: e.applyMatrix4 is not a function` after migrating from 
+`@cognite/3d-viewer` to `@cognite/reveal`, the problem is usually caused by missing or 
+outdated ThreeJS.
+
+To fix this issue, install or update ThreeJS using the 
+[`three`-package](https://www.npmjs.com/package/three). This should match the version
```
maksnester, line 157: That's very bad actually. Because the version used by reveal is very internal thing. Perhaps we should just use three.js as a common dependency instead of declaring it as a peer.

maksnester (approved):
And again. If you want changes immediately available you have to edit the versioned docs as well. There are not a lot of changes in master in comparison to 1.1, perhaps, we can just use the replacement script from `documentation/package.json`

`yarn replace-1.x`

PR #1152, opened 2020-09-28
fix(viewer): avoid direct dependency on sdk; loose three.js peer dependency version

maksnester:
1) dependency on cognite/sdk - we shouldn't use HttpError to avoid having @cognite/sdk module as a direct dependency for the reveal. Well, it's arguable, but it's a quick fix for this problem https://github.com/cognitedata/cdf-hub/pull/687/files#r489204315. 

And, by the way, the guys from cognite-sdk-js mark [HttpError with `@hidden` annotation](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/core/src/httpClient/httpError.ts#L5), perhaps it shouldn't be used outside (but it is used outside of sdk, not only in reveal). 

2) I set 0.117.x as the version range for the three.js to avoid warning when 117.1 is installed instead of 117.0 in projects that use reveal.

But should we move three.js under common dependencies? It’s a dependency. We use it directly and even re-export it. So… I don’t really understand why we should mark it as a peer dependency (except that it was peer dependency in the old 3d viewer).


larsmoa:
> 1. dependency on cognite/sdk - we shouldn't use HttpError to avoid having @cognite/sdk module as a direct dependency for the reveal. Well, it's arguable, but it's a quick fix for this problem https://github.com/cognitedata/cdf-hub/pull/687/files#r489204315.
> 
> And, by the way, the guys from cognite-sdk-js mark [HttpError with `@hidden` annotation](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/core/src/httpClient/httpError.ts#L5), perhaps it shouldn't be used outside (but it is used outside of sdk, not only in reveal).
> 

Ok, thx for raising this issue. I think we just should make our own `HttpError`.


> 1. I set 0.117.x as the version range for the three.js to avoid warning when 117.1 is installed instead of 117.0 in projects that use reveal.

Great.

> 
> But should we move three.js under common dependencies? It’s a dependency. We use it directly and even re-export it. So… I don’t really understand why we should mark it as a peer dependency (except that it was peer dependency in the old 3d viewer).

Probably. I thought I'd replicated how the old viewer did it but I'm probably wrong. I don't know what the right approach here is so I trust your judgment on this one 🥇

larsmoa (approved):

maksnester:
> I don't know what the right approach here is so I trust your judgment on this one 🥇

@larsmoa Well, I don't know upfront if it will have some side effects, but I guess we just should try because it makes sense to mark it as a common dependency. So I'll do another PR for that.

PR #1153, opened 2020-09-29
chore(viewer): move three.js under dependencies instead of peer

maksnester:
Tried to test it in console with installed old viewer and old three version. Looks ok, but there are some issues with old three + old 3d-viewer version. Since the old three version doesn't have types ts complains on the code where types for the old three are needed, for example on that code:

```
model.matrix
````

since matrix is inherited from THREE.Object3D ts complains that there is no such property (because it doesn't know where to find it). I think the right thing to do here is to update 3d-viewer with dependency on types for the old threejs (will submit PR).


larsmoa:
Should doc be updated to reflect the change, @maksnester ?

maksnester:
@larsmoa IDK really. I think it should just work and nothing to mention here. 

Also, in 3d-viewer, in its different versions, three.js was changed from peer to common dependency (now it's a common dependency in the [latest version](https://github.com/cognitedata/3d-viewer/blob/v6.1.4/package.json)).

I moved common issues section under 1.x for now, because it's still relevant for 1.1 but it shouldn't be relevant anymore after the next release.

maksnester:
@larsmoa converted to draft for now as I did couple more tests with 3d-viewer + reveal in console and not really sure what is the best approach here. 

I think when both 3d-viewer and reveal depend on threejs as a common dependency. It's the most accurate in terms of type strictness and it shows a lot of places where types are actually incompatible between three versions. But that's a hassle to fix all these issues.

maksnester:
@larsmoa I decided to leave all as it is after some tests in the console. Maybe console is just a bad example because there we have some issues with types between 3d-viewer and Reveal and their very different threejs versions. But anyway, I think it's better to keep things as they are for now.

PR #1165, opened 2020-10-02
Add endpoint for setting the visibility filter of a given point class

christjt:


viewer/src/datamodels/pointcloud/types.ts
```diff
@@ -42,3 +42,17 @@ export enum PotreePointSizeType {
   Adaptive = Potree.PointSizeType.ADAPTIVE,
   Fixed = Potree.PointSizeType.FIXED
 }
+
+export enum PotreePointClassType {
+  Unclassified = 1,
```
larsmoa, line 47:
```suggestion
export enum PotreePointClassType {
  Default = 0,
  Unclassified = 1,
```

larsmoa (approved):
Approved, but not merging because I did some changes I want @christjt to have a look at.

viewer/src/datamodels/pointcloud/PointCloudNode.ts
```diff
@@ -92,6 +92,28 @@ export class PointCloudNode extends THREE.Group {
...
   }
 
+  /**
+   * Sets a visible filter on points of a given class.
+   * @param pointClass The point class type to which the filter should be applied. This can either
+   * be one of the well known classes from {@link WellKnownPointClassTypes} or a number for user defined
+   * classes.
+   * @param visible Boolean flag that determines if the point class type should be visible or not.
+   */
+  setClassVisible(pointClass: number | WellKnownPointClassTypes, visible: boolean): void {
```
maksnester: I suggest to name it like `setClassVisibility` because now it feels like a function that makes a certain class just visible, but it accepts visibility param. 

Or maybe it's just me 🤷‍♂️

viewer/src/datamodels/pointcloud/types.ts
```diff
@@ -42,3 +42,82 @@ export enum PotreePointSizeType {
   Adaptive = Potree.PointSizeType.ADAPTIVE,
   Fixed = Potree.PointSizeType.FIXED
 }
+
+
+/**
+ * ASPRS well known point class types.
+ * @see {@link http://www.asprs.org/wp-content/uploads/2019/03/LAS_1_4_r14.pdf} (page 30)
+ */
+export enum WellKnownPointClassTypes {
```
maksnester: Does it relative to Potree.PointColorType.CLASSIFICATION? Should it extend it or something?

larsmoa: `Potree.PointColorType.CLASSIFICATION` determines how to color the model, this enum is used to control what points are visible.

viewer/src/datamodels/pointcloud/PointCloudNode.ts
larsmoa: @maksnester : kind of agree, but what would you call `isClassVisible`?

larsmoa: Also, maybe its just me from my days in .NET, but I feel Visibility should be an enum and not a bool. Leaning towards just keeping it, mainly because I think `isClassVisible` should match naming and that `isClassVisibility` doesn't make sense and that `getClassVisibility: boolean` seems a bit off to me.

viewer/src/datamodels/pointcloud/types.ts
maksnester: Perhaps I just posted the wrong enum, but doesn't potree have something related to it? If not OK – nothing to discuss here. 

Also, actually, enum extension is not something supported in typescript, so...

viewer/src/datamodels/pointcloud/PointCloudNode.ts
maksnester: Agree that visibility doesn't sound well with a boolean param.

examples/src/pages/SimplePointcloud.tsx
```diff
@@ -100,10 +100,22 @@ export function SimplePointcloud() {
...
         );
-      }
+      }      
       scene.add(pointCloudNode);
       revealManager.on('loadingStateChanged', setLoadingState);
 
+      const classesGui = gui.addFolder('Class filters');
+      const enabledClasses: { [clazz: number]: boolean } = {};
+      console.log(pointCloudNode.getClasses());
+      for (const clazz of pointCloudNode.getClasses()) {
```
maksnester, line 110: @larsmoa I know it's internal but still... do we really need to access the node instead of using public API here? Enum is exported. You can get its values and keys. This is the code I'm currently using in the console to get a clean label/value map.

```js
 Object.keys(WellKnownPointClassTypes)
      .filter(key => Number.isNaN(Number(key))) // filter out enum numbers that stored as keys
      .reduce((acc, key) => {
        acc[key] = WellKnownPointClassTypes[key];
        return acc;
      }, {})
```

Don't really insist on changes here, but just in case posted it here.

viewer/src/datamodels/pointcloud/PointCloudNode.ts
```diff
@@ -92,6 +94,58 @@ export class PointCloudNode extends THREE.Group {
...
 
+  /**
+   * Sets a visible filter on points of a given class.
+   * @param pointClass ASPRS classification class code. Either one of the well known 
+   * classes from {@link WellKnownAsprsPointClassCodes} or a number for user defined classes.
+   * @param visible Boolean flag that determines if the point class type should be visible or not.
+   * @throws Error if the model doesn't have the class given.
+   */
+  setClassVisible(pointClass: number | WellKnownAsprsPointClassCodes, visible: boolean): void {
+    if (!this.hasClass(pointClass)) {
```
maksnester, line 105: I think it would be handy to have some method that will return available classes as a subset of WellKnownAsprsPointClassCodes.

viewer/src/public/migration/CognitePointCloudModel.ts
```diff
@@ -94,6 +94,50 @@ export class CognitePointCloudModel extends THREE.Object3D implements CogniteMod
...
+  setClassVisible(pointClass: number | WellKnownAsprsPointClassCodes, visible: boolean): void {
+    this.pointCloudNode.setClassVisible(pointClass, visible);
+  }
+
+  /**
+   * Determines if points from a given class are visible.
+   * @param pointClass ASPRS classification class code. Either one of the well known 
+   * classes from {@link WellKnownAsprsPointClassCodes} or a number for user defined classes.
+   * @return true if points from the given class will be visible.
+   * @throws Error if the model doesn't have the class given.
```
maksnester, line 114: Shouldn't we just be tolerant about that and ignore it if there is no such class? No bad things should happen in that case, right? So why we should throw errors?

examples/src/pages/SimplePointcloud.tsx
larsmoa: Yes, the enum only has a few 'pre-defined' codes, the user can define custom values so we can't enumerate all of them.

viewer/src/public/migration/CognitePointCloudModel.ts
larsmoa: Isn't it a bit strange to be allowed to turn something on if its not there?

viewer/src/datamodels/pointcloud/PointCloudNode.ts
larsmoa: What would you use this for? I think it's important that we handle all codes equally as it's really just a list of reserved values, other values are also valid.

viewer/src/public/migration/CognitePointCloudModel.ts
maksnester: Well, it's just a design decision to make. Maybe it doesn't make much sense to allow it. Can't tell you any meaningful use case for now.

viewer/src/datamodels/pointcloud/PointCloudNode.ts
maksnester: For the filtering in the console. No reason to show all possible (WellKnown) values if there are only a few of them used in a model. But you just said that we can't enumerate all of them anyway if they are custom. So I'm not sure about it anymore.

larsmoa: I think we should show all present values in Console. Usually its 5-15 different classes in a file. See http://potree.org/potree/examples/classifications.html for example

PR #1178, opened 2020-10-05
chore(refac): add CadModelUpdateHandler.budget to allow setting budget externally

larsmoa:
Enables setting CAD model budget runtime. Not exposed publicly yet.


j-bjorne (approved):
Looks pretty good, consideration of the future would be to instead of having these Subjects and observable as part of this big rx file to split it up a bit. You could then say have a, `CameraInputController` class which would look the following way.

```

type CameraInput = {
  camera: THREE.PerspectiveCamera;
  cameraInMotion: boolean;
};

class CameraInputController {
 
  private readonly _cameraSubject: Subject<THREE.PerspectiveCamera> = new Subject();

  public void update(camera: THREE.PerspectiveCamera) {
    this._cameraSubject.next(camera);  
  }

  public Observable<CameraInput> observable() {
    return combineLatest([
        this._cameraSubject.pipe(auditTime(500)),
        this._cameraSubject.pipe(auditTime(250), emissionLastMillis(600))
      ]).pipe(map(makeCameraInput)),
  }

  private makeCameraInput([camera, cameraInMotion]: [THREE.PerspectiveCamera, boolean]): 
  CameraInput {
    return { camera, cameraInMotion };
  }
}
```

Then in the updatehandler you send it in as a dependency and just observe on it :)

But yeah, that is maybe for a later PR if you think it looks like a good idea.

PR #1219, opened 2020-10-16
doc: add documentation with example for clipping planes

larsmoa:
Depends on #1224


documentation/docs/examples/clipping.mdx
```diff
@@ -0,0 +1,79 @@
...
+useful to isolate interesting parts of the model and hide obscuring geometry. Reveal supports having multiple slicing planes
+in arbitrary position and orientation. When slicing planes are active,
+the amount of geometry loaded might be reduced since culled geometry doesn't need
+to be loaded (or rather loaded in a lower detail level).
+
+<video preload="auto" autoPlay="autoplay" loop controls>
+  <source src={useBaseUrl('/img/cad-clipping.webm')} />
+</video>
+
+:::warning
```
maksnester, line 20: There is no warning https://v2.docusaurus.io/docs/markdown-features#calloutsadmonitions

Perhaps it should be:
```suggestion
:::caution
```

documentation/docs/examples/clipping.mdx
```diff
@@ -0,0 +1,79 @@
...
+### Single, fixed slicing plane
+
+Slicing planes are defined by providing zero or more [`THREE.Plane`](https://threejs.org/docs/#api/en/math/Plane) 
+to `Cognite3DViewer.setSlicingPlanes`. The slicing planes are activated for all models and
+replaces any active slicing planes.
+
+Typically, [setFromNormalAndCoplanarPoint](https://threejs.org/docs/#api/en/math/Plane.setFromNormalAndCoplanarPoint)
+is used to create slicing planes. This function accepts a direction that defines the orientation of
+the plane and a point on the plane. For example will the following
+code create a slicing plane that only accepts geometry below `Y = 55`:
```
maksnester, line 37:
```suggestion
the plane and a point on the plane. For example, the following
code will create a slicing plane that only accepts geometry below `Y = 55`:
```

documentation/docs/examples/clipping.mdx
```diff
@@ -0,0 +1,79 @@
...
+const point = new THREE.Vector3(0, 55, 0);
+const plane = new THREE.Plane().setFromNormalAndCoplanarPoint(orientation, point);
+viewer.setSlicingPlanes([plane]);
+```
+
+### Animated slicing plane
+
+Planes can be manipulated runtime and effects will have effect in the viewer. Note however that rendering must be 
+triggered after updating the plane before the change will have visual effect. This can be done using 
+`Cognite3DViewer.forceRerender()`.
```
maksnester, line 49:
```suggestion
Planes can be manipulated runtime and effects will have an effect on the viewer. Note however that rendering must be
triggered after updating the plane before the change will have a visual effect. This can be done using 
`Cognite3DViewer.forceRerender()`.
```

documentation/docs/examples/clipping.mdx
```diff
@@ -0,0 +1,79 @@
...
+}, 1000 / 60);
+```
+:::tip
+Updating `THREE.Plane`-instances will not trigger data loading. Typically you will need to trigger data loading if
+you do substantial changes to the slicing planes. In these cases, use `Cognite3DViewer.setSlicingPlanes`. The example 
+above performs substantial changes to the planes without loading data. In a real application it's recommended to 
+not load data during animation, but initiate data after an animation is completed.
+:::
+
+### CAD geometry filter
```
maksnester, line 79: unexpectedly ended here (I know I draft thought, but anyway)

documentation/docs/examples/clipping.mdx
```diff
@@ -0,0 +1,79 @@
...
+  const dt = (performance.now() - start) / 1000.0
+  const s = 0.5 * (Math.sin(Math.PI * dt / 10.0) + 1.0);
+  point.y = bbox.max.y - (bbox.max.y - bbox.min.y) / 2.0 * s;
+  plane.setFromNormalAndCoplanarPoint(orientation, point);
+
+  viewer.forceRerender();
+}, 1000 / 60);
+```
+:::tip
+Updating `THREE.Plane`-instances will not trigger data loading. Typically you will need to trigger data loading if
```
maksnester, line 73: I read this text multiple times to understand it, perhaps we just need some proper example here? I mean, the one where you do things right as you described here:

> Typically you will need to trigger data loading if you do substantial changes to the slicing planes.

Because currently the example looks cool, but it leaves me without knowledge on how to make things right. And it requires some patience to understand what is written in that tip, but basically, the tip just says - "hey, the example looks cool, but don't do that in real code, do X instead". So maybe we can have the X as an additional example here?

documentation/sidebars.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
         'examples/cad-nodefiltering',
         'examples/cad-preview',
         'examples/pointcloud',
+        'examples/clipping',
```
maksnester, line 35:
```suggestion
        'examples/clipping',
        'examples/pointcloud',
```

not sure about the proper ordering for the examples, but I think it makes sense to separate pointclouds and cad instead of mixing them? Or we might want to have the items AZ-sorted here, what do you think?

documentation/docs/examples/clipping.mdx
maksnester: Just realized that "the X" example is the first one after checking the docs preview 🙄

documentation/docs/examples/clipping.mdx
larsmoa: Interestingly, warning actually works (but caution is better here)

documentation/docs/examples/clipping.mdx
larsmoa: I _think_ this is ok actually. If you're sure I'm wrong, please scream.

documentation/sidebars.js
larsmoa: Yeah, I kinda though clipping would be something generic that worked for both CAD and point cloud but turns out it doesn't work for point clouds yet...

documentation/docs/examples/clipping.mdx
larsmoa: Updated a bit - hopefully more clear now :)

maksnester:
@larsmoa there are some conflicts now.

documentation/docs/demos/Cognite3DViewerDemo.tsx
```diff
@@ -43,6 +44,7 @@ export default function Cognite3DViewerDemo({
     addModel({ modelId, revisionId });
 
     window.viewer = viewer;
+    (window as any).DragControls = DragControls;
```
maksnester, line 47: @larsmoa  that's here again instead of the custom scope 🤷‍♂️

maksnester (changes requested):

documentation/docs/demos/Cognite3DViewerDemo.tsx
```diff
@@ -11,6 +11,7 @@ import {
 
 import { CanvasWrapper } from '@site/src/components/styled';
 import { DemoProps } from '@site/src/components/DemoProps';
+import { DragControls } from 'three/examples/jsm/controls/DragControls';
```
maksnester, line 14:
```suggestion
```

documentation/docs/examples/clipping.mdx
```diff
@@ -70,42 +71,77 @@ setInterval(() => {
...
+  viewer.cameraControlsEnabled = true;
+});
+```
+
+### CAD geometry filter
+
+A special type of slicing is to provide a "geometry filter" when loading the model. A geometry filter restricts the 
+area to load geometry for and is specified when the model is created.
+
+```js
```
maksnester, line 133: is it intentionally non-runnable?

maksnester (approved):
Approved to avoid another review and removed auto-merge, but please check my comments.

documentation/docs/examples/clipping.mdx
larsmoa: Yes, because it must be applied on creation which we don't support now.

PR #1244, opened 2020-10-22
feat: report application ID to MixPanel and improve events

larsmoa:
Send application ID to MixPanel to be able to see what apps use Reveal. Also, clean up events to avoid duplicates and track when user has navigated to keep a track of the activity per session.

### DIFF
--- a/viewer/src/public/RevealManager.ts
+++ b/viewer/src/public/RevealManager.ts
@@ -173,6 +173,7 @@ export class RevealManager<TModelIdentifier> {
       {
         moduleName: 'RevealManager',
         methodName: 'addModel',
+        type,
         options: { nodeAppearanceProvider }
       },
       modelIdentifier
--- a/viewer/src/public/RevealManager.ts
+++ b/viewer/src/public/RevealManager.ts
@@ -13,7 +13,7 @@ import {
 } from './types';
 import { Subscription, combineLatest, asyncScheduler } from 'rxjs';
 import { map, share, filter, observeOn, subscribeOn } from 'rxjs/operators';
-import { trackError, trackLoadModel } from '@/utilities/metrics';
+import { trackError, trackEvent, trackLoadModel, trackNavigation } from '@/utilities/metrics';
 import { NodeAppearanceProvider, CadNode } from '@/datamodels/cad';
 import { RenderMode } from '@/datamodels/cad/rendering/RenderMode';
 import { EffectRenderManager } from '@/datamodels/cad/rendering/EffectRenderManager';
@@ -87,6 +87,7 @@ export class RevealManager<TModelIdentifier> {
       this._lastCamera.zoom = camera.zoom;
 
       this._cadManager.updateCamera(camera);
+      trackNavigation({ moduleName: 'RevealManager', methodName: 'update' });
     }
   }
 
--- a/viewer/src/public/RevealManager.ts
+++ b/viewer/src/public/RevealManager.ts
@@ -155,11 +155,13 @@ export class RevealManager<TModelIdentifier> {
   ): void {
     switch (event) {
       case 'loadingStateChanged':
-        this.eventListeners.loadingStateChanged.filter(x => x !== listener);
+        this.eventListeners.loadingStateChanged = this.eventListeners.loadingStateChanged.filter(x => x !== listener);
         break;
 
       case 'nodeIdToTreeIndexMapLoaded':
-        this.eventListeners.sectorNodeIdToTreeIndexMapLoaded.filter(x => x !== listener);
+        this.eventListeners.sectorNodeIdToTreeIndexMapLoaded = this.eventListeners.sectorNodeIdToTreeIndexMapLoaded.filter(
+          x => x !== listener
+        );
         break;
 
       default:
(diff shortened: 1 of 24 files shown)


viewer/src/public/RevealManager.ts
```diff
@@ -51,6 +53,15 @@ export class RevealManager<TModelIdentifier> {
     this._pointCloudManager = pointCloudManager;
     this._effectRenderManager = new EffectRenderManager(this._cadManager.materialManager);
     this.initLoadingStateObserver(this._cadManager, this._pointCloudManager);
+    this._updateSubject = new Subject();
+    this._updateSubject
+      .pipe(
+        auditTime(5000),
+        tap(() => {
+          trackNavigation({ moduleName: 'RevealManager', methodName: 'update' });
```
maksnester, line 61: It’s a bit confusing that we send method name update which doesn’t exist. I suggest to create that method and put tracking updates there.

viewer/src/public/createRevealManager.ts
```diff
@@ -48,7 +48,8 @@ export function createRevealManager<T>(
   client: ModelDataClient<T>,
   revealOptions: RevealOptions = {}
 ): RevealManager<T> {
-  initMetrics(revealOptions.logMetrics !== false, project, {
+  const applicationId = client.getApplicationIdentifier();
+  initMetrics(revealOptions.logMetrics !== false, project, applicationId, {
     moduleName: 'createRevealManager',
     methodName: 'createRevealManager',
     constructorOptions: omit(revealOptions, ['internal'])
```
maksnester, line 55: I guess we can remove omitting here since you disabled metrics for examples. And just pass options without omitting.

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -422,19 +407,6 @@ export class Cognite3DViewer {
    * ```
    */
   async addPointCloudModel(options: AddModelOptions): Promise<CognitePointCloudModel> {
-    trackLoadModel(
```
maksnester, line 425: Why don’t you want to track models loading anymore? Just wondering. Useless?

viewer/src/utilities/networking/types.ts
```diff
@@ -47,4 +47,9 @@ export interface ModelDataClient<TModelIdentifier>
     ModelCameraConfigurationProvider<TModelIdentifier>,
     JsonFileProvider,
     BinaryFileProvider,
-    HttpHeadersProvider {}
+    HttpHeadersProvider {
+  /**
+   * Returns an identifier that can be used to identify the application Reveal is used in.
```
maksnester, line 52: Does identifier tell more than route? Why to have app Id at all when we have url tracker?

viewer/src/utilities/metrics.ts
```diff
@@ -4,7 +4,7 @@
 
 import mixpanel from 'mixpanel-browser';
 
-type TrackedEvents = 'init' | 'loadModel' | 'error';
+type TrackedEvents = 'init' | 'construct3dViewer' | 'loadModel' | 'error' | 'navigated';
```
maksnester, line 7: Sorry if I just missed it (reviewing from phone) but how init differs from construct and why we need navigate ? What does it do at all? Navigate feels like app level metric if it’s about routes. 🤷‍♂️

maksnester: Aha, navigation is session update. A bit tricky :)

viewer/src/public/RevealManager.ts
j-bjorne: Technically it is tracking update, but throttled every 5000ms. Update is a function further down where _updateSubject.next() is called.

viewer/src/utilities/metrics.ts
j-bjorne: do we have a better word?

viewer/src/utilities/metrics.ts
```diff
@@ -65,3 +68,7 @@ export function trackError(error: Error, eventProps: EventProps) {
     ...eventProps
   });
 }
+
+export function trackNavigation(eventProps: EventProps) {
```
j-bjorne, line 72: Like mentioned, can be confused with Camera navigation

j-bjorne (approved):
Approved with some nitpick comments

viewer/src/public/RevealManager.ts
larsmoa: Yeah, it's a bit confusing - but the code sets up a RxJS issue that is _triggered_ from `update()` so the event comes from the `update`-function. Makes sense?

viewer/src/public/migration/Cognite3DViewer.ts
larsmoa: No - we're tracking them also in `RevealManager` so this just caused duplicated events for the same models.

viewer/src/utilities/networking/types.ts
larsmoa: It's the `CogniteClient`-`appId` provided on construction of the client SDK. Maybe it's a bit overlapping, but I was thinking you could have an app that had many different URLs (e.g. staging, dev, prod etc)

PR #1250, opened 2020-10-23
chore: add rule for detecting unused import in eslint

larsmoa:


viewer/.eslintrc.js
```diff
@@ -14,7 +14,8 @@ module.exports = {
     ecmaVersion: 2020, // Allows for the parsing of modern ECMAScript features
     sourceType: 'module' // Allows for the use of imports
   },
-  plugins: ['header', '@typescript-eslint', 'jsdoc'],
+  plugins: ['header', '@typescript-eslint', 'jsdoc', 'unused-imports'],
```
maksnester, line 17: Let’s set it as warn. It will be less annoying then, but precommit hook still will catch it

maksnester (approved):
LGTM but I suggest to use warn instead of error

maksnester:
Meh, can’t see labels on pr from phone screen

viewer/.eslintrc.js
larsmoa: The thing is that compile will fail in CI if there are unused imports, @maksnester  - so I actually think it should be an error (if we don't change that in test compile).

PR #1252, opened 2020-10-23
fix: avoid sending data to MixPanel that can be used to identify users

larsmoa:
Avoid sending user identifiable information to MixPanel to comply with GDPR.


maksnester (approved):
@larsmoa approved but dropped a message to the slack discussion.

PR #1285, opened 2020-10-28
chore: update ThreeJS to 0.122.0

larsmoa:


examples/package.json
```diff
@@ -36,7 +36,7 @@
     "react-dom": "^17.0.1",
     "react-router-dom": "^5.2.0",
     "styled-components": "^5.2.0",
-    "three": "link:../viewer/node_modules/three"
+    "three": "^0.122.0"
```
maksnester, line 39: @larsmoa why to remove the link? We just will run into conflicts between versions as we had before.

maksnester (changes requested):
https://github.com/cognitedata/reveal/pull/1285/files#r514138383

cc @larsmoa

examples/package.json
larsmoa: @maksnester : oops, should we have a similar link in documentation btw?

maksnester: Sure, we link reveal and sdk already. Probably make sense to do the same for three.js too.

maksnester (approved):

PR #1287, opened 2020-10-28
chore: retry preview documentation deploy if fails

larsmoa:
Experiment for now - will see if this works like I hope.


maksnester (approved):
I really don't understand much from this workflow, so I just trust your judgment on that :D Though from the beginning I told that it would be easier/better to set up pr-previews in our cognite infrastructure instead of (ab)using GitHub pages for that. But it would also take some time to figure out how exactly to do that.

PR #1293, opened 2020-11-02
feat: expose CAD budget in Cognite3dViewer to allow controlling model budget.

larsmoa:


viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -119,6 +120,28 @@ export class Cognite3DViewer {
...
+  }
+
+  /**
+   * Sets the current budget for downloading geometry for CAD models. Note that this
+   * budget is shared between all added CAD models and not a per-model budget.
+   * @version New in 1.2.0
+   */
+  public set cadBudget(budget: CadModelBudget) {
+    // Note! Type here differes from the one in RevealManager to expose a documentated
+    // type. This should map 1:1 with type in RevealManager
```
maksnester, line 141: Why RevealManager can't use the same documented type instead? 1:1 map described in comment feels unreliable to me.

maksnester (changes requested):

viewer/src/datamodels/cad/CadModelBudget.ts
```diff
@@ -22,8 +24,16 @@ export type CadModelSectorBudget = {
   readonly maximumNumberOfDrawCalls: number;
 };
 
-export const defaultCadModelSectorBudget = {
-  highDetailProximityThreshold: 10,
-  geometryDownloadSizeBytes: 35 * 1024 * 1024,
-  maximumNumberOfDrawCalls: 1000
-};
+export const defaultCadModelSectorBudget = isMobileOrTablet()
```
maksnester, line 27:
```suggestion
export const defaultCadModelSectorBudget: CadModelBudget = isMobileOrTablet()
```

Here should be a type. CadModelBudget or CadModelSectorBudget not really sure.

viewer/src/public/migration/Cognite3DViewer.ts
larsmoa: Because `RevealManager` is not allowed to import from `migration` - we're considering `migration` something that can import anything, but can't be imported outside `migration` - to keep things tidy...

Not sure how to have proper API reference docs without resorting to this hack... Any idea, @maksnester ?

maksnester: > we're considering migration something that can import anything, but can't be imported outside migration - to keep things tidy...

@larsmoa to be honest, it doesn't make any sense to me... I also raised related question in my [other PR](https://github.com/cognitedata/reveal/pull/1296).

Let's face the facts: 

* you have a type that belongs to public API (you use it in Cognite3DViewer)
* you want 100% the same type in RevealManager - why you can't just import it? I would agree that probably it's a bad idea to import experimental API into the public. But why not to import the public to experimental? 🤷‍♂️

maksnester: I can be missing something, but it looks like it would be perfectly fine to use CadModelBudget from migration/types everywhere. Which is public and documented.

larsmoa: If RevealManager imports from migration and migration wrapper imports RevealManager it would be a circular dependency between the packages... Feels wrong to me - is htis considered ok in JavaScript?

maksnester: @larsmoa No, circular deps are not ok anywhere. But I don't really understand why it should happen. Perhaps, some types/files rearranging must be done?

Anyway. I'm fine with skipping it for now since it's not that easy and not so urgent.

maksnester (approved):

PR #1307, opened 2020-11-17
fix: support for picking on touch devices

larsmoa:


maksnester (commented):
Nothing from my side, but I think @christjt should take a look also.

christjt (approved):

PR #1312, opened 2020-11-21
feat: allow developers to override anti-aliasing mode and enable multisampling with WebGL2

larsmoa:
Enable multisample antialiasing (MSAA) on WebGL 2 through an option in `Cognite3DViewer`. This yields much better image quality at some minor performance cost. For now it's not enabled by default (other than in the documentation site).

With only FXAA:
image

With MSAAx4 + FXAA:
image


viewer/src/__tests__/types.ts
```diff
@@ -0,0 +1,12 @@
+/*!
+ * Copyright 2020 Cognite AS
+ */
+
+/**
+ * Retrieves the property type of a single property of a type.
+ */
+export type PropType<TObj, TProp extends keyof TObj> = TObj[TProp];
```
maksnester, line 8: I think there is something built-in in ts for that.

https://stackoverflow.com/a/41075862/2670121

Just tried it. `id` is a number so I got a type error as expected. 

image

larsmoa: Isn't that exactly what the `PropType` above does? I just feel it's a bit more easy to read, but it might just be me

maksnester:
@christjt that's for you 🙂

viewer/src/__tests__/types.ts
maksnester: @larsmoa I don't mind you keep it if you like it. I just probably wouldn't ever use it, but it's totally up to you.

christjt (approved):
LGTM 👌

larsmoa:
Will await merge untill you're changes are merged @christjt - here be conflicts

PR #1314, opened 2020-11-30
[REV-61] fix: switch pointSizeType to Fixed when pointSize setter is called. Add public accessories for pointSizeType

maksnester:
Before

`model.pointSize = 10` has no effect (you can verify it [here](https://cognitedata.github.io/reveal-docs/docs/examples/pointcloud#point-size))

After

It's now possible to update `pointSizeType` to make `pointSize` setter actually apply new size.


christjt (approved):

maksnester:
We discussed that it should wait for further investigation about why the `pointSize` doesn't "just work" now.

larsmoa:
Underlying root issue will be fixed by #1570, making this void.

PR #1323, opened 2020-12-15
fix: bug causing assertion when root sector is missing F3D geometry [REV-69]

larsmoa:


viewer/src/datamodels/cad/sector/culling/TakenSectorTree.ts
```diff
@@ -41,7 +41,9 @@ export class TakenSectorTree {
       };
       return true;
     });
-    this.setSectorLod(sectorRoot.id, LevelOfDetail.Simple);
+    if (sectorRoot.facesFile.fileName !== null) {
```
maksnester, line 44:
```suggestion
    if (sectorRoot.facesFile.fileName) {
```
Hopefully, it can be only null, but I think it's a rule of thumb to check for empty values with a simple falsy-check condition. You ditch out `undefined` and `''` in this way. Feels safer.

maksnester (approved):
approved but with a suggested change

PR #1324, opened 2020-12-15
feat: fitCameraToModel now uses K-means clustering for determining bounds of CAD models

larsmoa:
This avoids many cases where fitToModel() zooms out to an "empty view" because there are junk geometry far away from the main model.

Examples (left is before, right after):

image

### DIFF
--- a/examples/src/pages/Migration.tsx
+++ b/examples/src/pages/Migration.tsx
@@ -122,7 +122,11 @@ export function Migration() {
           addModel({
             modelId: guiState.modelId,
             revisionId: guiState.revisionId,
-          })
+          }),
+        fitToModel: () => {
+          const model = cadModels[0] || pointCloudModels[0];
+          viewer.fitCameraToModel(model);
+        }
       };
 
       const settingsGui = gui.addFolder('settings');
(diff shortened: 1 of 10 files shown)


maksnester (approved):
@larsmoa It looks good to me, however, I think it might be a good idea to set that boolean flag as `true` by default. Meaning we will change the default behavior to fix the issues with huge bounding boxes. And in rare cases when this behavior is not needed - people can pass `false` explicitly. 

I just don't know if people that use reveal very carefully read our release notes and will do any changes. I will use it for cdf for sure, but I think for most projects it will just remain unfixed.

larsmoa:
> @larsmoa It looks good to me, however, I think it might be a good idea to set that boolean flag as `true` by default. Meaning we will change the default behavior to fix the issues with huge bounding boxes. And in rare cases when this behavior is not needed - people can pass `false` explicitly.
> 
> I just don't know if people that use reveal very carefully read our release notes and will do any changes. I will use it for cdf for sure, but I think for most projects it will just remain unfixed.

Actually that might break automatic near/far planes of camera and the function is considerably slower than the previous one. I think this is something we want to make explicit at least for now.

The user rarely will use this function, but will use `Cognite3dViewer.fitToModel` which sets the parameter to true and will have the fix applied without changing the code.

maksnester:
@larsmoa yes, you're right, I just didn't read through the code carefully. Didn't realize at first that you changed getModelBoundingBox, not fitCamera itself.

PR #1326, opened 2020-12-16
chore: report NPM package size in CI builds

larsmoa:


maksnester (approved):
Cool. 

But it's a pity that we can't reuse the same build though and have to run the same build again for that action.

Maybe we should remove the common build:prod step at all then? In case of failure we should get the error from that new size step. 

@larsmoa I removed auto-merge label

PR #1329, opened 2020-12-17
chore: retry visual tests on failure to make it less prone to errors

larsmoa:


maksnester (approved):
@larsmoa I actually started to work on visual tests yesterday and also added retries but on tests level. But PR is not ready yet. I just started on tests yesterday. Added HTTPS support for them and retries, changed run scripts a bit. Perhaps I also found the source of flakiness (loading div detection - noticed errors that were pointing on it), but that's not for sure at all yet. 

Anyway, just saying that I'm working on it too, but I don't mind if we merge it until I'll have it done.

PR #1331, opened 2020-12-18
chore: fix condition for retrying preview doc deploy

larsmoa:


.github/workflows/preview-documentation.yml
```diff
@@ -96,7 +96,7 @@ jobs:
 
       - name: Retry deploy 🚀 - pull
         working-directory: reveal-docs-preview
-        if: failure()
+        if: steps.deploy.outcome=='failure'
```
maksnester, line 99: @larsmoa Do you have some idea about how that `failure()` works at all? I've seen it in UI tests workflow and was wondering if it's something specific action related or should be available in general...

maksnester (approved):

PR #1341, opened 2021-01-06
[REV-60] refactor ui tests, add very basic pointcloud test

maksnester:
* removed retries of the whole CI job in favor of test case retry
* split single file with all the tests (aka Testable.tsx) to lots of `*TestPage.tsx` files under `examples/src/pages/e2e`. They contain only a small amount of code that is necessary to perform a test. 
* added a very basic pointcloud test (just renders) - _I think it's not that useful at the moment, the model is barely visible. @larsmoa IDK why there are so few points in that model that you sent me, but I can't affect on it :( If you can, please replace it with some better model someday._ 

I hope it will be easier to add new tests now in comparison with writing everything in one big file with all the tests. Let me know if there is something confusing or missing.


larsmoa:
Adding do-not-merge label until 1.3.1 has been released for good measure

larsmoa (changes requested):
A very welcome change - great! However, all seems to produce a blank result.

Also, I'm thinking if we should simplify `TestViewer` be a bare minimum DOMElement that is injected into the test pages. Currently we are only able to test non-Cognite3DViewer-stuff, so I think this approach maybe would enable testing that too? Ofc it would lead to some code duplication in viewer initialisation, but that could be extracted to a utility function. This isn't something that necessary needs to happen right away, but I think its the next step

maksnester:
> However, all seems to produce a blank result.

That's because `loadingStateChange` event is broken after the latest revert. So actually tests caught that :) 

> Currently we are only able to test non-Cognite3DViewer-stuff

I also unsure why revealManager was used instead of cognite3dviewer but I believe it's because cognite3dviewer doesn't (wasn't?) support local models. I just tried to keep tests 100% the same as they were. But yes, I agree that it make sense to use Cognite3DViewer more. 

@larsmoa

larsmoa:
The point about local models is a good one, will have to think about how to solve that...

larsmoa (approved):

PR #1362, opened 2021-01-24
feat: utility functions to manage HTML overlays in Cognite3dViewer [REV-81]

larsmoa:


viewer/tsconfig.json
```diff
@@ -26,6 +26,6 @@
     // do not add path aliases. They'll end up in generated declaration files.
   },
   // include is also used for eslint to understand which files should be linted
-  "include": ["*.js", "*.ts", "src/**/*", ".eslintrc.js", "script", "src/@types", "src/utilities/disposeAttributeArrayOnUpload.ts"],
+  "include": ["*.js", "*.ts", "src/**/*", ".eslintrc.js", "script", "src/@types", "src/utilities/disposeAttributeArrayOnUpload.ts", "tools"],
```
maksnester, line 29: but it should be included in src/**/* are you sure it affects anything?

viewer/jest.config.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     '^.+\\.tsx?$': 'ts-jest'
   },
   testRegex: '(.*\\.test\\..*|\\.(test|spec|Test))\\.tsx?$',
-  moduleDirectories: ['node_modules'],
+  moduleDirectories: ['node_modules', '<rootDir>', '<rootDir>/src', '<rootDir>/src/tools'],
```
maksnester, line 14:
```suggestion
```

https://jestjs.io/docs/en/configuration#moduledirectories-arraystring

viewer/tsconfig.json
maksnester, line 29:
```suggestion
  "include": ["*.js", "*.ts", "src/**/*", ".eslintrc.js", "script"],
```

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1042,6 +1045,82 @@ export class Cognite3DViewer {
...
+  /**
+   * Creates a tool for attaching HTML elements to a 3D position and updates it's
+   * position/visibility as user moves the camera. This is useful to create HTML
+   * overlays to highlight information about key positions in the 3D model.
+   *
+   * Attached elements *must* have CSS style 'position: absolute'. It's also recommended
+   * in most cases to have styles 'pointerEvents: none' and 'touchAction: none' to avoid
+   * interfering with 3D navigation. Consider also applying 'transform: translate(-50%, -50%)'
+   * to anchor the center of the element rather than the top-left corner. In some cases the
+   * `zIndex`-attribute is necessary for the element to appear on top of the viewer.
```
maksnester, line 1057: you described z-index but don't set it below. I would actually suggest having as little CSS as possible in examples. Especially if you provide tips in plain text. Or vice versa, move the plain text to be comments for code. That will give you less inconsistency between duplicated information. I would say it's better to use code comments (easier to copy-paste the same to the docs) and doesn't require additional explanation.

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1042,6 +1045,82 @@ export class Cognite3DViewer {
...
+  /**
+   * Registers a new tool. This makes the tool get updates from the viewer.
+   * The tool will be unregistered when it's diposed.
+   * @param tool Tool to register.
+   */
+  private registerTool(tool: Cognite3DViewerTool): void {
+    if (this._tools.has(tool)) {
+      throw new Error(`Tool ${tool} is already registered`);
+    }
+    const onDisposed: DisposedDelegate = () => this.unregisterTool(tool);
```
maksnester, line 1095: why do you need that DisposedDelegate type at all?

viewer/src/tools/HtmlOverlayTool.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    // Act & Assert
+    expect(() => helper.add(htmlElement, position)).toThrowError();
+  });
+
+  test('notifyRendered() places overlays correctly', () => {
+    // Arrange
+    const helper = new HtmlOverlayTool(canvasContainer, renderer, camera);
+    const htmlElement = document.createElement('div');
+    htmlElement.style.position = 'absolute';
+    expect(htmlElement.style.top).toBeEmpty();
```
maksnester, line 47: looks like assertions against jsdom

documentation/docs/examples/cad-2doverlay.mdx
```diff
@@ -13,118 +13,146 @@ that follows the object it's attached to. Note that these overlays always will b
...
+  const pos = new THREE.Vector3(
+      bounds.min.x + (bounds.max.x - bounds.min.x) * s[0],
+      bounds.min.y + (bounds.max.y - bounds.min.y) * s[1],
+      bounds.min.z + (bounds.max.z - bounds.min.z) * s[2]
+    );
+  overlays.add(createOverlay(), pos);
+}
+
+// Dispose all overlays after 10 seconds
+setTimeout(() => overlays.dispose(), 10000);
```
maksnester, line 157: if the viewer got disposed would overlays be disposed too by the viewer? I think they should be. And I think it would be convenient if disposal would destroy all created HTML elements.

maksnester (changes requested):
@larsmoa  at least requesting changes for suggestions to be committed. Config changes seem irrelevant. 

How did you fix your circular dependency issue btw? What was wrong? I don't think you made it work by changing those configs. 

About the new API. I think it's an interesting choice to manage additional tools, to avoid Viewer cluttering. Good job. Though in my opinion, I think biderectional connection for disposal might be good thing. Not only tools should notify the viewer that they are disposed, but the viewer should destroy those tools which are not relevant anymore after the viewer is disposed.

larshagencognite:
> @larshagencognite at least requesting changes for suggestions to be committed. Config changes seem irrelevant.

Did you mean to tag another Lars?

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1042,6 +1045,82 @@ export class Cognite3DViewer {
...
+      throw new Error(`Tool ${tool} is not registered`);
+    }
+    tool.off('disposed', onDisposed);
+    this._tools.delete(tool);
+  }
+
+  /**
+   * Notifies all registered tools that there has been rendered a frame.
+   */
+  private notifyToolsAboutRendering() {
```
maksnester, line 1118: As a second thought on the API, does it make sense for viewer to know about tools in that way? I mean, maybe instead tools should know about the viewer and subscribe on its 'rendered' or 'disposed' or other events?

maksnester:
@larshagencognite yes, sorry, I edited the comment but your quick reaction on mentions in PRs is impressive 🙂 👍

viewer/src/public/migration/Cognite3DViewer.ts
larsmoa: Maybe its a bit overkill, but I like the idea that you can change the signature one place and have it be updated automatically or result in at least better error messages for references that doesn't work.

documentation/docs/examples/cad-2doverlay.mdx
```diff
@@ -13,118 +13,146 @@ that follows the object it's attached to. Note that these overlays always will b
...
+:::note
+There should be a limited number of HTML overlays as it can impact performance negatively if there's too many.
+:::
+
 <DemoWrapper name="Cognite3DViewerDemo" />
 
 Attaching HTML elements to 3D objects/positions requires the following operations:
 
-1. Create a HTML element and place it as a child of the viewer DOM-element (i.e. `Cognite3DViewer.domElement`)
+1. Create a HTML element and apply style `position: absolute` to the element. There also some other recommended
```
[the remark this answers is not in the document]
larsmoa: Both are actually correct :) https://english.stackexchange.com/a/25746

viewer/src/public/migration/Cognite3DViewer.ts
larsmoa: > @larsmoa at least requesting changes for suggestions to be committed. Config changes seem irrelevant.
> 
> How did you fix your circular dependency issue btw? What was wrong? I don't think you made it work by changing those configs.

No, the configs was things I tested and forgot to remove - sorry. I fixed the error using `madge` to create a graph of dependencies and found some circular stuff. Never really understood exactly what the issue was, but fixed by changing imports to ensure the circual deps was removed.

> 
> About the new API. I think it's an interesting choice to manage additional tools, to avoid Viewer cluttering. Good job. Though in my opinion, I think biderectional connection for disposal might be good thing. Not only tools should notify the viewer that they are disposed, but the viewer should destroy those tools which are not relevant anymore after the viewer is disposed.

viewer/src/public/migration/Cognite3DViewer.ts
larsmoa: This has now been reversed - tools know about viewer, not the other way around.

documentation/docs/examples/cad-2doverlay.mdx
```diff
@@ -13,118 +13,152 @@ that follows the object it's attached to. Note that these overlays always will b
 they will not be hidden when the object is obscured by other objects.
 To achieve this it's necessary to use custom 3D objects.
 
+:::note
+There should be a limited number of HTML overlays as it can impact performance negatively if there's too many.
```
christjt, line 17: It is not clear to me if this implies that one should limit the number of HTML elements added to a HTMLOverlayTool, or number of HTMLOverlayTool instances (or both).

viewer/src/tools/Cognite3DViewerToolBase.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+import { assertNever } from '../utilities/assertNever';
+import { Cognite3DViewerTool, DisposedDelegate } from '../public/migration/Cognite3DViewerTool';
+
+/**
+ * Base class for tools attaching to a {@see Cognite3DViewer}.
+ * @internal
+ */
+export abstract class Cognite3DViewerToolBase implements Cognite3DViewerTool {
```
christjt, line 12: This just looks off, to me. base is implementing not base? Shouldn't it be` Cognite3DViewerTool implements Cognite3DViewToolBase`?

christjt (changes requested):
Looks really nice @larsmoa. Love the de-coupling from the viewer, and seems really easy to use. Well done!

documentation/docs/examples/cad-2doverlay.mdx
larsmoa: Clarified.

viewer/src/tools/Cognite3DViewerToolBase.ts
larsmoa: Interface removed

christjt (approved):

viewer/webpack.config.js
```diff
@@ -31,6 +31,7 @@ module.exports = env => {
     mode: development ? 'development' : 'production',
     entry: {
       index: './src/index.ts',
+      tools: './src/tools/index.ts',
```
maksnester, line 34:
```js
import { foo } from '@cognite/reveal/tools'
```
Should we have some discussion about this thing? I don't understand the benefits of this. 

These tools will be just a subset of higher-level APIs for the reveal, I'm not sure there is a point to split them into a different package of any kind.

You're basically creating a package inside a package but without proper versioning (versioning is directly bound to reveal, so it's not really a different package or something). It is ok for experimental since nobody really needs it except us, but it feels dubious to me having that for a public API.

A simple alternative for the sake of having lots of tools with nice autocomplete and types – you might create some factory, e.g.

```js
import { RevealTools, Cognite3DViewer } from '@cognite/reveal'

const overlays = RevealTools.create('htmlOverlay', { viewer }); 
// ...
overlays.add(createOverlay(), pos);
```

I would also ask @j-bjorne and @christjt to give some input on that.

viewer/src/tools/HtmlOverlayTool.ts
```diff
@@ -0,0 +1,216 @@
...
+  /**
+   * Registers a HTML overlay that will be updated on rendering.
+   *
+   * @param htmlElement
+   * @param position3D
+   * @param options
+   */
+  add(htmlElement: HTMLElement, position3D: THREE.Vector3, options: HtmlOverlayOptions = {}) {
+    this.ensureNotDisposed();
+    if (htmlElement.style.position !== 'absolute') {
```
maksnester, line 122: It will give a false alarm in case if you have some HTML element styled with css (not style property). In that case, you need computed styles.

```css
.label { 
  position: absolute;
  /* other styles */
}
```

```js
createOverlayHtmlElement() {
  let el = document.createElement('div');
  el.classList.add('label');
  return el
}

**Review by maksnester — CHANGES_REQUESTED** [review]

File: viewer/webpack.config.js (hunk 11 above)
Comment (larsmoa) [reply viewer/webpack.config.js:34]: How do `rxjs` do this? Here we import operators through `rxjs/operators`, other classes through `rxjs`.

If we are to ditch tools, I think we only will go for `import { HtmlOverlayTool } from '@cognite/reveal'` for now and see if we want to add a factory later.

File: viewer/src/tools/HtmlOverlayTool.ts (hunk 12 above)
Comment (larsmoa) [reply viewer/src/tools/HtmlOverlayTool.ts:122]: Oh, thx

File: viewer/webpack.config.js (hunk 11 above)
Comment (maksnester) [reply viewer/webpack.config.js]: rxjs terrifies me as well 😂  

but yeah, I agree some people do this and it is an option, just trying to understand real pros and cons of that approach.

What's for 

> import { HtmlOverlayTool } from '@cognite/reveal'

I don't agree, because we know upfront that we'll have more tools like this, and you can't just add the factory later and remove that possibility of import things (because of breaking changes). And keep 2 ways of doing things probably is not something you want.

Comment (maksnester) [reply viewer/webpack.config.js]: @larsmoa my biggest concern about that subentry was code duplication between entries, so I built viewer and examples app, then explored with the source-map-explorer tool. I can't see any code duplication for now between entries, all looks fine. 

The only issue I've seen is that types aren't picked up by IDE. I can see viewer build output is `/dist/tools/index.d.ts` + `/dist/tools.js` so if you want to go with sub-entry, it's probably better to move index ts out of the `/src/tools` to have generated js and its types aligned.

image

File: viewer/src/tools/HtmlOverlayTool.test.ts [hunk 13]
```diff
@@ -0,0 +1,154 @@
...
+      behindFarPlaneElement,
+      expect.any(THREE.Vector2),
+      expect.any(THREE.Vector3),
+      expect.anything()
+    );
+    expect(options.positionUpdatedCallback).toHaveBeenCalledWith(
+      withinNearAndFarPlaneElement,
+      expect.any(THREE.Vector2),
+      expect.any(THREE.Vector3),
+      expect.anything()
```
maksnester, line 133: 😮  but you don't assert anything (pun not intended) here

https://jestjs.io/docs/en/expect#expectanything

Or am I just misunderstanding what it does here?

viewer/src/tools/HtmlOverlayTool.ts
```diff
@@ -0,0 +1,220 @@
...
+};
+
+type HtmlOverlayElement = {
+  position3D: THREE.Vector3;
+  options: HtmlOverlayOptions;
+};
+
+/**
+ *
+ */
```
maksnester, line 30:
```suggestion
```

maksnester (commented):
Added a couple of docs nitpicks and one ts related suggestion for events.

documentation/docs/examples/cad-2doverlay.mdx
```diff
@@ -13,118 +13,152 @@ that follows the object it's attached to. Note that these overlays always will b
...
-    overlay.style.display = 'block';
-    overlay.style.position = 'absolute';
-  } else {
-    // Outside the screen
-    overlay.style.display = 'none';
-  }
+It's also possible to apply styling based on the position of the 
+overlay, e.g. to vary the opacity of the overlay based on distance 
+from the camera. To do this, provide an `options`-argument with a 
+`positionUpdatedCallback` to `Cognite3DViewer.attachHtmlOverlay`. This
```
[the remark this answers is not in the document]
larsmoa: Oh thx - old stuff. Will fix.

viewer/src/tools/HtmlOverlayTool.test.ts
larsmoa: it asserts that the function is called with `withinNearAndFarPlaneElement` as the first argument - i.e. it expects that the callback is triggered for all registered elements. I agree it was a bit cumbersome way to verify it though

maksnester: oh God, I totally missed `toHaveBeenCalledWith(`, sorry

maksnester: oh God, I just totally missed `toHaveBeenCalledWith(`, sorry

maksnester (approved):
I'm actually fine with that being merged, just consider applying suggestions here and there. 

Nicely done 👍

PR #1369, opened 2021-02-02
use cognite-cdn and update docs (no jsdeliver anywhere)

maksnester:


documentation/docs/installation.mdx
```diff
@@ -60,89 +71,16 @@ revealEnv.publicPath = `/reveal-worker/`;
...
-  ]
-}>
-<TabItem value="bash">
-
-```bash
-PUBLIC_PATH="https://static.server/parser-worker/" yarn build:prod
-```
-</TabItem>
-<TabItem value="win">
+Below are few more convenient options for copying if you don't like
```
larsmoa, line 77: Don't like what?

larsmoa (changes requested):
One sentence seems to end mid-sentence. Also, examples are broken due to missing parser worker (need to deploy to new CDN?)

documentation/docs/installation.mdx
maksnester: it has the code snippet right after the sentence, but I rephrased it a bit

image

looks ok?

larsmoa (approved):

PR #1372, opened 2021-02-05
feat: add ssao shading to render pipeline

christjt:


viewer/src/glsl/post-processing/pure-depth-ssao.frag
```diff
@@ -0,0 +1,107 @@
...
+  float verticalSampleDepth = verticalSampleCondition ? d2 : d4;
+  vec2 verticalSampleUv = verticalSampleCondition ? uv2 : uv4;
+
+  vec3 viewPos = viewPosFromDepth(sampleDepth, vUv);
+  
+  vec3 viewPos1 = (horizontalSampleCondition == verticalSampleCondition) ? viewPosFromDepth(horizontalSampleDepth, horizontalSampleUv) : viewPosFromDepth(verticalSampleDepth, verticalSampleUv);
+  vec3 viewPos2 = (horizontalSampleCondition == verticalSampleCondition) ? viewPosFromDepth(verticalSampleDepth, verticalSampleUv) : viewPosFromDepth(horizontalSampleDepth, horizontalSampleUv);
+
+  return normalize(cross(viewPos1 - viewPos, viewPos2 - viewPos));
+}
```
larsmoa, line 63: Should this and `viewPosFromDepth` be glslify-imports? Not sure - you decide.

viewer/src/glsl/post-processing/pure-depth-ssao.frag
```diff
@@ -0,0 +1,107 @@
...
+
+  vec3 bitangent = cross(viewNormal, tangent);
+
+  mat3 TBN = mat3(tangent, bitangent, viewNormal);
+
+  float occlusion = 0.0;
+
+  for (int i = 0; i < MAX_KERNEL_SIZE; i++){
+    
+    if(i >= numberOfSamples){
```
larsmoa, line 85: Isn't `MAX_KERNEL_SIZE` and `numberOfSamples` equal? My guess is that the loop can be unrolled a tiny bit more effectively if the `if` is avoided

viewer/src/glsl/post-processing/pure-depth-ssao.frag
```diff
@@ -0,0 +1,107 @@
...
+    vec4 offset = projMatrix * vec4(sampleVector, 1.0);
+    offset.xyz /= offset.w;
+    offset.xyz = offset.xyz * 0.5 + 0.5;
+
+    float realDepth = texture2D(tDepth, offset.xy).r;
+    vec3 realPos = viewPosFromDepth(realDepth, offset.xy);
+
+    float rangeCheck = smoothstep(0.0, 1.0, sampleRadius / length(viewPosition - realPos));
+
+    occlusion += (realPos.z >= sampleVector.z + bias ? 1.0 : 0.0) * rangeCheck;
```
larsmoa, line 101: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/step.xhtml would avoid conditional here

larsmoa (commented):
Looks great - a few minor comments :)

viewer/src/glsl/post-processing/pure-depth-ssao.frag
christjt: My philosophy on this is that it should be imports when it is needed / used by more than one shader. Or if the shader is so complex that having the abstraction will make it easier to understand

viewer/src/glsl/post-processing/pure-depth-ssao.frag
christjt: Woops, this was trash from earlier iteration where I had a stop condition since I didn't add the define

viewer/src/glsl/post-processing/pure-depth-ssao.frag
christjt: I don't think that the ternary operator used in this way actually creates branching. And even so, isn't this how the step operator is implemented?

larsmoa: Actually I don't think we're at the step where we should dive into if `step()` is faster than ternary operator so disregard this comment.

larsmoa:
SSAO doesn't seem to have any effect on iOS:

Desktop:
image

iOS:
image

I'm fine with disabling this on mobile devices if necessary

larsmoa (approved):

PR #1381, opened 2021-03-04
feat: exploded view tool

christjt:


viewer/src/utilities/index.ts
```diff
@@ -14,6 +14,7 @@ export { LocalModelIdentifier, CdfModelIdentifier } from './networking/types';
 export { createOffsetsArray } from './arrays';
 export { Box3 } from './Box3';
 export { HtmlOverlayTool as HtmlOverlayHelper } from '../tools/HtmlOverlayTool';
+export { ExpandAssetTool } from '../tools';
```
larsmoa, line 17: This should not be exported here - it belongs to `tools.ts` (I've removed the `HtmlOverlayTool` export in a separate PR #1383 )

viewer/src/tools/ExpandAssetTool.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Copyright 2021 Cognite AS
+ */
+import * as THREE from 'three';
+import { Cognite3DModel } from '../public/migration/Cognite3DModel';
+import { Cognite3DViewerToolBase } from './Cognite3DViewerToolBase';
+
+export class ExpandAssetTool extends Cognite3DViewerToolBase {
+  private _cadModel: Cognite3DModel;
+  private _treeBoundingBoxdata!: Promise<{ treeIndex: number; direction: THREE.Vector3; transform: THREE.Matrix4 }[]>;
+  private _rootTreeIndex: number;
```
larsmoa, line 11:
```suggestion
  private readonly _cadModel: Cognite3DModel;
  private _treeBoundingBoxdata!: Promise<{ treeIndex: number; direction: THREE.Vector3; transform: THREE.Matrix4 }[]>;
  private readonly _rootTreeIndex: number;
```

documentation/versioned_docs/version-1.x/examples/cad-explode.mdx
```diff
@@ -0,0 +1,60 @@
...
+The constructor of this tool needs to be supplied with a [tree index](../concepts.md) which is the tree index of the root of the asset, and an instance of the `Cognite3DModel` that the asset belongs to.
+
+:::note
+For performance reasons, the root tree index must have a subtree size of less than 1000, else it will throw an error.
+:::
+
+<DemoWrapper name="Cognite3DViewerDemo" />
+
+`ExpandAssetTool` contains two instance methods: `expand` and `reset`.
+`expand` takes a float describing how far the individual nodes should be expanded (this will very based on the initial vector between the node and root).
```
larsmoa, line 20:
```suggestion
`expand` takes a float describing how far the individual nodes should be expanded (this will vary based on the initial vector between the node and root).
```

documentation/versioned_docs/version-1.x/examples/cad-explode.mdx
```diff
@@ -0,0 +1,60 @@
...
+`reset` resets the tool and any transform that has been applied to any nodes.
+
+:::note
+In addition, an instance of the `ExpandAssetTool` also contains a promise: `readyPromise`, which indicates when the tool is ready to be used.
+The reason you have to await is that initially the tool will make API calls to determine the individual locations of the nodes in the subtree.
```
larsmoa, line 25:
```suggestion
`ExpandAssetTool` contains a promise: `readyPromise`, which indicates when the tool is ready to be used.
This indicates if the tool has cached necessary data from the Cognite Fusion API to determine the individual locations of the nodes in the subtree.
```

larsmoa (commented):
Looks good! I think we should call it "exploded" though - not expanded, as this is the technical term used (see https://en.wikipedia.org/wiki/Exploded-view_drawing).

Also, I'm a bit reluctant towards using "asset" since it might be interpreted as the tool needs asset mappings to work, while plain nodes are just fine. I don't really have any great alternatives here - but the use of the word can be reduced (e.g. call the tool `ExplodedViewTool`)

PR #1401, opened 2021-03-20
improvement: mark discarded sectors as hidden to avoid traversal

larsmoa:
This has close to zero actual effect on render performance.


christjt:
Why do we even care about this anymore? If it is discarded, then there is no geometry, right?

larsmoa:
@christj: In practice it doesn't really matter, but lets say Sector 0/1 is discarded and that the subtree of 0/1 is really deep. This fix would avoid traversing the full subtree by bailing out early.

christjt (approved):

PR #1407, opened 2021-03-23
improvement: replace huge CTM LRU cache with slim most frequently used cache to save memory

larsmoa:
Before:
`{consumedSectorHit: 516, consumedSectorMiss: 3423, ctmHit: 797, ctmMiss: 176}`
Cache size: 123.435 Kb

After:
`{consumedSectorHit: 516, consumedSectorMiss: 3423, ctmHit: 708, ctmMiss: 265}`
Cache size: 10.791 Kb

### DIFF
--- a/viewer/src/utilities/MostFrequentlyUsedCache.ts
+++ b/viewer/src/utilities/MostFrequentlyUsedCache.ts
@@ -21,11 +21,7 @@ export class MostFrequentlyUsedCache<TKey, TValue> {
   }
 
   set(key: TKey, value: TValue): boolean {
-    if (this._cache.has(key)) {
-      this._cache.set(key, value);
-      return true;
-    } else if (this._capacity < this._cache.size) {
-      // Still room
+    if (this._cache.has(key) || this._capacity < this._cache.size) {
       this._cache.set(key, value);
       return true;
     } else {
--- a/viewer/src/utilities/MostFrequentlyUsedCache.ts
+++ b/viewer/src/utilities/MostFrequentlyUsedCache.ts
@@ -38,6 +38,20 @@ export class MostFrequentlyUsedCache<TKey, TValue> {
     }
   }
 
+  remove(key: TKey): boolean {
+    this._retrieves.delete(key);
+    if (this._cache.has(key)) {
+      this._cache.delete(key);
+      return true;
+    }
+    return false;
+  }
+
+  clear() {
+    this._retrieves.clear();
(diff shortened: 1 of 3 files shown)


viewer/src/utilities/MostFrequentlyUsedCache.ts
```diff
@@ -0,0 +1,69 @@
...
+      const keysForRemoval = keys
+        .map(x => ({ key: x, retrivalCount: this._retrieves.get(x) || 0 }))
+        .sort((a, b) => a.retrivalCount - b.retrivalCount)
+        .slice(0, this._cache.size - this._capacity)
+        .map(x => x.key);
+      for (const key of keysForRemoval) {
+        this._cache.delete(key);
+      }
+    }
+  }
```
christjt, line 68:
```suggestion
  private ensureWithinCapacity(): void {
    if (this._capacity >= this._cache.size) return;
    const keys = Array.from(this._cache.keys());
    // Figure out what to remove
    const keysForRemoval = keys
      .map(x => ({ key: x, retrivalCount: this._retrieves.get(x) || 0 }))
      .sort((a, b) => a.retrivalCount - b.retrivalCount)
      .slice(0, this._cache.size - this._capacity)
      .map(x => x.key);
    for (const key of keysForRemoval) {
      this._cache.delete(key);
    }
  }
```

viewer/src/utilities/MostFrequentlyUsedCache.ts
```diff
@@ -0,0 +1,69 @@
...
+      this._cache.set(key, value);
+      return true;
+    } else {
+      // TODO 2020-12-05 larsmoa: Very inefficient way to set value.
+      // We often set a value and discard it a moment later because its not
+      // imporant. Fix this.
+      this._cache.set(key, value);
+      this.ensureWithinCapacity();
+      return this._cache.has(key);
+    }
```
christjt, line 38:
```suggestion
    if (this._cache.has(key) || this._capacity < this._cache.size) {
      this._cache.set(key, value);
      return true;
    } else {
      // TODO 2020-12-05 larsmoa: Very inefficient way to set value.
      // We often set a value and discard it a moment later because its not
      // imporant. Fix this.
      this._cache.set(key, value);
      this.ensureWithinCapacity();
      return this._cache.has(key);
    }
```

christjt (changes requested):

christjt (approved):

PR #1423, opened 2021-04-16
chore: rename repo access token

maksnester:
🤞


maksnester:
we should get rid of access tokens like that

PR #1424, opened 2021-04-19
feat: when a geometry filter is active, cull away (most) geometry outside the filter

larsmoa:
This increases rendering performance and reduces memory footprint.

### DIFF
--- a/parser-worker/createDetailedGeometry.ts
+++ b/parser-worker/createDetailedGeometry.ts
@@ -91,10 +91,10 @@ export async function createDetailedGeometry(
         for (let i = 0; i < fileMeshIndices.length; i++) {
           const meshIdx = meshIndices[fileMeshIndices[i]];
           const treeIndex = treeIndices[meshIdx];
-          const instanceMatrix = instanceMatrices.slice(meshIdx * 16, meshIdx * 16 + 16);
+          const instanceMatrix = instanceMatrices.subarray(meshIdx * 16, meshIdx * 16 + 16);
           instanceMatrixBuffer.set(instanceMatrix, i * 16);
           treeIndicesBuffer[i] = treeIndex;
-          const color = colors.slice(meshIdx * 4, meshIdx * 4 + 4);
+          const color = colors.subarray(meshIdx * 4, meshIdx * 4 + 4);
           colorBuffer.set(color, i * 4);
         }
         instancedMeshes.push({
--- a/parser-worker/createDetailedGeometry.ts
+++ b/parser-worker/createDetailedGeometry.ts
@@ -2,13 +2,11 @@
  * Copyright 2020 Cognite AS
  */
 
-import * as THREE from 'three';
 import { InstancedMesh, InstancedMeshFile, ParseCtmResult, ParseSectorResult, SectorGeometry, TriangleMesh } from "./types";
 
 export async function createDetailedGeometry(
   i3dFile: ParseSectorResult,
-  ctmFiles: Map<string, ParseCtmResult>,
-  geometryClipBox: THREE.Box3 | null
+  ctmFiles: Map<string, ParseCtmResult>
 ): Promise<SectorGeometry> {
   const { instanceMeshes, triangleMeshes } = i3dFile;
 
--- a/parser-worker/createDetailedGeometry.ts
+++ b/parser-worker/createDetailedGeometry.ts
@@ -2,11 +2,13 @@
  * Copyright 2020 Cognite AS
  */
 
+import * as THREE from 'three';
 import { InstancedMesh, InstancedMeshFile, ParseCtmResult, ParseSectorResult, SectorGeometry, TriangleMesh } from "./types";
 
 export async function createDetailedGeometry(
   i3dFile: ParseSectorResult,
-  ctmFiles: Map<string, ParseCtmResult>
+  ctmFiles: Map<string, ParseCtmResult>,
+  geometryClipBox: THREE.Box3 | null
 ): Promise<SectorGeometry> {
   const { instanceMeshes, triangleMeshes } = i3dFile;
 
--- a/viewer/src/datamodels/cad/rendering/EffectRenderManager.ts
+++ b/viewer/src/datamodels/cad/rendering/EffectRenderManager.ts
@@ -547,8 +547,12 @@ export class EffectRenderManager {
     camera: THREE.PerspectiveCamera,
     target: THREE.WebGLRenderTarget | null = this._normalRenderedCadModelTarget
   ) {
-    this._renderer.setRenderTarget(target);
-    this._renderer.render(this._cadScene, camera);
+    try {
+      this._renderer.setRenderTarget(target);
+      this._renderer.render(this._cadScene, camera);
+    } catch (error) {
+      debugger;
+    }
   }
 
   private renderInFrontCadModels(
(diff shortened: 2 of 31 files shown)


parser-worker/createDetailedGeometry.ts
```diff
@@ -91,10 +91,10 @@ export async function createDetailedGeometry(
         for (let i = 0; i < fileMeshIndices.length; i++) {
           const meshIdx = meshIndices[fileMeshIndices[i]];
           const treeIndex = treeIndices[meshIdx];
-          const instanceMatrix = instanceMatrices.slice(meshIdx * 16, meshIdx * 16 + 16);
+          const instanceMatrix = instanceMatrices.subarray(meshIdx * 16, meshIdx * 16 + 16);
```
larsmoa, line 94: This is an optimization, subarray() is a lot faster than slice():
https://jsbench.me/56knoajd9a/1

viewer/src/datamodels/cad/rendering/EffectRenderManager.ts
```diff
@@ -548,8 +548,12 @@ export class EffectRenderManager {
     camera: THREE.PerspectiveCamera,
     target: THREE.WebGLRenderTarget | null = this._normalRenderedCadModelTarget
   ) {
-    this._renderer.setRenderTarget(target);
-    this._renderer.render(this._cadScene, camera);
+    try {
+      this._renderer.setRenderTarget(target);
+      this._renderer.render(this._cadScene, camera);
+    } catch (error) {
+      debugger;
```
christjt, line 555: Lets not have this here :P

viewer/src/datamodels/cad/rendering/primitives.test.ts
```diff
@@ -70,7 +70,7 @@ describe('createPrimitives', () => {
   });
 
   test('no primitives doesnt return any nods', () => {
-    const nodes = Array.from(createPrimitives(emptySector, materials, bounds));
+    const nodes = Array.from(createPrimitives(emptySector, materials, bounds, null));
```
christjt, line 73: Sending in null to a function feels bad. Any way we can have a default here?

viewer/src/datamodels/cad/sector/CachedRepository.ts
```diff
@@ -129,20 +129,21 @@ export class CachedRepository implements Repository {
...
-        this._modelDataTransformer.transform(),
+        map(parsedSector => this._modelDataTransformer.transformSector(parsedSector, wantedSector.geometryClipBox)),
         this.nameGroup(wantedSector),
         map(group => ({ ...wantedSector, group: group.sectorMeshes, instancedMeshes: group.instancedMeshes })),
         shareReplay(1),
         take(1)
-      )
-    );
+      );
+    /*);*/
```
christjt, line 146: Does this do something magic?

viewer/src/datamodels/cad/sector/ModelStateHandler.ts
```diff
@@ -38,6 +38,12 @@ export class ModelStateHandler {
   }
 
   updateState(consumedSector: ConsumedSector) {
+    if (consumedSector === undefined) {
+      debugger;
+    }
+    if (this._sceneModelState === undefined || this._sceneModelState[consumedSector.blobUrl] === undefined) {
+      debugger;
+    }
```
christjt, line 46:
```suggestion
```

viewer/src/datamodels/cad/sector/SimpleAndDetailedToSector3D.ts
```diff
@@ -23,41 +20,32 @@ export class SimpleAndDetailedToSector3D {
...
+          this.materialManager.getModelMaterials(parsedSector.blobUrl)!,
+          geometryClipBox
+        );
+
+      case LevelOfDetail.Discarded:
+        throw new Error('Cannot transform discarded sector');
+
+      default:
+        assertNever(parsedSector.levelOfDetail);
+    }
```
christjt, line 49: Thank you...

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -473,11 +473,16 @@ export class Cognite3DViewer {
...
-      revisionId
-    });
+    const { modelId, revisionId, geometryFilter } = options;
+    const cadNode = await this._revealManager.addModel(
+      'cad',
+      {
+        modelId,
+        revisionId
+      },
+      undefined,
```
christjt, line 483: any way we can void sending in undefined here?

viewer/src/datamodels/cad/sector/CadModelClipper.ts
```diff
@@ -0,0 +1,90 @@
...
+    // Create a clipped sector tree
+    const root = model.scene.root;
+    const newRoot = clipSector(root, this._geometryClipBox);
+    if (newRoot === undefined) {
+      throw new Error('No sectors inside provided geometry clip box');
+    }
+
+    // Create a new clipped scene
+    const sectorMap = new Map<number, SectorMetadata>();
+    traverseDepthFirst(newRoot, x => {
```
christjt, line 29: I assume x is sector or sectorMetadata here, just call it that instead of x

viewer/src/datamodels/cad/rendering/createSimpleGeometryMesh.ts
```diff
@@ -0,0 +1,105 @@
...
+  const attributeByteValues = new Uint8Array(attributeValues.buffer);
+  const filteredByteValues = filterPrimitivesOutsideClipBoxByBaseBoundsAndInstanceMatrix(
+    attributeByteValues,
+    primitiveAttributes,
+    baseBoundingBox,
+    geometryClipBox
+  );
+  if (filteredByteValues.byteLength === 0) {
+    // No data, just skip
+    return undefined;
```
christjt, line 48: Does it make sense to return an empty mesh here instead of undefined?

christjt (changes requested):
Well done! Overall looks very nice. Just a few nits + some questions

viewer/src/datamodels/cad/rendering/primitives.test.ts
larsmoa: Fixed a default argument - I don't want to e.g. use an infinitly big bbox to represent "no filter" because that would cause unnecessary clipping operations that wouldn't cause any data reduction. And undefined is just as bad imo.

viewer/src/datamodels/cad/sector/CachedRepository.ts
larsmoa: Fixed

viewer/src/public/migration/Cognite3DViewer.ts
larsmoa: Fixed by creating a `AddCadModelOptions` type in `RevealManager.

viewer/src/datamodels/cad/rendering/createSimpleGeometryMesh.ts
larsmoa: By using `undefined` it's never added to the scene so there is no traversal before rendering (e.g. for pre-render sorting in ThreeJS). Probably no noticeable performance gain, but feels like the right thing to do for me.

christjt (approved):

PR #1425, opened 2021-04-19
chore: don't generate a new distinct_id for each session to avoid high MixPanel costs

larsmoa:


viewer/src/utilities/metrics.ts
```diff
@@ -77,10 +82,7 @@ export function initMetrics(logMetrics: boolean, project: string, applicationId:
   // Reset device ID (even if we don't send it)
   mixpanel.reset();
 
-  // Use a random identifier because we want to don't track users over multiple sessions to not
-  // violate GDPR. This overrides "distinct_id".
-  const randomIdentifier = generateUuidv4();
-  mixpanel.identify(randomIdentifier);
+  mixpanel.identify(mixpanelDistinctId);
```
maksnester, line 85: https://developer.mixpanel.com/docs/javascript-full-api-reference#mixpanelidentify

I don't think we need it at all.

viewer/src/utilities/metrics.ts
```diff
@@ -26,8 +26,13 @@ let globalLogMetrics = true;
...
   VERSION,
   project: 'unknown',
-  application: 'unknown'
+  application: 'unknown',
+  // Use a random identifier because we want to don't track users over multiple sessions to not
+  // violate GDPR.
+  sessionId: generateUuidv4()
 };
+// Don't identify users in MixPanel to avoid GDPR problems
+const mixpanelDistinctId = 'reveal-single-user';
```
maksnester, line 35:
```suggestion
  application: 'unknown',
};
```

I don't think we really need it. Mixpanel already assigns a random id to a user. https://help.mixpanel.com/hc/en-us/articles/360041039771-Getting-Started-with-Identity-Management#identification-examples

viewer/src/utilities/metrics.ts
maksnester: also, in this file, we have 

```
    disable_cookie: true,
    disable_persistence: true,
```

in the init config. I would say we should remove it. It will solve our problem with the same user recorded multiple times. I don't think it violates anything because we don't identify a user. We only set a cookie (you can also write to localStorage with the same effect) to understand whether it's the same user or not, but we don't know or track anything about the user.

maksnester (changes requested):

viewer/src/utilities/metrics.ts
maksnester: I also checked a few other setups for mixpanel in cognite:

* [route-tacker](https://github.com/cognitedata/cdf-hub-route-tracker/blob/ce00b8791a671e43991bf3315370d0b80905bba1/src/tracker/tracker.ts#L41) - mostly default settings
* [some generic FE metrics pkg](https://github.com/cognitedata/frontend/blob/master/packages/browser/commonjs/metrics/src/metrics.ts) also can't see that they disable anything by default. Aren't we overreact to GDPR here? 

I also read a bit more on the subject... In terms of GDPR mixpanel is a Data Processor, meaning they don't care about your settings to be compatible with GDPR, we (as Data Controller) have full responsibility for collecting user data, mixpanel just gives us a tool to collect data. BUT, we don't provide access to our apps to random people (random people who for some reason use open-sourced reveal can opt-out, it is an app developer responsibility here). I guess our users have to accept some agreement to use our services, which likely should have some point about data collection, that's why we don't have some annoying cookie consent in CDF for example (that's just my assumption). 

Anyway, found some nice [article](https://medium.com/@mschiefelbein/gdpr-and-analytics-tools-51c69fe14dfc) on the matter. Some highlights.

> Identifiable data: GDPR defines personal data as anything that can be used to identify a user — an image, a photo, an email address or an IP.

> Basically, you are the data controller and need to make sure that you:
    1.Collect only the data you need.
    2. Have a clear opt-in and consent process. Users should be made aware of what is collected and given a choice.
    3. Provide a means to control the data. Users should be able to access and request deletion of the data, and be kept informed about any breaches.

I think 2 and 3 are solved since CDF is not a public service and users should sign some agreement to access it?

Also, found a question about anonymous data. https://law.stackexchange.com/questions/29783/gdpr-for-dummies-can-i-track-anonymous-data-can-i-track-how-many-people-accept/29967

> The principles of data protection should therefore not apply to anonymous information, namely information which does not relate to an identified or identifiable natural person or to personal data rendered anonymous in such a manner that the data subject is not or no longer identifiable. This Regulation does not therefore concern the processing of such anonymous information, including for statistical or research purposes.

TL;DR – I'm not a legal expert, but I don't think an anonymous identifier used by default violates anything.

larsmoa: Well, remember that 1) Reveal is an _open_ library, it can be used by anyone with e.g. the publicdata tenant. Cognite is not in control of who the users are and we are not in position to ask for consent. This is why we introduced a random user ID.

The default UUID from MixPanel is stored in cookies, and as far as I understand, will require consent. The problem is that the UUID identifies what user is doing something, and given this UUID, usage history and patterns (e.g. project loaded, when models was loaded, application models was loaded with) it is in theory possible to guesstimate who the _actual_ user is.

It might be that we don't breach GDPR even if we keep the default UUID, but my understanding after taking to a legal expert is that it's a bit borderline and I think therefore we should play it safe.

Note that applications using Reveal (e.g. 3D management app) won't need to adhere to this in the same way.

maksnester: > it's a bit borderline

Well, I'm clearly not a legal expert so I would just avoid further discussions on the matter 🤷‍♂️

maksnester (approved):

PR #1430, opened 2021-04-19
chore: disable clippy lint rule

larsmoa:


f3df/src/lib.rs
```diff
@@ -1,4 +1,8 @@
-#![allow(unused_imports)]
+#![allow(
+    unused_imports,
```
maksnester, line 2: But why do we allow unused imports? Is it a good idea? 🤔

maksnester (approved):

PR #1447, opened 2021-04-26
chore(build): install deps before build when making a release

larsmoa:


maksnester (commented):
@larsmoa it is already handled in bump script, just in case you change `yarn.lock` after running yarn so it will be committed before version bump.

maksnester (approved):
though if you need it for some reason feel free to merge then

larsmoa:
> though if you need it for some reason feel free to merge then

Ah, will skip this then.

PR #1455, opened 2021-05-06
feat: add support for transparency in custom objects

larsmoa:
Combined with regular geometry (0.7 opacity):
image

Combined with ghosted geometry (0.3 opacity):
image

Combined with highlighted object (0.3 opacity):
image


christjt (approved):
Very nice! When making this I for some reason thought this would be very difficult. Well done :D

PR #1456, opened 2021-05-06
feat: Add axis view tool

christjt:


viewer/src/tools/AxisView/AxisViewTool.ts
```diff
@@ -0,0 +1,321 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import * as THREE from 'three';
+import ComboControls from '@cognite/three-combo-controls';
+import TWEEN from '@tweenjs/tween.js';
+import _ from 'lodash';
```
maksnester, line 8: please avoid the import of the whole lodash, it increases bundle size. Use `import foo from 'lodash/foo'` instead. We should add an eslint rule for that :)

maksnester (commented):
I haven't really reviewed, that's clearly not for me to review 😄

viewer/src/tools/AxisView/AxisViewTool.ts
```diff
@@ -0,0 +1,321 @@
...
+          throw new Error(`Unknown corner position for Axis Cross: Corner = ${position.corner}`);
+      }
+    }
+
+    function isAbsolute(position: Absolute | Relative): position is Absolute {
+      return (position as Absolute).xAbsolute !== undefined && (position as Absolute).yAbsolute !== undefined;
+    }
+  }
+
+  public tryClick(xScreenPos: number, yScreenPos: number, moveCameraOnSuccess = true): boolean {
```
larsmoa, line 93: I think this can (and should) be avoided by listening to click/pointer events on `viewer.domElement`. When the axis tool is clicked, use `event.preventDefault()` to avoid that the viewer handles the click. By doing this, users would not have to call this function.

viewer/src/tools/AxisView/AxisViewTool.ts
```diff
@@ -0,0 +1,321 @@
...
+import ComboControls from '@cognite/three-combo-controls';
+import TWEEN from '@tweenjs/tween.js';
+import _ from 'lodash';
+
+import { Cognite3DViewer } from '../../migration';
+import { Cognite3DViewerToolBase } from '../Cognite3DViewerToolBase';
+import { AxisBoxConfig, defaultAxisBoxConfig, AxisBoxFaceConfig, Corner, Absolute, Relative } from './types';
+
+export class AxisViewTool extends Cognite3DViewerToolBase {
+  private readonly _layoutConfig: AxisBoxConfig;
```
larsmoa, line 15:
```suggestion
  private readonly _layoutConfig: Required<AxisBoxConfig>;
```
https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype

By doing this you can avoid some `!` in the code below.

viewer/src/tools/AxisView/AxisViewTool.ts
```diff
@@ -0,0 +1,321 @@
...
+        camera.setRotationFromQuaternion(tmpRotation);
+      })
+      .start(TWEEN.now())
+      .onStart(() => {
+        cameraControls.enabled = false;
+      })
+      .onComplete(() => {
+        cameraControls.setState(camera.position, cameraTarget);
+        cameraControls.enabled = true;
+      });
```
larsmoa, line 318: Store `enabled` before tweening and reset on completion, rather than explicitly setting to true.

viewer/src/tools/AxisView/types.ts
```diff
@@ -0,0 +1,105 @@
...
+    xNegativeFace?: AxisBoxFaceConfig;
+    yPositiveFace?: AxisBoxFaceConfig;
+    yNegativeFace?: AxisBoxFaceConfig;
+    zPositiveFace?: AxisBoxFaceConfig;
+    zNegativeFace?: AxisBoxFaceConfig;
+  };
+  compass?: AxisBoxCompassConfig;
+};
+
+export type Absolute = {
```
larsmoa, line 21: `AbsolutePosition`

viewer/src/tools/AxisView/types.ts
```diff
@@ -0,0 +1,105 @@
...
+  xAbsolute: number;
+  yAbsolute: number;
+};
+
+export type Relative = {
```
larsmoa, line 26: `RelativePosition`

viewer/src/tools/AxisView/types.ts
```diff
@@ -0,0 +1,105 @@
...
+};
+
+export enum Corner {
+  TopRight,
+  TopLeft,
+  BottomLeft,
+  BottomRight
+}
+
+export const defaultAxisBoxCompassConfig: AxisBoxCompassConfig = {
```
larsmoa, line 55:
```suggestion
export const defaultAxisBoxCompassConfig: Required<AxisBoxCompassConfig> = {
```

viewer/src/tools/AxisView/types.ts
```diff
@@ -0,0 +1,105 @@
...
+  ringLabel: 'N',
+  labelDelta: Math.PI,
+  fontSize: undefined,
+  fontColor: new THREE.Color(0xff0000),
+  tickColor: new THREE.Color(0x949494)
+};
+
+export const defaultFaceConfig: AxisBoxFaceConfig = {
```
larsmoa, line 63:
```suggestion
export const defaultFaceConfig: Required<AxisBoxFaceConfig> = {
```

viewer/src/tools/AxisView/types.ts
```diff
@@ -0,0 +1,105 @@
...
+  label: '',
+  fontSize: undefined,
+  fontColor: new THREE.Color(0x333333),
+  outlineSize: undefined,
+  outlineColor: new THREE.Color(0x333333),
+  faceColor: new THREE.Color(0x949494)
+};
+
+export const defaultAxisBoxConfig: AxisBoxConfig = {
```
larsmoa, line 72:
```suggestion
export const defaultAxisBoxConfig: Required<AxisBoxConfig> = {
```

viewer/src/tools/AxisView/AxisViewTool.ts
```diff
@@ -0,0 +1,321 @@
...
+
+    this._screenPosition = new THREE.Vector2();
+    this._boxFaceGeometry = new THREE.PlaneGeometry(0.85, 0.85, 1, 1);
+
+    this._raycastCamera = new THREE.OrthographicCamera(-1, 1, 1, -1, -1, 1);
+    this._raycaster = new THREE.Raycaster();
+
+    this._viewer = viewer;
+
+    this._layoutConfig = _.merge(_.cloneDeep(defaultAxisBoxConfig), config);
```
larsmoa, line 40:
```suggestion
    this._layoutConfig = _.{ ...defaultAxisBoxConfig, ...config };
```
will *almost* do the same, but won't create a deep clone, but do you really need that here?

viewer/src/tools/AxisView/AxisViewTool.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  private readonly _boxFaceGeometry: THREE.PlaneGeometry;
+
+  private readonly _viewer: Cognite3DViewer;
+
+  private readonly _interactiveObjects: THREE.Mesh[];
+  private readonly _raycastCamera: THREE.OrthographicCamera;
+  private readonly _raycaster: THREE.Raycaster;
+
+  private _screenPosition: THREE.Vector2;
```
larsmoa, line 25:
```suggestion
  private readonly _screenPosition: THREE.Vector2;
```

viewer/src/tools/AxisView/types.ts
```diff
@@ -0,0 +1,105 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import * as THREE from 'three';
+
+export type AxisBoxConfig = {
+  size?: number;
+  position?: Absolute | Relative;
```
larsmoa, line 9: Clever!

viewer/src/tools/AxisView/AxisViewTool.ts
christjt: Okey, thanks for the tip. How is this different from importing * from Three?

larsmoa (changes requested):
This is very nice! Really lifts the experience.

A few comments/requests:
- The axis tool is placed to the bottom left before moving to the selected corner
- AxisBoxConfig and other config types are not exported
- I don't think `tryClick` is necessary - see code in comment

viewer/src/public/RevealManager.ts
```diff
@@ -233,6 +233,14 @@ export class RevealManager<TModelIdentifier> {
     }
   }
 
+  public addUiObject(object: THREE.Group, screenPos: THREE.Vector2, width: number, height: number) {
```
larsmoa, line 236: `addUiObject` accepts `THREE.Group`, `removeUiObject` accepts `THREE.Object3D`. I think it makes sense to allow `Object3D`. It feels a bit odd that position is `Vector2` while size is two numbers, align types.

viewer/src/tools/AxisView/AxisViewTool.ts
larsmoa: It doesn't really I belive @christjt, https://discourse.threejs.org/t/tree-shaking-three-js/1349/42

maksnester: yes, it doesn't. The only difference is that we for some reason export the whole THREE from Reveal so it doesn't matter if import it as a whole.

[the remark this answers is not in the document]
maksnester: > import { merge } from 'lodash';

That's, unfortunately, won't work too. Only `import foo from lodash/foo` is the option here. 

see also https://github.com/cognitedata/cognite-sdk-js/pull/386 where we tried several things in js sdk.

viewer/src/tools/AxisView/types.ts
christjt: This actually has some fields that intentionally could be undefined (f.ex font size will be automatically estimated based on size of component if not set).

viewer/src/tools/AxisView/types.ts
christjt: This actually has some fields that intentionally could be undefined (f.ex font size will be automatically estimated based on size of component if not set).

larsmoa: I believe that changing the field from e.g. `fontSize?: number` to `fontSize?: number | undefined` fixes this, @christjt

larsmoa (approved):

PR #1459, opened 2021-05-07
fix: refac non-standard array initialization that fails on Safari

larsmoa:


viewer/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -79,7 +74,12 @@ void updateFragmentColor(
     } else if (renderMode == RenderTypeGeometryType) {
         vec2 cap = normal.xy * 0.5 + 0.5;
         vec3 mc = texture2D(matCapTexture, cap).rgb * 1.5;
-        gl_FragColor = vec4(geometryTypeColors[geometryType] * mc, color.a);
+        vec3 geometryColor = 
+            isGeometryType(geometryType, 1) * vec3(1.0, 0.0, 0.0) + // Quads
```
christjt, line 78: Can't you just inline-if this?

PR #1468, opened 2021-05-21
chore: allow using apiKey to authenticate in the Migration example

larsmoa:


examples/src/pages/Migration.tsx
```diff
@@ -39,6 +39,7 @@ export function Migration() {
       const geometryFilterInput = urlParams.get('geometryFilter');
       const geometryFilter = createGeometryFilter(geometryFilterInput);
       const baseUrl = urlParams.get('baseUrl') || undefined;
+      const apiKey = urlParams.get('apiKey') || undefined;
```
maksnester, line 42: @larsmoa Hm... wouldn't it be better to use the `.env` file to put your API key? I don't think it's something frequently changing to put it in the URL. Or is it? 

Speaking of API keys, we could use them for snapshot tests as well and test against real models 🤔

larsmoa: good point regarding testing.

You are probably right regarding its bad using in URL, but it's really convinient sometimes when I get a temp key for a client project

PR #1473, opened 2021-05-26
improvement: better handling of errors in web workers

larsmoa:
Previously, all queued work would be queued for the first worker, rather than distributed among all works. This could cause situations where one worker had a long queue of tasks, while the other workers were idle.


viewer/src/utilities/workers/WorkerPool.ts
```diff
@@ -74,15 +75,22 @@ export class WorkerPool {
   }
 
   async postWorkToAvailable<T>(work: WorkDelegate<T>): Promise<T> {
-    let targetWorker = this.workerList[0];
-    for (const worker of this.workerList) {
-      if (worker.activeJobCount < targetWorker.activeJobCount) {
-        targetWorker = worker;
+    const targetWorker = this.workerList.reduce((bestWorker, candidate) => {
+      if (bestWorker === undefined || bestWorker.activeJobCount > candidate.activeJobCount) {
```
maksnester, line 79: To be honest I don't understand what's the difference here between the previous form

```
    let targetWorker = this.workerList[0];
    for (const worker of this.workerList) {
      if (worker.activeJobCount < targetWorker.activeJobCount) {
        targetWorker = worker;
      }
    }
```

and the current one. In both, you just iterate over all workers and reasign the chosen one. Though I like your reduce form more. 

---

Also, I don't understand why `bestWorker === undefined` check is needed? You initialised it with this.workerList[0] already, so it can't be undefined?

viewer/src/utilities/workers/WorkerPool.ts
```diff
@@ -74,15 +75,22 @@ export class WorkerPool {
...
+      return bestWorker;
+    }, this.workerList[0]);
+
     targetWorker.activeJobCount += 1;
-    const result = await work(targetWorker.worker);
-    targetWorker.activeJobCount -= 1;
+    const result = await (async () => {
+      try {
+        return await work(targetWorker.worker);
+      } finally {
```
maksnester, line 89: that's a really good catch 👍

maksnester (approved):
@larsmoa I removed auto-merge in case you'll want to add changes

viewer/src/utilities/workers/WorkerPool.ts
```diff
@@ -35,10 +35,11 @@ export class WorkerPool {
       const newWorker = {
         // NOTE: As of Comlink 4.2.0 we need to go through unknown before RevealParserWorker
         // Please feel free to remove `as unknown` if possible.
-        worker: (Comlink.wrap(this.createWorker()) as unknown) as RevealParserWorker,
+        worker: (wrap(this.createWorker()) as unknown) as RevealParserWorker,
         activeJobCount: 0,
         messageIdCounter: 0
       };
+      // newWorker.worker.getVersion().then(x => console.log(`Worker ${i} version:`, x));
```
maksnester, line 42:
```suggestion
```

viewer/src/utilities/workers/WorkerPool.ts
larsmoa: You are right - I misread the code and thought it would always pick the first worker when there was a queue.

PR #1485, opened 2021-05-27
docs: add notes about releasing parser-worker; remove three.js mention from docs

maksnester:
Speaking of threejs, is there any reason to not have it updated to the latest version @christjt ? It's just would be nice to catch any possible issue with having `@types/three` in reveal as a dependency as well
.


larsmoa:
> Speaking of threejs, is there any reason to not have it updated to the latest version @christjt ? It's just would be nice to catch any possible issue with having `@types/three` in reveal as a dependency as well
> .

It's a separate task that requires some code changes. I will do this after this is merged

larsmoa (approved):

PR #1492, opened 2021-05-28
chore: cleanup some TODOs

larsmoa:


christjt (approved):
I'm going to assume you actually checked the different cases (such as scene.json with NaN bounding boxes etc.)

PR #1513, opened 2021-06-03
chore: replace gl-matrix with ThreeJS

larsmoa:


viewer/src/datamodels/cad/cameraconfig.ts
```diff
@@ -1,47 +1,46 @@
...
+  target: THREE.Vector3;
   near: number;
   far: number;
 }
 
 export function suggestCameraConfig(rootSector: SectorMetadata): SuggestedCameraConfig {
-  const averageMin = vec3.create();
-  const averageMax = vec3.create();
+  const averageMin = new THREE.Vector3(0, 0, 0);
+  const averageMax = new THREE.Vector3(0, 0, 0);
```
christjt, line 18: I believe `new THREE.Vector3(0, 0, 0);` is equivalent to just:
```
new THREE.Vector3();
```

christjt (changes requested):
LGTM, only one tiny nit

christjt (approved):

PR #1517, opened 2021-06-06
feat: add Cognite logo watermark

larsmoa:
🤮


viewer/src/utilities/CogniteLogoOverlay.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.el = document.createElement('div');
+    this.el.title = `Cognite Reveal ${process.env.VERSION}`;
+    this.el.style.position = 'absolute';
+    this.el.style.bottom = '0';
+    this.el.style.left = '0';
+
+    this._overlay = document.createElement('div');
+    this.updateStyle(viewerBackgroundColor);
+    this._overlay.className = determineCssClass(new THREE.Color('black'));
+    this.el.appendChild(this._overlay);
```
maksnester, line 35: I don't really see why do we need 2 elements instead of one. I suggest to remove from CSS styles `margin` and `pointer-events` and merge 2 divs (`this.el`, `this._overlay`) into 1.

Also notice that you can't have `title` functional (where you display reveal version on hover) and `pointer-events: none;` at the same time. So you have to choose between displaying the version on hover and the logo being ignored for clicks. See video record (notice that I use only 1 div, your current implementation despite of `pointer-evenets: none` being set prefers to show `title` because `pointer-events` work for the `this.el`)

https://user-images.githubusercontent.com/4989157/120997885-dc91ff00-c787-11eb-8e1c-273d89ac9a06.mov

Despite I really like the idea about showing the version number, I would choose to ignore pointer events, because it's closer to watermark behaviour. What's for version number, you can write it under the logo using some tiniest font if you really want to keep it.

viewer/src/public/migration/Cognite3DViewer.ts
```diff
@@ -305,7 +309,7 @@ export class Cognite3DViewer {
       model.dispose();
     }
     this._models.splice(0);
-    this.spinner.dispose();
+    this._loadingSpinner.dispose();
```
maksnester, line 312: shouldn't the `dispose` call for the logo element be here?

maksnester (changes requested):
Suggested some changes, but looks great in general

PR #1518, opened 2021-06-07
Add sketch of glTF support

haakonflatval-cognite:


haakonflatval-cognite:
At this point, one case of primitives (boxes, to be specific) has been shown to successfully be placed in the glTF together with the triangle data, read and transformed by Reveal and finally visualized correctly. 

The path forwards for adding more primitives is straight-forward, but a bit laborious. We will need to add transform-methods into  viewer/src/datamodels/cad/sector/primitiveTransformers.ts to transform all of the "raw" primitives that are found in the i3d-file (listed together with their data layout [here](https://github.com/cognitedata/i3df-specification/blob/master/specification/i3df-specification.yaml)) into the primitives used in rendering in Reveal, listed [here](https://github.com/cognitedata/reveal/blob/31fc1ac5900f787b620f1131d2f37dd77469a73d/i3df/src/renderables.rs#L399). For this, we can use the [Rust implementation](https://github.com/cognitedata/reveal/tree/master/i3df/src/renderables) as reference. We will need to implement 30 such functions in total.
Finally, it needs to be tied together using the code in viewer/src/datamodels/cad/sector/CadSectorParser.ts.

haakonflatval-cognite:
This is probably also a good point to mention that the work on encoding instance mesh info into the glTF also remains, but has been planned for.

larsmoa:
Can this be closed now @haakonflatval-cognite @christjt ?

haakonflatval-cognite:
Closing in favor of #1588, as @christjt  mentioned

PR #1522, opened 2021-06-08
doc: remove @version info about stuff being new in 2.0.0

larsmoa:


maksnester (changes requested):
@larsmoa we don't need 1.x annotations in 2.0 release as well

maksnester (approved):
great, nailed it 👍

PR #1526, opened 2021-06-09
chore: remove Cognite3DModel.loadingHints

larsmoa:


maksnester (approved):
great idea to remove stale APIs and all the methods with NotSupportedInMigrationWrapper error

PR #1540, opened 2021-06-21
refac: make Cognite3DViewer.getIntersectionFromPixel async [REV-176]

larsmoa:
Make function async to enable async implementations in the future without breaking the API.


maksnester (approved):
since it's a breaking change it probably should be mentioned in the migration guide.

PR #1541, opened 2021-06-21
improvement: replace pink and purple outline colors with green and red [REV-171]

larsmoa:


viewer/src/utilities/types.ts
```diff
@@ -25,6 +25,14 @@ export class CogniteColors {
   public static readonly LightGray = new THREE.Color('rgb(242, 241, 240)');
 }
 
+/**
+ * Some additional colors to supplement {@link CogniteColors}.
```
maksnester, line 29: hm... that looks a bit excessive to have 2 colour sets. don't you think so @larsmoa ?

```
/** * Colors from the Cognite theme. */
export class CogniteColors {
```

What the motivation to have that extra set here instead of adding new colors directly to CogniteColros?

maksnester (changes requested):

viewer/src/utilities/types.ts
larsmoa: Cognite colors are company standard colors, so I didn't want to touch them

maksnester: Yeah, but I don't understand how RevealColors that you created are not the same "company standard colors"...

maksnester (approved):
I really see no sense in that additional color sets, it adds confusion into codebase in my opinion, but I don't want to block you with that, so approved, but consider merging those sets into one.

viewer/src/utilities/types.ts
larsmoa: The CogniteColors are from some design firm that has created a visual profile for Cognite. In this profile is a list of colors. CogniteColors reflects that list, and I dont want to add new stuff to it...

PR #1544, opened 2021-06-22
fix: add assert as a direct dependency to make sure it's available in client projects

larsmoa:
We import 'assert', but the module is only a indirect dev dependency so it will fail for clients that have build setups that doesn't include the assert module.


haakonflatval-cognite (approved):
Assuming package.json was the only change done manually, it looks good :+1:

PR #1545, opened 2021-06-22
fix: remove @internal from serialize() implementations

larsmoa:
This issue causes import to cause compile errors because serialize() is exported in base class, but not implementations.


haakonflatval-cognite (approved):
LGTM :+1:

PR #1548, opened 2021-06-23
doc: minor release procedure doc improvement

larsmoa:


README.md
```diff
@@ -189,7 +184,9 @@ targeting the master branch.
 1. Go to [https://github.com/cognitedata/reveal/releases/new](https://github.com/cognitedata/reveal/releases/new)
 1. Under the "Tag version" field, look for the newly-created tag, e.g. `@cognite/reveal@2.1.0`
 1. Specify the same release title as the tag name.
-1. Write the changes that new version brings. Get inspired by done tasks from your spring board. 
+1. Write the changes that new version brings. Get inspired by done tasks from your sprint board. Note! 
+   Remember to specify the correct ThreeJS version required by the released version - you can find this in package.json
```
maksnester, line 188: @larsmoa I don't think we should do it since three is now a common dependency and people can just use our export in case of any conflicts, but up to you.

maksnester (approved):

PR #1551, opened 2021-06-23
improvement: CTM cache will no longer trigger many requests for the same file

larsmoa:


maksnester (commented):
I see how you changed the setting of a cached item which is a promise now, but I don't see that you touched any code related to retrieving that promise from the cache... 🤔

maksnester (approved):
ok, I figured it out, LGTM 🙂

viewer/src/datamodels/cad/sector/CachedRepository.ts
```diff
@@ -164,11 +164,13 @@ export class CachedRepository implements Repository {
...
-    // TODO 2021-05-05 larsmoa: Move retry to getBinaryFile()
-    const buffer = await this._modelSectorProvider.getBinaryFile(modelBlobUrl, filename);
-    const parsedCtm = await this._modelDataParser.parseCTM(new Uint8Array(buffer));
-    this._ctmFileCache.set(cacheKey, parsedCtm);
-    return parsedCtm;
+    const ctmPromise = (async () => {
+      const buffer = await this._modelSectorProvider.getBinaryFile(modelBlobUrl, filename);
+      const parsedCtm = await this._modelDataParser.parseCTM(new Uint8Array(buffer));
+      return parsedCtm;
+    })();
```
maksnester, line 171:
```suggestion
    const ctmPromise = this._modelSectorProvider.getBinaryFile(modelBlobUrl, filename)
      .then((buffer) => this._modelDataParser.parseCTM(new Uint8Array(buffer)))
    ;
```

though I think a good old promise chain looks better than self invoking function

PR #1554, opened 2021-06-28
chore: add workflow to cleanup docs previews

maksnester:
image

That's how the current docs preview repo looks like. It's huge, I think that's why it's slow, not because of git history. I'll do manual cleanup there, but I hope that little workflow will remove each PR preview once it merged.

UPD: @larsmoa actually I got a second thought on it, and maybe history still matters the most here. If something ever were committed to the git it remains in the git, so I have doubts that this cleanup will help anyhow 🙁  but let's see still


larsmoa (approved):
Nice!

PR #1555, opened 2021-06-28
fix: not able to load same model twice [REV-132]

larsmoa:
Add the ability to load the same model twice, or to add-remove-add the same model.


viewer/src/datamodels/cad/sector/rxSectorUtilities.ts
```diff
@@ -2,105 +2,8 @@
...
-
-/**
- * Creates a RxJS operator for loading sectors given camera, budget etc input.
- * @param sectorRepository Repository to store sectors in
- * @param sectorCuller Culler used to prioritize sectors for loading
- * @param modelStateHandler Holds state about what sectors are loaded
- * @param collectStatisticsCallback Callback to collect statistics on how much data is loaded
- * @returns RxJS operator
- */
-export function handleDetermineSectorsInput(
```
christjt, line 28: VERY NICE

christjt (changes requested):
Looks VERY good! Love the refactoring done here. Very clean and waaay more understandable. Only some minor recommendations to make it even more concise.

larsmoa:
Thx for good input @christjt - I've done some changes and I also noticed that sectors wasn't made visible on screen until a full batch was loaded. I fixed this and did some minor improvements.

christjt (approved):

PR #1562, opened 2021-06-30
fix: geometry filters will now only affect the model they are assosciated with

larsmoa:
Before geometry filters where global and would affect all loaded models and geometry.

Note that support for clipIntersection was removed as this currently isn't compatible with geometry filters.


documentation/docs/examples/combine-models.mdx
```diff
@@ -38,3 +38,50 @@ viewer.addModel({
   viewer.fitCameraToModel(pointCloudModel);
 });
 ```
+
+## Use geometry filters to extract parts of a model
+
+Geometry filters allows only loading geometry within a specified box. By loading the same model with different geometry
+filters several subsections of the same model can be loaded in the same viewer.
```
christjt, line 45:
```suggestion
filters, several subsections of the same model can be loaded in the same viewer.
```

examples/src/pages/Migration.tsx
```diff
@@ -576,10 +576,10 @@ function createGeometryFilterStateFromBounds(bounds: THREE.Box3, out: { center:
   return out;
 }
 
-function createGeometryFilterFromState(state: { center: THREE.Vector3, size: THREE.Vector3, enabled: boolean }):
+function createGeometryFilterFromState(state: { center: THREE.Vector3, size: THREE.Vector3 }):
  { boundingBox: THREE.Box3, isBoundingBoxInModelCoordinates: true } | undefined {
   state.size.clamp(new THREE.Vector3(0,0,0), new THREE.Vector3(Infinity, Infinity, Infinity));
-  if (!state.enabled || state.size.equals(new THREE.Vector3(0,0,0))) {
+  if (state.size.equals(new THREE.Vector3(0,0,0))) {
```
christjt, line 582:
```suggestion
  if (state.size.equals(new THREE.Vector3())) {
```

christjt (changes requested):
LGTM, minor nits

christjt (approved):

PR #1564, opened 2021-06-30
chore: migrate viewer to yarn 2 workspaces

christjt:


documentation/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -25,7 +20,7 @@ function NavLink({
   activeClassName = 'navbar__link--active',
   prependBaseUrlToHref,
   ...props
-}: NavLinkProps) {
+}: any) {
```
larsmoa, line 23: Why `any`?

documentation/src/theme/NavbarItem/DefaultNavbarItem.tsx
```diff
@@ -64,7 +59,7 @@ function NavItemDesktop({
   position,
   className,
   ...props
-}: DesktopOrMobileNavBarItemProps) {
+}) {
```
larsmoa, line 62: Why remove return type?

viewer/packages/camera-manager/tests/cameraControls.test.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+
+import * as THREE from 'three';
+
+describe('Camera Controls', () => {
+  test('dummy', () => {
+    const asd = new THREE.Vector3();
+    expect(asd).toEqual(new THREE.Vector3());
+  });
+});
```
larsmoa, line 12: Add proper test (or remove)

larsmoa (changes requested):

larsmoa:
- Remove package-a
- yarn serve in viewer is broken
- Update README.md

README.md
```diff
@@ -22,13 +22,13 @@ All information below is for developers.
...
+To test the viewer, you need to build the viewer and then start the examples. 
 
 ```
 cd viewer
-yarn install
-yarn build
-yarn serve
+yarn && yarn run build
+cd ../examples
+yarn && yarn run start
```
larsmoa, line 31: Not something for now, but could maybe examples be included as workspaces as well?

documentation/tsconfig.json
```diff
@@ -8,13 +8,14 @@
     "jsx": "react",
     "lib": ["DOM", "DOM.Iterable", "ES6"],
     "noEmit": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
larsmoa, line 11: We should add a TODO about the use of any to get rid of it some day

viewer/core/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "private": true,
+  "main": "index.js",
+  "types": "index.d.ts",
+  "scripts": {
+    "build": "yarn ws:build",
+    "test": "yarn ws:test --config ./../jest.config.js",
+    "lint": "yarn ws:lint"
+  },
+  "dependencies": {
+    "@reveal/camera-manager": "workspace:*"
```
larsmoa, line 12: What does "workspace:*" mean? Just look up in all workspaces?

viewer/packages/camera-manager/tests/cameraControls.test.ts
```diff
@@ -0,0 +1,35 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import * as THREE from 'three';
+import 'jest-extended';
+import ComboControls from '../src/ComboControls';
+
+describe('Camera Controls', () => {
```
larsmoa, line 9: I think the file should be renamed to ComboControls.test.ts and this change to
```suggestion
describe('ComboControls', () => {
```

Also - we've moved test to be inlined in src/ - can this be done here as well?

larsmoa (commented):
Looks great - looking forward to put this to use :) Added a few inline comments. Also, can you add a section to `viewer/README.md` on the structure of a package and how to set up new packages + how to add deps between different packages?

viewer/packages/camera-manager/tests/cameraControls.test.ts
christjt: I personally don't really like / see the reason to have tests inlined in /src since we are now in a much more confined scope. I felt that tests where in /src simply because it was impossible to find the test for a given file with the previous directory structure. Also conceptually it is strange to have tests in /src since... well... it isn't source code...

README.md
christjt: yes, but not in scope for this PR

viewer/packages/camera-manager/tests/cameraControls.test.ts
larsmoa: My main argument for having them side by side is that it becomes hard to move classes around when tests are in separate folders. I always felt this was a PITA in .NET where tests quickly would end up in mismatching folders or with mismatching filenames...

viewer/README.md
```diff
@@ -43,3 +43,65 @@ In case if you see an error like that:
 > Refused to load the script 'https://apps-cdn.cogniteapp.com/@cognite/reveal-parser-worker/1.1.0/reveal.parser.worker.js' because it violates the following Content Security Policy directive: "script-src 'self' https://localhost:* blob:"
 
 See the next steps in our [documentation](https://cognitedata.github.io/reveal-docs/docs/installation#installation-for-projects-with-content-security-policy).
+
+## Local Packages
+The Reveal viewer is structured using local packages.
+This allows you to test features in isolation and constrain the dependencies with logical barriers between features.
+Packages are located in the `/packages/` subfolder.
+There is no enforced structure of the packages and their layout and content will vary depending on what the package exposes to the rest of the system.
+Dependencies that are located in the root `package.json` will be shared and are accessible by any package without having to explicitly declare a depencency in the respective package's `package.json`.
```
larsmoa, line 52:
```suggestion
Dependencies that are located in the `viewer/package.json` will be shared and are accessible by any package without having to explicitly declare a dependency in the respective package's `package.json`.
```

viewer/README.md
```diff
@@ -43,3 +43,65 @@ In case if you see an error like that:
 > Refused to load the script 'https://apps-cdn.cogniteapp.com/@cognite/reveal-parser-worker/1.1.0/reveal.parser.worker.js' because it violates the following Content Security Policy directive: "script-src 'self' https://localhost:* blob:"
 
 See the next steps in our [documentation](https://cognitedata.github.io/reveal-docs/docs/installation#installation-for-projects-with-content-security-policy).
+
+## Local Packages
+The Reveal viewer is structured using local packages.
+This allows you to test features in isolation and constrain the dependencies with logical barriers between features.
+Packages are located in the `/packages/` subfolder.
+There is no enforced structure of the packages and their layout and content will vary depending on what the package exposes to the rest of the system.
```
larsmoa, line 51: I think we however should at least suggest that it follows a standard, e.g. 
- `src/` - source files and tests
- `app/` - test application with `index.ts` as entry
- `index.ts` as main entry.
I just think it would benefit us if we had one fixed structure that we follow (the above is just a half-ass example that doesn't probably cover everything that we should "standarize")

viewer/packages/camera-manager/tests/cameraControls.test.ts
christjt: I see that this is a pain, but I still think that this is an effect of having a too deep / complex folder structure and that this will be much less of a problem going forward. If this assumption turns out to be false I will bite the bullet and inline all the tests.

larsmoa: The file and test description should _at least_ be renamed to match the file being tested

christjt: Agreed, will fix asap

christjt: Wait, this should already been done????

larsmoa (approved):

PR #1566, opened 2021-07-01
chore: migrate to v1 filterNodes endpoint

haakonflatval-cognite:


larsmoa (changes requested):
Looks good - but remove @experimental tag from SinglePropertyNodeCollection

larsmoa (approved):

PR #1576, opened 2021-08-05
feat: introduce LOD rendering of torus geometries to improve quality of certain primitives

larsmoa:
Before:
image
After:
image
(notice the silhouette of the big pipe)

### DIFF
--- a/viewer/src/utilities/three/BoundingBoxLOD.ts
+++ b/viewer/src/utilities/three/BoundingBoxLOD.ts
@@ -46,7 +46,6 @@ export class BoundingBoxLOD extends THREE.Object3D {
     if (levels.length > 0) {
       camPos.setFromMatrixPosition(camera.matrixWorld);
       const distanceToCamera = bounds.distanceToPoint(camPos) / cameraZoom;
-      console.log(distanceToCamera);
 
       this._currentLevel = 0;
       levels[0].object.visible = true;
--- a/viewer/src/utilities/three/BoundingBoxLOD.ts
+++ b/viewer/src/utilities/three/BoundingBoxLOD.ts
@@ -3,9 +3,19 @@
  */
 import * as THREE from 'three';
 
+// For BoundingBoxLOD.update()
+const updateVars = {
+  camPos: new THREE.Vector3(),
+  bounds: new THREE.Box3()
+};
+
+/**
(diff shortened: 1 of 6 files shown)


viewer/src/utilities/three/BoundingBoxLOD.ts
```diff
@@ -0,0 +1,68 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+import * as THREE from 'three';
+
+export class BoundingBoxLOD extends THREE.Object3D {
```
christjt, line 6: Really smart!

viewer/src/utilities/three/BoundingBoxLOD.ts
```diff
@@ -0,0 +1,68 @@
...
+          levels[i].object.visible = true;
+          this._currentLevel = i;
+        } else {
+          break;
+        }
+      }
+
+      for (let i = this._currentLevel + 1; i < levels.length; ++i) {
+        levels[i].object.visible = false;
+      }
```
christjt, line 65: If you reverse the sort order I think you can do this. This also assumes that all levels are initialized as `visible=false`

```js
    levels[this._currentLevel].visible = false;
    this._currentLevel = levels.findIndex(p => distanceToCamera >= p.distance);
    levels[this._currentLevel].visible = false;
```

Which looks cleaner in my opinion.

christjt (changes requested):
Overall looks really good. Clever to extend an object3D to create a LOD node.

viewer/src/utilities/three/BoundingBoxLOD.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  private readonly _updateVars = {
+    camPos: new THREE.Vector3(),
+    bounds: new THREE.Box3()
+  };
+  update(camera: THREE.Camera) {
+    const levels = this._levels;
+    const { camPos, bounds } = this._updateVars;
+    bounds.copy(this._boundingBox).applyMatrix4(this.matrixWorld);
+    const cameraZoom = camera instanceof THREE.PerspectiveCamera ? camera.zoom : 1.0;
```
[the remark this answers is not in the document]
larsmoa: We could, but this will just work - no reason to not have it work on non-perspective?

christjt (approved):

PR #1580, opened 2021-08-09
feat: upgrade to ThreeJS 0.131.0

larsmoa:


christjt:
Something is going wrong in CI also, you would should also update threejs in the documentation right?

christjt (approved):
LGTM!

PR #1581, opened 2021-08-09
fix: resolve warning about unresolved indirect dependency

larsmoa:
Fix warning `eslint-plugin-unused-imports@npm:0.1.3 doesn't provide @typescript-eslint/parser (p155d7), requested by @typescript-eslint/eslint-plugin` during `yarn` in `viewer/`


christjt:
Seems strange that we have to have eslint things as a direct dependency. What is the reason we need to move it from a dev-dependency?

PR #1582, opened 2021-08-12
feat: Add Geomap tool to support models to be placed on maps

pramodcog:


viewer/core/src/tools/Geomap/GeomapTool.ts
```diff
@@ -0,0 +1,98 @@
...
+    this._map = new GEOTHREE.MapView(GEOTHREE.MapView.PLANAR, this._currentMapProvider, this._currentHeightMapProvider);
+
+    var coords = GEOTHREE.UnitsUtils.datumsToSpherical(59.9016426931744, 10.607235872426175);
+    var bound = this._viewer.models[0].getModelBoundingBox();
+    this._map.position.set(-coords.x, bound.min.y, coords.y);
+    this._viewer.addObject3D(this._map);
+    this._map.updateMatrixWorld(true);
+  }
+
+  public SetMapProvider(idx : number) {
```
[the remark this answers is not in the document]
larsmoa: maybe MapProvider needs to be something that allows for providing API-keys. Not sure how to best do this

documentation/docs/examples/geomaptool.mdx
```diff
@@ -0,0 +1,86 @@
...
+  }
+}
+
+new GeomapTool(viewer, mapConfig);
+```
+
+Here Satellite Maps
+
+```jsx runnable
+// import * as GEOTHREE from 'geo-three';
```
larsmoa, line 73:
```suggestion
```

documentation/versioned_docs/version-2.x/examples/geomaptool.mdx
```diff
@@ -0,0 +1,86 @@
...
+- Several map providers namely, Bing maps, Mapbox and Here maps
+- Different map mode like Planar, Spherical & Height 
+- Models can be placed on specific geolocation using latitude and longitude values
+
+<DemoWrapper name="Cognite3DViewerDemo" />
+
+## Basic example
+
+```jsx runnable
+// import * as GEOTHREE from 'geo-three';
```
larsmoa, line 24:
```suggestion
```

documentation/versioned_docs/version-2.x/examples/geomaptool.mdx
```diff
@@ -0,0 +1,86 @@
...
+## Select different Map providers
+
+In Geomap tool, you can select different map providers.
+- Bing Maps
+- Mapbox
+- Here Maps
+- Openstreet Maps
+
+:::note
+Openstreet maps would require to byepass CORS.
```
larsmoa, line 50:
```suggestion
OpenStreetMap requires bypassing CORS.
```
Add some details on what this means and why - link to some other doc?

documentation/versioned_docs/version-2.x/examples/geomaptool.mdx
```diff
@@ -0,0 +1,86 @@
...
+new GeomapTool(viewer, mapConfig);;
+```
+
+## Select different Map providers
+
+In Geomap tool, you can select different map providers.
+- Bing Maps
+- Mapbox
+- Here Maps
+- Openstreet Maps
```
larsmoa, line 47:
```suggestion
- OpenStreetMap
```

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Cognite3DViewer } from '@reveal/core';
+import { APIKeys } from './MapConfig';
+
+export class Geomap {
+
+    private _viewer: Cognite3DViewer;
+    private _map: any;
+    private _mapProvider: any;
+
+    public CreateMap(viewer: Cognite3DViewer, mapConfig: any) {
```
larsmoa, line 15: Use camelCase

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,56 @@
...
+        this._viewer = viewer;
+        this.GetMapProvider(mapConfig);
+        this._map = new GEOTHREE.MapView(GEOTHREE.MapView.PLANAR, this._mapProvider, this._mapProvider);
+        this._viewer.addObject3D(this._map);
+    }
+
+    private GetMapProvider(mapConfig: any) {
```
larsmoa, line 22: Use camelCase

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,56 @@
...
+            case 'BingMap':
+                this._mapProvider = new GEOTHREE.BingMapsProvider( mapConfig.mapAPIKey || APIKeys.DEV_BING_API_KEY, mapConfig.mapMode);
+                break;
+            case 'HereMap':
+                this._mapProvider = new GEOTHREE.HereMapsProvider(mapConfig.mapAPIKey || APIKeys.DEV_HEREMAPS_APP_ID, mapConfig.mapAPICode || APIKeys.DEV_HEREMAPS_APP_CODE, mapConfig.mapMode, mapConfig.mapScheme, mapConfig.format, mapConfig.mapResSize);
+                break;
+            case 'MapBoxMap':
+            default:
+                this._mapProvider = new GEOTHREE.MapBoxProvider(mapConfig.mapAPIKey || APIKeys.DEV_MAPBOX_API_KEY, mapConfig.id || "mapbox/satellite-streets-v10", mapConfig.mapMode, mapConfig.format);
+                break;
```
larsmoa, line 33:
```suggestion
            case 'MapBoxMap':
                this._mapProvider = new GEOTHREE.MapBoxProvider(mapConfig.mapAPIKey || APIKeys.DEV_MAPBOX_API_KEY, mapConfig.id || "mapbox/satellite-streets-v10", mapConfig.mapMode, mapConfig.format);
                break;
            default:
                throw new Error('Must provide mapProvider');
```

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,56 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import * as GEOTHREE from 'geo-three';
+import { Cognite3DViewer } from '@reveal/core';
+import { APIKeys } from './MapConfig';
```
larsmoa, line 7: This file seems to have 4 spaces as tabs, the other files have 2 spaces as tabs.

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export class Geomap {
+
+    private _viewer: Cognite3DViewer;
+    private _map: any;
+    private _mapProvider: any;
```
larsmoa, line 13: Make readonly and avoid `any`

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,66 @@
...
+DEV_MAPBOX_API_KEY = "pk.eyJ1IjoicHJhbW9kLXMiLCJhIjoiY2tzb2JkbXdyMGd5cjJubnBrM3IwMTd0OCJ9.jA9US2D2FRXUlldhE8bZgA",
+DEV_HEREMAPS_APP_ID = "HqSchC7XT2PA9qCfxzFq",
+DEV_HEREMAPS_APP_CODE = "5rob9QcZ70J-m18Er8-rIA",
+DEV_BING_API_KEY = "AuViYD_FXGfc3dxc0pNa8ZEJxyZyPq1lwOLPCOydV3f0tlEVH-HKMgxZ9ilcRj-T",
+}
+
+/**
+ * Map Configuration of {@link GeomapTool}.
+ */
+export type MapConfig = {
```
larsmoa, line 17: This type should probably be split into "common" values (such as mapProvider) and a specific config for each mapProvider. See https://github.com/cognitedata/reveal/blob/387663e45aae9c56366d5e5389dffd1722c5b20d/examples/src/pages/e2e/TestViewer.tsx#L49 for an example on how to achieve this using union types.

larsmoa (changes requested):

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+}
+
+/**
+ * Here Map Tiles Image Format
+ */
+export enum HereMapImageFormat {
+  PNG = 'png',
+	PNG8 = 'png8',
+	JPG = 'jpg'
+}
```
larsmoa, line 109:
```suggestion
export enum HereMapImageFormat {
  PNG = 'png',
  PNG8 = 'png8',
  JPG = 'jpg'
}
```
Use spaces, not tabs. Note! There are other cases of this in the same file which should be fixed too

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+   isVectorMap?: boolean;
+};
+
+/**
+ * Latitude, Longitude position.
+*/
+export type LatLongPosition = {
+latitude: number;
+longitude: number;
+};
```
larsmoa, line 158:
```suggestion
export type LatLongPosition = {
  latitude: number;
  longitude: number;
};
```

viewer/tsconfig.json
```diff
@@ -12,6 +12,8 @@
     "noUnusedParameters": true,
     // Strict Type-Checking Options
     "strict": true,
+	"strictNullChecks": false,
+	"noImplicitAny": false,
```
larsmoa, line 16: Why? We should avoid this if we can

viewer/packages/tools/src/Geomap/GeomapTool.ts
```diff
@@ -0,0 +1,34 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import { Cognite3DViewer } from '@reveal/core';
+import { Cognite3DViewerToolBase } from '../Cognite3DViewerToolBase';
+import { MapConfig } from './MapConfig';
+import { Geomap } from './Geomap';
+
+export class GeomapTool extends Cognite3DViewerToolBase {
```
larsmoa, line 10: Add JSDoc to this class.

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+  HEREMAP = 'HereMap',
+  MAPBOXMAP = 'MapBoxMap'
+}
+
+/**
+ * Map data for Mapbox
+*/
+export enum MapBoxMode {
+  STYLE = 100,
+  MAP_ID = 101
```
larsmoa, line 28: Add some JSDoc about what this is.

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+ */
+/**
+ * Supported map API Keys
+ */
+export enum APIKeys {
+DEV_MAPBOX_API_KEY = "pk.eyJ1IjoicHJhbW9kLXMiLCJhIjoiY2tzb2JkbXdyMGd5cjJubnBrM3IwMTd0OCJ9.jA9US2D2FRXUlldhE8bZgA",
+DEV_HEREMAPS_APP_ID = "HqSchC7XT2PA9qCfxzFq",
+DEV_HEREMAPS_APP_CODE = "5rob9QcZ70J-m18Er8-rIA",
+DEV_BING_API_KEY = "AuViYD_FXGfc3dxc0pNa8ZEJxyZyPq1lwOLPCOydV3f0tlEVH-HKMgxZ9ilcRj-T",
+}
```
larsmoa, line 12: Remove this fallback - it should only be used in docs/examples.

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+  imageFormat?: HereMapImageFormat;
+  size?: number;
+};
+
+export type MapBoxConfig = {
+  id?: MapBoxStyle;
+  mode?: MapBoxMode;
+  tileFormat?: MapBoxImageFormat;
+  useHDPI?: boolean;
+};
```
larsmoa, line 128: Add JSdoc to the different fields

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Maps Configuration of {@link GeomapTool}.
+ */
+export type MapConfig = BingMapConfig & HereMapConfig & MapBoxConfig & {
```
larsmoa, line 133: Hmm - this will force user to provide all settings from BingMap, HereMap and MapBox all the time. 

Have you tried changing to `(BingMapConfig | HereMapConfig | MapBoxConfig) & { ... }`? I can help out if you need some help getting this to work.

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,158 @@
...
+   */
+  provider: string;
+  /**
+   * Map provider API Key
+   */
+  APIKey: string;
+  /**
+   * Latitude, Longitude position
+   */
+  latlong?: LatLongPosition;
```
larsmoa, line 145: Shouldn't this be required?

larsmoa (changes requested):

viewer/packages/tools/src/Geomap/MapConfig.ts
pramodcog: I had `(BingMapConfig | HereMapConfig | MapBoxConfig) & { ... }`, but it was not able to access variables within the other Types. eg. if BingMapConfig contains `type: string` there was an error with "Not defined", similarly other variables from Types HereMapConfig & MapBoxConfig.
It would help if you can suggest how to got over this error

viewer/tsconfig.json
pramodcog: `geo-three` library/package causes error when viewer is build without these option. This option not required by our tool/examples but by the library

larsmoa:
I have a few new comments after checking out the code:
- Rename occurrences of `MapBox` to `Mapbox` (since Mapbox uses that writing)
- Enum values should be PascalCase, not UPPER_CASE
- `Geomap.getMapProvider()` sets internal state - get functions should not modify state. Instead, just return the create provider.
- I don't think you need to store `_mapProvider` in Geomap - get rid of if and just provide it directly to the `MapView`
- I don't think it makes sense to have `GeomapTool.setMapProvider`. Changing the map provider will require a new config - and the appCode-argument doesn't make sense to certain map providers (right?). We don't need to be able to dynamically changing map provder - it's fine if that requires explicitly re-creating the tool
- I don't think it makes sense to have `GeomapTool.setLatLong'. If this is part of the config just have it 
- There's a lot of optional member values in the `MapConfig`-type. If these are truly optional these is fine - if they are required, remove the '?' from the members.

pramodcog:
> I have a few new comments after checking out the code:
> 
> * Rename occurrences of `MapBox` to `Mapbox` (since Mapbox uses that writing)
> * Enum values should be PascalCase, not UPPER_CASE
> * `Geomap.getMapProvider()` sets internal state - get functions should not modify state. Instead, just return the create provider.
> * I don't think you need to store `_mapProvider` in Geomap - get rid of if and just provide it directly to the `MapView`
> * I don't think it makes sense to have `GeomapTool.setMapProvider`. Changing the map provider will require a new config - and the appCode-argument doesn't make sense to certain map providers (right?). We don't need to be able to dynamically changing map provder - it's fine if that requires explicitly re-creating the tool
> * I don't think it makes sense to have `GeomapTool.setLatLong'. If this is part of the config just have it
> * There's a lot of optional member values in the `MapConfig`-type. If these are truly optional these is fine - if they are required, remove the '?' from the members.

- Renamed of `MapBox` to `Mapbox`
- Enum variables names are set to PascalCase
- `GeomapTool.setMapProvider` is provided for examples to show different map options
- `_mapProvider` in Geomap removed.
- 'GeomapTool.setLatLong` is removed and latitiude and longitude should set in config paramters.
- Option in `MapConfig` for different Types are optional for couple of Maps, the once which is need values to be passed are made `required`.

viewer/packages/tools/src/Geomap/GeomapTool.ts
```diff
@@ -0,0 +1,30 @@
...
+    this._maps = new Geomap(this._viewer, config);
+  }
+
+  public setMapProvider(provider: MapProviders, apiKey: string, appCode?: string, id?: MapboxStyle) {
+    this._maps.setMapProvider(provider, apiKey, appCode, id);
+  }
+
+  public dispose(): void {
+    super.dispose();
+  }
```
larsmoa, line 29: Do we need this function? I think the tool can have a static provider and don't need the ability to set this runtime - only on construction. This function seems a bit off and I don't like that it has a MapboxStyle-argument which only makes sense if provider is Mapbox.

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,77 @@
...
+        mapProvider = new GEOTHREE.HereMapsProvider(
+          mapConfig.APIKey,
+          mapConfig.appCode,
+          mapConfig.style,
+          mapConfig.scheme,
+          mapConfig.imageFormat,
+          mapConfig.size
+        );
+        break;
+      case MapProviders.MapboxMap:
```
larsmoa, line 43: Ref comment in GeomapTool - I dont think we need this function.

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export class Geomap {
+  private readonly _viewer: Cognite3DViewer;
+  private _map: MapView;
+
+  constructor(viewer: Cognite3DViewer, mapConfig: MapConfig) {
+    this._viewer = viewer;
+    const mapProvider = this.getMapProvider(mapConfig);
+    this._map = new GEOTHREE.MapView(GEOTHREE.MapView.PLANAR, mapProvider, mapProvider);
+    this._viewer.addObject3D(this._map);
```
larsmoa, line 18: return directly, no need to store in local var first.

larsmoa (changes requested):
Getting there! See some comments, and I realized there is a function we could add which would be very helpful:

GeomapTool.latLongToWorldCoordinates() - a function that accepts latLong and returns a THREE.Vector3 which 3D coordinates

viewer/packages/tools/src/Geomap/GeomapTool.ts
pramodcog: Creating providers objects static to use them would take more memory (3 provider objects) than creating provider at runtime (1 provider object). I assume the users will only use one map (provider), so it would be easier to maintain single provider.
As mentioned earlier `GeomapTool.setMapProvider` is useful for only example to showcase all 3 map providers. If we do not want the 3 option in examples, we can remove it, but it won't be useful for the user.

viewer/packages/tools/src/Geomap/Geomap.ts
pramodcog: `this._map` is needed for `setMapProvider()` & also to position the map according to the Latitude & Longitude values.

viewer/tsconfig.json
larsmoa: Hmm, should be possible to exclude modules from checks - can you checkout https://www.typescriptlang.org/tsconfig#exclude?

larsmoa: > * `GeomapTool.setMapProvider` is provided for examples to show different map options

I think we just can create a new tool instance for this case and dispose the old one (?)

documentation/versioned_docs/version-2.x/examples/geomaptool.mdx
```diff
@@ -0,0 +1,107 @@
...
+- 'imageFormat': Map image format.
+  - HereMapImageFormat.PNG, HereMapImageFormat.PNG8, HereMapImageFormat.JPG.
+
+```jsx runnable
+// import { GeomapTool, MapProviders, HereMapType, HereMapScheme, HereMapImageFormat } from '@cognite/reveal/tools';
+
+const mapConfig = {
+  provider: MapProviders.HereMap,
+  APIKey: "HqSchC7XT2PA9qCfxzFq",
+  appCode: "5rob9QcZ70J-m18Er8-rIA",
```
larsmoa, line 97: Can you hide the API-keys in customScope.ts? So you define e.g. `HereMapApiKey` an `HereMapAppCode` in customScope.ts (and similar for the other examples) and use this constant instead? See https://github.com/cognitedata/reveal/blob/945c4d190707793f769e370e50a3ab8d75372e49/documentation/versioned_docs/version-2.x/examples/cad-3dobjects.mdx#L64-L64 for a similar example.

This makes finding the API keys harder which will reduce chance of people just copying these.

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,193 @@
...
+ */
+export enum MapboxMode {
+  /**
+   * Access the map data using a map style.
+   */
+  Style = 100,
+  /**
+   * Access the map data using a map id.
+   */
+  Map_Id = 101
```
larsmoa, line 25: Can you elaborate what this means? Maybe link Mapbox doc if thats meaningful

larsmoa (changes requested):

viewer/packages/tools/src/Geomap/MapConfig.ts
```diff
@@ -0,0 +1,198 @@
...
+   */
+  APIKey: string;
+  /**
+   * Latitude, Longitude position
+   */
+  latlong: LatLongPosition;
+  /**
+   * Is Vector Map?
+   */
+  isVectorMap?: boolean;
```
larsmoa, line 189: What does this do? It seems to be unused, so remove.

larsmoa (changes requested):
Added a few new comments. Can you also read through unresolved comments from earlier?

Also, I noticed the following:
- You've added requestRedraw(), but the problem isn't only for one frame. It's about new map frames being downloaded and ready for visualization, but redraw isn't automatically triggered - it requires the user to move the camera. See movie:
https://user-images.githubusercontent.com/6741854/133994484-92276179-fb11-4be3-8717-f5ceb42a139f.mov
- Regular maps looks really lousy with SSAO - we need to fix this before release (but not part of this PR)

viewer/packages/tools/src/Geomap/MapConfig.ts
pramodcog: If `true`, than vector map is loaded instead of satellite

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,69 @@
...
+        mapProvider = new GEOTHREE.BingMapsProvider(mapConfig.APIKey, mapConfig.type);
+        break;
+      case MapProviders.HereMap:
+        mapProvider = new GEOTHREE.HereMapsProvider(
+          mapConfig.APIKey,
+          mapConfig.appCode as string,
+          mapConfig.style as string,
+          mapConfig.scheme as string,
+          mapConfig.imageFormat as string,
+          mapConfig.size as number
```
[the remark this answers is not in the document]
pramodcog: I faced an errors when we had earlier config structure, now I think it is not needed as structure is changed, will remove it

viewer/packages/tools/src/Geomap/MapConfig.ts
larsmoa: Hmm - does all providers support this or is it a per-provider setting?

viewer/packages/tools/src/Geomap/MapConfig.ts
pramodcog: Added details & link to mapbox documentation

documentation/docs/examples/geomaptool.mdx
```diff
@@ -0,0 +1,141 @@
...
+  latlong: {
+      latitude: 59.9016426931744,
+      longitude: 10.607235872426175
+  }
+}
+new GeomapTool(viewer, mapConfig);
+```
+## OpenStreet Maps
+- No configuration is needed for OpenStreet Maps.
+- 'APIKey' - blank string as no API key is needed for OpenStreet Map.
```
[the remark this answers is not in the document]
pramodcog: Removed APIKey from common

documentation/versioned_docs/version-2.x/examples/geomaptool.mdx
```diff
@@ -0,0 +1,141 @@
...
+  - Free up to 50000 web map loads. Create Mapbox account from https://account.mapbox.com/auth/signup/ and generate Access token to use them in your application.
+  - Licensing details can be found at https://www.mapbox.com/pricing/.
+- OpenStreet map
+  - It does not require any keys to access its tile map images.
+  - Please check lincensing details at https://wiki.osmfoundation.org/wiki/Licence.
+
+## Mapbox Maps
+Configuration of the Mapbox are listed below which can be used to set different option for the maps.
+- 'mode': Map tile access mode.
+  - MapboxMode.Style & MapboxMode.Map_Id.
```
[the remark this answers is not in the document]
pramodcog: https://github.com/cognitedata/reveal/pull/1582#discussion_r712435500

viewer/packages/tools/src/Geomap/MapConfig.ts
pramodcog: It is provider dependent

pramodcog:
@larsmoa On `viewer.requestRedraw()` to be called for the tiles/images of map downloaded either we should call it in `requestAnimationFrame(refreshMaps)` or in `setInterval()` as the geo-three library does not provide a seems of knowing if all required map images/tiles are downloaded (It is an async call which is added as texture image).

Let me know your thoughts on this

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -21,14 +21,20 @@ export class Geomap {
...
+    var intervalId = setInterval(() => {
+      this._viewer.requestRedraw();
     }, 100);
 
-      setTimeout(() => {
-        clearInterval(intervalId);
-      }, 20000);
+    setTimeout(() => {
+      clearInterval(intervalId);
+    }, timeOut);
```
larsmoa: This will get problematic when camera is moved over a period of time. `cameraChange` will trigger many times (lets say 100) and every time you set up a new `setInterval`. Eventually there will be 100 timers requesting redraw. This is no big problem, but is not necessary so I think we should avoid it.

Instead, keep any running `intervalId` as a member variable, and make sure you call `clearInterval` before setting up any new interval.

Other than that it looks good :)

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -28,12 +29,15 @@ export class Geomap {
...
+    if(this._intervalId == 0) {
+      this._intervalId = setInterval(() => {
+        this._viewer.requestRedraw();
+      }, 100);
+    }
 
     setTimeout(() => {
-      clearInterval(intervalId);
+      clearInterval(this._intervalId);
+      this._intervalId = 0;
```
larsmoa: If you have a long changes of camera movement, this will only cause the _first_ movement to have any effect. I'd rather call `clearInterval` every time `_intervalId !== 0` to ensure the schedule occurs for the last camera movement event, not the first.

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -10,6 +10,7 @@ import { LatLongPosition, MapConfig, MapProviders } from './MapConfig';
 export class Geomap {
   private readonly _viewer: Cognite3DViewer;
   private _map: GEOTHREE.MapView;
+  private _intervalId: any = 0;
```
larsmoa, line 13: Tip: instead of using any you can use `ReturnType<typeof setInterval>`.

pramodcog: We can use to equal/assign `ReturnType<typeof setInterval>` type, I used the same but could not get through it to access the value as number, so made it as `any`.

viewer/packages/tools/src/Geomap/Geomap.ts
pramodcog: `clearInterval` is called when `timeOut` value is triggered and `_intervalId == 0` or `_intervalId > 0`. If we do not want to miss huge camera movements than I can reduce the `timeOut` value for camera Movement to 1000/1500 milli seconds, which would narrow down the error

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,87 @@
...
+      this._intervalId = setInterval(() => {
+        this._viewer.requestRedraw();
+      }, 100);
+    }
+
+    setTimeout(() => {
+      clearInterval(this._intervalId);
+      this._intervalId = 0;
+    }, timeOut);
+  }
```
larsmoa, line 42: I think you've misunderstood me. Imagine this:
1. cameraChanged is triggered. `this._intervalId` is set and an interval timer is started, requesting redraw every 100 ms
2. cameraChanged is triggered again, before the first interval timer is stopped. A new timer is started, and `this._intervalId` is overwritten.

After a while the `setTimeout` call from 1) is fired. `clearInterval(this._intervalId)` clears the timer from 2, but the timer from 1. *will never be cleared* - running for all eternity. 

Now add that cameraChanged will be triggered _alot_ and it's clear that things will go quite badly.

Instead, try something like: (note - not tested, so might need some adjustments)

```suggestion
  private startRequestRedrawTimer(timeout: number) {
    this.clearRequestRedrawTimer();
    
    this._intervalId = setInterval(() => {
      this._viewer.requestRedraw();
    }, 200);
    this._timeoutId = setTimeout(() => {
      this.clearRequestRedrawTimer();
    }, timeout);
  }
  
  private clearRequestRedrawTimer() {
    if (this._intervalId !== 0) {
      clearInterval(this._intervalId);
      clearTimeout(this._timeoutId);
      this._intervalId = 0;
      this._timeoutId = 0;
    }
  }
```

viewer/packages/tools/src/Geomap/Geomap.ts
```diff
@@ -0,0 +1,87 @@
...
+    return mapProvider;
+  }
+
+  public latLongToWorldCoordinates(latLong: LatLongPosition) {
+    return GEOTHREE.UnitsUtils.datumsToSpherical(latLong.latitude, latLong.longitude);
+  }
+
+  public dispose(): void {
+    this._viewer.removeObject3D(this._map);
+    this._viewer.off('cameraChange', () => {});
```
larsmoa, line 85: This won't unsubscribe anything. See https://github.com/cognitedata/reveal/blob/master/viewer/packages/tools/src/DebugCameraTool.ts#L23-L33 for an example of how to do this correctly.

larsmoa (changes requested):

viewer/packages/tools/src/Geomap/Geomap.ts
pramodcog: @larsmoa the above existing code will have following workflow.

1. `cameraChanged` is triggered and `this._intervalId` is set and an interval timer is started, requesting redraw every 100 ms.
2. Again `cameraChanged` is triggered, but this time `this._intervalId` value is not change because `if(this._intervalId == 0)` is false and `setInterval()` is not called until previous `setInterval()` is cleared which will happen when `timeOut` duration is completed.
3. But `setTimeout()` should have been under the `if(this._intervalId == 0)` is avoid unnecessary `clearInterval()`.

Let me know if this needs any change

larsmoa (approved):

PR #1588, opened 2021-08-27
feat: add sector parser package and functionality for loading .glb sectors

christjt:
This is a work in progress and does not need to be reviewed quite yet, but feel free to give it a try if you are interested. Also #1518 can be closed in favor of this PR (sorry @haakonflatval-cognite 😥).


christjt:
More tests and README.md are coming, but it is ready for review :D

viewer/core/src/glsl/base/constructMatrix.glsl
```diff
@@ -1,12 +1,9 @@
-// TODO consider fixing Three.js so it is possible to pass a mat4:
-// see https://stackoverflow.com/questions/38853096/webgl-how-to-bind-values-to-a-mat4-attribute
-
 mat4 constructMatrix(vec4 column_0, vec4 column_1, vec4 column_2, vec4 column_3) {
```
larsmoa, line 1: Add comment about why you ditch W-component.

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -0,0 +1,165 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+import * as THREE from 'three';
+import { MeshGPUInstancing, InstancedMesh } from '@gltf-transform/extensions';
+import { Node, NodeIO, Document, Primitive } from '@gltf-transform/extensions/node_modules/@gltf-transform/core';
```
larsmoa, line 6:
```suggestion
import { Node, NodeIO, Document, Primitive } from '@gltf-transform/core';
```

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -0,0 +1,165 @@
...
+    });
+
+    return buffs;
+  }
+
+  private processNode(node: Node): [RevealGeometryCollectionType, THREE.BufferGeometry] | null {
+    const instanced = node.getExtension<InstancedMesh>(MeshGPUInstancing.EXTENSION_NAME)!;
+    const mesh = node.getMesh()!;
+
+    if (!instanced && !mesh) return null;
```
larsmoa, line 56: When does this happen? Is this ignoring stuff we don't support?

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -0,0 +1,165 @@
...
+        setQuadGeometry(geometry);
+        break;
+      case RevealGeometryCollectionType.TrapeziumCollection:
+        setTrapeziumGeometry(geometry);
+        break;
+      case RevealGeometryCollectionType.TorusSegmentCollection:
+        setTorusGeometry(geometry);
+        break;
+      default:
+        throw new Error(`${RevealGeometryCollectionType[primitiveCollectionName]} is not supported`);
```
larsmoa, line 162: Tip: `assertNever(primitiveCollectionName)` is nice to use hear as it will fail on compile if we add a new type that we forget to add a case for.

viewer/packages/sector-parser/app/testMaterials.ts
```diff
@@ -0,0 +1,275 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+import * as THREE from 'three';
+
+import matCapTextureImage from '../../../core/src/datamodels/cad/rendering/matCapTextureData';
+import { sectorShaders } from '../../../core/src/datamodels/cad/rendering/shaders';
```
larsmoa, line 7: Importing internals from core seems a bit dodgy, even for a test app. Is it a good idea to separate rendering stuff to a new package, e.g. 'cad-rendering'?

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -0,0 +1,165 @@
...
+  QuadCollection,
+  TorusSegmentCollection,
+  TrapeziumCollection,
+  NutCollection,
+  TriangleMesh,
+  InstanceMesh
+}
+
+export default class GltfSectorParser {
+  private readonly _io: NodeIO;
```
larsmoa, line 33: I'd rather call this _gltfParser or something

larsmoa (changes requested):

viewer/packages/sector-parser/src/GltfSectorParser.ts
christjt: Sadly enough this will brake it. Not sure if it is the lib that is braking or what...

viewer/packages/sector-parser/app/testMaterials.ts
christjt: Yeah, I agree with this. But was great having this test-app when developing. I agree that importing internals is not ideal, but the alternative makes it quite hard to further develop because it is very valuable to see what is going on. I really don't want to give up the ability to run an app here.

Best approach in my opinion would be to have a package containing shaders, matcaptexture and other essentials that could be imported here. There are downsides to this though, lets discuss in person.

viewer/packages/sector-parser/src/GltfSectorParser.ts
christjt: In practice, never because the backend will not produce empty nodes. It is just a check (actually it should probably be an assert).

viewer/packages/sector-parser/app/testMaterials.ts
larsmoa: Did we decide not to do this?

viewer/packages/sector-parser/app/testMaterials.ts
```diff
@@ -0,0 +1,276 @@
...
+import * as THREE from 'three';
+
+// TODO: Fix dependencies such that test app doesn't depend on core internals
+import matCapTextureImage from '../../../core/src/datamodels/cad/rendering/matCapTextureData';
+import { sectorShaders } from '../../../core/src/datamodels/cad/rendering/shaders';
+
+const matCapTexture = new THREE.Texture(matCapTextureImage);
+matCapTexture.needsUpdate = true;
+
+export function createInstancedMeshMaterial(): THREE.ShaderMaterial {
```
larsmoa, line 13: Not something we should fix now probably, but it would be nice if the material definitions was shared with Reveal and the test app. Doable?

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  public parseSector(data: ArrayBuffer) {
+    const document: Document = this._gltfReader.readBinary(data);
+
+    const defaultScene = document.getRoot().getDefaultScene();
+    const buffs: [RevealGeometryCollectionType, THREE.BufferGeometry][] = [];
+    defaultScene!.traverse((node: Node) => {
+      const res = this.processNode(node);
+      if (res) buffs.push(res);
```
larsmoa, line 49: Rename `res` to something that better describes what processNode returns.

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -0,0 +1,176 @@
...
+    });
+
+    return buffs;
+  }
+
+  private processNode(node: Node): [RevealGeometryCollectionType, THREE.BufferGeometry] | null {
```
larsmoa, line 55: Check out named tuples which would improve readability: https://www.typescriptlang.org/play?ts=4.0.2#example/named-tuples

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -0,0 +1,176 @@
...
+    const instanced = node.getExtension<InstancedMesh>(MeshGPUInstancing.EXTENSION_NAME)!;
+    const mesh = node.getMesh()!;
+
+    if (this.isEmptyNode(instanced, mesh)) return null;
+
+    const bufferGeometry = instanced ? new THREE.InstancedBufferGeometry() : new THREE.BufferGeometry();
+
+    const geometryType = RevealGeometryCollectionType[node.getName() as keyof typeof RevealGeometryCollectionType];
```
larsmoa, line 63: dafuc - comments are sometimes nice

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    this.setInstancedAttributes(instanced, bufferGeometry);
+
+    return [geometryType, bufferGeometry];
+  }
+
+  private isEmptyNode(instanced: InstancedMesh, mesh: Mesh) {
+    return !instanced && !mesh;
+  }
```
larsmoa, line 84: When will this be true? since InstanceMesh never is  "undefined" (at least according to function declaration), this doesnt make sense to me.

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -0,0 +1,176 @@
...
+    });
+
+    return buffs;
+  }
+
+  private processNode(node: Node): [RevealGeometryCollectionType, THREE.BufferGeometry] | null {
+    const instanced = node.getExtension<InstancedMesh>(MeshGPUInstancing.EXTENSION_NAME)!;
+    const mesh = node.getMesh()!;
+
+    if (this.isEmptyNode(instanced, mesh)) return null;
```
larsmoa, line 59: Avoid the ! and use a plain if to verify that instanced and mesh is not undefined/null, e.g.
```suggestion
    const instanced = node.getExtension<InstancedMesh>(MeshGPUInstancing.EXTENSION_NAME);
    const mesh = node.getMesh()!;

    if (!instanced && !mesh) return null;
```

viewer/packages/sector-parser/tests/GltfSectorParser.test.ts
```diff
@@ -0,0 +1,187 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import fs from 'fs';
+
+import 'jest-extended';
+import GltfSectorParser, { RevealGeometryCollectionType } from '../src/GltfSectorParser';
+
+describe('Gltf Sector Parser', () => {
```
larsmoa, line 10: Still hate that test description doesn't match class under test....

viewer/packages/webpack.config.js
```diff
@@ -31,9 +36,18 @@ module.exports = env => {
           use: {
             loader: 'ts-loader',
             options: {
-              configFile: path.resolve(__dirname, '../tsconfig.json')
+              configFile: path.resolve(__dirname, '../tsconfig.json'),
+              compilerOptions: {
+                noUnusedLocals: false,
+                noUnusedParameters: false
+              }
```
larsmoa, line 43: Why do we need to change this?

larsmoa (changes requested):

viewer/packages/sector-parser/app/testMaterials.ts
christjt: Absolutely, only reason this doesn't work like this now is that the place / package where these material definitions should be doesn't exist yet.

viewer/packages/webpack.config.js
christjt: Since this is the webpack config for the test app, it is VERY annoying that you are unable to run your app if you are in the middle of creating some feature and not everything is in use (yet).

larsmoa (approved):

PR #1590, opened 2021-08-31
fix: issue causing HtmlOverlayTool to misplace elements when 3D viewer was scaled on large screens

larsmoa:

PR #1596, opened 2021-09-03
fix: issue causing HtmlOverlayTool to misplace elements when 3D viewer was scaled on large screens

larsmoa:
Backport of #1590 for 2.0


christjt (approved):

christjt:
NOTE: Please tend to the changes and fix #1590 before merging this. I simply approved since it is a cherrypick of another PR.

PR #1598, opened 2021-09-03
refac: create cad-parsers package

haakonflatval-cognite:


haakonflatval-cognite:
Examples do, in fact, work

haakonflatval-cognite:
Just an addendum: `test-utilities` (formerly `core/src/__testutilities__`) has been put in the root folder of `viewer`, and is not considered a package (i.e. it does not have a `package.json`). It is only used by tests, and each test file using it references the relevant files directly by path.

larsmoa:
Some things:
- cad-parsers/package.json depends on skmeans, but @types/skmeans is in viewer/package.json - move this together (@christjt - should we have per-package deps or should they all be in root package.json?)
- Go through the index.ts exports - we should try to reduce the number of exports more. I belive some of the types only are used in tests. Not sure how to handle these, but if we can avoid exporting the types it would be great.
- `traverseDepthFirst` is very simple and it doesn't make sense it lives in cad-parsers. I think we simply should duplicate it, or a bit more sustainable create a utility-package for these types of general tools
- Does 'transformCameraConfiguration` belong in cad-parsers?

Note that I think it's ok to ditch support we have in render-debug-widget.ts which relies some of the stuff inside cad-parsers.

viewer/packages/cad-parsers/index.ts
```diff
@@ -0,0 +1,18 @@
...
+export { MetadataRepository } from './src/metadata/MetadataRepository';
+export { CadModelMetadataRepository } from './src/metadata/CadModelMetadataRepository';
+export { CadModelMetadata } from './src/metadata/CadModelMetadata';
+export { SectorSceneFactory } from './src/utilities/SectorSceneFactory';
+export {
+  SectorScene,
+  BinaryFileProvider,
+  File3dFormat,
+  ModelDataClient,
+  BlobOutputMetadata
```
larsmoa, line 15: Seems to only be used outside cad-parsers so I dont think it belongs in this package.

haakonflatval-cognite: I see that ModelDataClient is used by `cad-parsers/src/metadata/CadModelMetadataRepository.ts`, and that it again relies on BinaryFileProvider and JsonFileProvider.. BlobOutputMetadata can be removed, I think

examples/src/utils/renderer-debug-widget.ts
```diff
@@ -411,7 +411,7 @@ function updateWantedSectorOverride(
     const acceptedSimple = filterSectorNodes(quadsFilter, root);
     const acceptedDetailed = filterSectorNodes(detailedFilter, root);
     const wanted: reveal.WantedSector[] = [
-      ...cadNode.sectorScene
+      ...(cadNode.sectorScene as reveal.cadParsers.SectorScene)
```
larsmoa, line 414: Why is this necessary now?

viewer/core/src/datamodels/cad/CadModelFactory.test.ts
```diff
@@ -4,11 +4,10 @@
 import * as THREE from 'three';
 
 import { Mutable } from '../../utilities/reflection';
-import { createCadModelMetadata } from '../../__testutilities__/createCadModelMetadata';
-import { generateSectorTree } from '../../__testutilities__/createSectorMetadata';
+import { generateSectorTree, createCadModelMetadata } from '../../../../test-utilities';
```
larsmoa, line 7: Should test-utilities be a package? Not sure how that will work dependency-wise though.

haakonflatval-cognite: I discussed a bit with Christopher, I ended up with not having it as a package. Instead it is just a folder without a package.json that is referenced from all the tests that uses it. Open to suggestions for how to do this better

examples/src/utils/renderer-debug-widget.ts
haakonflatval-cognite: Had problems compiling without it. May be that the problem is local. I'll change it and see if it works

haakonflatval-cognite: Update: It did not work. likely because the type of the argument variable in `map()` is not directly exposed through reveal. Although I don't see why that should be a problem :man_shrugging:

larsmoa: Why are cadParsers exposed though?

viewer/core/src/datamodels/cad/CadModelFactory.test.ts
larsmoa: Ok, lets leave it as is

larsmoa:
Approved, but rename PR to something a bit more informative (and preferably use e.g. "refac: [description]") pattern.

larsmoa (approved):

PR #1608, opened 2021-09-14
Create package for CAD geometry loaders

haakonflatval-cognite:


christjt (changes requested):
Overall looks really good. A step in the correct direction for sure!

Just a small NIT as a requested change.

haakonflatval-cognite:
Visual regression tests are still failing, but I blame that on Christopher

haakonflatval-cognite:
Big concern, by the way: Right now part of the styling API resides in cad-geometry-parsers, while the rest is back in core. The reason is some weird dependencies, especially that some "NodeCollection" classes depend on Cognite3DViewer

christjt (approved):

PR #1619, opened 2021-09-22
feat: add Cognite3DModel.modelUnit and modelUnitToMetersFactor

larsmoa:


viewer/core/src/public/migration/Cognite3DModel.ts
```diff
@@ -39,6 +40,35 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
...
+   * - Inches
+   * - Yards
+   * - Miles
+   * - Mils
+   * - Microinches
+   * Note that coordinates in Reveal always are converted to meters using {@see modelUnitToMetersFactor}.
+   * @version New since 2.1
+   */
+  get modelUnit(): string {
+    return this.cadNode.cadModelMetadata.scene.unit;
```
christjt, line 60: This could potentially also return empty string, which should also be indicated in the jsdoc.

christjt (approved):
Looks good, minor comment

viewer/core/src/public/migration/Cognite3DModel.ts
christjt: Also, this should actually be returned to the user as an enum, right? Or are we afraid of new measurement units to be added?

larsmoa: Added a union type for units.

viewer/core/src/public/migration/types.ts
```diff
@@ -16,6 +16,23 @@ export type Color = {
   g: number;
   b: number;
 };
+
+/**
+ * Units supported by {@link Cognite3DModel}.
+ */
+export type WellKnownUnit =
```
christjt, line 23: Rename to something like MeasurementUnit or something similar? WellKnownUnit seems very subjective, well known to who?

christjt (approved):

viewer/core/src/public/migration/types.ts
larsmoa: I think that `WellKnown` is a quite well established naming scheme to mark that it's stuff that is well known to a certain API. Some examples:
https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/enum
https://docs.microsoft.com/en-us/dotnet/api/system.security.principal.wellknownsidtype?view=net-5.0
https://docs.microsoft.com/en-us/dotnet/api/?term=WellKnown

PR #1620, opened 2021-09-22
fix: when two models was loaded and one was ghosted the other model was invisible

larsmoa:
Resolves ACEPC-110


viewer/core/src/datamodels/cad/rendering/EffectRenderManager.ts
```diff
@@ -508,6 +509,16 @@ export class EffectRenderManager {
       result.hasBackElements = result.hasBackElements || hasBackElements;
       result.hasInFrontElements = result.hasInFrontElements || hasInFrontElements;
       result.hasGhostElements = result.hasGhostElements || hasGhostElements;
+    });
+
+    // Split scenes based on what render stages we need
+    const { hasBackElements, hasInFrontElements, hasGhostElements } = result;
+    this._rootSectorNodeBuffer.forEach(rootSectorNodeData => {
```
haakonflatval-cognite, line 516: So the bug fix was to compute `hasBackElements` etc. for every `rootSectorNodeData` and then using the accumulated result for determining what to show?

larsmoa: Yes

haakonflatval-cognite (approved):
Looks good :+1:

PR #1622, opened 2021-09-23
fix: optimize floatBitSubset shader

christjt:


viewer/core/src/glsl/math/floatBitsSubset.glsl
```diff
@@ -1,17 +1,9 @@
...
-      if(i >= a) break;
-      float backBits = pow(2.0, lsbif + float(i));
-      outNumber += (mod(inNumber, backBits * 2.0) - mod(inNumber, backBits)) / denominator;
-    }
-    return outNumber;
+    float r = float(fromLeastSignificantBitIndex);
+    float l = float(toMostSignificantBitIndex);
+
+    float bitShift = pow(2.0, r);
+    return mod(((inNumber - mod(inNumber, bitShift)) / bitShift), pow(2.0, l - r));
```
larsmoa, line 6: Very cool, but I think it would be good if you doc this with the example you showed

larsmoa (approved):

PR #1623, opened 2021-09-23
feat: add support for limiting loaded geometry by "render cost"

larsmoa:
Note that the default budget for render cost is Infinity, so this doesn't affect the default behaviour.


christjt (approved):
LGTM

PR #1627, opened 2021-09-28
fix:handle non-determinant depth values after resizing when MSAA is enabled causing custom objects to be invisible

christjt:
I am not proud of this…

### DIFF
--- a/viewer/core/src/glsl/post-processing/outline-detect-combine.frag
+++ b/viewer/core/src/glsl/post-processing/outline-detect-combine.frag
@@ -72,8 +72,10 @@ void main() {
   float customDepth = texture2D(tCustomDepth, vUv).r;
   float ghostDepth = texture2D(tGhostDepth, vUv).r;
 
-  // This is a hack to make sure that all textures are initialize
-  // If a texture is unused, it will have a clear value of 0.0
+  // This is a hack to make sure that all textures are initialized
+  // If a texture is unused, it will have a clear value of 0.0.
+  // Without this we've seen issues with MSAA where resizing render targets
+  // causes depth to cleared to either 1 or 0 depending on the device/browser
   customDepth = customDepth > 0.0 ? customDepth : 1.0; 
   backDepth = backDepth > 0.0 ? backDepth : 1.0;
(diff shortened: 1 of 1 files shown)


viewer/core/src/glsl/post-processing/outline-detect-combine.frag
```diff
@@ -71,6 +71,11 @@ void main() {
   float backDepth = texture2D(tBackDepth, vUv).r;  
   float customDepth = texture2D(tCustomDepth, vUv).r;
   float ghostDepth = texture2D(tGhostDepth, vUv).r;
+
+  customDepth = customDepth > 0.0 ? customDepth : 1.0; 
+  backDepth = backDepth > 0.0 ? backDepth : 1.0; 
+  ghostDepth = ghostDepth > 0.0 ? ghostDepth : 1.0;
+  frontDepth = frontDepth > 0.0 ? frontDepth : 1.0;
```
larsmoa, line 78: add a comment explaining this (yes - it should have been there in the code you removed too, but hey - never to late to improve)

larsmoa:
image

viewer/core/src/glsl/post-processing/outline-detect-combine.frag
```diff
@@ -71,6 +71,17 @@ void main() {
   float backDepth = texture2D(tBackDepth, vUv).r;  
   float customDepth = texture2D(tCustomDepth, vUv).r;
   float ghostDepth = texture2D(tGhostDepth, vUv).r;
+
+  // This is a hack to make sure that all textures are initialize
+  // If a texture is unused, it will have a clear value of 0.0
```
larsmoa, line 76:
```suggestion
  // This is a hack to make sure that all textures are initialized
  // If a texture is unused, it will have a clear value of 0.0.
  // Without this we've seen issues with MSAA where resizing render targets
  // causes depth to cleared to either 1 or 0 depending on the device/browser
```

larsmoa (approved):
Approved - but optimally I think we should try to write visual tests to avoid regressions for these types of issues in the future. Since I'm not able to reproduce the issue I wont make an attempt on this now...

PR #1635, opened 2021-10-04
feat: camera controls update with click target change

Savokr:
Update for camera controls in Cognite3DViewer. Update includes:

- Change of camera target when user clicks on any object
- Different algorithm of changing camera target when zooming with mouse wheel
- Prevent going inside objects when zooming with mouse wheel
- Small fixes


viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -78,6 +79,8 @@ export default class ComboControls extends EventDispatcher {
   private _accumulatedMouseMove: Vector2 = new Vector2();
   private domElement: HTMLElement;
   private target: Vector3 = new Vector3();
+  private viewTarget: Vector3 = new Vector3();
```
christjt, line 82: all these fields should have an `_` prefix

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -274,7 +286,7 @@ export default class ComboControls extends EventDispatcher {
     // @ts-ignore event.wheelDelta is only part of WebKit / Opera / Explorer 9
     if (event.wheelDelta) {
       // @ts-ignore event.wheelDelta is only part of WebKit / Opera / Explorer 9
-      delta = -event.wheelDelta / 40;
+      delta = -event.wheelDelta / 40; // Why division by 40?
```
christjt, line 289: I don't know, why? Either add a TODO with an actionable, or remove the comment all together. This comment does not add anything.

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -500,12 +513,13 @@ export default class ComboControls extends EventDispatcher {
...
     const moveForward = keyboard.isPressed('w') ? true : keyboard.isPressed('s') ? false : undefined;
+
     if (moveForward !== undefined) {
-      this.dolly(0, 0, this.getDollyDeltaDistance(moveForward, keyboardDollySpeed * speedFactor));
+      this.dolly(0, 0, this.getDollyDeltaDistance(moveForward, keyboardDollySpeed * speedFactor), true);
       this.firstPersonMode = true;
     }
 
-    // pan
+    // pan (actually not seems like a pan movement according to google) should we rename it?
```
christjt, line 522: Sure

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -583,37 +591,58 @@ export default class ComboControls extends EventDispatcher {
...
+        radius = minDistance;
+        if (dynamicTarget) {
+          // push targetEnd forward
+          reusableCamera.getWorldDirection(cameraDirection);
+          targetEnd.add(cameraDirection.normalize().multiplyScalar(Math.abs(deltaDistance)));
+        } else {
+          // stops camera from moving forward only if target became close to scroll target
+          if (this.scrollTarget.clone().sub(this.target).length() < 0.2) deltaDistance = distToTarget - radius;
+        }
       }
```
christjt, line 610: Try to clean this up if you can to avoid so many indent-blocks. Either by precomputing state, or extracting into private functions (or both).

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -583,37 +591,58 @@ export default class ComboControls extends EventDispatcher {
...
 
-    const distFromRayOrigin = -deltaDistance * ratio;
+      // Here we use the law of sines to determine how far we want to move the target.
+      // Direction is always determined by scrollTarget-target vector
+      const tarSTarVec = reusableVector3.subVectors(this.scrollTarget, this.target).normalize(),
+        camTarVec = new Vector3().subVectors(this.target, this.camera.position),
+        camSTarVec = new Vector3().subVectors(this.scrollTarget, this.camera.position);
+
+      const tarCamSTarAngle = camTarVec.angleTo(camSTarVec),
+        tarSTarCamAngle = tarSTarVec.negate().angleTo(camSTarVec.negate());
```
christjt, line 619: Try to find some better names for these variables. Long variable names is ok if needed. Better to be expressive than save storage space for the script.

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -583,37 +591,58 @@ export default class ComboControls extends EventDispatcher {
...
+
+      let targetOffsetDistance = deltaDistance * (Math.sin(tarCamSTarAngle) / Math.sin(tarSTarCamAngle));
+
+      const ratio = Math.abs(targetOffsetDistance / deltaDistance);
```
christjt, line 623: what kind of ratio is this?

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -583,37 +591,58 @@ export default class ComboControls extends EventDispatcher {
...
 
-    sphericalEnd.radius = radius;
+      if (ratio > 4)
+        if (ratio > 10) targetOffsetDistance *= 0.1;
+        else targetOffsetDistance *= 0.4;
```
christjt, line 627: Fix magic numbers

viewer/packages/camera-manager/src/Keyboard.ts
```diff
@@ -12,8 +12,11 @@ const keyMap: { [s: string]: string } = {
   39: 'right',
   40: 'down',
   65: 'a',
+  66: 'b',
+  67: 'c',
   68: 'd',
   69: 'e',
+  70: 'f',
```
christjt, line 19: Were these added for a specific reason, or just for the test-app?

viewer/core/src/datamodels/cad/rendering/EffectRenderManager.ts
```diff
@@ -186,7 +186,7 @@ export class EffectRenderManager {
     this._customObjectRenderTarget.depthTexture.format = THREE.DepthFormat;
     this._customObjectRenderTarget.depthTexture.type = THREE.UnsignedIntType;
 
-    this._compositionTarget = new THREE.WebGLRenderTarget(0, 0, { stencilBuffer: false });
+    this._compositionTarget = new THREE.WebGLRenderTarget(0, 0, { stencilBuffer: false }); // Final depth buffer
```
christjt, line 189: Not sure if this comment is correct / makes sense. This rendertarget outputs a color-buffer as well.

christjt (changes requested):
All-in-all it looks good. Have a few code style suggestions and nits that you can take a look at. This is just an initial look at the code and I will take another look where I try out the feature as well as taking a deeper look at the solution.

Really great that we finally are attempting to iterate on the camera controls. It is something that has been bugging our clients (and me 😂) for a long time.

examples/yarn.lock
```diff
@@ -1264,13 +1264,6 @@
   dependencies:
     comlink "4.3.1"
 
-"@cognite/reveal-parser-worker@1.2.0":
```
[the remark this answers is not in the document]
Savokr: Don't know either 🙂 . I can add it back

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -162,6 +163,22 @@ export class Cognite3DViewer {
...
+  private changeTarget = async (event: any) => {
+    const { offsetX, offsetY } = event;
+
+    const intersection = await this.getIntersectionFromPixel(offsetX, offsetY);
+    if (intersection !== null) {
+      this.setCameraTarget(intersection.point, true);
+    } else {
+      const boBo = this._models[0].getModelBoundingBox();
+      this.setCameraTarget(boBo.getCenter(new THREE.Vector3()), true);
+    }
```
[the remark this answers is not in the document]
Savokr: Forgot to remove 'else' case, we probably want to to do just nothing when we click not on the model.

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -162,6 +163,22 @@ export class Cognite3DViewer {
...
+   * @param event MouseEvent that contains pointer location data.
+   */
+  private changeTarget = async (event: any) => {
+    const { offsetX, offsetY } = event;
+
+    const intersection = await this.getIntersectionFromPixel(offsetX, offsetY);
+    if (intersection !== null) {
+      this.setCameraTarget(intersection.point, true);
+    } else {
+      const boBo = this._models[0].getModelBoundingBox();
```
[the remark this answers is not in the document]
Savokr: Removed usage of bounding box, so won't be a problem 🙂

examples/yarn.lock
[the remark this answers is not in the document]
Savokr: Oh, it is still in the yarn.lock but a little higher. Check lines 1260. Maybe it was duplicated for some reason and removed automatically by yarn.

larsmoa:
Haven't done a full review, only some testing and I think we're heading in the right direction, but:
- This _really_ needs to be configurable and I believe it in a the first release should be *opt-in*. Next we can consider *opt-out* and finally we can consider removing the old mode. This is important and must be in place before a merge as we cannot rely on having click-listener for this to work.
- I dont think the "zoom to cursor" works well for this. Consider when user clicks on an object in the corner of the screen. Target is changed and object will be in the center of the screen. The mouse cursor of course stays in the corner. User will often want to zoom in on the clicked object and uses mouse scroll to do so. However, since the mouse cursor is not in the center, the user ends up zooming towards another point than the object clicked. To me it doesn't feel natural that I have to move the mouse cursor again after clicking an object...

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -603,17 +617,90 @@ export default class ComboControls extends EventDispatcher {
...
+    const targetCameraScrollTargetAngle = cameraToTargetVec.angleTo(cameraToScrollTargetVec),
+      targetScrollTargetCameraAngle = targetToScrollTargetVec.negate().angleTo(cameraToScrollTargetVec.negate());
+
+    let targetOffsetDistance =
+      deltaDistance * (Math.sin(targetCameraScrollTargetAngle) / Math.sin(targetScrollTargetCameraAngle));
+
+    const targetOffsetToDeltaratio = Math.abs(targetOffsetDistance / deltaDistance);
+
+    // if target movement is too fast we want to slow it down a bit
+    const downscaleCoefficient = targetOffsetToDeltaratio > 4 ? (targetOffsetToDeltaratio > 10 ? 0.4 : 0.2) : 1;
```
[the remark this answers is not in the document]
Savokr: Fixed, hopefully looks good now

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1186,22 +1285,113 @@ export class Cognite3DViewer {
...
+      this.controls.setState(this.getCameraPosition(), target);
+      return;
+    }
+
+    const { camera, raycaster, _minDefaultAnimationDuration, _maxDefaultAnimationDuration } = this;
+
+    if (duration == null) {
+      const distance = target.distanceTo(this.controls.getState().target);
+      duration = distance * 125; // 125ms per unit distance
+      duration = Math.min(Math.max(duration, _minDefaultAnimationDuration), _maxDefaultAnimationDuration);
```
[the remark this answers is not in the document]
Savokr: Tested on different models and looks ok for me. Also, this value is based on distance to the camera, so, from my understanding should behave the same in all models because when they are imported in reveal they have the same scale.

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1186,22 +1285,113 @@ export class Cognite3DViewer {
...
-    const { camera } = this;
+    const { camera, raycaster, _minDefaultAnimationDuration, _maxDefaultAnimationDuration } = this;
 
     if (duration == null) {
       const distance = position.distanceTo(camera.position);
-      duration = distance * 125; // 250ms per unit distance
-      duration = Math.min(Math.max(duration, 600), 2500); // min duration 600ms and 2500ms as max duration
+      duration = distance * 125; // 125ms per unit distance
+      duration = Math.min(Math.max(duration, _minDefaultAnimationDuration), _maxDefaultAnimationDuration);
     }
```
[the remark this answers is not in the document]
Savokr: Extracted to calculateDefaultDuration function

viewer/core/src/public/migration/types.ts
```diff
@@ -87,6 +87,31 @@ export interface Cognite3DViewerOptions {
...
+   * Default is true.
+   */
+  useScrollTargetControls?: boolean;
+
+  /**
+   * Enables or disables change of camera target on mouse click. New target is then set to the point of the model under current cursor position.
+   *
+   * Default is false.
+   */
+  useOnClickTargetChange?: boolean;
```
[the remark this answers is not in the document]
Savokr: Finally working properly in runtime

examples/src/pages/Migration.tsx
```diff
@@ -89,6 +90,13 @@ export function Migration() {
       viewer = new Cognite3DViewer(viewerOptions);
       (window as any).viewer = viewer;
 
+      const controlsOptions: CameraControlsOptions = {
+        onClickTargetChange: true,
+        zoomToCursor: 'scrollTarget'
+      }
+
+      viewer.setCameraControlsMode(controlsOptions);
```
larsmoa, line 98: This looks really nice. Could you also add some stuff to the Migration UI to allow setting different options so it's easy to test out the different modes there?

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -115,11 +116,24 @@ export class Cognite3DViewer {
     disposed: new EventTrigger<DisposedDelegate>()
   };
 
+  private readonly _cameraControlsOptions = {
+    canInterruptAnimations: false,
+    useScrollTargetControls: false,
+    useOnClickTargetChange: false,
+    scrollEventsAdded: false,
+    clickEventsAdded: false
+  };
```
larsmoa, line 125: Why can't we just use `CameraControlsOptions` here?

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -155,6 +169,125 @@ export class Cognite3DViewer {
       new THREE.Vector3()
     )
   };
+  /**
+   * Convert pixel coordinates of the cursor to [-1,1]^2 coordinates.
+   * @param pixelX
+   * @param pixelY
+   */
+  private convertPixelCoordinatesToNormalized = (pixelX: number, pixelY: number) => {
```
larsmoa, line 177:
```suggestion
  private convertPixelCoordinatesToNormalized(pixelX: number, pixelY: number) {
```

larsmoa: We use "standard function" and not "lambda functions" (unsure what the naming of this is)

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -210,9 +343,11 @@ export class Cognite3DViewer {
...
     this._renderer.localClippingEnabled = true;
 
-    this._automaticNearFarPlane = options.automaticCameraNearFar !== undefined ? options.automaticCameraNearFar : true;
-    this._automaticControlsSensitivity =
-      options.automaticControlsSensitivity !== undefined ? options.automaticControlsSensitivity : true;
+    this._automaticNearFarPlane = options.automaticCameraNearFar ?? true;
+    this._automaticControlsSensitivity = options.automaticControlsSensitivity ?? false;
+    //this._canInterruptAnimations = options.canInterruptAnimations ?? false;
+    //this._useScrollTargetControls = options.useScrollTargetControls ?? false;
+    //this._useOnClickTargetChange = options.useOnClickTargetChange ?? false;
```
larsmoa, line 350:
```suggestion
```

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -233,10 +368,13 @@ export class Cognite3DViewer {
     this.scene = new THREE.Scene();
     this.scene.autoUpdate = false;
 
+    //if (this._cameraControlsOptions.useOnClickTargetChange || this._cameraControlsOptions.useScrollTargetControls) this.addOtherControlsModes();
```
larsmoa, line 371:
```suggestion
```

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -233,10 +368,13 @@ export class Cognite3DViewer {
...
+
     this.controls = new ComboControls(this.camera, this.canvas);
     this.controls.dollyFactor = 0.992;
-    this.controls.minDistance = 1.0;
+    this.controls.minDistance = 0.15;
     this.controls.maxDistance = 100.0;
+    //this.controls.useScrollTarget = this._cameraControlsOptions.useScrollTargetControls;
```
larsmoa, line 377:
```suggestion
```

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -457,6 +595,54 @@ export class Cognite3DViewer {
...
+      case 'scrollTarget':
+        this._cameraControlsOptions.useScrollTargetControls = true;
+        this.controls.useScrollTarget = true;
+
+        if (!this._cameraControlsOptions.scrollEventsAdded) {
+          this.setupOtherControlsModes();
+        }
+        break;
+      default:
+        break;
```
larsmoa, line 634: use `assertNever(controlsOptions.zoomToCursor)` to ensure you handle all valid enum values.

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -457,6 +595,54 @@ export class Cognite3DViewer {
...
+   * @param controlsOptions JSON object with camera controls options.
+   */
+  setCameraControlsMode(controlsOptions: CameraControlsOptions) {
+    this._cameraControlsOptions.useOnClickTargetChange =
+      controlsOptions?.onClickTargetChange ?? this._cameraControlsOptions.useOnClickTargetChange;
+    this._cameraControlsOptions.canInterruptAnimations =
+      controlsOptions?.canInterruptAnimations ?? this._cameraControlsOptions.canInterruptAnimations;
+
+    switch (controlsOptions?.zoomToCursor) {
+      case 'disable':
```
larsmoa, line 609:
```suggestion
      case 'disable':
      case undefined:
```
When undefined, you want to have it disabled - currently no change happens if this is not defined (which I think is not what we want).

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -457,6 +595,54 @@ export class Cognite3DViewer {
     }
   }
 
+  /**
+   * Sets camera controls mode anything allowed in CameraControlsOptions type.
+   * @param controlsOptions JSON object with camera controls options.
+   */
+  setCameraControlsMode(controlsOptions: CameraControlsOptions) {
+    this._cameraControlsOptions.useOnClickTargetChange =
+      controlsOptions?.onClickTargetChange ?? this._cameraControlsOptions.useOnClickTargetChange;
```
larsmoa, line 604: `controlsOptions` can never be undefined, so no reason to use `?` here

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -155,6 +169,125 @@ export class Cognite3DViewer {
...
+      .add(this.camera.position);
+
+    return farPoint;
+  };
+
+  /**
+   * Adds or removes event listeners for additional features of camera controls.
+   * @param removeListeners
+   */
+  private setupOtherControlsModes = (removeListeners?: boolean) => {
```
larsmoa, line 209: The whole setupOtherControlsModes() and how this is handled is a bit confusing - what is "other controls modes"? Can this be changed to e.g. `setupControls(mode: ...)` ?

viewer/core/src/public/migration/types.ts
```diff
@@ -176,6 +175,36 @@ export interface AddModelOptions {
...
 }
 
+export type CameraControlsOptions = {
+  /**
+   * Enables or disables the usage of "zoom-to-cursor" mechanics based on either lerping in the direction of the cursor or using so-called scroll target notion.
+   * When mouse wheel scroll is initiated scroll target is set to the point on the model where cursor is hovering over. Then when zooming is happening camera
+   * target moves towards scroll target with speed proportional to zooming speed. This prevents the camera to go through the model when using mouse navigation,
+   * keyboard navigation still allows to go through the model.
+   *
+   * Default is true.
```
larsmoa, line 185:
```suggestion
   * Default is 'disable'.
```

viewer/core/src/public/migration/Cognite3DViewer.ts
larsmoa, line 609:
```suggestion
      case undefined:
      case 'disable':
```
Handle the default case.

viewer/core/src/public/migration/Cognite3DViewer.ts
larsmoa: I notice that we don't follow this standard in ComboControls.ts - this is because it's been migrated from another repository. We should probably change this class as well though (but I dont think you should bother to do that in this PR)

larsmoa (changes requested):

viewer/core/src/public/migration/Cognite3DViewer.ts
Savokr: Will change that

examples/src/pages/Migration.tsx
Savokr: Added

viewer/core/src/public/migration/Cognite3DViewer.ts
Savokr: Change everything to normal notation and notated that we want to change ComboControls in another PR

viewer/core/src/public/migration/Cognite3DViewer.ts
Savokr: Yup

viewer/core/src/public/migration/Cognite3DViewer.ts
Savokr: Using them now

larsmoa:
Usability testing of the different modes (this is without looking at the code, because thats the context our users will be in):

`scrollToTarget`:
- Nice feeling when zooming into a object
- I think a better name would be `zoomToCursor`
- Confusing when zooming back out. I'd expect that I ended up in the original location when I zoom in and then out again without moving cursor - i.e. that I would zoom "along the same vector" when zooming in and out.
scroll-target-zoom-in-zoom-out
t

`basicLerp`:
- Not 100% sure whats going on - what controls zoom speed?
- Is this what the current mode is?
- Not sure about naming here

`disabled`:
- Zoom to target, regardless of where the cursor is?
- Maybe call this `zoomToTarget`

Target on click:
- Works well, nice

Controls actions stop animations:
- I don't understand what this does, can't notice any effect
- Thinking we should drop this one and just choose the value that makes most sense

viewer/core/src/public/migration/types.ts
```diff
@@ -191,6 +190,29 @@ export interface AddModelOptions {
...
+   * target moves towards scroll target with speed proportional to zooming speed. This prevents the camera to go through the model when using mouse navigation,
+   * but keyboard navigation still allows to go through the model.
+   *
+   * Default is 'zoomPastCursor'.
+   *
+   */
+  mouseWheelAction?:
+    | 'zoomToTarget' // mouse wheel scroll zooms just to the current target (center of the screen) of the camera
+    | 'zoomPastCursor' // mouse wheel scroll zooms in the direction of the ray coming from camera through cursor screen position, allows going through objects
+    | 'zoomToCursor'; // mouse wheel scroll zooms towards the point on the model which cursor is hovering over, doesn't allow going through objects
```
larsmoa, line 206: The descriptions you have inline here are very useful and should be placed in the `/** ... */`-doc above so it's also accessible to the users.

viewer/packages/camera-manager/app/index.ts
```diff
@@ -3,12 +3,18 @@
...
 import ComboControls from '../';
+import Keyboard from '../src/Keyboard';
 
 let renderer: THREE.WebGLRenderer;
 let camera: THREE.PerspectiveCamera;
 let scene: THREE.Scene;
 let controls: ComboControls;
 let sphere: THREE.Mesh;
+let keyboard: Keyboard;
+let cuState: {
```
larsmoa, line 14: rename

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -131,17 +142,17 @@ export default class ComboControls extends EventDispatcher {
...
+      _sphericalEnd,
+      _deltaTarget,
       handleKeyboard,
       enableDamping,
       dampingFactor,
       EPSILON,
-      targetFPS,
+      _targetFPS,
       enabled
     } = this;
```
larsmoa, line 157: This isn't important, but for the next time I'd change this into 
```
const {
      _camera: camera,
      _target: target,
      _targetEnd: targetEnd,
      _spherical: spherical,
...
    } = this;
```
to avoid all the changes below.

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -492,119 +517,124 @@ export default class ComboControls extends EventDispatcher {
...
+    const camera = this._camera as OrthographicCamera;
     camera.zoom *= 1 - deltaDistance;
     camera.zoom = MathUtils.clamp(camera.zoom, this.minZoom, this.maxZoom);
     camera.updateProjectionMatrix();
   };
 
-  private readonly dollyPerspectiveCamera = (x: number, y: number, deltaDistance: number) => {
-    const { dynamicTarget, minDistance, raycaster, reusableVector3, sphericalEnd, targetEnd, camera, reusableCamera } =
-      this;
+  private readonly calculateTargetOfssetLerp = (
```
larsmoa, line 610:
```suggestion
  private readonly calculateTargetOffsetLerp = (
```

larsmoa: ...and update usages

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -613,45 +643,162 @@ export default class ComboControls extends EventDispatcher {
...
+    return targetOffset;
+  };
+
+  // Function almost equal to mapLinear except it is behaving the same as clamp outside of specifed range
+  private readonly clampedMap = (value: number, xStart: number, xEnd: number, yStart: number, yEnd: number) => {
+    if (value < xStart) value = yStart;
+    else if (value > xEnd) value = yEnd;
+    else value = MathUtils.mapLinear(value, xStart, xEnd, yStart, yEnd);
+
+    return value;
```
larsmoa, line 662:
```suggestion
    return MathUtils.clamp(MathUtils.mapLinear(value, xStart, xEnd, yStart, yEnd), yStart, yEnd);
```

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -613,45 +643,162 @@ export default class ComboControls extends EventDispatcher {
...
+  };
+
+  private readonly calculateTargetOfssetScrollTarget = (deltaDistance: number, cameraDirection: THREE.Vector3) => {
```
larsmoa, line 665:
```suggestion
  private readonly calculateTargetOffsetScrollTarget = (deltaDistance: number, cameraDirection: THREE.Vector3) => {
```

larsmoa (changes requested):

viewer/packages/camera-manager/src/ComboControls.ts
Savokr: Didn't know about that one, will use in the future

viewer/packages/camera-manager/src/ComboControls.ts
Savokr: Refactored these guys...

viewer/packages/camera-manager/src/ComboControls.ts
Savokr: Seems to be wrong, because clamp function will set the `value` to `xStart`, not `yStart` when it's less than `xStart` (same for upper bound).  This was the reason why I decided to make it a separate function.

larsmoa (approved):
I think this looks very good now :) 

We need some proper documentation on this however, but do that in a separate PR (it can probably wait for CameraManager)

christjt (approved):

PR #1639, opened 2021-10-05
fix: pin comment-on-pr package to 1.2 to fix duplicated comments about preview docs

christjt:
Last fix was incorrect and now created a new comment on each run.

### DIFF
--- a/.github/workflows/preview-documentation.yml
+++ b/.github/workflows/preview-documentation.yml
@@ -56,7 +56,7 @@ jobs:
           keep_files: false
 
       - name: Add comment about preview URL
-        uses: unsplash/comment-on-pr@v1.3.0
+        uses: unsplash/comment-on-pr@v1.2.0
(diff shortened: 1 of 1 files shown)


.github/workflows/preview-documentation.yml
```diff
@@ -56,12 +56,11 @@ jobs:
           keep_files: false
 
       - name: Add comment about preview URL
-        uses: unsplash/comment-on-pr@v1.3.0
+        uses: unsplash/comment-on-pr@v1.2.0
```
haakonflatval-cognite, line 59: Is there any reason for choosing version 1.2 instead of `master`, as we had before?

christjt: Because that is where the bug is 😂

haakonflatval-cognite (approved):
Great success :ok_hand:

PR #1652, opened 2021-10-12
fix: setDefaultNodeApperance(Hidden) + styled node collection with Highlighted nodes to result in all objects to be hidden

larsmoa:
This resolves an issue when doing this:
```
model.setDefaultModelAppearance(DefaultNodeAppearance.Hidden);
model.assignStyledNodeCollection(collection, DefaultNodeAppearance.Highlighted);
```
In this case all nodes would remain hidden, since `Highlighted` didn't set the visible flag.


haakonflatval-cognite:
For the record - I'm not going rogue, Lars said it could be merged

PR #1671, opened 2021-10-26
fix: transparent custom objects was not blended with CAD when background color was overridden

larsmoa:
Fixes https://cognitedata.atlassian.net/browse/ACEPC-116


christjt (approved):
LGTM! There are failing visual tests though :(

Issue #1672, opened 2021-10-27
`worldToNormalizedViewportCoordinates` does not support devices with a non 1-to-1 pixel ratio


hayer:
The current implementation of `worldToNormalizedViewportCoordinates` seems to not support pixel ratios different from 1.

```
const pixelRatio = 1; // renderer.getPixelRatio();
...

const x = (position.x + 1) / (scaleX * pixelRatio * 2);
const y = (-position.y + 1) / (scaleY * pixelRatio * 2);
```

haakonflatval-cognite:
We ended up addressing this issue with PR #1700. I don't know what specific issue (if any) you experienced with the previous implementation, but it was hopefully addressed by that PR.

PR #1673, opened 2021-10-27
fix: erroneous conversion when converting from world space to screen space and device pixel ratio is not 1

hayer:
Possible fix for #1672 .


pramodcog:
@hayer can you update the failing unit test, so that we can take next steps to move forward in the process

hayer:
> @hayer can you update the failing unit test, so that we can take next steps to move forward in the process

Please see my reply to @larsmoa regarding the return values. I can update the tests, but at the moment I am a little confused about what they should return.

larsmoa:
Replaced by #1700

PR #1665, opened 2021-10-21
feat: add support for gltf in the streaming pipeline

christjt:
As an intermidiate step I have created a PassThroughCuller which simply schedules all sectors for loading. Can be replaced by new sector culler.

PR #1629, opened 2021-09-29
feat: new sector culler for the v9 file format sector tree

larsmoa:

PR #1674, opened 2021-10-27
experiment: GLTF with new sector culler

larsmoa:
Combination of #1665 and #1629


christjt:
Can we close this @larsmoa ?

PR #1677, opened 2021-11-02
improvement: better shading for ghosted geometry and better render performance

larsmoa:
Disable transparency for ghost mode to improve rendering performance. Improve shading for ghost mode geometry to make it look nicer without transparency.

Before:
image

After:
image


viewer/packages/rendering/src/glsl/base/updateFragmentColor.glsl
```diff
@@ -36,8 +36,8 @@ void updateFragmentColor(
         gl_FragColor = vec4(albedo.rgb * mc.rgb * 1.7, color.a);
     } else if (renderMode == RenderTypeGhost) {
         float amplitude = max(0.0, dot(normal, vec3(0.0, 0.0, 1.0)));
-        vec3 albedo = min(vec3(0.8) * (0.4 + 0.6 * amplitude), 1.0);
-        gl_FragColor = vec4(albedo, 0.2);
+        float s = 0.4 + 0.6 * amplitude;
+        gl_FragColor = vec4(s, s, s, 0.3);
```
christjt, line 40: What are these changes exactly? Is is equivilant?

christjt:
```suggestion
        gl_FragColor = vec4(vec3(s), 0.3);
```

christjt (changes requested):
Looks like a nice change. Not sure what the clients will think as I assume they are accustomed to the "old" way of ghosting, so I guess we will have to wait and see. 

Also, in `materials.ts` we set a `transparent` flag for all out shader materials. I guess these can be removed now?

viewer/packages/rendering/src/glsl/base/updateFragmentColor.glsl
larsmoa: It's changing the shading to basically use the same shading as normal geometry, just in grayscale and without matcap.

christjt (approved):

PR #1678, opened 2021-11-03
fix: Cognite3DModel duplicates node collections that get reassigned style

haakonflatval-cognite:


christjt (changes requested):
Looks good, can you add a visual test as well?

larsmoa (approved):
Great, but mention what the issue is in the title (makes it easier to make a changelog). I suggest that you use the format "fix: [something that the user experienced]"

larsmoa:
> Looks good, can you add a visual test as well?

Not sure I agree that we need one we have unit test?

christjt (approved):

PR #1679, opened 2021-11-03
Fix off-by-one-error in IndexSet

haakonflatval-cognite:
In index set inversion, specifically

Wish that we had just used the first-inclusive-last-exclusive convention for NumericRanges to begin with :/


christjt (approved):

larsmoa:
Would love a unit test and a description of what fails

PR #1680, opened 2021-11-04
feat: New camera manager class for better controls customisation and configuration

Savokr:
Move of all camera related functionality to a separate CameraManager class with possibility of easier configuration/extension of camera controls.


viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,457 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import * as THREE from 'three';
+import TWEEN from '@tweenjs/tween.js';
+import { ComboControls } from './ComboControls';
+import { CallbackData } from './types';
+
+export class CameraManager extends THREE.EventDispatcher {
```
larsmoa, line 10: Why does this inherit EventDispatcher?

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,457 @@
...
+        )
+    };
+
+    /**
+   * Calculates new target when raycaster doesn't have any intersections with the model.
+   * @param cursorPosition Cursor position for desired calculations.
+   * @param cursorPosition.x
+   * @param cursorPosition.y
+   */
+  private calculateMissedRaycast = (cursorPosition: { x: number; y: number }, modelsBB: THREE.Box3): THREE.Vector3 => {
```
larsmoa, line 56:
```suggestion
  private calculateMissedRaycast(cursorPosition: { x: number; y: number }, modelsBB: THREE.Box3): THREE.Vector3  {
```
Why usually use this way of defining functions. No biggie, but nice to standardise (if there aren't any good reasons why we should do something else)

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,457 @@
...
+import { ComboControls } from './ComboControls';
+import { CallbackData } from './types';
+
+export class CameraManager extends THREE.EventDispatcher {
+    public controls: ComboControls;
+    
+    private _camera: THREE.PerspectiveCamera;
+    private _domElement: HTMLElement;
+    
+    private modelRaycast: (x: number, y: number) => Promise<CallbackData>;
```
larsmoa, line 16:
```suggestion
    private readonly _modelRaycastCallback: (x: number, y: number) => Promise<CallbackData>;
```

viewer/packages/camera-manager/src/types.ts
```diff
@@ -0,0 +1,37 @@
...
+        point: THREE.Vector3;
+        /**
+         * Distance from the camera to the intersection.
+         */
+        distanceToCamera: number;
+    } | null;
+    /**
+     * Bounding box for all models on the scene
+     */
+    modelsBB: THREE.Box3;
```
larsmoa, line 36: Rename to `modelsBoundingBox`

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: I thought that it could be similar to how ComboControls throw 'cameraChange' event and because of that inherit EventDispatcher class. Probably can remove it for now

viewer/packages/camera-manager/src/types.ts
Savokr: Done

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Changed to normal

viewer/core/src/public/migration/InputManager.ts
```diff
@@ -0,0 +1,3 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
```
larsmoa, line 3: Remove this empty file for now.

viewer/packages/camera-manager/src/CameraManager.ts
larsmoa: Yeah, remove it - not used currently.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,534 @@
...
+import { assertNever } from '@reveal/utilities';
+
+const DefaultCameraControlsOptions: Required<CameraControlsOptions> = {
+  zoomToCursor: 'basicLerp',
+  onClickTargetChange: false,
+  canInterruptAnimations: false
+};
+
+export class CameraManager extends THREE.EventDispatcher {
+  public controls: ComboControls;
```
larsmoa, line 18: I don't think this should be public. Also add `readonly`

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,534 @@
...
+  private _cameraControlsOptions: Required<CameraControlsOptions> = { ...DefaultCameraControlsOptions };
+
+  public automaticNearFarPlane: boolean = true;
+  public automaticControlsSensitivity = false;
+
+  private isDisposed = false;
+  private readonly _animationDuration: number = 600;
+  private readonly _minDefaultAnimationDuration: number = 600;
+  private readonly _maxDefaultAnimationDuration: number = 2500;
+  private readonly _minDistanceDefault: number = 0.1;
```
larsmoa, line 34: Make the defaults `static readonly` and use `CamelCase`

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,534 @@
...
+   */
+  private async changeTarget(event: MouseEvent) {
+    const { offsetX, offsetY } = event;
+
+    const { x, y } = this.convertPixelCoordinatesToNormalized(offsetX, offsetY);
+
+    const callbackData = await this._modelRaycastCallback(offsetX, offsetY);
+
+    const newTarget =
+      callbackData?.intersection?.point ?? this.calculateMissedRaycast({ x, y }, callbackData.modelsBoundingBox);
```
larsmoa, line 108:
```suggestion
    const newTarget =
      callbackData.intersection?.point ?? this.calculateMissedRaycast({ x, y }, callbackData.modelsBoundingBox);
```
`callbackData` is never `undefined` (?)

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,534 @@
...
+
+    this.setCameraTarget(newTarget, true);
+  }
+
+  /**
+   * Changes controls scroll target based on current cursor position.
+   * @param event MouseEvent that contains pointer location data.
+   */
+  private async changeScrollTarget(event: any) {
```
larsmoa, line 117: Avoid using `any` - use the actual type.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,534 @@
...
+
+      this.controls.enableKeyboardNavigation = false;
+      this.changeTarget(e);
+    };
+
+    const onWheel = async (e: any) => {
+      const timeDelta = wheelClock.getDelta();
+      timeAfterClick += clickClock.getDelta();
+
+      if (timeAfterClick > 3) newTargetUpdate = false;
```
larsmoa, line 166: What is "3"? Seconsd?

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,534 @@
...
+
+      const wantNewScrollTarget = startedScroll && !newTargetUpdate && e.deltaY < 0;
+
+      if (wantNewScrollTarget) {
+        startedScroll = false;
+
+        this.changeScrollTarget(e);
+      } else {
+        if (timeDelta > 0.1) startedScroll = true;
+      }
```
larsmoa, line 176:
```suggestion
      } else if (timeDelta > 0.1) {
        startedScroll = true;
      }
```

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,534 @@
...
+  setCameraTarget(target: THREE.Vector3, animated: boolean = false): void {
+    if (this.isDisposed) {
+      return;
+    }
+
+    const animationTime = animated ? this._animationDuration : 0;
+    this.moveCameraTargetTo(target, animationTime);
+  }
+
+  /** @private */
```
larsmoa, line 270:
```suggestion
```

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,534 @@
...
+  private readonly calculateDefaultDuration = (distanceToCamera: number) => {
```
larsmoa, line 271:
```suggestion
  private calculateDefaultDuration(distanceToCamera: number): number {
```

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,534 @@
...
+    }
+
+    if (duration === 0) {
+      this.controls.setState(this._camera.position, target);
+      return;
+    }
+
+    const { _camera, _raycaster } = this;
+
+    if (duration === undefined) duration = this.calculateDefaultDuration(target.distanceTo(_camera.position));
```
larsmoa, line 364:
```suggestion
    duration = duration ??this.calculateDefaultDuration(target.distanceTo(_camera.position));
```

viewer/packages/camera-manager/src/CameraManager.ts
larsmoa: Use this style for function definitions.

larsmoa (changes requested):

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Yes, but not sure about that mechanics and how it feels. It is preventing zoom to cursor functionality for 3 seconds after target change click was initiated. Do you think it's bad idea or we can keep it and add other parameter for it?

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Should be, yeah

viewer/packages/camera-manager/src/CameraManager.ts
larsmoa: Hmm, it sounds a bit odd and that it easily could lead to confusion.

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Oh, it is undefined if raycast misses the model actually, so should be with question mark.

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Removed, finally

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Using it in the viewer to access controls methods, so I would say that it is reasonable to have them public

Savokr: Finally done

viewer/packages/utilities/src/events/InputHandler.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export class InputHandler {
+  private readonly domElement: HTMLElement;
+
+  private readonly _events = {
+    click: new EventTrigger<PointerEventDelegate>(),
+    hover: new EventTrigger<PointerEventDelegate>()
+  };
+
+  private setupEventListeners() {
```
christjt, line 19: Private methods should be declared last.

christjt (changes requested):
Overall a really good change in the correct direction! Great to see a lot of the business logic being moved out of the `Cognite3DViewer`. Most comments are quite minor, and overall this looks good. Still missing the abstraction for the manager (should be accessed through an interface).

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,584 @@
...
+  /**
+   * Changes controls target based on current cursor position.
+   * @param event MouseEvent that contains pointer location data.
+   */
+  private async changeTarget(event: MouseEvent) {
+    const { offsetX, offsetY } = event;
+
+    const { x, y } = this.convertPixelCoordinatesToNormalized(offsetX, offsetY);
+
+    const callbackData = await this._modelRaycastCallback(offsetX, offsetY);
```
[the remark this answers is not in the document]
Savokr: renamed to modelRaycastData

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,584 @@
...
+    const { offsetX, offsetY } = event;
+    const { x, y } = this.convertPixelCoordinatesToNormalized(offsetX, offsetY);
+
+    const callbackData = await this._modelRaycastCallback(offsetX, offsetY);
+
+    const newScrollTarget =
+      callbackData?.intersection?.point ?? this.calculateMissedRaycast({ x, y }, callbackData.modelsBoundingBox);
+
+    this.controls.setScrollTarget(newScrollTarget);
+  }
```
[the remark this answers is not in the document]
Savokr: Created a new function called calculateNewTarget that replaces both of them. Only addition is that now camera target and controls 'scrollTarget' are set inside onClick and onWheel method.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,569 @@
...
+  setCameraControlsState(controlsState: ControlsState): void {
+    this._controls.setState(controlsState.position, controlsState.target);
+  }
+
+  setCameraControlsOptions(controlsOptions: CameraControlsOptions): void {
+    this._cameraControlsOptions = { ...CameraManager.DefaultCameraControlsOptions, ...controlsOptions };
+
+    this.teardownControls();
+    this.setupControls();
+  }
```
larsmoa, line 292: Try to keep get-functions next to set-functions so it's easier to find them

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,569 @@
...
+
+  private calculateDefaultDuration(distanceToCamera: number): number {
+    let duration = distanceToCamera * 125; // 125ms per unit distance
+    duration = Math.min(
+      Math.max(duration, CameraManager.DefaultMinAnimationDuration),
+      CameraManager.DefaultMaxAnimationDuration
+    );
+
+    return duration;
```
larsmoa, line 310: Move to bottom of the class

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,584 @@
...
+  outBuffer = outBuffer || [
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3(),
+    new THREE.Vector3()
+  ];
```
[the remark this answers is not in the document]
larsmoa: I'll up with `range(0, 8).map(_ => new THREE.Vector3())`

https://lodash.com/docs/2.4.2#range

viewer/packages/camera-manager/src/types.ts
```diff
@@ -0,0 +1,89 @@
...
+   */
+  onClickTargetChange?: boolean;
+};
+
+export type ControlsState = {
+  position: THREE.Vector3;
+  target: THREE.Vector3;
+};
+
+export interface RevealCameraControls extends THREE.EventDispatcher {
```
larsmoa, line 35: I don't think we need this interface (at least not yet).

I think we should place the abstraction layer at `CameraManager`, and have users implement their own manager instead of drop in controls. This gives more flexibility, but also will require us to add more tooling to enable users to implement this class.

I'm very open for discussion on this topic though...

larsmoa (changes requested):

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Ok, missed it for some reason

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,584 @@
...
+    return this.controls.getState();
+  }
+
+  getCameraControls(): ComboControls {
+    return this.controls;
+  }
+
+  setCameraControlsState(controlsState: ControlsState): void {
+    this.controls.setState(controlsState.position, controlsState.target);
+  }
```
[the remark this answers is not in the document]
Savokr: Changed most of them to setters/getters but I think that for state it looks better as a method.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,569 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import * as THREE from 'three';
+import TWEEN from '@tweenjs/tween.js';
+import { ComboControls } from './ComboControls';
+import { CallbackData, CameraControlsOptions, CameraChangeData, PointerEventDelegate, ControlsState } from './types';
+import { assertNever, EventTrigger, InputHandler } from '@reveal/utilities';
+import { RevealCameraControls } from '..';
```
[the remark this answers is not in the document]
Savokr: Yeah, it was auto-imported by VSCode, did't want to do that intentionally.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,584 @@
...
+        break;
+      case 'zoomToCursor':
+        this.controls.setScrollTarget(this.controls.getState().target);
+        this.controls.useScrollTarget = true;
+        this.controls.zoomToCursor = true;
+        break;
+
+      default:
+        assertNever(this._cameraControlsOptions.mouseWheelAction);
+    }
```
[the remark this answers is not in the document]
Savokr: Extracted it into handleMouseWheelAction function with slight modification.

viewer/packages/camera-manager/src/types.ts
```diff
@@ -0,0 +1,88 @@
...
+   */
+  setViewTarget: (target: THREE.Vector3) => void;
+
+  /**
+   * Sets scroll target (used for different scrolling mechanics) for controls.
+   */
+  setScrollTarget: (target: THREE.Vector3) => void;
+}
+
+export type CallbackData = {
```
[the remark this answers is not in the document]
Savokr: Renamed to `CameraManagerCallbackData`

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,569 @@
...
+
+  private isDisposed = false;
+
+  private readonly _modelRaycastCallback: (x: number, y: number) => Promise<CallbackData>;
+  private _onClick: ((event: MouseEvent) => void) | undefined = undefined;
+  private _onWheel: ((event: MouseEvent) => void) | undefined = undefined;
+
+  private static readonly DefaultAnimationDuration = 600;
+  private static readonly DefaultMinAnimationDuration = 600;
+  private static readonly DefaultMaxAnimationDuration = 2500;
```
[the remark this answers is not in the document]
Savokr: Reduced them by half

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,584 @@
...
+      })
+      .onComplete(() => {
+        if (this.isDisposed) {
+          return;
+        }
+        this._domElement.removeEventListener('pointerdown', stopTween);
+      })
+      .start(TWEEN.now());
+    tween.update(TWEEN.now());
+  }
```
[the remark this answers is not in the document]
Savokr: Added two function that are shared between `moveCameraTargetTo` and `moveCameraTo`.

viewer/packages/camera-manager/src/types.ts
```diff
@@ -0,0 +1,89 @@
...
+   *
+   */
+  mouseWheelAction?: 'zoomToTarget' | 'zoomPastCursor' | 'zoomToCursor';
+  /**
+   * Enables or disables change of camera target on mouse click. New target is then set to the point of the model under current cursor position.
+   *
+   * Default is false.
+   *
+   */
+  onClickTargetChange?: boolean;
```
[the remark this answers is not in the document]
Savokr: Changed

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,569 @@
...
+    raycastFunction: (x: number, y: number) => Promise<CallbackData>
+  ) {
+    this._camera = camera;
+    this._domElement = domElement;
+    this._inputHandler = new InputHandler(domElement);
+    this._modelRaycastCallback = raycastFunction;
+    this._controls = new ComboControls(camera, domElement);
+    this._controls.dollyFactor = 0.992;
+    this._controls.minDistance = 0.15;
+    this._controls.maxDistance = 100.0;
```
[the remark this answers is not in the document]
Savokr: Found out maxDistance is never used inside ComboControls 🙂, removed it and minDistance setting since it is always adjusted in runtime. Default dollyFactor from ComboControls is also feeling alright.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,569 @@
...
+    return { x, y };
+  }
+
+  /**
+   * Calculates new target when raycaster doesn't have any intersections with the model.
+   * @param cursorPosition Cursor position for desired calculations.
+   * @param cursorPosition.x
+   * @param cursorPosition.y
+   */
+  private calculateMissedRaycast(
```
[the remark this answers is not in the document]
Savokr: Renamed to `calculateNewTargetWithoutModel`

viewer/packages/camera-manager/src/types.ts
Savokr: I think I agree about that now. I'll create a possible interface for `CameraManager` and we'll see how it will go. Another thought that I had was creating a base CameraManager class with some "obvious" methods (like fitCameraToModel/BoundingBox) that users can override, so that modification could be easier if you only want to add/change a few methods.

viewer/core/src/public/migration/Cognite3DViewer.test.ts
```diff
@@ -224,14 +224,16 @@ describe('Cognite3DViewer', () => {
...
       mouseWheelAction: 'zoomToTarget'
     });
 
     // Assert
     const newCameraControlsOptions = viewer.getCameraControlsOptions();
 
-    expect(newCameraControlsOptions.onClickTargetChange).not.toEqual(originalCameraControlsOptions.onClickTargetChange);
+    expect(newCameraControlsOptions.changeCameraTargetOnClick).not.toEqual(
+      originalCameraControlsOptions.changeCameraTargetOnClick
+    );
```
larsmoa, line 236: This test doesn't look very useful, it basically just checks if a setter owrks. Not sure there's any good way to make it better though

viewer/packages/camera-manager/src/types.ts
larsmoa: Not sure what value this interface adds now - do we want to keep it? My suggestion is to keep the interface for now but dont change the return type in `Cognite3dViewer` and don't expose the interface.

viewer/packages/utilities/src/events/InputHandler.ts
```diff
@@ -0,0 +1,120 @@
...
+    const { domElement } = this;
+
+    const maxMoveDistance = 8;
+    const maxClickDuration = 250;
```
larsmoa, line 24: Move to class constants.

viewer/packages/utilities/src/events/InputHandler.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    let pointerDown = false;
+    let pointerDownTimestamp = 0;
+    let validClick = false;
+
+    const startOffset = new Vector2();
+
+    const onHoverCallback = debounce((e: MouseEvent) => {
+      this._events.hover.fire(clickOrTouchEventOffset(e, domElement));
```
larsmoa, line 33: Move this implementation and other event handler implementations below to member functions to make `setupEventListeners` smaller

viewer/packages/utilities/src/events/InputHandler.ts
larsmoa: Move private functions to the after all public functions

larsmoa (changes requested):

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,584 @@
...
+ * Copyright 2021 Cognite AS
+ */
+
+import * as THREE from 'three';
+import TWEEN from '@tweenjs/tween.js';
+import { ComboControls } from './ComboControls';
+import { CallbackData, CameraControlsOptions, CameraChangeData, PointerEventDelegate, ControlsState } from './types';
+import { assertNever, EventTrigger, InputHandler } from '@reveal/utilities';
+
+export class CameraManager {
```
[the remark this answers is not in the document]
Savokr: Not sure what to do with that. Can extract some methods with event listeners but it will require passing domElement as an argument, so not sure whether it will look better. As for creating some other classes, we could add Animations class but probably in other PR.

viewer/core/src/public/migration/Cognite3DViewer.test.ts
Savokr: Could be a check for added eventListeners maybe, can look into that in other PR.

viewer/packages/camera-manager/src/types.ts
Savokr: Changed to returning `ComboControls` everywhere

viewer/packages/utilities/src/events/InputHandler.ts
Savokr: Others require the context of this function so moved only parts of them but looks better I think. Not sure about number of arguments, though...

viewer/packages/camera-manager/src/CameraManager.ts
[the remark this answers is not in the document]
larsmoa: There's a few possible low hanging fruit her:
- NearFarPlaneHelper
- CameraAnimationHelper

Since this is moved from viewer its ok if we do this in a separate PR to me, but it would make things nicer.

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -41,14 +41,13 @@ function getPinchInfo(domElement: HTMLElement, touches: TouchList) {
...
-export default class ComboControls extends EventDispatcher {
+export class ComboControls extends EventDispatcher {
   public enabled: boolean = true;
   public enableDamping: boolean = true;
   public dampingFactor: number = 0.25;
   public dynamicTarget: boolean = true;
   public minDistance: number = 0.1;
-  public maxDistance: number = Infinity;
-  public dollyFactor: number = 0.98;
+  public dollyFactor: number = 0.99;
```
larsmoa, line 50: Why is maxDistance removed and dollyFactor introduced?

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,584 @@
...
+    duration = Math.min(
+      Math.max(duration, CameraManager.DefaultMinAnimationDuration),
+      CameraManager.DefaultMaxAnimationDuration
+    );
+
+    return duration;
+  }
+}
+
+function getBoundingBoxCorners(bbox: THREE.Box3, outBuffer?: THREE.Vector3[]): THREE.Vector3[] {
```
larsmoa, line 567: Can this be replaced with `visitBox3CornerPoints` which we have in the utilities-package?

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -77,7 +76,7 @@ export default class ComboControls extends EventDispatcher {
   public zoomToCursor = true;
   public minDeltaRatio = 1;
   public maxDeltaRatio = 8;
-  public minDeltaDownscaleCoefficient = 0.1;
+  public minDeltaDownscaleCoefficient = 0.05;
```
larsmoa, line 79: Why change?

I don't think behaviour should change from this PR.

larsmoa (changes requested):

viewer/packages/camera-manager/README.md
```diff
@@ -60,7 +60,6 @@ controls.dynamicTarget = false; // possible to zoom past the target (will move t
 controls.dollyFactor = 0.98; // zoom factor (when zooming one step the distance to the target will be distance = oldDistance * dollyFactor)
 
 controls.minDistance = 1; // minimum distance to the target (see also dynamicTarget)
-controls.maxDistance = Infinity; // maximum distance to the target
```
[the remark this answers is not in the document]
Savokr: This value was never used inside `ComboControls` class. 🙂   (It was like that long before my changes)

viewer/packages/camera-manager/src/ComboControls.ts
Savokr: Reason for removal of maxDistance is described above, dollyFactor default is taken from the value that was before set every time ComboControls were created in Cognite3DViewer. Also, this difference in dolly factor (0.01) affects controls in very tiny way, so doesn't really matter here. It is now just defaulted to the value that it was always used with before.

viewer/packages/camera-manager/src/ComboControls.ts
Savokr: Sure, will change to previous value. Was experimenting on this value for zooming from long distances. Will change it in a different PR, probably the one about controls sensitivity and small fixes for zooming speed.

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: I checked `getBox3CornerPoints` method and it is essentially doing the same thing (only the order of corners is different but it shouldn't matter in most (all?) cases) but every function call it is creating a new array. However, implementation in `getBoundingBoxCorners` work on present array if it is passed. Since call is happening every frame we need to stick to last implementation. We can try just to remove method from CameraManager and change implementation in utils if it is ok.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,584 @@
...
+    // See https://stackoverflow.com/questions/8101119/how-do-i-methodically-choose-the-near-clip-plane-distance-for-a-perspective-proj
+    if (this.isDisposed) {
+      return;
+    }
+    if (!this.automaticControlsSensitivity && !this.automaticNearFarPlane) {
+      return;
+    }
+
+    const { cameraPosition, cameraDirection, corners } = this._updateNearAndFarPlaneBuffers;
+    getBoundingBoxCorners(combinedBbox, corners);
```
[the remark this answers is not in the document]
Savokr: Passed all needed values as arguments. Also, storing these values as globals is required because of every frame update for of them.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -0,0 +1,590 @@
...
+    return { tween, stopTween };
+  }
+
+  private calculateCameraFar(
+    near: number,
+    cameraPosition: THREE.Vector3,
+    cameraDirection: THREE.Vector3,
+    corners: Array<THREE.Vector3>
+  ): number {
+    const { nearPlane, nearPlaneCoplanarPoint } = this._updateNearAndFarPlaneBuffers;
```
larsmoa, line 404: It's a bit sketchy to fetch vars from buffers from another "vars buffer". Move the stuff you need to a new buffer 
`_calculateCameraFarBuffers` or something so it's clear where the vars are used.

larsmoa (approved):
See comment

larsmoa (approved):

viewer/packages/camera-manager/src/CameraManager.ts
larsmoa: Maybe you at least can implement getBox3CornerPoints using visitBox3CornerPoints?

Savokr: Not sure what you mean here, there is no method called `visitBox3CornerPoints`, you meant `getBox3CornerPoints`? If yes, then you suggest using it for `getBoundingBoxCorners` or vice versa?

larsmoa: Not sure the exact naming but there was a function in the utilities package I thought could be used here to implement the function

larsmoa: Anyway, minor. It's OK as is.

christjt (approved):

PR #1681, opened 2021-11-04
fix: correct SSAO blur pass

haakonflatval-cognite:
So that it is an actual seperated kernel


haakonflatval-cognite:
castberg_new_blur

haakonflatval-cognite:
Old blur:
castberg_old_blur
New blur: 
castberg_new_blur

haakonflatval-cognite:
This just in: I compared separated blur passes and full blur. The test was running the blur algorithms 1000 times alternating between two textures / render targets. In addition, we called `readPixels` right before the first and before the last timestamp, reading from the last render target. Average times for these 1000 iterations over about 40 frames were:
- Two-pass separated Gauss filter: 103 ms
- Single-pass full blur Gauss filter: 48 ms
- Original (wrong) quasi-separated Gauss filter: 49 ms

haakonflatval-cognite:
The code necessary to reproduce the data given above has been pushed to the branch `hflatval/ssao-blur-benchmarking-backup`

viewer/packages/rendering/src/glsl/post-processing/ssao-blur-combine.frag
```diff
@@ -1,7 +1,5 @@
-// Copyright Cognite (C) 2019 Cognite
+// Copyright Cognite (C) 2021 Cognite
 //
-// Efficient Gaussian blur based on technique described by Daniel Rákos in
-// http://rastergrid.com/blog/2010/09/efficient-gaussian-blur-with-linear-sampling/
```
christjt, line 4: I think we should leave the reference, as it is still kinda based on this. And is also a good resource for understanding the shader.

christjt (approved):
Overall looks good to me.

PR #1682, opened 2021-11-05
feat: allow placing the loading indicator in any corner and control opacity

larsmoa:


viewer/core/src/public/migration/types.ts
```diff
@@ -62,6 +62,20 @@ export interface Cognite3DViewerOptions {
...
   renderTargetOptions?: { target: THREE.WebGLRenderTarget; autoSetSize?: boolean };
 
+  /**
+   * Style the loading indicator.
+   */
+  loadingIndicatorStyle?: {
+    /**
+     * What corner the spinner should be placed in. Defaults top topLeft.
+     */
+    placement: 'topLeft' | 'topRight' | 'bottomLeft' | 'bottomRight';
```
christjt, line 72: What is the benefit of this vs. an enum?

viewer/core/src/utilities/Spinner.ts
```diff
@@ -6,10 +6,18 @@ import css from './spinnerStyles.css';
 import svg from '!!raw-loader!./spinnerCogniteLogo.svg';
 import * as THREE from 'three';
 
+import { assertNever } from '@reveal/utilities';
+
+export type Corner = 'topLeft' | 'topRight' | 'bottomLeft' | 'bottomRight';
```
christjt, line 11: What is the benefit of this vs. an enum?

christjt (approved):
LGTM, a little question only.

viewer/core/src/utilities/Spinner.ts
larsmoa: It matches `Cognite3dViewerOptions` and how we've done it there - e.g. for MSAA and SSAO

viewer/core/src/utilities/Spinner.ts
```diff
@@ -38,12 +46,43 @@ export class Spinner {
...
     this.el.className = Spinner.classnames.base;
+    this.el.classList.add(Spinner.classnames.topLeft);
     this.el.innerHTML = svg;
 
     parent.style.position = 'relative';
     parent.appendChild(this.el);
   }
 
+  set placement(placement: Corner) {
+    this.el.classList.remove(
```
pramodcog, line 57: `this.el.classList.remove()` for any non existing class works fine but this might fire an error on IE 10. If we are not supporting IE 10, than this is not an issue :)

pramodcog (approved):

viewer/core/src/public/migration/types.ts
larsmoa: This matches the options in Cognite3DViewer, i.e. how we've done this for antialiasing and ssao

PR #1685, opened 2021-11-08
feat: replace gltf sector parser with custom made parser

christjt:
This was done in order to improve performance. Chrome performance profiling shows  about a 5x performance increase in parsing sectors. (this will of course vary depending on the sector and its contents).

~~NB: instanced triangle meshes are not included (work in progress, and will be a separate PR).~~


viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -2,224 +2,289 @@
...
   }
 
-  public parseSector(data: ArrayBuffer): TypedGeometry[] {
-    const document: Document = this._gltfReader.readBinary(data);
+  public parseSector(data: ArrayBuffer) {
+    const headers = this._glbMetadataParser.parseGlbMetadata(data);
 
-    const defaultScene = document.getRoot().getDefaultScene();
-    const typedGeometryBuffers: TypedGeometry[] = [];
+    const json = headers.json;
```
larsmoa, line 52:
```suggestion
```

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -2,224 +2,289 @@
...
   constructor() {
-    this._gltfReader = new NodeIO();
-    this._gltfReader.registerExtensions([MeshGPUInstancing]);
+    this._glbMetadataParser = new GlbMetadataParser();
   }
 
-  public parseSector(data: ArrayBuffer): TypedGeometry[] {
-    const document: Document = this._gltfReader.readBinary(data);
+  public parseSector(data: ArrayBuffer) {
+    const headers = this._glbMetadataParser.parseGlbMetadata(data);
```
larsmoa, line 50:
```suggestion
```

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -2,224 +2,289 @@
...
+    const defaultSceneNodeIds = json.scenes[json.scene].nodes;
+
+    defaultSceneNodeIds
+      .map(nodeId => json.nodes[nodeId])
+      .forEach(node => {
+        const processedNode = this.processNode(node, headers, data)!;
+
+        if (processedNode === undefined) {
+          return;
+        }
```
larsmoa, line 68:
```suggestion
        const processedNode = this.processNode(node, headers, data)!;
        if (processedNode === undefined) {
          return;
        }
```

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -2,224 +2,289 @@
...
+      glbHeaderData,
+      instancingExtension: instancingExtension,
+      meshId,
+      data
+    };
 
     switch (geometryType) {
       case RevealGeometryCollectionType.InstanceMesh:
-        instanceId = this.processInstancedMesh(mesh, instanced, bufferGeometry);
+        //   instanceId = this.processInstancedMesh(mesh, instanced, bufferGeometry);
```
larsmoa, line 97: I don't think it's a good idea to merge this PR before we have instance meshes and feature parity with as is (?)

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -2,224 +2,289 @@
...
-        accessor.getElementSize()
-      );
+    const json = glbHeaderData.json;
 
-      bufferGeometry.setAttribute(`a${attributeName}`, instancedBufferAttribute);
-    });
-  }
+    const uniqueBufferViews = [
+      ...new Set(Object.values(attributes).map(accessorId => json.accessors[accessorId].bufferView))
+    ];
```
larsmoa, line 209: Not 100% sure on whats going on here, but isn't it a bit excessive to create a Set if theres only one element?

viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -2,224 +2,289 @@
...
+
+      assert(elementSize !== undefined);
+
+      const interleavedBufferAttribute = new THREE.InterleavedBufferAttribute(
+        interleavedBuffer,
+        size,
+        (accessor.byteOffset ?? 0) / elementSize
+      );
+      const transformedAttributeName = transformAttributeName(attributeName);
+      bufferGeometry.setAttribute(transformedAttributeName, interleavedBufferAttribute);
```
larsmoa, line 260:
```suggestion
      const interleavedBuffer = typedArrayMap[accessor.componentType];
      const size = this.COLLECTION_TYPE_SIZES.get(accessor.type);
      assert(size !== undefined);
      const elementType = this.DATA_TYPE_BYTE_SIZES.get(accessor.componentType);
      assert(elementType !== undefined);
      const elementSize = elementType.BYTES_PER_ELEMENT;
      assert(elementSize !== undefined);

      const interleavedBufferAttribute = new THREE.InterleavedBufferAttribute(
        interleavedBuffer,
        size,
        (accessor.byteOffset ?? 0) / elementSize
      );
      const transformedAttributeName = transformAttributeName(attributeName);
      bufferGeometry.setAttribute(transformedAttributeName, interleavedBufferAttribute);
```
To me, it doesn't make the code more readable with all of these blank lines, but just a suggestion

viewer/packages/sector-parser/src/reveal-glb-parser/GlbMetadataParser.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+import assert from 'assert';
+import { GlbHeaderData, GltfJson } from '../types';
+
+export class GlbMetadataParser {
+  private _textDecoder: TextDecoder;
+
+  private readonly GLB_HEADER_BYTE_SIZE = 12;
+  private readonly CHUNK_HEADER_BYTE_SIZE = 8;
```
larsmoa, line 12: Make static

viewer/packages/sector-parser/src/reveal-glb-parser/GlbMetadataParser.ts
```diff
@@ -0,0 +1,75 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import assert from 'assert';
+import { GlbHeaderData, GltfJson } from '../types';
+
+export class GlbMetadataParser {
+  private _textDecoder: TextDecoder;
```
larsmoa, line 9:
```suggestion
  private readonly _textDecoder: TextDecoder;
```

viewer/packages/sector-parser/tests/GltfSectorParser.test.ts
```diff
@@ -5,7 +5,8 @@
 import fs from 'fs';
 
 import 'jest-extended';
-import GltfSectorParser, { RevealGeometryCollectionType } from '../src/GltfSectorParser';
+import { GltfSectorParser } from '../src/GltfSectorParser';
+import { RevealGeometryCollectionType } from '../src/types';
 
 describe('GltfSectorParser Tests', () => {
```
larsmoa, line 11:
```suggestion
describe('GltfSectorParser', () => {
```
I think we should keep the naming scheme we have or switch to a new for all tests and not use several. Also, I've been thinking of if we should do this instead:
```
describe(GltfSectorParser.name, () => {
```
Which would avoid names getting out of sync.

larsmoa (changes requested):
Looks good, see some comments. 

However, I'm skeptical to merging before we have instance meshes - why not wait? Essentially we'd break existing functionality if we brought this in (even if its not master)

viewer/packages/sector-parser/src/GltfSectorParser.ts
christjt: Yeah, wasn't 100% satisfied with this. What I am trying to do is get all unique values of a given list of numbers. Basically want Linq `Distinct` method here. Any suggestions?

haakonflatval-cognite: From what I can understand, `Object.values()` returns a list. It is funneled into a `Set` for deduplication

viewer/packages/sector-parser/src/GltfSectorParser.ts
christjt: I don't agree with this, and prefer the space when reading. Not obsessed with this though so can change if you are very opinionated about this.

viewer/packages/sector-parser/src/GltfSectorParser.ts
larsmoa: All you are trying to do is to ensure that all values are equal right? In this case you'd only need to iterate through them and ensure they are all equal to the first element.

larsmoa (approved):

PR #1688, opened 2021-11-09
feat: add Cognite3dModel.updateStyledNodeCollection() and styledNodeCollections

larsmoa:
Also disallow adding the same node collection twice.

### DIFF
--- a/viewer/core/src/public/migration/Cognite3DModel.ts
+++ b/viewer/core/src/public/migration/Cognite3DModel.ts
@@ -34,11 +34,6 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
     return this.cadNode.nodeTransformProvider;
   }
 
-  /** @internal */
-  get styledNodeCollections(): { nodes: NodeCollectionBase; appearance: NodeAppearance }[] {
-    return this._styledNodeCollections;
-  }
-
   /**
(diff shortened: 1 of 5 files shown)


viewer/core/src/public/migration/Cognite3DModel.ts
```diff
@@ -152,25 +154,50 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase {
...
+    }
+
+    this._styledNodeCollections[index].appearance = newAppearance;
+    this.cadNode.nodeAppearanceProvider.assignStyledNodeCollection(nodeCollection, newAppearance);
+  }
+
   /**
    * Removes styling for previously added styled collection, resetting the style to the default (or
    * the style imposed by other styled collections).
    * @param nodeCollection   Node collection previously added using {@link assignStyledNodeCollection}.
```
haakonflatval-cognite, line 192: Should we have a `@throws` clause in this comment as well?

haakonflatval-cognite (approved):
Looks good! Small suggestion only

viewer/core/src/public/migration/Cognite3DModel.ts
larsmoa: Fixed

PR #1700, opened 2021-11-16
fix: correct worldToNormalizedCoordinates to handle pixel ratio properly

haakonflatval-cognite:
Turns out the best way to handle pixel ratio is to not try too hard.

Addresses #1672


haakonflatval-cognite:
It seems like we can get away from the pixel ratio mess by not considering pixel ratio at all. This has been tested to work correctly when setting pixel ratio manually in Cognite3DViewer.

larsmoa:
Can we add some unit tests and maybe a visual test for this?

haakonflatval-cognite:
There are already a number of unit tests for `worldToNormalizedCoordinates`. But a visual test might be an idea. I'll look into it

larsmoa:
Also, add reference to issue and a more comprehensive title/description of the issue.

viewer/packages/tools/src/HtmlOverlay/HtmlOverlayTool.test.ts
```diff
@@ -78,15 +78,16 @@ describe(HtmlOverlayTool.name, () => {
...
     // Act
     helper.add(htmlElement, position);
     helper.forceUpdate();
 
     // Assert
     expect(htmlElement.style.visibility).toBe('visible');
-    expect(htmlElement.style.top).toBe(`${renderer.domElement.height / 2}px`);
-    expect(htmlElement.style.left).toBe(`${renderer.domElement.width / 2}px`);
+    expect(htmlElement.style.left).toBe(`${renderSize.x / 2}px`);
+    expect(htmlElement.style.top).toBe(`${renderSize.y / 2}px`);
```
larsmoa, line 90: Why change the test? Why is center of dom element not valid? I'm a bit scared that we change the verification here

larsmoa (approved):

PR #1702, opened 2021-11-17
fix: bug causing changes to custom objects not to cause matrix to be updated

larsmoa:


viewer/packages/rendering/src/rendering/EffectRenderManager.ts
```diff
@@ -610,8 +610,14 @@ export class EffectRenderManager {
   }
 
   private renderCustomObjects(scene: THREE.Scene, camera: THREE.PerspectiveCamera) {
-    this._renderer.setRenderTarget(this._customObjectRenderTarget);
-    this.renderStep('customobjects', scene, camera);
+    const originalAutoUpdate = scene.autoUpdate;
+    try {
+      scene.autoUpdate = true;
+      this._renderer.setRenderTarget(this._customObjectRenderTarget);
```
pramodcog, line 616: Is there a scenario where setting custom render target or `renderStep` throw an exception?

larsmoa: Yes, renderStep can throw @pramodcog

pramodcog (approved):
Looks good to me

PR #1705, opened 2021-11-19
fix: allow users to turn off metrics logging

### LINKED ISSUE #1699: logMetrics = false doesnt actually disable usage statistics in Cognite3dViewer
We are using the Cognite3dViwer an passing the argument logMetric = false to the viewer, however the code will still log usage statistics.  Documented code says: "Might be used to disable usage statistics." 
But logMetrics seems to be dropped when createRevealManagerOptions is called

Is there anyway to actually disable usage statistics today? If not can this be something we can expect in a next release?

haakonflatval-cognite:
MetricsLogger must be truly global (i.e. not just static) to avoid issue where tools and internals of Cognite3DViewer would access two different MetricsLogger instances.

Known limitation with current approach:
If a web page uses multiple Reveal viewers and _any_ of these has `logMetrics` set to `true`, then _all_ the viewers will log metrics.

Fixes #1699


larsmoa (approved):
Nice - add a reference to the issue that triggered this fix before merging.

PR #1707, opened 2021-11-20
feat: add support for ignoring clipping planes through the node styling API

larsmoa:


larsmoa:
This will also break with geometry filter where user would expect that this wasn't applied to geometry outside the clip box. 

Closing for now.

PR #1714, opened 2021-11-23
fix: bump potree-core to fix point cloud issue on Firefox

haakonflatval-cognite:
We encountered problems running anything point cloud related in Firefox. Turns out it was because of a custom GLSL-definition of the `round()` function in `potree-core`.

This has now been fixed in `potree-core` 1.5.1.


larsmoa (approved):

larsmoa (changes requested):
Fix better title/description of PR so it's clear _why_ it's being done, not just _what_. See tech talk hosted last week on title/description

larsmoa (approved):

PR #1720, opened 2021-11-25
improvement: transfer buffers from parser web worker to main thread

larsmoa:
This avoids copying buffers, saving some time when transfering them.

See https://github.com/GoogleChromeLabs/comlink#comlinktransfervalue-transferables-and-comlinkproxyvalue

Note! This PR doesn't really change anything as we will need to release a new parser-worker before it has effect. In order to test this, run it locally and follow procedures in parser-worker/README.md for testing local worker.

### DIFF
--- a/examples/yarn.lock
+++ b/examples/yarn.lock
@@ -1252,10 +1252,15 @@
     exec-sh "^0.3.2"
     minimist "^1.2.0"
 
-"@cognite/potree-core@1.5.0", "@cognite/potree-core@^1.5.0":
+"@cognite/potree-core@^1.5.0":
   version "1.5.0"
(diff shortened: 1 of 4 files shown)


haakonflatval-cognite (approved):
Looks good to me :+1: Have you made any measurements to see if this is actually faster?

larsmoa:
> Looks good to me 👍 Have you made any measurements to see if this is actually faster?

No, I suspect this will have slim to none actual effect that users actually will feel. The reason is that after the main thread gets the data, it needs to run through the buffers again and restructure data. This is the most costly part of the current parsing and doesn't improve by this fix.

However, it certainly doesn't hurt...

PR #1721, opened 2021-11-26
fix: ensure all doc examples in 2.x work properly

haakonflatval-cognite:
This is in order to fix recent issues seen with live code examples being broken in the documentation: https://cognitedata.atlassian.net/browse/REV-258

NB: There's no knowing if this will actually work :/

### DIFF
--- a/documentation/package.json
+++ b/documentation/package.json
@@ -33,10 +33,9 @@
   "dependencies": {
     "@cognite/reveal": "link:../viewer/dist",
     "@cognite/reveal-1.x": "npm:@cognite/reveal@^1.5.5",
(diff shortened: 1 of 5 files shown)


larsmoa (approved):
How many of the problems in docs do we think this fixes?

haakonflatval-cognite:
> How many of the problems in docs do we think this fixes?

2 now :stuck_out_tongue:

christjt (approved):

haakonflatval-cognite:
At this point, the errors related to THREE are seen to work properly in the local test environment. The `instanceof` errors needs a new release, as they require remote repository to be updated. The errors only appear when using 2.x, which uses the remote repo (they are not reproducible with the local repo, for some God forsaken reason).

PR #1722, opened 2021-11-29
fix: too little geometry was culled during load when a geometry filter was active

larsmoa:
In a regression introduced in #1513, geometry filter culling of geometry didn't correctly work, causing too much clipped geometry to be loaded (and processed/rendered), reducing performance and visual quality.

See below for a comparison of the same view point (the blue box is the active geometry filter, but we disable clipping filters to also show loaded geometry outside this box).

Before fix:
image
2420 draw calls

After fix:
image
670 draw calls

Fixes REV-259.

### DIFF
--- a/viewer/packages/cad-geometry-loaders/src/sector/sectorUtilities.ts
+++ b/viewer/packages/cad-geometry-loaders/src/sector/sectorUtilities.ts
@@ -49,7 +49,7 @@ export function consumeSectorDetailed(
 ): { sectorMeshes: AutoDisposeGroup; instancedMeshes: InstancedMeshFile[] } {
   const bounds = metadata.bounds;
 
-  if (geometryClipBox !== null && geometryClipBox.containsBox(bounds)) {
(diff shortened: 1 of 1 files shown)


viewer/packages/cad-geometry-loaders/src/sector/sectorUtilities.ts
```diff
@@ -129,3 +129,23 @@ export function findSectorMetadata(root: SectorMetadata, sectorId: number): Sect
   }
   return foundSector;
 }
+
+/**
+ * Like THREE.Box3.containsBox(), but with fuzziness added.
```
christjt, line 134: Can you explain a little more here what is going on?

christjt (changes requested):

viewer/packages/cad-geometry-loaders/src/sector/sectorUtilities.ts
larsmoa: Added explanation and improved function name

christjt (approved):

PR #1723, opened 2021-12-01
release: @cognite/reveal-parser-worker@1.3.0

larsmoa:
New parser worker that uses "transfer" to move binary data between parser and main process.


haakonflatval-cognite (approved):
Sick feat, yo

PR #1732, opened 2021-12-01
fix: proper handling of fast 'clicks' and movement of the mouse at the same time

Savokr:
Previous code for handling such situations was just totally wrong 🙂. Also, fixed small issue with bounding box usage when camera target was changed.

### DIFF
--- a/documentation/yarn.lock
+++ b/documentation/yarn.lock
@@ -1224,9 +1224,9 @@
     skmeans "^0.11.3"
 
 "@cognite/reveal-2.x@npm:@cognite/reveal@2.x.x":
-  version "2.2.1"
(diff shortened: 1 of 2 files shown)


larsmoa (approved):
Looks good, but did some testing and found the max move distance to be too low, can you increase this a bit? Not sure what a good number is, but check google

Savokr:
> Looks good, but did some testing and found the max move distance to be too low, can you increase this a bit? Not sure what a good number is, but check google

Couldn't find anything meaningful in google. Increased maxMoveDistance to 8, should be fine now, I think. All clicks that I would think as clicks are handled properly.

PR #1733, opened 2021-12-01
feat: adds support for minDiagonalLength sector metadata

larsmoa:
Add support for new metadata field "minDiagonalLength". 

The plan was to use this to prioritise low detail sectors, but this turns out not to work well yet. Adding the field and will consider how to use this going forward.


haakonflatval-cognite (approved):
Alright, so not in use right now, but probably okay to include it since it is present in the optimizer output

PR #1738, opened 2021-12-02
feat: set proper update range for partial geometry updates

christjt:
Before:
image

After:
image

### DIFF
--- a/viewer/packages/rendering/src/GeometryBatchingManager.ts
+++ b/viewer/packages/rendering/src/GeometryBatchingManager.ts
@@ -61,13 +61,9 @@ export class GeometryBatchingManager {
       }
 
       const { defragBuffer, mesh } = geometry;
-      defragBuffer.remove(batchId);
+      const { updateRange } = defragBuffer.remove(batchId);
 
-      const instanceAttributes = this.getAttributes(mesh.geometry, THREE.InterleavedBufferAttribute);
-
-      instanceAttributes.forEach(namedAttribute => {
-        namedAttribute.attribute.needsUpdate = true;
-      });
+      this.updateInstanceAttributes(mesh, updateRange);
 
       mesh.count -= instanceCount;
     });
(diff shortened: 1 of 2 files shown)


viewer/packages/rendering/src/GeometryBatchingManager.ts
```diff
@@ -107,18 +103,51 @@ export class GeometryBatchingManager {
...
+    if (oldCount === -1) {
+      attribute.data.updateRange = { offset: newOffset, count: newCount };
+      attribute.data.needsUpdate = true;
+      return;
+    }
+
+    const offset = Math.min(oldOffset, newOffset);
+    const count = Math.max(oldOffset + oldCount, newOffset + newCount) - offset;
+
+    attribute.data.updateRange = { offset, count };
```
haakonflatval-cognite, line 148: The math looks correct, but I'm just wondering: How does `updateRange` work for `InterleavedBufferAttribute`? Does it update the portion of the attribute you specified, plus everything in-between that resides in the same buffer, which may belong to other attributes?

haakonflatval-cognite (approved):
Looks good, just a small question

viewer/packages/rendering/src/GeometryBatchingManager.ts
```diff
@@ -107,18 +103,51 @@ export class GeometryBatchingManager {
...
+  ) {
     this.getAttributes(mesh.geometry, THREE.InterleavedBufferAttribute).forEach(namedAttribute => {
-      namedAttribute.attribute.data.needsUpdate = true;
+      const attribute = namedAttribute.attribute;
+      this.setUpdateRange(attribute, updateRange);
+      attribute.data.needsUpdate = true;
     });
   }
 
+  private setUpdateRange(
```
larsmoa, line 126: I think this should be renamed to `extendUpdateRange`

larsmoa (changes requested):

larsmoa (approved):

PR #1739, opened 2021-12-07
feat: V9 geometry filter and primitive attribute read/write-utilities

haakonflatval-cognite:
The old geometry filter does not work for the new CAD format.

This implements a new geometry filterer for tackling V9.


viewer/packages/cad-parsers/src/cad/V9GeometryFilterer.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  const filteredGeometry = V9GeometryFilterer.filterGeometry(bufferGeometry, collectionType, clipBox);
+
+  const newPrimitives = parseInterleavedGeometry(primitiveType, filteredGeometry);
+
+  expect(newPrimitives).toHaveLength(1);
+  expect(newPrimitives[0]).toContainEntries(Object.entries(primitives[0]));
+}
+
+describe('V9GeometryFilterer accepts empty clipBox', () => {
```
larsmoa, line 36: We use `describe` as the 'category', and `test` to describe what's being tested. I think this test should group all tests together in one category.

viewer/packages/cad-parsers/src/cad/V9GeometryFilterer.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  Box,
+  getCollectionType
+} from '../../../../test-utilities/src/primitives/primitiveTypes';
+
+import {
+  createPrimitiveInterleavedGeometry,
+  parseInterleavedGeometry
+} from '../../../../test-utilities/src/primitives/primitiveThreeBuffers';
+
+function assertSecondCulledAwayFromBoxAtX10Yneg10Z0(primitives: any[], primitiveType: PrimitiveType) {
```
larsmoa, line 21: This function doesn't only assert - it also performs the test. Either rename the function or split into a function that performs the filtering and one that does the assert.

viewer/packages/cad-parsers/src/cad/V9GeometryFilterer.ts
```diff
@@ -0,0 +1,163 @@
...
+ * Copyright 2021 Cognite AS
+ */
+
+import { assertNever, TypedArray, TypedArrayConstructor } from '@reveal/utilities';
+import { RevealGeometryCollectionType } from '@reveal/sector-parser';
+import { computeBoundingBoxFromEllipseValues } from '../utilities/computeBoundingBoxFromAttributes';
+import { filterPrimitivesOutsideClipBox } from './filterPrimitivesCommon';
+import * as THREE from 'three';
+
+export class V9GeometryFilterer {
```
larsmoa, line 11: Not sure it makes sense to have this as a class since it only has static functions.

viewer/packages/cad-parsers/src/cad/V9GeometryFilterer.ts
```diff
@@ -0,0 +1,163 @@
...
+    filterType: new (...args: any[]) => T
+  ): Map<string, T> {
+    return new Map(
+      Object.entries(geometry.attributes)
+        .filter(namedAttribute => namedAttribute[1] instanceof filterType)
+        .map(nameAttributePair => [nameAttributePair[0], nameAttributePair[1] as T])
+    );
+  }
+
+  static filterGeometry(
```
larsmoa, line 28: It should be clearer that this filters geometry outside the provided clip box.

viewer/packages/cad-parsers/src/cad/V9GeometryFilterer.ts
```diff
@@ -0,0 +1,163 @@
...
+      const r2 = interleavedAttributeMap.get('_verticalRadius')!.getX(index);
+      const height = interleavedAttributeMap.get('_height')!.getX(index);
+
+      const centerAttribute = interleavedAttributeMap.get('_center')!;
+      center.set(centerAttribute.getX(index), centerAttribute.getY(index), centerAttribute.getZ(index));
+
+      return computeBoundingBoxFromEllipseValues(r1, r2, height, center, out);
+    };
+
+    return this.filterOnValues(interleavedAttributeMap, computeCallback, clipBox);
```
larsmoa, line 137: There is no `this` in static context so god knows how/if this works.

larsmoa (commented):
I feel there's a lot of code just to be able to test things here - could it be simplified somehow? You probably test things very thoroughly with the current approach, but I fear that the tests soon become very hard to maintain. The test utilities could probably be used for a lot of other tests though, so it's not black and white. 

Aim for as much shared code between the V8 filtering and V9. Is there a way we could share more code by abstracting parts of the V8 filtering? Not sure - just something we should check out.

haakonflatval-cognite:
This does not handle instance meshes nor triangle meshes sophisticatedly at all. Each sector seems to have only one triangle mesh, which is just a collection of all triangle meshes in the sector merged together. This is either included in its entirety, or discarded (if it has a specified bounding box and does not intersect the clip box entirely. Instance meshes are treated likewise.

viewer/packages/cad-parsers/src/cad/filterPrimitivesCommon.ts
```diff
@@ -0,0 +1,35 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
+import * as THREE from 'three';
+
+export function filterPrimitivesOutsideClipBox(
+  attributesByteValues: Uint8Array,
```
larsmoa, line 8: Maybe just pass in `ArrayBuffer` instead? And create `Uint8Array` and `Float32Array` views within the function as required?

viewer/packages/cad-parsers/src/cad/filterPrimitivesV9.test.ts
```diff
@@ -0,0 +1,365 @@
...
+  Nut,
+  getCollectionType
+} from '../../../../test-utilities/src/primitives/primitiveTypes';
+
+import {
+  createPrimitiveInterleavedGeometry,
+  parseInterleavedGeometry
+} from '../../../../test-utilities/src/primitives/primitiveThreeBuffers';
+
+function testSecondFilteredAwayFromBoxAtX10Yneg10Z0(primitives: any[], primitiveType: PrimitiveType) {
```
larsmoa, line 30: Can `primitives` be changed to something more exact?

Thinking of something in the lines of
```

type Primitive = Box | Circle | Nut | ...;
function testSecondFilteredAwayFromBoxAtX10Yneg10Z0(primitives: Primitive[], primitiveType: PrimitiveType) {
```

I know it really doesn't matter, but it makes the code 2% more readble

viewer/packages/cad-parsers/src/cad/filterPrimitivesV9.test.ts
```diff
@@ -0,0 +1,365 @@
...
+      for (let i = 0; i < array0.length; i++) {
+        expect(array0[i]).toBeCloseTo(array1[i]);
+      }
+    } else {
+      expect(primitives[0][key]).toBeCloseTo(newPrimitives[0][key] as number);
+    }
+  }
+}
+
+describe('filterPrimitivesV9 filters primitives correctly', () => {
```
larsmoa, line 62:
```suggestion
describe(filterPrimitivesV9.name, () => {
```

A neat trick I recently discovered to ensure tests are not stuck with stale names

larsmoa: BTW, test gets 10/10 ⭐

viewer/packages/cad-parsers/src/cad/filterPrimitivesV9.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  return createNewBufferGeometry(newArray, geometryBuffer, interleavedAttributes);
+}
+
+const _views = new Map<number, TypedArrayConstructor>([
+  [1, Uint8Array],
+  [4, Float32Array]
+]);
+
+const quadBoundingBox = new THREE.Box3(new THREE.Vector3(-0.5, -0.5, -1e-4), new THREE.Vector3(0.5, 0.5, -1e-4));
```
larsmoa, line 84: What is the epsilon? Also, should the max be positive z?
```suggestion
const quadBoundingBox = new THREE.Box3(new THREE.Vector3(-0.5, -0.5, -1e-4), new THREE.Vector3(0.5, 0.5, 1e-4));
```

viewer/packages/cad-parsers/src/cad/primitiveGeometries.ts
```diff
@@ -121,6 +121,7 @@ export const { torusLodGeometries, torusGeometryBoundingBox } = (() => {
 
   return {
     torusLodGeometries,
+    // TODO, hflatval: I believe this is wrong, torusLodGeometries are just planes near the origin
```
larsmoa, line 124:
```suggestion
    // TODO 2021-12-10 hflatval: I believe this is wrong, torusLodGeometries are just planes near the origin
```
I love TODOs with dates (if they get one year old they will never be fixed so then we can consider to remove them)

larsmoa: This is a plain bug - so more than a TODO. We should either fix or remove support for filtering of torus for now.

viewer/test-utilities/src/primitives/primitiveAttributes.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+}
+
+export function getShouldNormalize(type: AttributeComponentType): boolean {
+  switch (type) {
+    case 'byte':
+      return true;
+    default:
+      return false;
+  }
```
larsmoa, line 29:
```suggestion
  return type === 'byte';
```

viewer/test-utilities/src/primitives/primitiveAttributes.ts
```diff
@@ -0,0 +1,172 @@
+/*!
+ * Copyright 2021 Cognite AS
+ */
+
```
larsmoa, line 4: The functions in this file deserves some tsdoc

viewer/test-utilities/src/primitives/primitiveThreeBuffers.ts
```diff
@@ -0,0 +1,102 @@
...
+  getShouldNormalize
+} from './primitiveAttributes';
+import { writePrimitiveToBuffer } from './primitiveWrite';
+import { readPrimitiveFromBuffer } from './primitiveRead';
+import { PrimitiveType } from './primitiveTypes';
+
+import * as THREE from 'three';
+import { TypedArray } from '../../../packages/utilities';
+
+export function createPrimitiveInterleavedGeometry(name: PrimitiveType, primitiveDescs: any[]): THREE.BufferGeometry {
```
larsmoa, line 18: Replace `any[]` with a union type (if possible?), e.g. `type Primitive = Box | Circle | ...`

viewer/test-utilities/src/primitives/primitiveAttributes.ts
larsmoa: Clearly mark the functions that is meant for outside consumption. Maybe create a `primitives/index.ts` that exports the relevant functions? A lot of exported functions here and not easy to know which functions are relevant for use in the tests.

larsmoa (changes requested):
Generally looks good - see comments.

One important improvement I think needs to be made is to structure the test utilities you've created to make it easier to use from the outside. Currently there's a ton of free functions and it's not easy to know what files are relevant. Consider creating a separate `primitives/internal`-folder or something to hide away implementation details and add a `primitives/index.ts` with relevant exports.

larsmoa:
And improve PR title

viewer/packages/cad-parsers/src/cad/filterPrimitivesV9.ts
haakonflatval-cognite: The epsilon is just .. An epsilon. I could put it in a variable called `epsilon`. Would that make it better? :thinking:

viewer/packages/cad-parsers/src/cad/filterPrimitivesV9.test.ts
haakonflatval-cognite: `filterPrimitivesV9` is not a class, so I suppose this can't be done?

viewer/packages/cad-parsers/src/cad/filterPrimitivesV9.ts
haakonflatval-cognite: Edit - created a constant `epsilon`. Don't know what else to do

viewer/packages/cad-parsers/src/cad/filterPrimitivesV9.test.ts
larsmoa: Yes, that will work for functions too

viewer/packages/cad-parsers/src/cad/filterPrimitivesV9.ts
larsmoa: I agree it's not a big change - the change from two negative Z-values to one positive matters a lot though :)

larsmoa: What I wondered about was why there is a need to have an epsilon here.

larsmoa (approved):

PR #1741, opened 2021-12-09
fix: clipping planes are not applied to all geometry types for v9 models

larsmoa:


viewer/packages/rendering/src/CadMaterialManager.ts
```diff
@@ -262,3 +269,25 @@ function applyToModelMaterials(materials: Materials, callback: (material: THREE.
   callback(materials.triangleMesh);
   callback(materials.simple);
 }
+
+function isRawShaderMaterial(material: THREE.Material): material is THREE.RawShaderMaterial {
```
haakonflatval-cognite, line 273: Is it recent that we started using `RawShaderMaterial`?

haakonflatval-cognite (approved):
Looks good :+1:

viewer/packages/rendering/src/CadMaterialManager.ts
larsmoa: Yes, we just introduced this.

PR #1742, opened 2021-12-10
Merge master to release/3.0-alpha

larsmoa:


documentation/package.json
```diff
@@ -50,7 +50,7 @@
     "react-dom": "^16.8.4",
     "react-live": "^2.3.0",
     "styled-components": "^5.1.1",
-    "three": "0.133.0"
+    "three": "0.125.2"
```
Savokr, line 53: Why we are targeting three r125 here but not the latest version?

Savokr (approved):
Everything is good, just wondering about threejs version.

documentation/package.json
larsmoa: Thats a mistake

PR #1744, opened 2021-12-13
Disable torus segment filtering

haakonflatval-cognite:
Due to error in bounding box computation

### DIFF
--- a/viewer/packages/cad-parsers/index.ts
+++ b/viewer/packages/cad-parsers/index.ts
@@ -38,7 +38,7 @@ export {
   torusLodGeometries,
   boxGeometryBoundingBox,
   quadGeometryBoundingBox,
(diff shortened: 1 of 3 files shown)


viewer/packages/cad-parsers/index.ts
```diff
@@ -38,7 +38,7 @@ export {
   torusLodGeometries,
   boxGeometryBoundingBox,
   quadGeometryBoundingBox,
-  torusGeometryBoundingBox,
+  /* torusGeometryBoundingBox, */
```
larsmoa, line 41: All of these should be clearly commented with _why_ you've commented out the code

larsmoa (changes requested):

larsmoa (approved):

PR #1745, opened 2021-12-14
refac: v9 outputs will now be identified using the reveal-directory format like v8 outputs

larsmoa:
Optimizer will start outputting GLTF as 'reveal-directory' with version 9. This PR adds support for this, but ensures backward compatibiliy.


haakonflatval-cognite (approved):
Seems good :+1:

PR #1746, opened 2021-12-14
fix: html overlay clustering not always clustering correctly

haakonflatval-cognite:
Ref. conversations with Aize.

Sometimes cluster elements that result from HTML overlay clustering do not properly merge with other elements that may overlap them. This tries to fix it.

The fundamental reason is that when a cluster element is created from a bunch of overlapping HTML elements, the HTML elements are removed from the grid structure that detects collisions, but the clustered element is not added. Thus, if a new cluster element is created, which overlaps with a previous cluster element, it will not be detected. Same goes for a cluster element that overlaps other HTML elements which were not included in the collision triggering the creation of the cluster element.


haakonflatval-cognite:
I've laid out about the state of this PR here: https://cognitedata.atlassian.net/jira/software/c/projects/REV/boards/368?selectedIssue=REV-279 Work is discontinued for now.

PR #1747, opened 2021-12-16
fix: camera controls `zoomToCursor` mode was behaving wrong when zoomed on new target very slowly 

Savokr:
Changes in this PR:

- Fixed 'zoomToCursor' mode behaving incorrectly when changing to new `scrollTarget` 
- Fixed zooming speed problems in all controls modes for mouse wheels without acceleration 
- Fixed issues when zooming very closely in 'zoomToCursor' mode


viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -711,20 +709,25 @@ export default class ComboControls extends EventDispatcher {
...
     if (radius < minDistance) {
       this._temporarilyDisableDamping = true;
 
       // stops camera from moving forward only if target became close to scroll target
-      if (_scrollTarget.distanceTo(_target) < minDistance) {
-        radius = minDistance;
+      if ((_scrollTarget.distanceTo(_target) < minDistance) || (radius <= 0)) {
         deltaTargetOffsetDistance = 0;
+        
+        radius = distToTarget;
```
larsmoa, line 723: Shouldn't radius be `radius = Math.max(distToTarget, minDistance);` ?

Savokr: In that case it would lead to the same problem of slight movement in cases when it equals minDistance. Because when you change the radius to some new value you also need to offset the target accordingly (for correct zooming movement). That part of calculating proper target offset was the initial problem I was struggling with.

viewer/core/cad.ts
```diff
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2021 Cognite AS
```
haakonflatval-cognite, line 2: I don't think these should be part of this PR - I asked Lars about this last week: https://cognitedata.slack.com/archives/G0115JQ8JM8/p1641199767434200

haakonflatval-cognite (changes requested):

viewer/core/cad.ts
Savokr: Oh, ok. I think linter did it. Because I haven't touched dates in all these files. Should I reconfigure it or leave as is?

larsmoa: This should be fixed :/ I updated the linter to avoid this, I think you simply can rollback commit d827b9100bf6640338d793e932d9285b489bd629

Savokr: Yeah, fixed this one

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -526,25 +529,25 @@ export class CameraManager {
     const onWheel = async (e: any) => {
       const timeDelta = wheelClock.getDelta();
 
+      if (timeDelta > 0.08) scrollStarted = false;
```
haakonflatval-cognite, line 532: Magic number :eyes: What does it mean?

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -47,6 +47,7 @@ export class ComboControls extends EventDispatcher {
   public dampingFactor: number = 0.25;
   public dynamicTarget: boolean = true;
   public minDistance: number = 0.1;
+  public minZoomDistance: number = 0.1;
```
haakonflatval-cognite, line 50: Could this be set directly to the default value that is given in CameraManager, or is there a reason we would want them to be defined to separate values? Goes for all variables in this block

haakonflatval-cognite (changes requested):
Looks mostly good, just a couple of comments I'd like to have addressed :+1:

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: It's a magic number, yeah. Will extract it in a variable. Basically it is the smallest possible time between 2 distinct model raycasts, i.e. scrollTarget updates

viewer/packages/camera-manager/src/ComboControls.ts
Savokr: Yeah, can be the same as default inside CameraManager

haakonflatval-cognite (approved):
LGTM :+1:

PR #1750, opened 2021-12-30
feat: Upgrade shader to 3.0ES for webgl2

pramodcog:


documentation/package.json
```diff
@@ -50,7 +50,7 @@
     "react-dom": "^16.8.4",
     "react-live": "^2.3.0",
     "styled-components": "^5.1.1",
-    "three": "0.125.2"
+    "three": "0.135.0"
```
haakonflatval-cognite, line 53: If this works, this is a good idea - but do we know whether it works? The reason it is stuck on 125, is that the Reveal 1.x documentation depends on it. Have we tested whether they still work with this change?

haakonflatval-cognite:
Also - the changes to yarn.lock seem a bit excessive (13,000 lines added and removed). Could it be reduced? I think it should work by resetting it to what it is in `release/3.0-alpha` and then running `yarn` again.

pramodcog:
Working on new branch and cherry-pick only these changes, but it would be good if what ever is reviewable through the naked eye as there are huge changes

PR #1755, opened 2022-01-03
chore: update yarn.lock in examples

larsmoa:


haakonflatval-cognite:
When I checkout this branch and run `yarn` in the `examples` folder, it re-inserts all the dependencies that were removed in this PR. That shouldn't happen, right?

PR #1756, opened 2022-01-03
improvement: assignStyledNodeCollection now also handles update

larsmoa:
Removes the need for a seperate updateStyledNodeCollection() function

### DIFF
--- a/viewer/core/src/public/migration/Cognite3DViewer.ts
+++ b/viewer/core/src/public/migration/Cognite3DViewer.ts
@@ -497,7 +497,7 @@ export class Cognite3DViewer {
 
     this.isDisposed = true;
 
-    if (this.latestRequestId !== undefined) {
+    if (this.latestRequestId !== -1) {
       cancelAnimationFrame(this.latestRequestId);
     }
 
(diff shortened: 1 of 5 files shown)


viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1572,34 +1572,38 @@ export class Cognite3DViewer {
...
+          renderController.clearNeedsRedraw();
+          this.revealManager.resetRedraw();
+          this._clippingNeedsUpdate = false;
+          const renderTime = Date.now() - start;
+
+          this._events.sceneRendered.fire({ frameNumber, renderTime, renderer: this.renderer, camera: this.camera });
+        }
       }
+    } finally {
+      this.latestRequestId = requestAnimationFrame(this._boundAnimate);
```
haakonflatval-cognite, line 1606: What is the significance of moving this here? What is the purpose of `this.latestRequestId`?

haakonflatval-cognite (changes requested):
Looks good, just a comment I'd like answers for

viewer/core/src/public/migration/Cognite3DViewer.ts
larsmoa: Oh, that should not be part of this PR - fixing.

haakonflatval-cognite (approved):
Looks alright now :+1:

PR #1757, opened 2022-01-03
fix: error message listed same outputs twice

larsmoa:

PR #1759, opened 2022-01-04
fix: comments I forgot to commit from #1757

larsmoa:


haakonflatval-cognite (approved):
Ok, nice that they're coming around eventually :sweat_smile:

PR #1760, opened 2022-01-04
fix: msaa hides geometry

haakonflatval-cognite:
In a recent PR to `three.js` (https://github.com/mrdoob/three.js/pull/22558/files), multisampled depth would be ignored as part of some optimization


haakonflatval-cognite:
Am planning to add a visual test shortly

larsmoa (approved):

haakonflatval-cognite:
There was an attempt at adding a visual test, but no luck. It is considered a separate issue here: https://cognitedata.atlassian.net/browse/REV-291

PR #1761, opened 2022-01-04
improvement: simplify CAD model budget by only controlling "renderCost"

larsmoa:
\- Remove draw calls and max download size as it doesn't work well in V9
\- Simplified to only work on a single number which makes things more clear to users
\- Makes it easier to automatically adjust this in the future
\- Adjusted the default budget

### DIFF
--- a/viewer/packages/cad-geometry-loaders/src/sector/culling/takensectors/TakenV8SectorMap.ts
+++ b/viewer/packages/cad-geometry-loaders/src/sector/culling/takensectors/TakenV8SectorMap.ts
@@ -3,7 +3,7 @@
  */
 import { TakenV8SectorTree } from './TakenV8SectorTree';
 import { PrioritizedWantedSector, DetermineSectorCostDelegate, SectorCost, addSectorCost } from '../types';
-import { CadModelSectorBudget } from '../../../CadModelSectorBudget';
+import { CadModelBudget } from '../../../CadModelBudget';
 import { TakenSectorMapBase } from './TakenSectorMapBase';
 
 import { CadModelMetadata, V8SectorMetadata } from '@reveal/cad-parsers';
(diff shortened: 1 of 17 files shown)


viewer/test-utilities/src/createDetermineSectorInput.ts
```diff
@@ -18,9 +18,7 @@ export function createDetermineSectorInput(
     loadingHints: {},
     cameraInMotion: false,
     budget: budget || {
-      geometryDownloadSizeBytes: 20,
       highDetailProximityThreshold: 10,
-      maximumNumberOfDrawCalls: Infinity,
       maximumRenderCost: Infinity
```
haakonflatval-cognite, line 22: Should this one be less than infinite?

viewer/packages/cad-geometry-loaders/src/sector/culling/takensectors/TakenV8SectorMap.ts
```diff
@@ -66,13 +66,9 @@ export class TakenV8SectorMap extends TakenSectorMapBase {
...
 
-  isWithinBudget(budget: CadModelSectorBudget): boolean {
+  isWithinBudget(budget: CadModelBudget): boolean {
     const currentCost = this.totalCost;
-    return (
-      currentCost.downloadSize < budget.geometryDownloadSizeBytes &&
-      currentCost.drawCalls < budget.maximumNumberOfDrawCalls &&
-      currentCost.renderCost < budget.maximumRenderCost
-    );
+    return currentCost.renderCost < budget.maximumRenderCost;
```
haakonflatval-cognite, line 71: Just a general question - do all V8 sectors specify both estimated number of draw calls and render cost? Because this change will apply to both V8 and V9, right?

haakonflatval-cognite (approved):
Some comments, otherwise looks good :+1:

viewer/test-utilities/src/createDetermineSectorInput.ts
larsmoa: Well, it's just for testing - and behaviour is not changed from previous. Might be that it would have made sense with some other values, but I think its fine as is as I dont want to fix tests that rely on this.

viewer/packages/cad-geometry-loaders/src/sector/culling/takensectors/TakenV8SectorMap.ts
larsmoa: Yes, this is in place in v8 as well.

PR #1764, opened 2022-01-05
fix: Setting default node appearance to ghosted hides "default"-rendered geometry

haakonflatval-cognite:
The fix seems to be to properly add `mesh.userData.treeIndices`.

### DIFF
--- a/viewer/packages/rendering/src/GeometryBatchingManager.ts
+++ b/viewer/packages/rendering/src/GeometryBatchingManager.ts
@@ -242,8 +242,11 @@ export class GeometryBatchingManager {
     const typedArray = treeIndexAttribute.data.array as TypedArray;
     const instanceByteSize = treeIndexAttribute.data.stride * typedArray.BYTES_PER_ELEMENT;
 
-    const firstWholeInstanceIndex = Math.ceil(byteOffset / instanceByteSize);
-    const firstInvalidInstanceIndex = Math.floor((byteOffset + byteCount) / instanceByteSize);
+    assert(byteOffset % instanceByteSize === 0);
+    assert(byteCount % instanceByteSize === 0);
+
+    const firstWholeInstanceIndex = byteOffset / instanceByteSize;
+    const firstInvalidInstanceIndex = (byteOffset + byteCount) / instanceByteSize;
 
     for (let i = firstWholeInstanceIndex; i < firstInvalidInstanceIndex; i++) {
       const treeIndex = treeIndexAttribute.getX(i);
--- a/viewer/packages/rendering/src/GeometryBatchingManager.ts
+++ b/viewer/packages/rendering/src/GeometryBatchingManager.ts
@@ -61,7 +61,12 @@ export class GeometryBatchingManager {
       }
 
       const { defragBuffer, mesh } = geometry;
-      const { updateRange } = defragBuffer.remove(batchId);
+
+      const updateRange = defragBuffer.getRangeForBatchId(batchId);
+
+      this.removeTreeIndicesFromMeshUserData(mesh, updateRange);
+
+      defragBuffer.remove(batchId);
 
       this.updateInstanceAttributes(mesh, updateRange);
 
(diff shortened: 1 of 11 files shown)


haakonflatval-cognite:
Has currently implemented handling of `mesh.userData.treeIndices` in `GeometryBatchingManager`, which seems to fix the issue for primitives and instance meshes.
Will find fix for triangle meshes next.

viewer/packages/rendering/src/GeometryBatchingManager.ts
```diff
@@ -71,6 +76,12 @@ export class GeometryBatchingManager {
     this._sectorMap.delete(sectorId);
   }
 
+  private *getFloatsFromAttribute(attribute: THREE.InterleavedBufferAttribute): Iterable<number> {
+    for (let i = 0; i < attribute.count; i++) {
+      yield attribute.getX(i);
+    }
+  }
```
larsmoa, line 83: This will be quite slow, iterators/generators are slow. I suggest you just use a plain loop instead below.

See https://jsbench.me/eyky1pj2ih/1 for a comparison on performance

christjt: That sucks because the generator way of doing it is a lot more elegant imo. Didn't realize there was THAT much difference.

viewer/packages/rendering/src/GeometryBatchingManager.ts
```diff
@@ -201,6 +219,42 @@ export class GeometryBatchingManager {
...
+  ): void {
+    const { byteOffset, byteCount } = updateRange;
+
+    const interleavedAttributes = this.getAttributes(mesh.geometry, THREE.InterleavedBufferAttribute);
+
+    const treeIndexAttribute = this.getTreeIndexAttribute(interleavedAttributes);
+    const typedArray = treeIndexAttribute.data.array as TypedArray;
+    const instanceByteSize = treeIndexAttribute.data.stride * typedArray.BYTES_PER_ELEMENT;
+
+    const firstWholeInstanceIndex = Math.ceil(byteOffset / instanceByteSize);
```
christjt, line 249: Does it make sense to math.ceil here? I would think that if byteOffset is not divisible by instanceByteSize then the state of the buffer is invalid.

viewer/packages/rendering/src/GeometryBatchingManager.ts
```diff
@@ -201,6 +219,42 @@ export class GeometryBatchingManager {
...
+    const firstInvalidInstanceIndex = Math.floor((byteOffset + byteCount) / instanceByteSize);
```
christjt, line 250: Same here

christjt (changes requested):

viewer/packages/rendering/src/GeometryBatchingManager.ts
```diff
@@ -201,6 +219,42 @@ export class GeometryBatchingManager {
...
+  private getTreeIndexAttribute(
+    instanceAttributes: { name: string; attribute: THREE.InterleavedBufferAttribute }[]
+  ): THREE.InterleavedBufferAttribute {
+    let treeIndexAttribute: THREE.InterleavedBufferAttribute | undefined = undefined;
+    for (let i = 0; i < instanceAttributes.length; i++) {
+      if (instanceAttributes[i].name == 'a_treeIndex') {
+        treeIndexAttribute = instanceAttributes[i].attribute;
+      }
+    }
+
```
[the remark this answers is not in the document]
larsmoa: Wouldn't find (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) be better?

viewer/packages/rendering/src/GeometryBatchingManager.ts
larsmoa: I agree - code gets uglier, but it's just no way justifiable considering that this is in loading hotpath

viewer/packages/rendering/src/GeometryBatchingManager.ts
haakonflatval-cognite: I agree - putting in a couple of asserts instead.

christjt (approved):

PR #1765, opened 2022-01-06
feat: add continuous loading option for geometry streaming

christjt:


viewer/core/src/public/migration/types.ts
```diff
@@ -158,6 +158,12 @@ export interface Cognite3DViewerOptions {
    */
   customDataSource?: DataSource;
 
+  /**
+   * Allows for controlling if geometry streaming should be halted when
+   * the camera is moving.
+   */
+  continuousModelStreaming?: boolean;
```
larsmoa, line 165: I think we should add something about performance being impacted by this and it's not recommended to enable this for low end devices. Also, add information about this being default false.

larsmoa (changes requested):

larsmoa (approved):

PR #1767, opened 2022-01-07
fix: Added Temporary API keys missing in Docs for all map providers

pramodcog:
Added temporary API keys for all map providers for testing in Docs


christjt (approved):
LGTM

larsmoa (changes requested):
Store keys in customScope.ts as we do today to avoid people easily copy-pasting our API keys by accident.

larsmoa (approved):

larsmoa:
how should we ensure we catch expiry going forward @pramodcog ?

PR #1768, opened 2022-01-07
fix: composite overlays was not removed when calling clear() or dispose()

larsmoa:
Fixes an issue causing composite HTML elements from HtmlOverlayTool to not be removed from DOM when calling HtmlOverlayTool.dispose()


Savokr (approved):
Looks good to me

PR #1769, opened 2022-01-07
test: migrate visual tests to V9

haakonflatval-cognite:
Also add three tests that stay at V8


haakonflatval-cognite:
Almost all images had to be recreated because of a bug in the V8 model causing it to have a bounding box extended too much in one dimension, which positions the camera weirdly. Also, we suspect that the normals on some surfaces of some primitives are flipped in the V9 model, but this should be addressed in the shaders at a later point in time.

We might be able to keep the old visual test images by introducing the same bounding box bug in the V9 model, but it seems like the wrong way to go about this.

christjt (approved):
LGTM

PR #1770, opened 2022-01-11
fix: Externally added event listeners were not disposed of when disposing of Cognite3DViewer

Savokr:
When disposing of Cognite3DViewer externally added event listeners were not disposed of which caused trouble in dev-server hot reload.

### DIFF
--- a/viewer/core/src/public/migration/Cognite3DViewer.ts
+++ b/viewer/core/src/public/migration/Cognite3DViewer.ts
@@ -113,6 +113,7 @@ export class Cognite3DViewer {
     sceneRendered: new EventTrigger<SceneRenderedDelegate>(),
     disposed: new EventTrigger<DisposedDelegate>()
   };
+  private readonly _mouseHandler: InputHandler;
 
   private readonly _models: CogniteModelBase[] = [];
   private readonly _extraObjects: THREE.Object3D[] = [];
@@ -255,7 +256,8 @@ export class Cognite3DViewer {
       );
     }
     this.renderController = new RenderController(this.camera);
-
+    
+    this._mouseHandler = new InputHandler(this.domElement);
     this.startPointerEventListeners();
 
     this.revealManager.setRenderTarget(
@@ -350,6 +352,7 @@ export class Cognite3DViewer {
 
     this._events.disposed.fire();
     this.disposeOfAllEventListeners();
+    this._mouseHandler.dispose();
   }
 
   /**
@@ -1333,13 +1336,11 @@ export class Cognite3DViewer {
   }
 
   private readonly startPointerEventListeners = () => {
-    const mouseHandler = new InputHandler(this.domElement);
-
-    mouseHandler.on('click', e => {
+    this._mouseHandler.on('click', e => {
       this._events.click.fire(e);
     });
 
-    mouseHandler.on('hover', e => {
+    this._mouseHandler.on('hover', e => {
       this._events.hover.fire(e);
     });
   };
--- a/viewer/core/src/public/migration/Cognite3DViewer.ts
+++ b/viewer/core/src/public/migration/Cognite3DViewer.ts
@@ -11,7 +11,7 @@ import { LoadingState } from '@reveal/cad-geometry-loaders';
 
 import { defaultRenderOptions, SsaoParameters, SsaoSampleQuality, AntiAliasingMode } from '@reveal/rendering';
 
-import { assertNever, EventTrigger, InputHandler } from '@reveal/utilities';
+import { assertNever, EventTrigger, InputHandler, disposeOfAllEventListeners } from '@reveal/utilities';
 import { MetricsLogger } from '@reveal/metrics';
 
 import { worldToNormalizedViewportCoordinates, worldToViewportCoordinates } from '../../utilities/worldToViewport';
@@ -351,7 +351,7 @@ export class Cognite3DViewer {
     this.spinner.dispose();
 
     this._events.disposed.fire();
-    this.disposeOfAllEventListeners();
+    disposeOfAllEventListeners(this._events);
     this._mouseHandler.dispose();
   }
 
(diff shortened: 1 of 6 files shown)


viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1321,14 +1325,22 @@ export class Cognite3DViewer {
     return true;
   }
 
-  private readonly startPointerEventListeners = () => {
-    const mouseHandler = new InputHandler(this.domElement);
+  /**
+   * Method for deleting all external events that are associated with current instance of Cognite3DViewer.
+   */
+  private disposeOfAllEventListeners() {
+    const cognite3DViewerEvents = ['click', 'hover', 'cameraChange', 'sceneRendered', 'disposed'];
```
larsmoa, line 1332: Is there anything we can do to make this list dynamic based on the events we actually support? E.g. just loop over all member of `this._events`?

larsmoa: Or maybe we should introduce an `EventsHelper` class that can help us maintain this? I see that the code is duplicated to basically anywhere we have these listeners so might be worth creating an utility to avoid duplication.

larsmoa (changes requested):

viewer/core/src/public/migration/Cognite3DViewer.ts
Savokr: Yeah, we can just iterate through members of this._events, not sure why I decided not to do that 🥲. There are additional classes already present: EventTrigger and InputHandler. If we want to keep logic of `someClass.on/off('event', ...);` then we will have to extend each class to `EventsHelper`. If it's ok, I can create such class in another PR.

larsmoa: I was thinking to have a `EventsHelper`-class that contained all individual events (e.g. 'click', 'hover', etc), but I realize now that it doesn't make sense since that would work well with types. Might be that you can create a generic implementation of `disposeOfAllEventListeners()` though, and just call this from each of the classes where you now have implemented it?

Not sure what your proposal about "have to extend each class to EventsHelper", but interested in hearing details if you have an idea.

Savokr: Added common method for all classes

viewer/packages/utilities/src/events/InputHandler.ts
```diff
@@ -143,3 +146,12 @@ export class InputHandler {
     this._events.hover.fire(clickOrTouchEventOffset(e, this.domElement));
   }, 100);
 }
+
+/**
+ * Method for deleting all external events that are associated with current instance of a class.
+ */
+export function disposeOfAllEventListeners(eventList: any): void {
```
larsmoa, line 153: Add a proper type for `eventList`. I believe it would be something along the lines of 
```suggestion
export type EventCollection = { [eventName: string]: EventHandler<void> };
export function disposeOfAllEventListeners(eventList: EventCollection): void {
```

larsmoa (changes requested):
Good, but disposeOfAllEventListeners should be made type safe

viewer/packages/utilities/src/events/InputHandler.ts
Savokr: Seems like it won't work that simple, because eventList has different generic type in each event type field.
It could be `_events = {
    cameraChange: new EventTrigger<CameraChangeDelegate>(),
    click: new EventTrigger<PointerEventDelegate>(),
    hover: new EventTrigger<PointerEventDelegate>(),
    sceneRendered: new EventTrigger<SceneRenderedDelegate>(),
    disposed: new EventTrigger<DisposedDelegate>()
  }`. I don't think you can create a general type for such a thing, because list can also include or exclude some of these types of events.

larsmoa: What if you use `EventHandler<unknown>` instaed of `EventHandler<void>`?

Savokr: It says: `Type 'unknown' does not satisfy the constraint '(...args: any[]) => void'.` But with `any` it works fine:
`type EventCollection = { [eventName: string]: EventTrigger<any> };`.

larsmoa: I think we should try any then. Or maybe unknown => void works?

Savokr: No, not working like this but another option is just the general type from EventTrigger, i.e. `type EventCollection = { [eventName: string]: EventTrigger<(...args:any) => void> };`. This one also works fine.

larsmoa: The more specific the better, so I think the last suggestion you have looks good!

larsmoa: (might ber that `...args: unknown[]` works too - thats better than `...args:any[]` if it works)

Savokr: Works in type declaration but it is not automatically assigned to `_events` object, then every time it is required to write `_events as EventCollection`. So, I think any version is better in that case.

Savokr: Added TODO

viewer/packages/utilities/src/events/InputHandler.ts
```diff
@@ -9,6 +9,9 @@ import { Vector2 } from 'three';
 
 type PointerEventDelegate = (event: { offsetX: number; offsetY: number }) => void;
 
+//TODO savokr 17-01-2022: Change EventCollection type to be more specific by contstraining types of possible events (see draft PR).
+export type EventCollection = { [eventName: string]: EventTrigger<(...args: any[]) => void> };
```
larsmoa, line 13:
```suggestion
export type EventCollection = { [eventName: string]: EventTrigger<(...args: any[]) => void> };
```
This was actually better than I thought - I think this solution is ok. My draft PR maybe improves a tiny bit, but nothing substantial.

larsmoa (approved):

PR #1772, opened 2022-01-11
fix: correct asset mapping request

haakonflatval-cognite:


haakonflatval-cognite:
Woops, should have based this on master, BRB

PR #1773, opened 2022-01-11
fix: wrong model ID was sent to byids request

haakonflatval-cognite:


larsmoa:
Is this only 3.0?

haakonflatval-cognite:
> Is this only 3.0?

Yes, we don't fetch bounding boxes for the AssetNodeCollection in 2.x. 
As a sidenote - we do have a `getBoundingBoxesByNodeIds` method in `NodesCdfClient.ts` which we strictly speaking should be using here. But it will be hard to change now, since `AssetNodeCollection` is exposed to the user, so we can't the arguments of its constructor. (And, in addition, it seems hard to get the `NodesCdfClient` into the constructor from user space).

PR #1774, opened 2022-01-11
fix: issue causing prioritizedLoadingHint to have no effect

larsmoa:
Node style setting prioritizedForLoadingHint didn't have any effect. Exposed the issue using a unit test and fixed it.


haakonflatval-cognite (approved):
Looks good :+1:

PR #1780, opened 2022-01-13
fix: error when trackError() was called with undefined error from CadManager

larsmoa:
Sometimes happens when CadManager fails parsing a sector.


pramodcog (approved):
Looks good to me

PR #1782, opened 2022-01-14
feat: tool that adjusts CAD budget for improved framerate when camera is moving (experimental)

larsmoa:


haakonflatval-cognite:
Just so we're on the same wavelength - I have made the new glTF sector cache scale the cache size according to the budget. If this changes the budget continuously, the performance of the cache may become slow due to this resizing. We may discuss if there's another way of doing those things at the same time.

PR #1785, opened 2022-01-17
doc: remove version tags as they are not relevant in v3 docs

larsmoa:
These are used to indicate which minor version features were added, and since all of the 2.x features are available in 3.0 we can remove them.


haakonflatval-cognite (approved):
LGTM :+1:

PR #1786, opened 2022-01-17
improvement: collect metrics for what model versions are loaded

larsmoa:
Collect metrics about model version format to collect data on usage of v8 vs v9 CAD models.

### DIFF
--- a/viewer/core/src/public/RevealManager.ts
+++ b/viewer/core/src/public/RevealManager.ts
@@ -201,13 +201,6 @@ export class RevealManager {
     modelIdentifier: ModelIdentifier,
     options?: AddCadModelOptions
   ): Promise<PointCloudNode | CadNode> {
-    MetricsLogger.trackLoadModel(
-      {
-        type
-      },
-      modelIdentifier
-    );
-
     switch (type) {
       case 'cad': {
         return this._cadManager.addModel(modelIdentifier, options?.geometryFilter);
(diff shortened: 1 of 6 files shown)


viewer/core/src/public/RevealManager.ts
```diff
@@ -201,13 +201,6 @@ export class RevealManager {
...
     options?: AddCadModelOptions
   ): Promise<PointCloudNode | CadNode> {
-    MetricsLogger.trackLoadModel(
-      {
-        type
-      },
-      modelIdentifier
-    );
-
     switch (type) {
```
Savokr, line 204: Why it was removed from here entirely but not updated?

Savokr (approved):
Looks good for me, only one comment from above.

viewer/core/src/public/RevealManager.ts
```diff
@@ -201,13 +201,6 @@ export class RevealManager {
     modelIdentifier: ModelIdentifier,
     options?: AddCadModelOptions
   ): Promise<PointCloudNode | CadNode> {
-    MetricsLogger.trackLoadModel(
```
Savokr, line 204: Why it was removed from here entirely but not updated to a new format?

larsmoa: This was moved to `CadModelFactory` and `PointCloudManager`, because `RevealManager` doesn't know about formatVersion as this needs to be fetched from the server first. A bit annoying to have this two places, but acceptable.

PR #1787, opened 2022-01-17
feat: update Cognite SDK version to 7.0.0

haakonflatval-cognite:


haakonflatval-cognite:
The biggest change in the new SDK is that it now requires us to do much of the authentication ourselves. In our code, this means that the examples must be changed. There are multiple ways of doing this, I have just found one that works. We should perhaps discuss what would suit us best, but it is not critical, as it is only for internal use.

Outside users will need to do some changes to their SDK initialization as well. Thus, we should wait with merging this until we have a beta, or on some other convenient milestone.

The documentation will probably need these changes as well, but I'll wait for feedback on this before going there.

haakonflatval-cognite:
I have rewritten the auth bit, and it should now work just as smoothly as the previous one. 

Documentation might still need some treatment.

larsmoa:
We need to update docs to reflect this change and provide some guidance on how to do authentication (that is - link to Cognite SDK docs on the subject hopefully).

examples/src/utils/example-helpers.ts
```diff
@@ -84,19 +86,96 @@ type CredentialEnvironmentList = {
...
+  if (!environmentParam) {
+    return undefined;
+  }
+
+  const credentialEnvironmentList = JSON.parse(process.env.REACT_APP_CREDENTIAL_ENVIRONMENTS!) as CredentialEnvironmentList;
+
+  return credentialEnvironmentList.environments[environmentParam];
+}
+
+export async function createSDKFromEnvironment(
```
larsmoa, line 103: ☠️

larsmoa (approved):
Approved, but lets hold the merge for a while

PR #1788, opened 2022-01-17
refac: standardize event listeners to always accept one "event" argument

larsmoa:
Reduces flexibility in what an event looks like, but makes it easier to implement general functionality around events.

To me, less flexibility is acceptable - I believe that event handlers always should have maximum one argument with an object providing necessary context for the given event.

### DIFF
--- a/viewer/packages/tools/src/Cognite3DViewerToolBase.ts
+++ b/viewer/packages/tools/src/Cognite3DViewerToolBase.ts
@@ -3,7 +3,7 @@
  */
 
 import { assertNever, EventTrigger } from '@reveal/core/utilities';
-import { EventListener, EmptyEvent } from '@reveal/utilities';
+import { EmptyEvent } from '@reveal/utilities';
 
 /**
  * Base class for tools attaching to a {@see Cognite3DViewer}.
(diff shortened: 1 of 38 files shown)


viewer/packages/tools/src/Cognite3DViewerToolBase.ts
```diff
@@ -34,7 +35,7 @@ export abstract class Cognite3DViewerToolBase {
    * @param event
    * @param handler
    */
-  off(event: 'disposed', handler: () => void): void {
+  off(event: 'disposed', handler: EventListener<EmptyEvent>): void {
```
Savokr, line 38: Should it be `listener`, not `handler`?

viewer/packages/utilities/src/events/InputHandler.ts
```diff
@@ -30,28 +32,28 @@ export class InputHandler {
...
-        this._events.hover.subscribe(callback as PointerEventDelegate);
+        this._events.hover.subscribe(listener);
         break;
       default:
         assertNever(event);
     }
   }
 
-  off(event: 'click' | 'hover', callback: PointerEventDelegate): void {
+  off(event: 'click' | 'hover', callback: EventListener<PointerEvent>): void {
```
Savokr, line 49: Here as well.

larsmoa:
Leaning towards discarding this PR but leaving open for a bit more

PR #1789, opened 2022-01-18
fix: issue causing loading indicator to not update while camera was moving

larsmoa:
Dont trigger 'not loading' whenever camera moves as we now support loading while camera moves

### DIFF
--- a/viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
+++ b/viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
@@ -117,7 +117,6 @@ export class CadModelUpdateHandler {
 
   updateCamera(camera: THREE.PerspectiveCamera): void {
     this._cameraSubject.next(camera);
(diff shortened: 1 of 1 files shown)


viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
```diff
@@ -117,7 +117,6 @@ export class CadModelUpdateHandler {
 
   updateCamera(camera: THREE.PerspectiveCamera): void {
     this._cameraSubject.next(camera);
-    this._progressSubject.next(notLoadingState);
```
pramodcog, line 120: Do we know why this was added in the first place as it does not make any sense to block the loading indicator while the camera is updating

larsmoa: I think it was to send a clear message immediately to the user that loading stops when camera is moved, even though that wasn't entirely true always.

christjt (approved):
LGTM

PR #1790, opened 2022-01-19
fix: Overridden unit test WebGL1 shader console error

pramodcog:


viewer/test-utilities/src/setupJest.ts
```diff
@@ -31,6 +31,14 @@ const consoleWarn = console.warn.bind(console);
   }
 };
 
+const consoleError = console.error.bind(console);
+(console as any).error = (message?: any, ...optionalParams: any[]) => {
```
larsmoa, line 35: Add a comment about why we do this

larsmoa (changes requested):

larsmoa (approved):

PR #1791, opened 2022-01-19
improvement: track sector loading errors in MixPanel

larsmoa:
Allows us to monitor sector loading failure, like we do for v8

### DIFF
--- a/viewer/packages/sector-loader/src/GltfSectorRepository.ts
+++ b/viewer/packages/sector-loader/src/GltfSectorRepository.ts
@@ -7,8 +7,10 @@ import { ConsumedSector, V9SectorMetadata, WantedSector, LevelOfDetail } from '@
 import { BinaryFileProvider } from '@reveal/modeldata-api';
 import { CadMaterialManager } from '@reveal/rendering';
 import { GltfSectorParser, ParsedGeometry, RevealGeometryCollectionType } from '@reveal/sector-parser';
-import { SectorRepository } from '..';
+import { MetricsLogger } from '@reveal/metrics';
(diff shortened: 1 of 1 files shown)


viewer/packages/sector-loader/src/GltfSectorRepository.ts
```diff
@@ -47,67 +49,72 @@ export class GltfSectorRepository implements SectorRepository {
...
+      return {
+        levelOfDetail: sector.levelOfDetail,
+        group: group,
+        instancedMeshes: [],
+        metadata: metadata,
+        modelIdentifier: sector.modelIdentifier,
+        geometryBatchingQueue: geometryBatchingQueue
+      };
+    } catch (error) {
+      MetricsLogger.trackError(error as Error, { moduleName: 'GltfSectorRepository', methodName: 'loadSector' });
```
haakonflatval-cognite, line 115: I noticed while working with the "not resuming loading after reconnect"-problem that the error passed to MetricsLogger from the V8SectorRepository always(?) was `undefined`. Have you checked whether the error you get here is more meaningful?

haakonflatval-cognite (approved):
Had a small comment I'd like to have addressed, otherwise looks good :+1:

viewer/packages/sector-loader/src/GltfSectorRepository.ts
haakonflatval-cognite: On closer inspection, the error seems to be the same as the one received in V8SectorRepository, which is just `undefined`. I'm fixing this in my own PR, so no action is required here.

PR #1792, opened 2022-01-19
Fix: documentation update for version 3.0

Savokr:
Added documentation for reveal 3.x as separate version in docusaurus.
Several questions that should be solved:
1. Correctly reprocess or manually upload models for version 3.x.
3. Do we want to remove the banner that says that 2.x is not actively maintained? (turned out to be not that easy as it is automated. Also, the only solution with our version of docusaurus is a little dirty hack)


haakonflatval-cognite (approved):
LGTM :+1:

larsmoa:
Don't merge until we've figured out what the problem with the reprocessed model is.

PR #1793, opened 2022-01-19
fix: fix nothing being loaded after regaining connection

haakonflatval-cognite:
Turns out it was because RXJS died from eating exceptions thrown at runtime. Some of these exceptions were from the MetricsLogger, which was passed undefined arguments. Others stemmed from `loadSectors` directly.

This fixes the problem for both V8 and V9.

An issue was found with V8 models

### DIFF
--- a/viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
+++ b/viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
@@ -124,10 +124,9 @@ export class SectorLoader {
     models: CadNode[]
   ): Promise<ConsumedSector>[] {
     const consumedPromises = batch.map(async wantedSector => {
-      const model = models.filter(
-        model => model.cadModelMetadata.modelIdentifier === wantedSector.modelIdentifier
-      )[0];
-      return model.loadSector(wantedSector)
+      const model = models.filter(model => model.cadModelMetadata.modelIdentifier === wantedSector.modelIdentifier)[0];
+      return model
+        .loadSector(wantedSector)
         .catch(error => {
           log.error('Failed to load sector', wantedSector, 'error:', error);
           // Ignore error but mark sector as discarded since we didn't load any geometry
--- a/viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
+++ b/viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
@@ -88,7 +88,7 @@ export class SectorLoader {
       const filteredSectors = await this.filterSectors(sectorCullerInput, batch, sectorCuller, progressHelper);
       const consumedPromises = this.startLoadingBatch(filteredSectors, progressHelper, cadModels);
       for await (const consumed of PromiseUtils.raceUntilAllCompleted(consumedPromises)) {
-        if (consumed.result) {
+        if (consumed.result !== undefined) {
           this._modelStateHandler.updateState(consumed.result);
           yield consumed.result;
         }
--- a/viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
+++ b/viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
@@ -97,8 +97,6 @@ export class SectorLoader {
   }
 
   private getSectorCuller(sectorCullerInput: DetermineSectorsInput): SectorCuller {
-    let sectorCuller: SectorCuller;
-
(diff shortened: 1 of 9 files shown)


haakonflatval-cognite:
Another minor issue has appeared:
- For V8 models, the message `Uncaught exception (in promise)` will appear in the console whenever a sector fails to load. This is harmless, but may be distracting. On the other hand, we currently don't tell the user that something is wrong with the downloading in any other way, but we may want to be more informative than this.

viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
```diff
@@ -88,8 +88,10 @@ export class SectorLoader {
       const filteredSectors = await this.filterSectors(sectorCullerInput, batch, sectorCuller, progressHelper);
       const consumedPromises = this.startLoadingBatch(filteredSectors, progressHelper, cadModels);
       for await (const consumed of PromiseUtils.raceUntilAllCompleted(consumedPromises)) {
-        this._modelStateHandler.updateState(consumed);
-        yield consumed;
+        if (consumed.result) {
```
larsmoa, line 91: Won't this fail horribly if the result valid, but is falsy? E.g. 0, false etc etc

viewer/packages/cad-geometry-loaders/src/utilities/PromiseUtils.ts
```diff
@@ -15,13 +25,24 @@ export class PromiseUtils {
    *
    * @param promises A list of promises to complete.
    */
-  static async *raceUntilAllCompleted<T>(promises: Promise<T>[]): AsyncIterable<T> {
+  static async *raceUntilAllCompleted<T>(promises: Promise<T>[]): AsyncIterable<PromiseResult<T>> {
```
larsmoa, line 28: Good upgrade to the function. Makes more sense with the new behaviour

viewer/packages/sector-loader/src/V8SectorRepository.ts
```diff
@@ -94,7 +94,9 @@ export class V8SectorRepository implements SectorRepository {
       }
     } catch (error) {
       this._consumedSectorCache.remove(cacheKey);
-      MetricsLogger.trackError(error as Error, { methodName: 'loadSector', moduleName: 'CachedRepository' });
+      if (error) {
+        MetricsLogger.trackError(error as Error, { methodName: 'loadSector', moduleName: 'CachedRepository' });
+      }
       throw error;
```
larsmoa, line 100: I think this should be handled within MetricsLogger - even though its very odd that someone would do `throw undefined`, `MetricsLogger` should be more resilient to handling whatever it gets without barfing.

larsmoa (changes requested):

viewer/packages/sector-loader/src/V8SectorRepository.ts
haakonflatval-cognite: I revisitted the code - it actually looks like `MetricsLogger` wasn't what caused the loading problem. So while this change doesn't really contribute to fixing the problem, it still fixes that `MetricsLogger` will print `undefined` in the console (as an error) if it is called with an undefined argument. And that message is very unhelpful to a user. 

I'll go forth in trying to find the source of the undefined error, as it seems like a better way to fix it.

viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
haakonflatval-cognite: `consumed` is supposed to be of the type `PromiseResult<ConsumedSector>`, and `result` will have type `ConsumedSector`. If it is falsy, it does not have that type, and I think it's fair to treat it as an error in that case.

viewer/packages/sector-loader/src/V8SectorRepository.ts
haakonflatval-cognite: I found the source - it was at the bottom, with the low-level fetch calls. I have made sure they return better errors.

viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
larsmoa: Ok, I still think being explicit about this is best - so I still prefer this:
```suggestion
        if (consumed.result !== undefined) {
```
up to you

larsmoa (approved):

larsmoa (approved):

PR #1794, opened 2022-01-20
doc: fix issue causing navigation to not work with interactable clipping plane example

larsmoa:
Resolves issue with the "Interactable clipping plane" example causing navigation to be blocked by the drag controller or the clip plane.

### DIFF
--- a/documentation/docs/examples/clipping.mdx
+++ b/documentation/docs/examples/clipping.mdx
@@ -115,8 +115,7 @@ controls.addEventListener('drag', () => {
 });
 
 // When we are hovering over a sphere we want to disable viewer controls
-controls.addEventListener('hoveron', () => { 
-  controls.activate();
+controls.addEventListener('hoveron', () => {
   viewer.cameraControlsEnabled = false;
   draggerMaterial.color.set('gray');
   viewer.requestRedraw();
@@ -124,7 +123,6 @@ controls.addEventListener('hoveron', () => {
 
 // And turn them back on when we stop hovering over it
 controls.addEventListener('hoveroff', () => {
-  controls.deactivate();
   viewer.cameraControlsEnabled = true;
   draggerMaterial.color.set('white');
   viewer.requestRedraw();
--- a/documentation/docs/examples/clipping.mdx
+++ b/documentation/docs/examples/clipping.mdx
@@ -116,6 +116,7 @@ controls.addEventListener('drag', () => {
 
 // When we are hovering over a sphere we want to disable viewer controls
 controls.addEventListener('hoveron', () => { 
+  controls.activate();
   viewer.cameraControlsEnabled = false;
   draggerMaterial.color.set('gray');
   viewer.requestRedraw();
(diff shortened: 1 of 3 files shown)


documentation/docs/examples/clipping.mdx
```diff
@@ -116,13 +116,15 @@ controls.addEventListener('drag', () => {
...
 controls.addEventListener('hoveron', () => { 
+  controls.activate();
   viewer.cameraControlsEnabled = false;
   draggerMaterial.color.set('gray');
   viewer.requestRedraw();
 });
 
 // And turn them back on when we stop hovering over it
 controls.addEventListener('hoveroff', () => {
+  controls.deactivate();
```
Savokr, line 127: After controls deactivation (i.e. after one drug or just 'hoveroff') you can't drug ball again and 'hoveron' event is not thrown because controls are deactivated. So we should avoid using `controls.deactivate()` I think. 
I fixed it last time when it was a different problem, not sure what broke it. Possibly the reason for controls block is inside CameraManager/ComboControls.

documentation/versioned_docs/version-2.x/examples/clipping.mdx
```diff
@@ -116,13 +116,15 @@ controls.addEventListener('drag', () => {
 
 // When we are hovering over a sphere we want to disable viewer controls
 controls.addEventListener('hoveron', () => { 
+  controls.activate();
```
Savokr, line 119: Same here

Savokr (changes requested):
Not a fix yet because it works only once 🙂 . I could also try to look into it as I was the last one who touched this example.

documentation/docs/examples/clipping.mdx
larsmoa: Any idea how to fix the issue properly @Savokr ?

Savokr: Thinking on it now. Tried to find a commit after which it was broken

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -418,12 +419,12 @@ export class ComboControls extends EventDispatcher {
...
-      window.removeEventListener('mousemove', onMouseMove);
-      window.removeEventListener('mouseup', onMouseUp);
+      window.removeEventListener('pointermove', onMouseMove);
+      window.removeEventListener('pointerup', onMouseUp);
     };
 
-    window.addEventListener('mousemove', onMouseMove, { passive: false });
-    window.addEventListener('mouseup', onMouseUp, { passive: false });
+    window.addEventListener('pointermove', onMouseMove, { passive: false });
+    window.addEventListener('pointerup', onMouseUp, { passive: false });
```
larsmoa, line 427: What does this do?

Savokr: They are responsible for mouse move during rotation. Or if you mean why are they changed, it's because porting initial event to `pointerdown` requires everything else to have the same type of event to work properly.

larsmoa (commented):
I want to approve, but I cant since I'm author. Feel free to approve @Savokr and merge

Savokr (approved):
Looks good for me

PR #1798, opened 2022-01-21
fix: certain models cause computer crashes

haakonflatval-cognite:
The `GeometryBatchingManager` maintains a number of buffer arrays representing instance meshes and different primitive types, stored in a map using some string ID as key. In some cases, the key for an instance mesh, which is always a number stored as a string, would coincide with the key for a primitive, which is also a number (but seemingly is thought not to be).

A fix is therefore to always make sure the primitive array buffer ID does not contain just a number.


larsmoa:
Can we add some assertions to GeometryBathcingManager to do some basic verification of the data we add to avoid these types of issues?

larsmoa (approved):

PR #1802, opened 2022-01-21
experiment: faster (?) preview doc build in CI

larsmoa:


haakonflatval-cognite (commented):
Will this turn off the CI-check for building docs in PRs? Won't that make it too easy for us to merge/release a revision with a failing doc build?

larsmoa:
> 

No, docs will still be built, but minification of source is disabled. I don't expect this to cause problems, minification should in theory not change behaviour of the code. It would be better to have it enabled of course, but I think it's worth the reduction in build time in CI.

.github/workflows/ci.yml
```diff
@@ -108,29 +108,6 @@ jobs:
...
-        run: yarn build:prod
-
-      - name: Install NPM dependencies
-        working-directory: documentation
-        run: yarn --immutable
-
-      - name: Build docs
-        working-directory: documentation
-        run: yarn run build
-
```
haakonflatval-cognite, line 133: Wait, so doc build is not disabled in CI? What does this do then?

larsmoa: Yes, it would disable _production builds_ of docs, but `preview docs` will still be built (and they are the same, just without minification enabled).

larsmoa: image

haakonflatval-cognite: Ah, I see

PR #1805, opened 2022-01-24
fix: remove-batch needs to update until end of defrag buffer

christjt:
When removing batch from defrag buffer it needs to extend all the way to the end of buffer since every consecutive batch is shifted down in the buffer.


haakonflatval-cognite (approved):
LGTM :+1: 
Is this what caused the missing geometry?

christjt:
> LGTM 👍 Is this what caused the missing geometry?

Yes

larsmoa (commented):
Looks unit testable - is it? Add test?

PR #1809, opened 2022-01-26
Trigger sector load on assigning priority [REV-295]

haakonflatval-cognite:
Causes setting prioritized hint for node collections to trigger sector loading so that the result becomes visible at once.

### DIFF
--- a/viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
+++ b/viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
@@ -70,7 +70,7 @@ export class CadModelUpdateHandler {
         this._loadingHintsSubject.pipe(startWith({} as CadLoadingHints)),
         this._budgetSubject.pipe(startWith(this._budget))
       ]).pipe(map(makeSettingsInput)),
-      this._prioritizedLoadingHintsSubject,
+      this._prioritizedLoadingHintsSubject.pipe(startWith(undefined as void)),
       combineLatest([
         this._cameraSubject.pipe(auditTime(500)),
         this._cameraSubject.pipe(auditTime(250), emissionLastMillis(600))
@@ -142,8 +142,6 @@ export class CadModelUpdateHandler {
     this._modelStateHandler.addModel(model.cadModelMetadata.modelIdentifier);
     this._modelSubject.next({ model, operation: 'add' });
     model.nodeAppearanceProvider.on('prioritizedAreasChanged', () => this.updatePrioritizedAreas());
-    // Ensure there is at least one call in the line for the model to load
-    this.updatePrioritizedAreas();
   }
 
   removeModel(model: CadNode): void {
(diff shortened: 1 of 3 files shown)


viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
```diff
@@ -138,6 +141,9 @@ export class CadModelUpdateHandler {
   addModel(model: CadNode): void {
     this._modelStateHandler.addModel(model.cadModelMetadata.modelIdentifier);
     this._modelSubject.next({ model, operation: 'add' });
+    model.nodeAppearanceProvider.on('prioritizedAreasChanged', () => this.updatePrioritizedAreas());
+    // Ensure there is at least one call in the line for the model to load
+    this.updatePrioritizedAreas();
```
haakonflatval-cognite, line 146: I experienced that without this, the model is not loaded at all, probably because rxjs requires there to be a signal sent for each input subject to trigger load. Is there a better way of fixing this?

christjt (approved):
Nice, LGTM

larsmoa (changes requested):
Unit testable?

viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
```diff
@@ -149,6 +155,10 @@ export class CadModelUpdateHandler {
     this._loadingHintsSubject.next(cadLoadingHints);
   }
 
+  updatePrioritizedAreas(): void {
+    this._prioritizedLoadingHintsSubject.next();
+  }
```
larsmoa, line 160: Does this need to be public?

viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
larsmoa: Hmm, not sure I understand. Without this the model isn't loaded at all? How has this worked before? Looks quite dodgy - wouldn't `_modelSubject.next()` trigger a change so that data is loaded?

larsmoa (changes requested):

viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
haakonflatval-cognite: Nope, fixed.

viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
haakonflatval-cognite: It is documented here: https://rxjs.dev/api/index/function/combineLatest
> combineLatest will actually wait for all input Observables to emit at least once
So it seems we need to emit at least once to make it load. It fits with my experience.
Anyway, I found a cleaner way, using `startWith` to create an initial value.

larsmoa: Great, startWith is bette r:)

larsmoa (approved):

PR #1812, opened 2022-01-26
doc: Proper handling of versioned reveal demos

Savokr:
Fixed proper reveal version for corresponding version of documentation. Turned out that we were always using master version of reveal in demos even with 2.x version of docs 🙃. You can check it on production documentation with new controls modes that can be set in console for 2.x version. Fixed it and added automatic replacement for future rebuilds.


documentation/package.json
```diff
@@ -25,23 +25,25 @@
...
-    "replace-3.x": "yarn replace-3.x:clean-versioned && yarn replace-3.x:copy-docs && yarn replace-3.x:update-sdk && yarn replace-3.x:update-reveal && yarn replace-3.x:update-runnable",
+    "replace-3.x": "yarn replace-3.x:clean-versioned && yarn replace-3.x:copy-docs && yarn replace-3.x:update-sdk && yarn replace-3.x:update-reveal && yarn replace-3.x:update-runnable && yarn replace-3.x:update-site-imports",
     "replace-latest-by-next": "yarn replace-3.x"
   },
   "dependencies": {
     "@cognite/reveal": "link:../viewer/dist",
     "@cognite/reveal-1.x": "npm:@cognite/reveal@1.x.x",
     "@cognite/reveal-2.x": "npm:@cognite/reveal@2.x.x",
-    "@cognite/reveal-3.x": "npm:@cognite/reveal@3.0.0-alpha.0",
+    "@cognite/reveal-3.x": "npm:@cognite/reveal@3.x.x",
```
larsmoa, line 46: Does this pick up alpha versions too?

Savokr: I assume, yes. When building locally it asked to pick a version from the list. Not sure how it was resolved on documentation deployment but it is using proper version.

Savokr: Just checked and actually it's a wrong version format for alphas. Changed it to correct one. By the way, good site to check proper version tag: https://semver.npmjs.com/

PR #1823, opened 2022-01-28
chore: upgrade three.js to r137

haakonflatval-cognite:
Biggest change was that `RGBFormat` is no longer a valid format for textures.


viewer/packages/rendering/src/transform/NodeTransformTextureBuilder.ts
```diff
@@ -60,9 +59,9 @@ export class NodeTransformTextureBuilder {
...
   private setOverrideIndex(treeIndex: number, transformIndex: number) {
     const data = this._transformOverrideIndexTexture.image.data;
-    data[treeIndex * 3 + 0] = (transformIndex + 1) >> 16;
-    data[treeIndex * 3 + 1] = (transformIndex + 1) >> 8;
-    data[treeIndex * 3 + 2] = (transformIndex + 1) >> 0;
+    data[treeIndex * 4 + 0] = (transformIndex + 1) >> 16;
+    data[treeIndex * 4 + 1] = (transformIndex + 1) >> 8;
+    data[treeIndex * 4 + 2] = (transformIndex + 1) >> 0;
     this._transformOverrideIndexTexture.needsUpdate = true;
   }
```
larsmoa, line 66: Do we now waste a byte? Should we use the last byte as well?

haakonflatval-cognite: Yes, we waste a byte. We could probably come up with something interesting to use it for :thinking:

larsmoa: or we could just simply use all 32 bytes for storing the transformIndex for now? Even if it seems like a very odd case that you'd actually need that many different transforms though

haakonflatval-cognite: I suggest that we don't. Since we use 32-bit floats to hold tree indices, Reveal will struggle with anything that has more than 2^24 nodes anyway, which is exactly the number of nodes we support with the transform index now. So it doesn't seem like a reasonable way to scale currently.

PR #1824, opened 2022-01-28
chore: only deploy preview docs from one PR concurrently

larsmoa:
This ensures that only one job is publishing preview docs at the same time, avoiding CI to fail because multiple PRs are deploying docs at the same time.

### DIFF
--- a/.github/workflows/preview-documentation.yml
+++ b/.github/workflows/preview-documentation.yml
@@ -4,9 +4,8 @@ on:
   pull_request:
     branches-ignore:
       - 'dependabot/**'
-
 jobs:
-  build-and-deploy:
+  build:
     name: Publish documentation preview
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/preview-documentation.yml
```diff
@@ -47,6 +46,12 @@ jobs:
           yarn install --immutable
           yarn build:fast
 
+  deploy:
```
haakonflatval-cognite, line 49: How does this solve the issue? Is this somehow guaranteed to be executed serially among all PRs?

larsmoa: Notice the `concurrency`-argument below. This ensure there only will be one execution of the step across all ongoing Github Actions.

haakonflatval-cognite: Alright, seems good :+1:

haakonflatval-cognite (approved):

PR #1828, opened 2022-02-01
feat: add clipping planes to viewer state

larsmoa:


viewer/core/src/utilities/ViewStateHelper.ts
```diff
@@ -14,11 +13,19 @@ import { NodeAppearance } from '@reveal/cad-styling';
...
   };
-  models: ModelState[];
+  models?: ModelState[];
+  clippingPlanes?: ClippingPlanesState[];
+};
+
+export type ClippingPlanesState = {
+  nx: number;
+  ny: number;
+  nz: number;
```
Savokr, line 27: Why we shouldn't use ordinary Vector3?

viewer/packages/tools/src/AxisView/AxisViewTool.test.ts
```diff
@@ -13,15 +13,16 @@ import { defaultAxisBoxConfig } from './types';
...
   let viewer: Cognite3DViewer;
 
-  const sdk = new CogniteClient({ appId: 'cognite.reveal.unittest' });
-  mockClientAuthentication(sdk);
-  const context = createGlContext(64, 64, { preserveDrawingBuffer: true });
-  const renderer = new THREE.WebGLRenderer({ context });
-  const domSize = { height: 480, width: 640 };
-
   beforeEach(() => {
+    const sdk = new CogniteClient({ appId: 'cognite.reveal.unittest' });
```
Savokr, line 20: From how it was written previously, I assume it could be inside `beforeAll`, not `beforeEach` because context could be the same for each test.

Savokr (approved):
Looks good, only have small comments about the test and format of the state.

viewer/core/src/utilities/ViewStateHelper.ts
larsmoa: `THREE.Vector3` would add a bunch of unnecessary stuff to the resulting JSON. The point here is to generate a small output as possible, so we use dedicated types.

viewer/packages/tools/src/AxisView/AxisViewTool.test.ts
larsmoa: I guess it could be in beforeAll. I generally don't use it a lot as beforeEach ensures that no state "bleeds" between test A and B and generally it's fast enough to just re-init for each test.

PR #1831, opened 2022-02-03
fix: Zooming to points not on the model was too fast in zoomToCursor mode

Savokr:
Changed algorithm of calculating new scroll target if raycast missed the model to account for current distance to point under cursor.


larsmoa:
Initial testing looks promising!

PR #1832, opened 2022-02-03
improvement: Remove WebGL1 reference

pramodcog:


haakonflatval-cognite (approved):
Looks good :+1: Have you checked whether all warnings are gone from the visual tests now?

Also, seems like I introduced some merge conflict with my upgrade to three 137 in the tests. I think you can just delete what I did and use what's here, if it works.

Could be nice to rebase everything on the newest master instead of keeping the history of irrelevant commits too, but I guess it doesn't matter as long as you squash and remove them from the commit message in the end :thinking:

pramodcog:
> Looks good 👍 Have you checked whether all warnings are gone from the visual tests now?
> 
> Also, seems like I introduced some merge conflict with my upgrade to three 137 in the tests. I think you can just delete what I did and use what's here, if it works.
> 
> Could be nice to rebase everything on the newest master instead of keeping the history of irrelevant commits too, but I guess it doesn't matter as long as you squash and remove them from the commit message in the end 🤔

Will squash & merge, so that history of irrelevant commits are not in the merge

PR #1833, opened 2022-02-04
Better suggested camera config

haakonflatval-cognite:
This influences how camera is configured on first model load. Inspired by this complaint: https://cognitedata.slack.com/archives/C8A0QGMLN/p1643968454118329

The reasoning is that the camera is now looking at the largest area of the bounding box, whereas earlier it would look at the smaller. Provided screenshots may explain better.

Before: 
bg1

After: 
bg0


haakonflatval-cognite:
In the newest update, camera angle is restricted to max. 30 degrees above (below) the horizon.

viewer/packages/rendering/src/cameraconfig.ts
```diff
@@ -13,7 +13,26 @@ export interface SuggestedCameraConfig {
...
+      xzProjection.x = 1.0;
+      xzProjection.y = 1.0;
+    }
+
+    xzProjection.normalize().multiplyScalar(Math.cos(MAX_VERTICAL_ANGLE));
+    direction.x = xzProjection.x;
+    direction.z = xzProjection.y;
+
+    direction.y = Math.sin(MAX_VERTICAL_ANGLE);
+  }
```
christjt, line 32: Concider inverting the if statement to reduce nesting. Also, the `xzProjection` can be covered using the ternary operator.

christjt (changes requested):
Looks good, only a minor NIT

christjt (approved):

haakonflatval-cognite:
Second test is confirmed to fail without the below-30-degree-horizon restriction.

haakonflatval-cognite:
Turns out this works less-than-well for real-life models. But so did the last one. Oh well, leaving it for now. (It is only in use internally, although it is exposed undocumentedly in `CadNode`.

PR #1834, opened 2022-02-04
doc: change examples to not use same app ID as docs

larsmoa:
- Documentation will now be identified as `reveal-docs-<version>` in Mixpanel.
- Change documentation examples to use a separate appId from the docs
- Also export REVEAL_VERSION from @cognite/reveal.


christjt (approved):
LGTM

PR #1842, opened 2022-02-08
improvement: add alternative unit names that doesn't end with S

larsmoa:
Add alternative spelling of units to be more resilient.


larsmoa:
> Looks good, but should we also ensure resilience towards e.g. lower-case spellings of the units?

Probably a good idea, yes...

viewer/core/src/datamodels/cad/CadModelFactory.ts
```diff
@@ -32,6 +32,7 @@ export class CadModelFactory {
 
   async createModel(externalModelIdentifier: ModelIdentifier, geometryFilter?: GeometryFilter): Promise<CadNode> {
     const metadata = await this._cadModelMetadataRepository.loadData(externalModelIdentifier);
+    console.log(metadata);
```
haakonflatval-cognite, line 35:
```suggestion
```

haakonflatval-cognite (approved):
Should remove the console log, otherwise looks good! :+1:

haakonflatval-cognite (changes requested):
Also, this was caught in the visual tests: "Uncaught (in promise) TypeError: Cannot read property 'toLowerCase' of undefined".

Seems like a null check is needed (although I don't really see how this passes the type check, which doesn't allow undefined's in the `getDistanceToMeterConversionFactor` function

haakonflatval-cognite (approved):

PR #1845, opened 2022-02-09
fix: defer download of web workers until needed

haakonflatval-cognite:
This simultaneously fixes the issue where the parser workers are downloaded even when only using v9.


larsmoa (approved):
LGTM, but:
1) see comment
2) rename PR to emphasise what is fixed (e.g. "improvement: avoid initialization of web workers before they are needed") - btw, chore is more used for CI/buildsetup etc

PR #1847, opened 2022-02-11
fix: iOS primitive rendering issue

pramodcog:


viewer/packages/rendering/src/glsl/sector/primitives/cone.frag
```diff
@@ -156,7 +157,6 @@ void main() {
       }
   #endif
 
-
-    float fragDepth = updateFragmentDepth(p, projectionMatrix);
-    updateFragmentColor(renderMode, color, v_treeIndex, normal, fragDepth, matCapTexture, GeometryType.Primitive);
+    updateFragmentDepth(p, v_projectionMatrix);
+    updateFragmentColor(renderMode, color, v_treeIndex, normal, gl_FragCoord.z, matCapTexture, GeometryType.Primitive);
```
larsmoa, line 161: Why use gl_FragCoord.z?

pramodcog: This is wrong, I think, I pushed while testing/identifying the issue, my bad!

hayer:
LGTM. Important fix for my team. @larsmoa could you please prioritize reviewing this? 👼🏻

larsmoa (changes requested):
Add `flat` for all of the "uniforms"

pramodcog:
> Add `flat` for all of the "uniforms"

I doubt that we could not use `flat` with uniform, let me try and check!

larsmoa:
> > Add `flat` for all of the "uniforms"
> 
> I doubt that we could not use `flat` with uniform, let me try and check!
> 
> Yup, it is not valid. Uniforms does not take `flat` qualifier

I mean that use `flat` for all the varyings that replace uniforms. Flat for uniforms doesnt make sense as its constant.

larsmoa (changes requested):
As mentioned previously, we have a lot of uniforms in many fragment shader - have you assessed what works and what doesn't? We need to be thorough on this!

viewer/packages/rendering/src/glsl/sector/primitives/cone.frag
```diff
@@ -125,13 +117,17 @@ void main() {
...
       ) {
         // Missed the other point too
         discard;
       }
     }
+    #if NUM_CLIPPING_PLANES > 0
+      if (isClipped(p, v_clippingPlanes)) {
+        discard;
+      }
+    #endif
```
larsmoa, line 130: Why is this moved?

pramodcog: `NUM_CLIPPING_PLANES` value is assigned by the threejs only when clipping is active, else its value is always `0`, so the earlier, the check for discarding the clipped pixel function call will have an compiler error as I am passing `v_clippingPlanes[NUM_CLIPPING_PLANES ]` as a parameter.

PR #1852, opened 2022-02-14
refac: Create camera manager interface that will allow users to provide their own CameraManager

Savokr:
Cherry picked version of https://github.com/cognitedata/reveal/pull/1848 

Working version of CameraManager interface with all necessary fixes in internal code and requested ability to provide custom camera, etc.


viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -240,6 +243,10 @@ export class ComboControls extends EventDispatcher {
     this.triggerCameraChangeEvent();
   };
 
+  get cameraAdditionalRotation(): Quaternion {
+    return this._additionalRotation;
+  }
```
larsmoa, line 248: Dafuck

viewer/packages/camera-manager/src/types.ts
```diff
@@ -61,7 +61,89 @@ export interface RevealCameraControls extends THREE.EventDispatcher {
...
+   * viewer.setCameraPosition(position);
+   * viewer.setCameraTarget(target);
+   * ```
+   */
+  setCameraPosition: (position: THREE.Vector3) => void;
+  /**
+   * @obvious
+   * @returns Camera's position in world space.
+   */
+  getCameraPosition: () => THREE.Vector3;
```
larsmoa, line 121: use plain function definitions, not lambda

larsmoa (changes requested):

documentation/versioned_docs/version-3.x/examples/clipping.mdx
```diff
@@ -116,14 +116,14 @@ controls.addEventListener('drag', () => {
...
   viewer.requestRedraw();
 });
 
 // And turn them back on when we stop hovering over it
 controls.addEventListener('hoveroff', () => {
-  viewer.cameraControlsEnabled = true;
+  viewer.cameraManager.cameraControlsEnabled = true;
   draggerMaterial.color.set('white');
   viewer.requestRedraw();
 });
```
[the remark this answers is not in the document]
Savokr: Oh, that's right

examples/src/pages/Migration.tsx
```diff
@@ -26,6 +26,7 @@ import { CameraUI } from '../utils/CameraUI';
 import { PointCloudUi } from '../utils/PointCloudUi';
 import { ModelUi } from '../utils/ModelUi';
 import { createSDKFromEnvironment } from '../utils/example-helpers';
+import { DefaultCameraManager } from '@cognite/reveal/packages/camera-manager';
```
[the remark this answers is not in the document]
Savokr: Oh, it was automatically imported and I haven't corrected that

viewer/packages/camera-manager/src/ComboControls.ts
Savokr: This is the key thing that got animation working. It is rotation that is applied after lookAt method was called. Value for it is set in `setCameraRotation` method and basically equal to the difference between desired rotation of the camera and the one we get after all manipulations in update function (lookAt method). So if we set rotation to the same value as the camera have right now, this additional rotation will become zero.

Savokr: So, renamed to cameraRawRotation and changed the way it is handled a little bit. (As we describe on Wednesday 16.02).

Savokr:
Fixed everything from comments and implemented setState method, so should be ready for another review. 
However, I thought about one problem that could arise when someone just want to pass custom `THREE.PerspectiveCamera` instance. The way our current `DefaultCameraManager` is implemented requires to pass `raycastFunction` in the constructor. It uses `Cognite3DViewer` data about the model to create proper raycast. So if someone just want to have custom camera without creating their own `CameraManager` it won't be possible with current implementation. 

My thoughts: we could either create an implementation-specific method in `DefaultCameraManager` to change camera in runtime so that people will be able to set it after `Cognite3DViewer` initialization or we can set `raycastFunction` in runtime and change the constructor of `DefaultCameraManager` a little bit.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -71,34 +97,31 @@ export class CameraManager {
...
+    raycastFunction: (x: number, y: number) => Promise<CameraManagerCallbackData>,
+    camera?: THREE.PerspectiveCamera
   ) {
-    this._camera = camera;
+    this._camera = camera ?? new THREE.PerspectiveCamera(60, undefined, 0.1, 10000);
+    // TODO savokr 28-10-2021: Consider removing default camera position initialization
+    this._camera.position.x = 30;
+    this._camera.position.y = 10;
+    this._camera.position.z = 50;
+    this._camera.lookAt(new THREE.Vector3());
```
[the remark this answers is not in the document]
Savokr: Removing it fails a few visual tests, so not sure what to do. We can either set it manually in visual tests or leave as is.

christjt:
Visual tests are failing, probably due to default camera position being removed (which is good). Just add it to the `Cognite3DTestViewer`, which should then apply to all visual tests.

viewer/core/src/public/migration/types.ts
```diff
@@ -76,31 +77,14 @@ export interface Cognite3DViewerOptions {
...
   /**
-   * When false, the sensitivity of the camera controls will not be updated automatically.
-   * This can be useful to better control the sensitivity of the 3D navigation.
-   *
-   * When not set, control the sensitivity of the camera using `viewer.cameraControls.minDistance`
-   * and `viewer.cameraControls.maxDistance`.
+   * Camera manager instance that is used for controlling the camera.
+   * It is responsible for all manipulations that are done with the camera,
+   * including animations and modification of state. Also, gives ability
+   * to provide custom THREE.PerspectiveCamera instance to Cognite3DViewer.
```
larsmoa, line 84:
```suggestion
   * to provide custom `THREE.PerspectiveCamera` instance to {@link Cognite3DViewer}.
```

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,58 @@
...
-
-  private _cameraControlsOptions: Required<CameraControlsOptions> = { ...CameraManager.DefaultCameraControlsOptions };
-
-  public automaticNearFarPlane = true;
-  public automaticControlsSensitivity = true;
-
+import { CameraChangeData, CameraState } from './types';
+/**
+ * Camera manager interface.
+ * */
```
larsmoa, line 8:
```suggestion
 */
```

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,58 @@
...
+export interface CameraManager {
   /**
-   * Reusable buffers used by functions in Cognite3dViewer to avoid allocations.
+   * Get THREE.PerspectiveCamera object for special cases. All manipulations of the camera should be done
+   * through other CameraManager methods.
+   * **/
```
larsmoa, line 13:
```suggestion
   * Returns the camera used for rendering in {@link Cognite3DViewer}. 
   * Note that the camera will not be modified directly by Reveal. Implementations must trigger the `cameraChange`-event whenever the
   * camera changes.
   */
```

It's up to implementations if they want to modify the camera directly, whats important is that Reveal wont do that and that cameraChanged is triggered when it is modified.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,58 @@
...
+  getCamera(): THREE.PerspectiveCamera;
+  /**
+   * Set camera's state
+   *
+   * @param state Camera state, all fields are optional.
```
larsmoa, line 18:
```
   * @param state Camera state, all fields are optional.
   * @param state.position ....
   * @param state.target ...
```

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,58 @@
...
-    return farPoint;
-  }
+  getCameraState(): Required<CameraState>;
 
   /**
-   * Calculates new camera target based on cursor position.
-   * @param event MouseEvent that contains pointer location data.
+   * Required for understadning of `cameraChanged` definition for other parts of Reveal
+   * @param event
+   * @param callback
```
larsmoa, line 37:
```suggestion
   * Subscribe to changes to the camera. This is used by Reveal react to changes to the camera.
   * @param event
   * @param callback
```

Add some descriptions to `event` and `callback` too.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,58 @@
...
-
-    return newTarget;
-  }
-
+  on(event: 'cameraChange', callback: CameraChangeData): void;
   /**
-   * Removes controls event listeners if they are defined.
+   * Required for understadning of `cameraChanged` definition for other parts of Reveal
+   * @param event
+   * @param callback
```
larsmoa, line 43: Change to something about "unsubscribe to event", current description doesn't say what this function does - it says what the event does (which doesn't make sense in the unsubscribe function)

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,58 @@
...
-        assertNever(controlsOptions.mouseWheelAction);
-    }
-  }
+  fitCameraToBoundingBox(boundingBox: THREE.Box3, duration?: number, radiusFactor?: number): void;
   /**
-   * Method for setting up camera controls listeners and values inside current controls class.
+   * Updates internal state of camera manager.
+   * @param deltaTime
+   * @param boundingBox
    */
```
larsmoa, line 52: This doesn't tell anything to implementors about what they should do in their implementations. This description must be detailed and explain in detail what the function is expected to do and what the arguments are.

viewer/packages/camera-manager/src/DefaultCameraManager.ts
```diff
@@ -0,0 +1,742 @@
...
+  CameraControlsOptions,
+  CameraChangeData,
+  PointerEventDelegate,
+  CameraState
+} from './types';
+import { CameraManager } from './CameraManager';
+import { assertNever, EventTrigger, InputHandler, disposeOfAllEventListeners } from '@reveal/utilities';
+import range from 'lodash/range';
+
+export class DefaultCameraManager implements CameraManager {
```
larsmoa, line 19: Add description about how the default camera manage behaves and what it supports.

viewer/packages/camera-manager/src/DefaultCameraManager.ts
```diff
@@ -0,0 +1,742 @@
...
+  enableKeyboardNavigation(): void {
+    this.keyboardNavigationEnabled = true;
+  }
+
+  /**z
+   * Disables camera movement by pressing WASD or arrows keys.
+   */
+  disableKeyboardNavigation(): void {
+    this.keyboardNavigationEnabled = false;
+  }
```
larsmoa, line 192: These are not necessary when we have the setter above, remove.

viewer/packages/camera-manager/tests/DefaultCameraManager.test.ts
```diff
@@ -0,0 +1,43 @@
+/*!
+ * Copyright 2022 Cognite AS
+ */
+
+import * as THREE from 'three';
+import { DefaultCameraManager } from '../src/DefaultCameraManager';
+import { InputHandler } from '@reveal/utilities';
+
+describe('DefaultCameraManager', () => {
```
larsmoa, line 9:
```suggestion
describe(DefaultCameraManager.name, () => {
```

This test could be extensively extended but I'm not sure it's worth it at the current stage.

viewer/packages/camera-manager/src/types.ts
```diff
@@ -61,6 +61,20 @@ export interface RevealCameraControls extends THREE.EventDispatcher {
...
+   */
+  position?: THREE.Vector3;
+  /**
+   * Camera target in world space.
+   * **/
+  target?: THREE.Vector3;
+  /**
+   * Camera local rotation in quaternion form.
+   */
+  rotation?: THREE.Quaternion;
```
larsmoa, line 76: Not sure how to handle this, BUT target and rotation are conflicting. What would it take to get rid of `target` and make that a purely internal thing in `CameraManager`-implementations?

larsmoa: Alternatively, should we define a precedence order - e.g. that rotation takes precedence over target when both are set?

larsmoa (changes requested):
Great leap forward and I love the super-simplistic nature of the `CameraManager`-interface which makes it easy to create simple implementations.

See a bunch of comments. I think the biggest question is around the `CameraState`-type and that `target` and `rotation` is conflicting, so we need to make a choice on how to handle that.

viewer/packages/camera-manager/src/types.ts
Savokr: That's how I implemented it in `DefaultCameraManager`. Target takes precedence when both are defined because I thought both of them are changing target, so explicit definition should have priority. Another option is to change rotation to only roll. That will remove the conflict as roll doesn't affect target-oriented rotation.

viewer/packages/camera-manager/tests/DefaultCameraManager.test.ts
Savokr: Totally agree, I could add some tests in another PR.

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -240,6 +245,10 @@ export class ComboControls extends EventDispatcher {
     this.triggerCameraChangeEvent();
   };
 
+  get cameraRawRotation(): Quaternion {
```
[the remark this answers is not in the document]
Savokr: Added

viewer/packages/camera-manager/src/types.ts
Savokr: Now, setting rotation and target at the same time throws an error

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,68 @@
...
-
-    return farPoint;
-  }
+  setCameraState(state: CameraState): void;
 
   /**
-   * Calculates new camera target based on cursor position.
-   * @param event MouseEvent that contains pointer location data.
+   * Get camera's state
    */
```
larsmoa, line 40: Add doc for each of the fields in `CameraState`, e.g.
```
@param state.position The camera position in world coordinates
```

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,68 @@
...
-  }
+  off(event: 'cameraChange', callback: CameraChangeData): void;
 
+  fitCameraToBoundingBox(boundingBox: THREE.Box3, duration?: number, radiusFactor?: number): void;
   /**
-   * Convert pixel coordinates of the cursor to [-1,1]^2 coordinates.
-   * @param pixelX
-   * @param pixelY
+   * Updates internal state of camera manager. Expected to update visual state of the camera
+   * as well as it's near and far planes if needed. Called every time screen is refreshed.
```
larsmoa, line 59: `Called every time screen is refreshed.` <- this is wrong, right? It's actually called more often than that?

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -94,6 +95,7 @@ export class ComboControls extends EventDispatcher {
   private readonly _spherical: Spherical = new Spherical();
   private _sphericalEnd: Spherical = new Spherical();
   private readonly _deltaTarget: Vector3 = new Vector3();
+  private readonly _rawCameraRotation: THREE.Quaternion = new Quaternion();
```
larsmoa, line 98:
```suggestion
  private readonly _rawCameraRotation = new Quaternion();
```

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: That's correct and one of the things that should be changed, I believe. Also, it's not every new rendered frame but every screen refresh with `requestAnimationFrame`.

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Are you sure we want `@param` tag even on return? I added it as: 
`@returns Camera state: position, target and rotation.
`

viewer/packages/camera-manager/src/CameraManager.ts
larsmoa: The doc should reflect what actually happens. I at least read this "every time the 3D view is re-rendered". A more precise version might be "Called in requestAnimationFrame-loop" and also emphasis that it's very important that the performance is high.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,75 @@
...
-    changeCameraTargetOnClick: false
-  };
-
-  private _cameraControlsOptions: Required<CameraControlsOptions> = { ...CameraManager.DefaultCameraControlsOptions };
-
-  public automaticNearFarPlane = true;
-  public automaticControlsSensitivity = true;
-
+import { CameraChangeData, CameraState } from './types';
+/**
```
larsmoa, line 6:
```suggestion

/**
```

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,75 @@
...
+ * Camera manager interface. It is responsible for all manipulations that are done with the camera,
+ * including animations and modification of state. Also, gives ability
+ * to provide custom `THREE.PerspectiveCamera` instance to {@link Cognite3DViewer}.
+ * For working implementation required to fire `cameraChange`-event when camera is changed.
+ * Default implementation is {@link DefaultCameraManager}.
+ */
```
larsmoa, line 12:
```suggestion
 * Interface for manager responsible for all manipulations to the camera,
 * including animations and modification of state. Implementations are responsible for
 * providing a `THREE.PerspectiveCamera` instance to {@link Cognite3DViewer}. Implementations
 * must trigger the `cameraChange`-event when camera is changed.
 * The default implementation is {@link DefaultCameraManager}.
 */
```

viewer/packages/camera-manager/src/DefaultCameraManager.ts
```diff
@@ -0,0 +1,686 @@
...
+  CameraChangeData,
+  PointerEventDelegate,
+  CameraState
+} from './types';
+import { CameraManager } from './CameraManager';
+import { assertNever, EventTrigger, InputHandler, disposeOfAllEventListeners } from '@reveal/utilities';
+import range from 'lodash/range';
+
+/**
+ * Default implementation of {@link CameraManager}. Uses {@link ComboControls} to control the camera.
```
larsmoa, line 20: Remove info about ComboControls now that we dont expose that anymore.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -1,604 +1,75 @@
...
-        controls.useScrollTarget = true;
-        controls.zoomToCursor = true;
-        break;
-      case undefined:
-        break;
-      default:
-        assertNever(controlsOptions.mouseWheelAction);
-    }
-  }
+  on(event: 'cameraChange', callback: CameraChangeData): void;
```
larsmoa, line 49: Huh, is `CameraChangeData` the callback definition? Or the data in the event?

I think it should be renamed to `CameraChangedEvent`

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Changed description to: 
` /**
   * Updates internal state of camera manager. Expected to update visual state of the camera
   * as well as it's near and far planes if needed. Called in `requestAnimationFrame`-loop. 
   * Reveal performance affects frequency with which this method is called.
   * @param deltaTime Delta time since last update in seconds.
   * @param boundingBox Global bounding box of the model(s) and any custom objects added to the scene.
   */
  `

examples/src/pages/e2e/cad/DefaultCognite3DViewerTestPageV8.tsx
```diff
@@ -2,15 +2,20 @@
...
 import { Cognite3DTestViewer } from '../Cognite3DTestViewer';
 
 function DefaultCognite3DViewerTestPageV8() {
   const modelUrl = 'primitives_v8';
 
-  return <Cognite3DTestViewer modelUrls={[modelUrl]} />;
+  return <Cognite3DTestViewer modelUrls={[modelUrl]} initializeCallback={(viewer:Cognite3DViewer) => {
+    viewer.cameraManager.setCameraState({position: new THREE.Vector3(30,10,50), 
+      target: new THREE.Vector3()});
+  }} />;
```
larsmoa, line 18:
```suggestion
  return <Cognite3DTestViewer modelUrls={[modelUrl]} initializeCallback={(viewer:Cognite3DViewer) => {
    viewer.cameraManager.setCameraState({position: new THREE.Vector3(30,10,50), 
      target: new THREE.Vector3()});
  }} />;
```
Instead of fixing every single usage of `Cognite3DTestViewer`, just add this into the implementation of `Cognite3DTestViewer`.

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -211,7 +213,10 @@ export class ComboControls extends EventDispatcher {
 
     _spherical.makeSafe();
     _camera.position.setFromSpherical(_spherical).add(_target);
-    _camera.lookAt(this.lookAtViewTarget ? this._viewTarget : _target);
+
+    if (this._rawCameraRotation.equals(new Quaternion()))
```
larsmoa, line 217: Avoid `new` here - it's in the hot path and we should be very careful about allocations. Declare a variable outside the function or see if there is some other way of doing the same check without allocation.

larsmoa: Also, add `{ ... }` for the if-else clauses

larsmoa (changes requested):
See comments

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Renamed

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Also, returning type describes how state looks, so I think it's ok to have such description

viewer/packages/camera-manager/src/ComboControls.ts
Savokr: Create the `isIdentityQuaternion` function and check using it

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -863,4 +879,8 @@ export class ComboControls extends EventDispatcher {
     _panVector.multiplyScalar(distance);
     _targetEnd.add(_panVector);
   };
+
+  private isIdentityQuaternion(q: THREE.Quaternion) {
+    return q.x === 0 && q.y === 0 && q.z === 0 && q.w === 1;
+  }
```
larsmoa, line 885: Cant this be used instead? https://threejs.org/docs/#api/en/math/Quaternion.identity

larsmoa (approved):
Approved, but see single comment

viewer/packages/camera-manager/src/ComboControls.ts
Savokr: This is a method to make some quaternion an identity, not to check whether it is equal to identity.

larsmoa:
> This pull request **introduces 8 alerts** when merging [1df6187](https://github.com/cognitedata/reveal/commit/1df618755984a213f724757f1bd43c976094ef75) into [a2ba4ad](https://github.com/cognitedata/reveal/commit/a2ba4adfaa7a1a62c633b6a49109041b9b39d9c3) - [view on LGTM.com](https://lgtm.com/projects/g/cognitedata/reveal/rev/pr-f89797add37acb0102477171415fcb9fea7e72d9)
> 
> **new alerts:**
> 
> * 7 for Property access on null or undefined
> * 1 for Invocation of non-function

@Savokr

PR #1854, opened 2022-02-16
fix: model format priority

haakonflatval-cognite:
Reveal is supposed to choose geometry file format depending on the priority given in CadModelMetadataRepository. Instead, it chooses the first output in the output list that matches one of the two formats (v8 / v9). 

In practice, this seems to virtually never be an issue since v9 is often (always?) first, but anyway.

### DIFF
--- a/viewer/packages/cad-parsers/src/metadata/CadModelMetadataRepository.test.ts
+++ b/viewer/packages/cad-parsers/src/metadata/CadModelMetadataRepository.test.ts
@@ -1,0 +1,112 @@
+/*!
+ * Copyright 2022 Cognite AS
+ */
+
+import { BlobOutputMetadata } from 'extensions/datasource';
+import { CadModelMetadataRepository } from './CadModelMetadataRepository';
+import { File3dFormat, ModelDataProvider, ModelMetadataProvider } from '@reveal/modeldata-api';
+
+import * as THREE from 'three';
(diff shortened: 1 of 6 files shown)


larsmoa (changes requested):
Great - but expose the issue using a unit test ;)

christjt (approved):

viewer/packages/cad-parsers/src/metadata/CadModelMetadataRepository.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      } else {
+        return v8SceneSectorMetadata;
+      }
+    },
+    getBinaryFile: async () => {
+      return new ArrayBuffer(1);
+    },
+    headers: {}
+  };
+}
```
larsmoa, line 79: Move these to inside `describe(...)`, maybe as `beforeEach-clause`.

larsmoa (approved):
Looks good, but I think the test should be slightly refactored - see comment

viewer/packages/cad-parsers/src/metadata/CadModelMetadataRepository.test.ts
haakonflatval-cognite: Discussed with Lars, agreed to just move the junk to after the `describe` section.

PR #1856, opened 2022-02-16
fix: GltfSectorRepository fails on small cache size

haakonflatval-cognite:
This is because the default number of sectors to remove on cleanup is 0 for small cache sizes, so that the cache remains full after cleanup.


larsmoa (changes requested):
Ensure that cache wont fail even if size is zero.

larsmoa (approved):

PR #1857, opened 2022-02-17
fix: near/far camera plane hiding some peripheral geometry

haakonflatval-cognite:
Fixes https://cognitedata.atlassian.net/browse/REV-320


viewer/core/src/public/migration/Cognite3DModel.ts
```diff
@@ -18,6 +18,7 @@ import { NumericRange } from '@reveal/utilities';
 import { MetricsLogger } from '@reveal/metrics';
 import { CadNode, NodeTransformProvider } from '@reveal/rendering';
 import { NodeAppearance, NodeCollection } from '@reveal/cad-styling';
+import { computeSceneBoundingBox } from '@reveal/utilities/src/computeSceneBoundingBox';
```
christjt, line 21: Expose this in the index file of utilities instead of importing it from .../src

viewer/packages/utilities/src/computeSceneBoundingBox.ts
```diff
@@ -0,0 +1,18 @@
+/*!
+ * Copyright 2022 Cognite AS
+ */
+
+import { SectorMetadata } from '@reveal/cad-parsers';
```
christjt, line 5: Utilities should not have a dependency to cad-parsers. Consume a list of bounds in `computeSceneBoundingBox` instead.

christjt (changes requested):

christjt (approved):
LGTM

PR #1858, opened 2022-02-18
improvement: use newer geometrybounds

haakonflatval-cognite:


viewer/packages/cad-geometry-loaders/src/sector/culling/ByScreenSizeSectorCuller.ts
```diff
@@ -91,21 +89,28 @@ function takeSectorsWithinBudget(
...
+  transformedBounds: new THREE.Box3()
+};
+
 function sortSectorsByPriority(
   modelsAndCandidateSectors: Map<CadModelMetadata, V9SectorMetadata[]>,
   weightFunctions: WeightFunctionsHelper,
-  transformedBounds: THREE.Box3,
   prioritizedAreas: PrioritizedArea[]
 ): { model: CadModelMetadata; sectorId: number; priority: number }[] {
+  const { transformedBounds } = _sortSectorsByPriorityVars;
```
christjt, line 101: Try to find a better name for `_sortSectorsByPriorityVars.` It sounds like a function. Also, why does this need to be global?

christjt (approved):

PR #1861, opened 2022-02-21
fix: primitives not rendering correctly on iOS

pramodcog:
Only occurs when WebGL is backed by Metal on iOS


viewer/packages/rendering/src/glsl/sector/primitives/cone.frag
```diff
@@ -14,12 +13,15 @@
...
 uniform sampler2D colorDataTexture;
-uniform sampler2D overrideVisibilityPerTreeIndex;
 uniform sampler2D matCapTexture;
 
 uniform vec2 treeIndexTextureSize;
 
 uniform mat4 projectionMatrix;
+uniform int renderMode;
+
+//Fix for iOS primitives not getting depth value
```
christjt, line 23:
```suggestion
//Fix for iOS primitives not getting depth value if included before all uniform are declared
```

christjt (approved):
Tiny suggestion, such that comments match. But other than that LGTM

viewer/packages/rendering/src/glsl/sector/primitives/eccentricCone.frag
```diff
@@ -13,12 +12,15 @@
...
 uniform sampler2D colorDataTexture;
-uniform sampler2D overrideVisibilityPerTreeIndex;
 uniform sampler2D matCapTexture;
 
 uniform vec2 treeIndexTextureSize;
 
 uniform mat4 projectionMatrix;
+uniform int renderMode;
+
+//Fix for iOS primitives not getting depth value if included before all uniform are declared
```
larsmoa, line 22:
```suggestion
// Note! Must be placed after all uniforms in order for this to work on iOS (REV-xyz)
```
Comment doesn't say anything about what can be done to this line and what cannot be done. Also add/reference a Jira issue with more details on the issue.

larsmoa (changes requested):

larsmoa (approved):

Issue #1851, opened 2022-02-14
Render size is possibly set wrong in Cognite3DViewer resizeIfNecessary


hkot1991:
I was looking into your resizeIfNecessary method in Cognite3DViewer and I think I found a bug.
According to threejs render documentation the renderers setSize method already handles the device pixel ratio: 
![image](https://user-images.githubusercontent.com/61226854/153819623-853aeffe-0668-434f-a0dd-a61a74eff3ff.png)

So the width and height passed to setSize in resizeIfNecessary will be multipled AGAIN with the device ratio inside setSize.
Making the render size way higher than it needs to be for cases where the device ratio is not 1. 

Looks to me like a bug, but please let me know if I'm wrong 🙏

christjt:
Thank you for reporting, @hkot1991. We will take a look 👍

christjt:
Been looking a little more into this and it seems like you are correct. Oddly enough, this seems NOT to be the case for when setting the size of render targets. So I believe most of the rendering in Reveal will not be affected by this (other than the very last render pass that gets rendered to the canvas). In any case we are tracking the issue internally and should produce a fix for this soon.

christjt:
Attempt at fix in: https://github.com/cognitedata/reveal/pull/1874. Just reopen this if you are still experiencing issues. And again, thanks for reporting the issue.

PR #1874, opened 2022-02-22
fix: incorrect usage of device pixel ratio

christjt:
Fix for #1851.

The issue is that when setting the size of WebGLRenderer, it will scale it by device pixel ratio. So it should not also be scaled manually. In practice it doesn't cause any issue since we don't ever set the device pixel ratio, but if the user provides their own WebGLRenderer, with DPR set, then the size will be scaled by DPR twice.


Savokr (approved):
Looks good 🤌

PR #1890, opened 2022-02-23
fix: cache result of "compute bounds of most geometry" to force deterministic results

christjt:
Fixes issue reported by Aize where certain models would "bounce around" when clicking fit camera to model. Reason for this is that `skmeans`, which we use to compute the bounding box of the model without noise, is not deterministic. The rubberband fix for this (since sadly it doesn't support seed https://github.com/solzimer/skmeans/issues/18) is to only compute once and cache the result.


viewer/packages/cad-parsers/src/utilities/SectorScene.ts
```diff
@@ -66,6 +67,28 @@ export class SectorSceneImpl implements SectorScene {
...
+    const accepted: SectorMetadata[] = [];
+    traverseDepthFirst(this.root, x => {
+      if (frustum.intersectsBox(x.subtreeBoundingBox)) {
+        accepted.push(x);
+        return true;
+      }
+      return false;
+    });
+    return accepted;
+  }
```
haakonflatval-cognite, line 89: Why is this moved?

haakonflatval-cognite (approved):
Looks good - just a comment

PR #1897, opened 2022-02-23
fix: Types public export and proper camera resizing for creation of custom CameraManager

Savokr:


viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1201,16 +1200,9 @@ export class Cognite3DViewer {
...
-    camera.updateProjectionMatrix();
-  } else if (camera instanceof THREE.OrthographicCamera) {
-    camera.left = -width;
-    camera.right = width;
-    camera.top = height;
-    camera.bottom = -height;
-  }
+function adjustCamera(camera: THREE.PerspectiveCamera, width: number, height: number) {
+  camera.aspect = width / height;
+  camera.updateProjectionMatrix();
```
Savokr, line 1205: This fixes proper adjustment of camera aspect ratio. Wasn't investigating too much into why this if statement is not working as expected. Camera is assigned a proper value but for some reason `camera instanceof THREE.PerspectiveCamera` is not true in that case. Since we are not supporting OrthographicCamera and `camera` property on `Cognite3DViewer` has `THREE.PerspectiveCamera` type, I thought removing this check should be ok.

christjt (approved):
LGTM

PR #1902, opened 2022-02-24
fix: sector loader example not handling bounds update / rename

christjt:


viewer/packages/sector-loader/app/index.ts
```diff
@@ -124,15 +124,15 @@ async function loadSectors(
...
   const sector = sceneJson.sectors[0];
-  sector.bounds = new THREE.Box3(sector.boundingBox.min, sector.boundingBox.max);
+  sector.subtreeBoundingBox = new THREE.Box3(sector.boundingBox.min, sector.boundingBox.max);
 
   const model = new THREE.Group();
 
   const modelIdentifier = output!.blobId.toString();
 
   await Promise.all(
     sceneJson.sectors.map(async (sector: any) => {
```
larsmoa, line 134: Can we use a concrete type instead of any?

haakonflatval-cognite:
I didn't really mean to push this, but I'll leave it here anyway. This commit addresses the type-unsafety Lars pointed out, while it rewrites what Christopher did.

larsmoa (approved):

PR #1918, opened 2022-02-25
Fix memory leaks

hkot1991:
I been looking into the memory usage of the viewer, and it looks like there are some listeners that were never removed, this made it so the cognite3DViewer was never removed from memory

I tested this with the example code, and used chrome developer tools snapshots
Upon re-entering an example and returning to main menu there would be multiple Cognite3dViewers in memory

Thus I have fixed the lingerring listeners I found while testing
There might still be more


hkot1991:
Any news on when this will be merged? Seems that the publish documentation is failing, probably because its a merge from a fork

Savokr:
Hi Hanne, thanks for the PR, I am working on it but it seems like memory leak issue you pointed out is not solved by this PR. From my investigation it is more about deleting model data from memory rather than unused event listeners. I am trying to fix the issue as it is quite important when someone wants to load/unload models because right now removing the model is not emptying the memory. 
Regarding documentation part in CI, we don't have a good option way to handle external PRs right now, so I guess the only solution is that I will create a new PR and cherry-pick your commits there, this should preserve the author but PR will be with my permissions.

christjt:
Closing this in favour of #2025. Thanks for doing the initial work (changes here are merged into the other branch)

PR #1948, opened 2022-03-04
Updated sdk-core peer dependency

hkot1991:
Looks like the dev dependency was updated but not the peer depencency. This pr fixes this


larsmoa (approved):

hkot1991:
@larsmoa I think you need to merge  without expecting the build to complete. It wont publish the documentation because its from a fork 😞 

Or you have to create an own pr with this change

PR #1949, opened 2022-03-04
chore: cleanup imports and exports

haakonflatval-cognite:
The changes done here were found in the following way
- Using `npx depcheck` in each package and in core. It prints out any unused dependency in the package.json and imports from packages not mentioned in package.json. It was used to find packages that unneededly depended on other packages, or when a package imported from itself. Using Bash, you can iterate through all packages by `cd`-ing to  `packages` and running 
```
for a in */; do echo $a; pushd $a; npx depcheck; popd; done
```
NOTE: It will report all external packages as missing because it does not read the package.json in `viewer`. Search for `@reveal` in the output to find relevant entries. It should be run again after fixing errors, as it only seems to report one instance of each missing package for each module. 

- Using the custom export checker script at https://github.com/cognitedata/reveal-export-checker. It parses the `index.ts` files for all packages and then reads through all imports in the packages to match up with the provided exports. It does not consider exports from `core` (and thus messes up results for `tools` which imports a lot from `core`), since it can't really decide what should be publicly exported and not. 

I have thus not touched anything that is exported publicly, meaning the index files in core and tools. In addition, I have not changed any exports from utilities, as we may want those to be visible in any case.


haakonflatval-cognite:
Additional question for reviewer: I have moved `createDetermineSectorInput` from `test-utilities` to the `cad-geometry-loaders` package, even though it is only used by tests. The tests that uses it are all in that package. Is this a reasonable thing to do? Should it be named something else (add a `.test` at the end etc.)?

christjt (approved):
Nice utility for checking deps! As for the `createDetermineSectorInput`, ideally all tests would be located in [package]/tests/ folder and then any utility specific to those tests can go in there. Nice to move the utility into the package if the package is the only thing that uses it. But as you have noticed, it is now unclear that it is simply a test utility and not part of the business logic. Having everything in a /tests/ folder would remedy that.

larsmoa:
Merge @haakonflatval-cognite ?

PR #1950, opened 2022-03-04
refactor: move primitive generation to cad parsers

haakonflatval-cognite:


haakonflatval-cognite:
Got some new problems with dependencies going the wrong way etc.. Will postpone work on this to when we have a more defined architecture overview.

PR #1959, opened 2022-03-11
feat: add point cloud classification UI

larsmoa:
Adds UI for filtering point clouds by classification to the example.

Note that filtering doesn't actually work. This is due to a bug in classification filtering thats not caused by this PR.


haakonflatval-cognite (approved):
Looks alright, although it's a bit hard to verify at the moment :sweat_smile:

Issue #1971, opened 2022-03-16
Problems with ClosedGeneralCylinder


vegasten:
We have some problems using the ClosedGeneralCylinder and i3df, and are experiencing some graphical issues. The bends in the pipes consists of multiple cylinders with cuts or slopes, but it seems like something is going wrong somewhere when rendering. We have tested the code on our side, and think we're doing everything correctly (even though it is hard to be sure). Have you experienced any problems with the ClosedGeneralCylinder?

The model in this screenshots is Huldra, which is available to the public: 
https://data.equinor.com/dataset/Huldra

![HVACBendsGlitch](https://user-images.githubusercontent.com/14194541/158583088-2aac4d5b-4320-420a-b172-fcdcda0de0a6.png)

The goal we want to achieve: 
![target](https://user-images.githubusercontent.com/14194541/158594806-c4dcb597-7d69-4f81-9108-db07288befb9.png)

One extra observation is that it seems like cylinders with radius set to 1 meter is rendered correctly. Correctly in the sense that it is rendering cylinders with 1 meter radius, not the cylinders we ultimately want to render:
![HVACBendsWIthRadius1](https://user-images.githubusercontent.com/14194541/158583658-132db717-08ba-4ba7-b0d5-f2e1019aa379.png)

christjt:
@vegasten I want to give the Huldra model a run. Is it OK for us to use for test purposes?

vegasten:
@christjt I need to look into how much I am allowed to share. Do you have an e-mail I can contact you on?

christjt:
Sure! christopher.tannum (at) cognite.com

christjt:
@vegasten Sorry this has taken so much time to verify / fix (we are focusing mainly on the gltf outputs lately). Do you have coordinates within the model for this exact case? (we have the Huldra model to test with).

espenrl:
I believe I found the solution. The following line

https://github.com/cognitedata/reveal/blob/0c19bcb20d13c658f634259e6c197887089a0533/i3df/src/renderables/cylinder.rs#L203

should be

```
let dist_from_b_to_ext_b = 1 + cylinder.radius * f32::tan(cylinder.slope_b);
```

The reason it works fine for `radius` equal to `1.0` is because then you get the same result.

PR #1972, opened 2022-03-16
Edit math for general closed cylinder

vegasten:
Regarding issue #1971 

I did some edits to the math for creating a general cylinder. I am really not sure about the edits, because they look a bit suspicious, but it fixed the rendering problems on my end.


christjt:
Thanks for contributing, @vegasten. I'll take a look and test with some test-models we have to make sure it doesn't break anything.

christjt:
Closing this due to Reveal dropping support for i3d/f3d based output from 4.0. If this problem persists with the GLTF outputs, please re-open.

PR #1982, opened 2022-03-22
feat: new render pipeline system

christjt:
Known things missing:

- SSAO and multisample is also done for mobile
-  A "blank" renderpass will still be executed for (back, ghost inFront and SSAO) regardless if it actually has any objects that get affected by this.

Known issue:
- macOS on intel with metal backend has rendering artifacts (issues with capturing depth).


viewer/packages/rendering/src/pipeline-executors/BasicPipelineExecutor.ts
```diff
@@ -0,0 +1,23 @@
+/*!
+ * Copyright 2022 Cognite AS
+ */
+
+import * as THREE from 'three';
+import { PipelineExecutor } from '../PipelineExecutor';
+import { RenderPipelineProvider } from '../RenderPipelineProvider';
+
+export class BasicPipelineExecutor implements PipelineExecutor {
```
larsmoa, line 9: I think a unit test would be good, even if it simple (to encourage any future changes to add tests...)

viewer/packages/rendering/src/pipeline-executors/StepPipelineExecutor.ts
```diff
@@ -0,0 +1,57 @@
+/*!
+ * Copyright 2022 Cognite AS
+ */
+
+import * as THREE from 'three';
+import { PipelineExecutor } from '../PipelineExecutor';
+import { RenderPipelineProvider } from '../RenderPipelineProvider';
+import { GpuTimer } from '../utilities/GpuTimer';
+
+export class StepPipelineExecutor implements PipelineExecutor {
```
larsmoa, line 10: Unit test

viewer/packages/rendering/src/render-pipelines/CadGeometryRenderPipeline.ts
```diff
@@ -0,0 +1,125 @@
...
+    renderer.setRenderTarget(this._cadGeometryRenderTargets.back);
+    yield this._cadGeometryRenderPasses.back;
+
+    renderer.setRenderTarget(this._cadGeometryRenderTargets.ghost);
+    yield this._cadGeometryRenderPasses.ghost;
+
+    renderer.setRenderTarget(this._cadGeometryRenderTargets.inFront);
+    yield this._cadGeometryRenderPasses.inFront;
+
+    this.pipelineTearDown(renderer);
```
larsmoa, line 65: Should this be in a `finally { ... }`-clause to ensure cleanup even if shit hits the fan?

viewer/packages/rendering/src/render-pipelines/CadGeometryRenderPipeline.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public dispose(): void {}
+
+  private pipelineSetup(renderer: THREE.WebGLRenderer) {
+    this._currentRendererState = {
+      autoClear: renderer.autoClear,
+      clearColor: renderer.getClearColor(new THREE.Color()),
+      clearAlpha: renderer.getClearAlpha()
+    };
```
larsmoa, line 75: Are you aware of `WebGLRendererStateHelper` that helps to store/restore state for WebGLRenerer?

viewer/packages/rendering/src/render-pipelines/DefaultRenderPipeline.ts
```diff
@@ -0,0 +1,190 @@
...
+  private readonly _renderTargetData: RenderTargetData;
+  private readonly _cadModels: IdentifiedModel[];
+  private readonly _customObjects: THREE.Object3D[];
+  private readonly _autoResizeOutputTarget: boolean;
+  private readonly _outputRenderTarget: THREE.WebGLRenderTarget;
+  private _currentRendererState: {
+    autoClear: boolean;
+    clearColor: THREE.Color;
+    clearAlpha: number;
+  };
```
larsmoa, line 30: `WebGLRendererStateHelper`?

viewer/packages/tools/src/AxisView/AxisViewTool.ts
```diff
@@ -6,6 +6,7 @@ import * as THREE from 'three';
 import cloneDeep from 'lodash/cloneDeep';
```
larsmoa, line 6: Why changes to AxisViewTool?

viewer/packages/utilities/src/counterMap.ts
```diff
@@ -16,7 +14,10 @@ export function incrementOrInsertIndex(indexMap: Map<number, number>, index: num
 
 export function decrementOrDeleteIndex(indexMap: Map<number, number>, index: number): void {
   const count = indexMap.get(index);
-  assert(count !== undefined);
+
+  if (count === undefined) {
+    return;
+  }
```
larsmoa, line 20: Uhm. Can/should this happen?

larsmoa (changes requested):
Huge leap forward! See comments. In general, a few more unit test never hurts ;) Also, are there visual tests we should add as a result of this PR?

viewer/packages/utilities/src/counterMap.ts
christjt: Yes it can, we have some state where the same sector gets discarded twice, or that a sector loads, and then before finish loading gets discarded which then causes this effect.

viewer/package.json
```diff
@@ -94,7 +94,7 @@
     "file-loader": "6.2.0",
     "gl": "5.0.0",
     "glslify-loader": "2.0.0",
-    "html-webpack-plugin": "5.5.0",
+    "html-webpack-plugin": "4.5.2",
```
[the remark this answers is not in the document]
christjt: the plugin for 5.5.0 doesn't actually work and expects webpack 5 (seems like they are following the semver).

viewer/packages/tools/src/AxisView/AxisViewTool.ts
christjt: Converted to be a normal custom object, no need for a UI object backdoor.

viewer/packages/rendering/src/pipeline-executors/StepPipelineExecutor.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      if (count === this._numSteps) {
+        this._renderer.setRenderTarget(null);
+        renderPass.render(this._renderer, camera);
+        break;
+      }
+      renderPass.render(this._renderer, camera);
+    }
+
+    this._gpuTimer.end();
```
haakonflatval-cognite, line 46:
```suggestion
      if (count === this._numSteps) {
        break;
      }
      renderPass.render(this._renderer, camera);
    }

    this._renderer.setRenderTarget(null);
    renderPass.render(this._renderer, camera);
    
    this._gpuTimer.end();
```

haakonflatval-cognite (approved):
Okay, then.

viewer/packages/modeldata-api/src/LocalModelMetadataProvider.ts
```diff
@@ -15,6 +15,9 @@ export class LocalModelMetadataProvider implements ModelMetadataProvider {
     if (!(modelIdentifier instanceof LocalModelIdentifier)) {
       throw new Error(`Model must be a ${LocalModelIdentifier.name}, but got ${modelIdentifier.toString()}`);
     }
+    if (modelIdentifier.localPath.startsWith('http')) {
```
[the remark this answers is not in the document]
christjt: @larsmoa Maybe you can answer this, as I believe you added it?

viewer/packages/rendering/src/RenderPass.ts
```diff
@@ -0,0 +1,9 @@
+/*!
+ * Copyright 2022 Cognite AS
+ */
+
+import * as THREE from 'three';
+
+export interface RenderPass {
```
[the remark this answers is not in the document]
christjt: Personally I would prefer I-prefix for interfaces, though it is not a typescript convention to do this. I think we should either do it for everything or nothing. In either case I don't want to end up in a situation some are and some are not. So I think I will leave this as is, and then, if we all agree, do a cleanup of all interfaces in the repo.

viewer/packages/rendering/src/pipeline-executors/StepPipelineExecutor.ts
christjt: This is not exactly equivalent as far as I can see. If the number of steps in the pipeline is less the the step size set, then the same render pass would be rendered twice with this code change.

viewer/core/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1164,7 +1169,11 @@ export class Cognite3DViewer {
         combinedBbox.union(bbox);
       }
     });
+
     this._extraObjects.forEach(obj => {
+      if (obj instanceof CognitePointCloudModel) {
```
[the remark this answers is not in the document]
christjt: obj is of type Object3D, so `.boundingBox` is not available. Kinda what setFromObject is trying to "fix". And even if we got that working, we would then need to traverse all the objects, as the bounding box is not guaranteed to cover children (parent could have bounding box size smaller than children theoretically).

viewer/packages/rendering/src/pipeline-executors/StepPipelineExecutor.ts
christjt: Unresolve if you disagree.

viewer/packages/modeldata-api/src/LocalModelMetadataProvider.ts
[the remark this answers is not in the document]
larsmoa: This was something I added in order to support local point clouds while testing the PR. Should probably been a separate one

viewer/packages/rendering/src/glsl/post-processing/blit.frag
```diff
@@ -0,0 +1,73 @@
...
+#if defined(OUTLINE)
+uniform sampler2D tOutlineColors;
+#endif
+
+in vec2 vUv;
+
+out vec4 fragColor;
+
+#if defined(SSAO_BLUR) 
+#pragma glslify: import('./gaussian-blur.glsl')
```
pramodcog, line 26: Can't have combine with with respective `uniform` declaration?

christjt: This would brake the "fix" you did since uniforms apparently needs to come before any function definition on iOS. This may of course have been fixed later, but even then it is cleaner to have all the uniforms defined before functions.

christjt: Also the checks here are only compiler checks, so it should not effect runtime performance.

viewer/packages/rendering/src/render-pipeline-providers/DefaultRenderPipelineProvider.ts
```diff
@@ -0,0 +1,186 @@
...
+  private readonly _blitToScreenMesh: THREE.Mesh;
+  private _rendererStateHelper: WebGLRendererStateHelper;
+
+  set renderOptions(renderOptions: RenderOptions) {
+    const { ssaoRenderParameters } = renderOptions;
+    this._ssaoPass.ssaoParameters = ssaoRenderParameters ?? defaultRenderOptions.ssaoRenderParameters;
+
+    const shouldAddFxaa = AntiAliasingMode[renderOptions.antiAliasing] === AntiAliasingMode[AntiAliasingMode.FXAA];
+
+    if (shouldAddFxaa === this._blitToScreenMaterial.defines.FXAA ?? false) {
```
larsmoa, line 42: Ehhhm, this was hard to read.

viewer/packages/utilities/src/SceneHandler.ts
```diff
@@ -0,0 +1,76 @@
...
+    this._customObjects.push(object);
+    this._scene.add(object);
+  }
+
+  public removeCustomObject(object: THREE.Object3D): void {
+    this.scene.remove(object);
+    remove(this._customObjects, object);
+  }
+
+  public dispose(): void {
```
larsmoa, line 53: I think this should be unit tested

larsmoa (approved):
Consider what of my comments should be addressed. Looks good!

PR #1987, opened 2022-03-23
fix: documentation version1.x by reverting threejs version

pramodcog:


christjt (approved):
Thanks, just double check that v1 doc works in the preview PR before merging.

PR #2019, opened 2022-03-31
fix: share() so `consumedSectorObservable` works

hayer:
Currently using the `Observable<ConsumedSector>` from `CadModelUpdateHandler` seems to make it _not work_ .. in lack of better description. The reason seems to be that it is not shared properly.

Proposing this as a fix.


christjt:
Hey, what exactly doesn't work? When getting the consumed sector observable, we do share it.

```
  consumedSectorObservable(): Observable<ConsumedSector> {
    return this._updateObservable.pipe(share());
  }
```

j-bjorne:
How are you accessing this Observable?

With the method

```
  consumedSectorObservable(): Observable<ConsumedSector> {
    return this._updateObservable.pipe(share());
  }
```

it shares this pipe the way you propose it. Do you have any code or snippet of yours to see why exactly it fails?

hayer:
```
        const revealManager = (<any>global3DViewer).revealManager;
        const cadManager = revealManager._cadManager;
        const cadModelUpdateHandler = cadManager._cadModelUpdateHandler as CadModelUpdateHandler;

        const consumedSectorObservable = cadModelUpdateHandler.consumedSectorObservable().subscribe(console.log);
```

Using `consumedSectorObservable()` without this change seems to make loading hang..?

christjt:
We don't see the need for this change and cannot reproduce the example. If you subscribe to the observer, then it seems to pass data just fine. Please let me know of the case you are trying to solve if this is a blocker.

PR #2021, opened 2022-03-31
fix: Docs reveal version for 2.x and controls modes example 

Savokr:
Fixed proper reveal version for 2.x documentation and controls mode example.


documentation/yarn.lock
```diff
@@ -10633,10 +10610,10 @@ rxjs@7.3.0, rxjs@^7.1.0:
   dependencies:
     tslib "~2.1.0"
 
-rxjs@7.5.5:
-  version "7.5.5"
-  resolved "https://registry.yarnpkg.com/rxjs/-/rxjs-7.5.5.tgz#2ebad89af0f560f460ad5cc4213219e1f7dd4e9f"
-  integrity sha512-sy+H0pQofO95VDmFLzyaw9xNJU4KTRSwQIGM6+iG3SypAtCiLDzpeG8sJrNCWn2Up9km+KhkvTdbkrdy+yzZdw==
+rxjs@7.5.4:
```
pramodcog, line 10613: Why the version got downgraded?

pramodcog (changes requested):

documentation/yarn.lock
Savokr: Not sure why, it was done by yarn automatically

larsmoa: Revert changes to yarn.lock to be sure... @Savokr

Savokr: @larsmoa It was updated specifically to update reveal version because previously it was 2.2.2 instead of 2.3.0

pramodcog: Revert only the `rxjs` changes in the yarn.lock

PR #2023, opened 2022-04-01
fix: Point cloud rendering issue in docs

pramodcog:
Will only work for 3.x once we release version 3.0.0 of Reveal.

_Note: 2.x docs will still not work for Point Cloud examples_


larsmoa (changes requested):
This doesn't work for 2.x and 3.x

larsmoa (approved):

PR #2024, opened 2022-04-01
fix: make nuts look flat

haakonflatval-cognite:
AKA make them use normal instance mesh shaders instead of the primitive specific shaders


larsmoa:
I think we should add a task for suggestions from @christjt, @haakonflatval-cognite  - and not do it right now.

larsmoa (approved):

christjt (approved):

haakonflatval-cognite:
The scope of this PR is extended. It now:
- Removes the `primitives/instanced.*` shaders in favor of the `instancedMesh.*` shaders
- Uses custom-built box with 8 vertices instead of THREE's builtin with 24 vertices. (We can now use only 8 vertices as we don't need to store normals per-vertex for the new shaders, but compute them with df/dx etc. in the fragment shader.)

PR #2025, opened 2022-04-01
fix: Memory leaking issues when model is removed from the scene

Savokr:
Unfortunately for now is a "bad" version where references to CadNode are not properly removed from CadModelUpdateHandler class.


larsmoa (changes requested):

viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
```diff
@@ -178,7 +180,8 @@ export class CadModelUpdateHandler {
             array.push(model);
             return array;
           case 'remove':
-            return array.filter(x => x.cadModelMetadata.modelIdentifier !== model.cadModelMetadata.modelIdentifier);
+            array = array.filter(x => x.cadModelMetadata.modelIdentifier !== model.cadModelMetadata.modelIdentifier);
+            return array;
```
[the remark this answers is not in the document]
Savokr: Because I was thinking that it could be a reason for unwanted reference, as old ray still holds model value inside. But probably it is erased in any case and replaced with new one, right?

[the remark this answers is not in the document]
Savokr: Also, in case of new model we are modifying passed array, not creating a new one, so we should probably do the same for both.

viewer/packages/camera-manager/src/ComboControls.ts
```diff
@@ -129,9 +129,9 @@ export class ComboControls extends EventDispatcher {
     window.addEventListener('pointerdown', this.onFocusChanged);
 
     this.dispose = () => {
```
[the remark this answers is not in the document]
Savokr: Removing listeners on disposal is quite important, so I think we can have changes in DefaultCameraManager. I will test tween animation changes but part with camera change is good for sure.

viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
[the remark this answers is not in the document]
larsmoa: Hmm, I dont think that would happen - the array will go out of scope. Also, you are not modifying the original `array` - only the local reference to it (i.e. you change the reference, not the content).

viewer/packages/camera-manager/src/DefaultCameraManager.ts
```diff
@@ -97,13 +97,15 @@ export class DefaultCameraManager implements CameraManager {
...
 
-    this._controls.addEventListener('cameraChange', event => {
-      const { position, target } = event.camera;
-      this._events.cameraChange.fire(position.clone(), target.clone());
-      this._nearAndFarNeedsUpdate = true;
-    });
+    this._controls.addEventListener('cameraChange', this.cameraChange);
   }
 
+  cameraChange = (event: any): void => {
```
larsmoa, line 103: Use plain functions. Also, should be private? Get rid of `any`.

viewer/packages/camera-manager/src/DefaultCameraManager.ts
```diff
@@ -97,13 +97,15 @@ export class DefaultCameraManager implements CameraManager {
     this._controls = new ComboControls(this._camera, domElement);
     this._controls.minZoomDistance = DefaultCameraManager.DefaultMinZoomDistance;
 
-    this._controls.addEventListener('cameraChange', event => {
-      const { position, target } = event.camera;
-      this._events.cameraChange.fire(position.clone(), target.clone());
-      this._nearAndFarNeedsUpdate = true;
-    });
+    this._controls.addEventListener('cameraChange', this.cameraChange);
```
larsmoa, line 100: I'm confused why using `this.cameraChange` as a listener works, as a it's unbound. Any idea?

viewer/packages/camera-manager/src/DefaultCameraManager.ts
```diff
@@ -406,34 +412,32 @@ export class DefaultCameraManager implements CameraManager {
...
   }
 
-  private createTweenAnimation(
-    from: any,
-    to: any,
-    duration: number
-  ): { tween: TWEEN.Tween; stopTween: (event: Event) => void } {
+  private stopTween: any;
+
+  private createTweenAnimation(from: any, to: any, duration: number): { tween: TWEEN.Tween } {
```
larsmoa, line 417: Can `any` be avoided?

viewer/core/src/public/migration/RenderController.ts
```diff
@@ -24,9 +24,7 @@ export default class RenderController {
     this._lastCameraRotation = new THREE.Euler();
     this._lastCameraZoom = 0;
 
-    window.addEventListener('focus', () => {
-      this.redraw(); // force rendering on focus
-    });
+    window.addEventListener('focus', this.redraw);
```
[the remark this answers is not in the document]
Savokr: Added binding of context for `redraw`, so should work

viewer/packages/rendering/src/CadMaterialManager.ts
```diff
@@ -114,6 +114,17 @@ export class CadMaterialManager {
     this.updateClippingPlanesForModel(modelIdentifier);
   }
 
+  deleteModelData(modelIdentifier: string): void {
```
larsmoa, line 117: Rename to `removeModelMaterials`

larsmoa (changes requested):

viewer/packages/camera-manager/src/DefaultCameraManager.ts
christjt, line 103: Is there a type we can use here instead of any?

christjt (changes requested):

viewer/packages/camera-manager/src/DefaultCameraManager.ts
Savokr: It's bounded because of lambda function declaration

viewer/packages/rendering/src/GeometryBatchingManager.ts
```diff
@@ -40,6 +40,22 @@ export class GeometryBatchingManager {
     this._sectorMap = new Map();
   }
 
+  public dispose(): void {
+    for (const sectorId of this._sectorMap.keys()) {
+      this.removeSectorBatches(sectorId);
+    }
+
+    for (const mesh of this._batchedGeometriesGroup.children) {
+      if (mesh instanceof THREE.Mesh || mesh instanceof THREE.InstancedMesh) {
```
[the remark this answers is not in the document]
Savokr: Oh, makes sense

viewer/packages/camera-manager/src/DefaultCameraManager.ts
Savokr: Will be in a new PR

viewer/packages/camera-manager/src/DefaultCameraManager.ts
Savokr: Will be in a new PR

Savokr: Will be in a new PR

viewer/packages/rendering/src/CadMaterialManager.ts
```diff
@@ -114,6 +114,17 @@ export class CadMaterialManager {
...
+    const modelData = this.materialsMap.get(modelIdentifier);
+    for (const mat of Object.values(modelData.materials)) {
```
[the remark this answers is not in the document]
Savokr: It doesn't work directly and also the same pattern is used in`materials.ts`

christjt (approved):

larsmoa (approved):

PR #2031, opened 2022-04-04
fix: let threejs update shared uniforms

christjt:
Fixes low detail not properly being rendered for v8. Also lets threejs handle all common uniform updates (model, view, matrix, etc.) which should also be a performance boost (though probably very minor).


pramodcog (approved):
LGTM

PR #2036, opened 2022-04-05
fix: continuous loading now only works if model is v9

christjt:


larsmoa (approved):
Approved, but see suggestion.

PR #2045, opened 2022-04-07
feat: Added support for clicked button data for Cognite3DViewer.on('click',...) method

Savokr:
Support for clicked button data for external click handling using Cognite3DViewer. Also includes small refactoring of EventDelegates so that they are in one place because before there were 3 exactly the same version of PointerEventDelegate in different places. Now all events related types are located in utilities/events and can be imported from there by any internal package.


examples/src/pages/Geomap.tsx
```diff
@@ -177,7 +177,7 @@ export function Geomap() {
       const selectedSet = new TreeIndexNodeCollection([]);
 
       new AxisViewTool(viewer);
-      viewer.on('click', async event => {
+      viewer.on('click', async (event: { offsetX: any; offsetY: any; }) => {
```
christjt, line 180: remove this

viewer/core/src/public/migration/types.ts
```diff
@@ -204,31 +204,26 @@ export { CameraConfiguration } from '@reveal/utilities';
...
  * @module @cognite/reveal
  * @see {@link Cognite3DViewer.on}.
  */
-export type SceneRenderedDelegate = (event: {
-  frameNumber: number;
-  renderTime: number;
-  renderer: THREE.WebGLRenderer;
-  camera: THREE.PerspectiveCamera;
-}) => void;
+export { SceneRenderedDelegate } from '@reveal/utilities';
```
christjt, line 226: Still don't understand why the core package is exporting things from utilities.

viewer/packages/camera-manager/src/CameraManager.ts
```diff
@@ -2,7 +2,8 @@
  * Copyright 2022 Cognite AS
  */
 
-import { CameraChangeDelegate, CameraState } from './types';
+import { CameraState } from './types';
+import { CameraChangeDelegate } from '@reveal/utilities';
```
christjt, line 6: It seems very strange to me that the camera change delegate is a part of utilities.

viewer/packages/utilities/src/events/types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Delegate for pointer events.
+ */
+export type PointerEventDelegate = (event: { offsetX: number; offsetY: number; button?: number }) => void;
+
+/**
+ * Delegate for camera update events.
+ */
+export type CameraChangeDelegate = (position: THREE.Vector3, target: THREE.Vector3) => void;
```
christjt, line 13: I feel like this belongs in the camera package. I am open for arguments though.

christjt (changes requested):
Overall looks good. Still not convinced that the `CameraChangeDelegate` should live in utilities.

viewer/core/src/public/migration/types.ts
Savokr: Now exported from top level in viewer

viewer/packages/camera-manager/src/CameraManager.ts
Savokr: Move to CameraManager package

PR #2074, opened 2022-04-19
fix: migration & geomap examples for implicit conversion error

pramodcog:


larsmoa:
I dont think this is a proper fix - the proper thing would be to expose the correct types here, right? Is this a problem in 3.0?

pramodcog:
Closing as typing issue was cleared when yarn run build is used

PR #2080, opened 2022-04-20
fix: add tsc alias as a webpack build hook

christjt:
Fixes [REV-352](https://cognitedata.atlassian.net/jira/software/c/projects/REV/issues/REV-352).


pramodcog (approved):
It works for me

PR #2127, opened 2022-05-12
fix: typo in unpackSectorVisibility() causing weight to only depend x position

larsmoa:
Contribution by @vero-so / Equinor


viewer/packages/cad-geometry-loaders/src/sector/culling/OrderSectorsByVisibilityCoverage.ts
```diff
@@ -442,7 +442,7 @@ export class GpuOrderSectorsByVisibilityCoverage implements OrderSectorsByVisibi
         const b = renderTargetBuffer[4 * i + 2];
         const distance = renderTargetBuffer[4 * i + 3]; // Distance stored in alpha
         if (r !== 255 || g !== 255 || b !== 255) {
-          const rx = (y - halfWidth) / halfWidth;
+          const rx = (x - halfWidth) / halfWidth;
 
           const sectorIdWithOffset = b + g * 255 + r * 255 * 255;
```
haakonflatval-cognite, line 447:
```suggestion
          const sectorIdWithOffset = b + g * 256 + r * 256 * 256;
```
Does this make more sense?

larsmoa: Does it? I dont think so - when we write 1.0 to e.g. `B` this means 255, not 256 (?) @haakonflatval-cognite

haakonflatval-cognite: Because if b = 255, g = 0, r = 0, we will end up with the same offset as if b = 0, g = 1, r = 0, using the old formula @larsmoa

vero-so: Yes, the regular conversion RGB to a single int is done via bitshifts R<<16 + G<<8 + B, so that would be the 256 and 256*256, as you say. But I guessed it works, since it is consistent with packIntToColor where you don't use the bit-shift (mod256), but mod255. Thus a channel max is 254, not 255. In the worst case, you are able to store 255^3+1 instead of 256^3-1 IDs, but maybe that's not a problem?
```
vec3 packIntToColor(float number) {
    float r = floor(number / (255.0 * 255.0)) / 255.0;
    float g = mod(floor(number / 255.0), 255.0) / 255.0;
    float b = mod(number, 255.0) / 255.0;
    return vec3(r, g, b);
}
```
Id    | (R,G,B)
254 | (0,0,254)
255 | (0,1,0)
256 | (0,1,1)

larsmoa: I see - you're right. @haakonflatval-cognite, I think we should do this in a separate PR - not this PR

haakonflatval-cognite (approved):
LGTM

PR #2128, opened 2022-05-12
improvement: disable SSAO and MSAA on mobile devices

larsmoa:
Merge after #1982


viewer/core/src/public/migration/restrictViewerOptionsToDevice.ts
```diff
@@ -0,0 +1,70 @@
...
+      return 'disabled';
+
+    case 'disabled':
+    case undefined:
+      return ssaoQualityHint;
+
+    default:
+      assertNever(ssaoQualityHint);
+  }
+}
```
christjt, line 70: We already have a `createRevealManagerOptions`, that has `determineAntiaAlisaingMode` and `determineSsaoParamateres`. It makes more sense to me that it should be handled here instead of as a pre-create options step.

viewer/core/src/public/migration/restrictViewerOptionsToDevice.ts
```diff
@@ -0,0 +1,70 @@
...
+    case 'msaa2':
+    case 'msaa4':
+    case 'msaa8':
+    case 'msaa16':
+    case 'msaa2+fxaa':
+    case 'msaa4+fxaa':
+    case 'msaa8+fxaa':
+    case 'msaa16+fxaa':
+      log.warn(`Anti-aliasing mode '${antiAliasingHint}' is not supported on mobile devices, falling back to 'fxaa'`);
+      return 'fxaa';
```
christjt, line 39: msaaXX (without fxaa) will fall through and return 'fxaa'. I suppose this is unintended?

christjt (changes requested):

christjt:
Should this target master, or render-manager branch?

viewer/core/src/public/migration/restrictViewerOptionsToDevice.ts
larsmoa: This was intended - but we can change if you'd like. My logic was that if user wanted msaa and we can't enable it, we should at least enable fxaa.

viewer/core/src/public/migration/restrictViewerOptionsToDevice.ts
larsmoa: We might - but I wanted to separate the concern and let `createRevealManagerOptions` simply do as it's told. Again, I'll change this if you think this is just extra noise and unnecessary separation.

christjt: In my opinion this would not brake SOC, it would also simplify the code (or at least share some code), as we are technically setting the options twice with the current approach. The way I see it is that `createRevealManagerOptions`s job is to do this filtering / defaulting of the options. It also becomes clearer when reading code (one point of altering the input options data vs. two).

larsmoa: Not sure I think it's better now, but changed.

christjt (approved):
LGTM! Thanks :D

PR #2131, opened 2022-05-12
fix: Removed metrics for CameraManager and CustomDataSource

Savokr:
Removed logging metrics for `CameraManager` and `CustomDataSource` because when they were defined on options passed to `Cognite3DViewer` it was trying to send the whole instance of the class which obviously failed and was causing stack overflow error.


larsmoa (changes requested):
I think we should switch to `pick` from lodash (https://lodash.com/docs/4.17.15#pick).

Also, I think we should address the fact that a error in metrics can cause infinite loops

christjt:
@larsmoa can we maybe just add this as a hotfix such that we can release a patch, and then address these issues properly after?

christjt (approved):
LGTM

larsmoa (approved):

PR #2139, opened 2022-05-18
fix: Memory leaks on disposal of Cognite3DViewer

Savokr:
When disposing of `Cognite3DViewer` models were not removed from the scene and `dispose` method of `Cognit3DModel` was not properly working.


examples/src/pages/Migration.tsx
```diff
@@ -98,6 +98,16 @@ export function Migration() {
...
 
+      // For disposal memory leaks tests
+      (window as any).createViewer = () => {
+        viewer = new Cognite3DViewer(viewerOptions);
+        (window as any).viewer = viewer;
+        viewer.addModel({
+          modelId: parseInt(urlParams.get('modelId') ?? '', 10),
+          revisionId: parseInt(urlParams.get('revisionId') ?? '', 10)
+        })
+      };
```
larsmoa, line 109: I dont think we should add this here.

viewer/packages/rendering/src/CadMaterialManager.ts
```diff
@@ -116,13 +116,16 @@ export class CadMaterialManager {
...
   removeModelMaterials(modelIdentifier: string): void {
     const modelData = this.materialsMap.get(modelIdentifier);
-    for (const mat of Object.values(modelData.materials)) {
-      mat.dispose();
-    }
 
-    this.materialsMap.delete(modelIdentifier);
-    modelData.nodeTransformTextureBuilder.dispose();
-    modelData.nodeAppearanceTextureBuilder.dispose();
+    if (modelData) {
```
larsmoa, line 120: Should it not be an exception if someone calls `removeModelMaterial()` with an invalid model identifier?

larsmoa (changes requested):

viewer/packages/cad-model/src/Cognite3DModel.ts
```diff
@@ -284,7 +284,8 @@ export class Cognite3DModel extends THREE.Object3D implements CogniteModelBase,
    * Cleans up used resources.
    */
   dispose(): void {
-    this.children = [];
+    this.cadNode.dispose();
+    this.clear();
```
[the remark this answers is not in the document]
Savokr: I was testing it and it seems that potree manages garbage collection inside itself. Right when the model is removed the memory is not freed but when another model is loaded it cleans everything from the last one. If we want to have cleaned model data right after removal, I can look into it in separate PR.

examples/src/pages/Migration.tsx
Savokr: Ok, can remove it

viewer/packages/rendering/src/CadMaterialManager.ts
Savokr: I guess, yeah

viewer/packages/cad-model/src/Cognite3DModel.ts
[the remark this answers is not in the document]
Savokr: Ok, tested it again and it seems that point cloud models have exactly the same issue with memory leaks because right now we are not disposing of anything plus `removePointCloud` method of `PotreeGroupWraper` is not implemented properly as it is not removing a few references of models inside itself. So this all things should be in another PR which I will create at some point because right now I think it is better to focus on other stuff.

viewer/packages/rendering/src/CadMaterialManager.ts
larsmoa: Throw exception?

larsmoa (changes requested):
Throw exception in removeModelMaterials

viewer/packages/rendering/src/CadMaterialManager.ts
Savokr: Throwing error is bad?

larsmoa: ? It's not, not in this function at least?

larsmoa: And if it is, why check that its "truthy"? Also, modelIdentifier==="" will evaluate to false

larsmoa (approved):

PR #2141, opened 2022-05-18
test: added test for Cognite3dViewer.dispose()

larsmoa:


viewer/core/src/public/migration/Cognite3DViewer.test.ts
```diff
@@ -75,6 +75,38 @@ describe('Cognite3DViewer', () => {
...
+    };
+    nock(/.*/)
+      .defaultReplyHeaders({ 'access-control-allow-origin': '*', 'access-control-allow-credentials': 'true' })
+      .get(/.*\/outputs/)
+      .twice() // the first one goes to determine model type
+      .reply(200, outputs);
+    nock(/.*/)
+      .defaultReplyHeaders({ 'access-control-allow-origin': '*', 'access-control-allow-credentials': 'true' })
+      .get(/.*\/scene.json/)
+      .reply(200, sceneJson);
```
christjt, line 98: Is it best to mock the http requests or mock the SDK, I don't imagine there are that many methods you have to mock. Also we don't really care about the sdk for this test, right?

christjt (changes requested):

viewer/core/src/public/migration/Cognite3DViewer.test.ts
larsmoa: I agree, but the files are retrieved using `fetch(...)` so we cannot mock the SDK, ref https://github.com/cognitedata/reveal/blob/larsmoa/remove-models-on-dispose/viewer/packages/modeldata-api/src/CdfModelDataProvider.ts#L46 @christjt

christjt (approved):

PR #2157, opened 2022-05-25
example: add UI for setting point size type

larsmoa:


examples/src/utils/PointCloudUi.ts
```diff
@@ -1,27 +1,28 @@
...
+import { Cognite3DViewer, CognitePointCloudModel, PotreePointColorType, PotreePointShape, PotreePointSizeType } from "@cognite/reveal";
 import * as dat from 'dat.gui';
 
 export class PointCloudUi {
-  constructor(viewer: Cognite3DViewer, ui: dat.GUI) {
-    const pointCloudParams = {
+  private readonly _viewer: Cognite3DViewer;
+  private readonly _params =
+    {
       pointSize: 0.15,
```
haakonflatval-cognite, line 8:
```suggestion
      pointSize: 1.0,
```
The point size is a little small when the size is at this value, and the camera is zoomed in (see picture). I suggest increasing to 1.0. But it's a matter of preference and balance, I suppose.
bg0

haakonflatval-cognite (approved):

PR #2159, opened 2022-05-25
fix: instanced mesh is not transformed by transform override

christjt:
Also transitively fixes the explode tool issue in the documentation.


pramodcog (approved):
Looks fine

larsmoa (approved):

PR #2163, opened 2022-05-27
improvement: Make typescript strict

haakonflatval-cognite:
Turns on the `strict` option in `tsconfig.json` and makes the build pass.

### DIFF
--- a/viewer/packages/sector-parser/src/GltfSectorParser.ts
+++ b/viewer/packages/sector-parser/src/GltfSectorParser.ts
@@ -59,7 +59,7 @@ export class GltfSectorParser {
     node: Node,
     glbHeaderData: GlbHeaderData,
     data: ArrayBuffer
-  ): Promise<ParsedGeometry> | undefined {
+  ): Promise<ParsedGeometry| undefined> {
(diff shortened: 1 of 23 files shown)


viewer/packages/sector-parser/src/GltfSectorParser.ts
```diff
@@ -240,7 +240,7 @@ export class GltfSectorParser {
     if (dracoCompression !== undefined) {
       const dracoBufferView = json.bufferViews[dracoCompression.bufferView];
 
-      const dracoMeshOffset = offsetToBinChunk + dracoBufferView.byteOffset;
+      const dracoMeshOffset = offsetToBinChunk + dracoBufferView.byteOffset!;
```
haakonflatval-cognite, line 243: Question for @christjt : Should `byteOffset` be an optional field? If it actually is optional, we should probably fix this in another way.

christjt: The gltf sharp writer simply will not write the byteOffset field into the gltf file if it has the value of 0. So for our case it is possibly undefined. Agree that it is ugly though and `byteOffset ?? 0` makes a lot more sense.

christjt (approved):

PR #2166, opened 2022-05-27
fix: clearing wrong render target

christjt:
Fixes the "white screen" issue when picking / mouse scrolling


pramodcog (approved):
LGTM! and I have checked if we have similar `clearColor()` else where and found it in PointCloudOctreePicker.ts. I guess this has not caused any issue, but would be good to re-confirm.

PR #2169, opened 2022-05-28
example: add fly to function to UI

larsmoa:
For easily creating videos


larsmoa:
> I do notice that it only changes position and not camera angle, but I suppose it has value on its own. LGTM

Angle? You mean rotation? Thats implicitly defined by target

PR #2173, opened 2022-05-30
chore: refactor CadNode and related into proper package

christjt:


viewer/packages/cad-geometry-loaders/package.json
```diff
@@ -8,9 +8,11 @@
     "lint": "yarn ws:lint"
   },
   "dependencies": {
+    "@reveal/cad-model": "workspace:*",
```
haakonflatval-cognite, line 11: Intuitively, I would sort-of have expected `cad-geometry-loaders` to be a dependency of `cad-model`, and not the other way around, mostly since `cad-model` contains `Cognite3DModel`, which feels like it should be somewhat near the top. What do you think? Are there reasons for why this could be better?

haakonflatval-cognite (changes requested):
Mostly good, but curious about a couple of things

viewer/packages/cad-geometry-loaders/package.json
christjt: I disagree with this. It doesn't make sense for the cad-model to have a dependency to the loading / streaming component. Why would it need to know anything about how it is loaded. I do however agree with Cognite3DModel conceptually being a "higher level" component in Reveal. And I think that a common API package makes sense that contains this. cad-model for me is cadNode and its surrounding. What do you think?

examples/yarn.lock
```diff
@@ -1263,10 +1263,10 @@
   version "0.0.0"
   uid ""
 
-"@cognite/sdk-core@^4.1.7":
```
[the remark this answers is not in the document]
christjt: Thanks 😇

haakonflatval-cognite (approved):

PR #2176, opened 2022-05-30
fix: change equality check of visiblity shader attribute to a less check

vero-so:
Hi, we encountered issues with the visibility attribute != 1.0 check on some machines. For some reason, the comparison turns out wrong on rather high-end GPUs than low end one.
To demonstrate the issue, I made a debug square within the simple test example. The coverage factors are hardcoded to 1,1,1; so that the discarding part based on the random function is "turned off" (if condition never meets), and we can observe the effect. The color is just the normal vector, passed from the vertex shader. You can observe some strange streak line patterns (non random) that caused by the numerics. If you change the equality test of the visibility to <0.999.., the issue will disappear.
I have also tried to upload the visibility attribute in its own buffer, and added highp keyword to it, but it did not help.
It would be great if you could consider this fix. Thanks.
visibility_without_fix
visibility_w_fix


christjt (changes requested):

viewer/packages/rendering/src/glsl/sector/sectorCoverage.frag
```diff
@@ -11,11 +11,11 @@ in lowp vec2 v_seed;
 in vec3 v_viewPosition;
 
 out vec4 outputColor;
-    
+
 const NodeAppearance dummyNodeAppearance = NodeAppearance(vec4(0.0), false, false, false);
 
 void main() {
-    if(v_visible != 1.0 || isClipped(dummyNodeAppearance, v_viewPosition)){
+    if((v_visible < 0.99999) || isClipped(dummyNodeAppearance, v_viewPosition)){
```
christjt, line 18: Or perhaps a little more idiomatic floating point comparison?
```suggestion
    const epsilon = 0.0001;
    var isVisible = abs(v_visible - 1.0) < epsilon;
    if(!isVisible || isClipped(dummyNodeAppearance, v_viewPosition)){
```

christjt (changes requested):

viewer/packages/rendering/src/glsl/sector/sectorCoverage.frag
vero-so: Ah, I just see this now.

vero-so: I commited an update (w small changes due to glsl syntax).

christjt (approved):

christjt:
Thanks for the contribution @vero-so, please update, then it should be ready to go.

vero-so:
@christjt I think I cannot merge the changes (or rebase) due to branch protection rules (I am not an authorized user), so you might need to do that. Thanks.

PR #2200, opened 2022-06-10
chore: bump cognite SDK and SDK-core

christjt:
# Description
Bump SDK version, current version seems to break in the browser due to certain nodejs dependencies. This causes issues when running documentation locally.

### DIFF
--- a/viewer/package.json
+++ b/viewer/package.json
@@ -64,8 +64,8 @@
   },
   "devDependencies": {
     "@azure/msal-browser": "2.24.0",
-    "@cognite/sdk": "7.5.1",
-    "@cognite/sdk-core": "^4.1.1",
+    "@cognite/sdk": "7.6.1",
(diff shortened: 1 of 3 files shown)


viewer/package.json
```diff
@@ -64,8 +64,8 @@
   },
   "devDependencies": {
     "@azure/msal-browser": "2.24.0",
-    "@cognite/sdk": "7.5.1",
-    "@cognite/sdk-core": "^4.1.1",
+    "@cognite/sdk": "7.6.1",
+    "@cognite/sdk-core": "4.4.0",
```
Savokr, line 68: Do we want to keep ^ for minor versions update?

Savokr (approved):
LGTM

viewer/package.json
christjt: No, don't trust them anymore. So need to validate even minor versions...

PR #2230, opened 2022-06-21
feat: upgrade to Webpack 5

haakonflatval-cognite:
# Description

Please include a summary of the change and which issue is fixed. 
Please also include relevant motivation and context. 
List any dependencies that are required for this change.

# Checklist:

- [ ] I have outlined any known defects / lacking cabilities in the contents of this PR.


haakonflatval-cognite:
Note: the visual tests seem to fail because the rendering engine specified in the puppeteer configuration is not available in the CI workers. It seems to fall back to the default ('--use-gl=egl`).

haakonflatval-cognite:
Current status:  The viewer and the examples seem to work fine with Webpack 5 now. The documentation is still not running properly. On my end it fails with the errors `__webpack_modules__[moduleId] is not a function` and `Uncaught ReferenceError: process is not defined`. 

The latest commit at the point of this comment is an attempt at converting `three` to a peer dependency (which would break the API, but can postpone release to 4.x). It is only provided for convenience if that path seems more feasible. The previous commit (https://github.com/cognitedata/reveal/pull/2230/commits/2b75e9beca0c416b43242b2446cfe8bd285f82fc) preserves the behavior of today's bundling (it doesn't bundle Three JS, but it will usually be installed by the client anyway). Both of these commits shows the same error in the browser when using the documentation.

christjt:
> Current status: The viewer and the examples seem to work fine with Webpack 5 now. The documentation is still not running properly. On my end it fails with the errors `__webpack_modules__[moduleId] is not a function` and `Uncaught ReferenceError: process is not defined`.
> 
> The latest commit at the point of this comment is an attempt at converting `three` to a peer dependency (which would break the API, but can postpone release to 4.x). It is only provided for convenience if that path seems more feasible. The previous commit ([2b75e9b](https://github.com/cognitedata/reveal/commit/2b75e9beca0c416b43242b2446cfe8bd285f82fc)) preserves the behavior of today's bundling (it doesn't bundle Three JS, but it will usually be installed by the client anyway). Both of these commits shows the same error in the browser when using the documentation.

Process not defined is something a have been expecting. Basically Webpack 5 does not polyfill Nodejs things (which process is). The "is not a function" thing I am not quite sure of though :(

christjt:
Also this thread still scares me: https://github.com/webpack/webpack/issues/11277, and was the original reason I gave up porting to webpack 5

christjt:
Closing this as it has been merged, well done!

Issue #2236, opened 2022-06-23
3.1.0 defaults to using Medium SSAO on mobile devices


Strepto:
https://github.com/cognitedata/reveal/blob/a4d35b882b458956c27c7f109d9f00bd52ff6163/viewer/packages/api/src/public/migration/renderOptionsHelpers.ts#L140-L142

There is a bug here where the default fallthrough of `restrictSsaoOptionBasedOnDevice` is the input value, and the input value might be `undefined`. This makes the switch here
https://github.com/cognitedata/reveal/blob/a4d35b882b458956c27c7f109d9f00bd52ff6163/viewer/packages/api/src/public/migration/renderOptionsHelpers.ts#L69-L72

just use the default value, which is "SSAO medium", and this makes performance on mobile unusable. Its easy to work around in the client by specifying the quality to `'disabled`, but it might be nice for new users fo the library to not have to specify the ssaoQualityHint. 

(Remark: i have looked at this for about 10 minutes and did not actually verify this is a fix)

Strepto:
Also:

The isMobileOrTablet check does not detect iPad when they have "Request Desktop Website" (Which is the default).

An ugly workaround could be: 
```
function iOS() {
  return [
    'iPad Simulator',
    'iPhone Simulator',
    'iPod Simulator',
    'iPad',
    'iPhone',
    'iPod'
  ].includes(navigator.platform)
  // iPad on iOS 13+ detection ('Request Desktop Website' workaround)
  || (navigator.userAgent.includes("Mac") && "ontouchend" in document) // Super hacky and will fail when Macs get touch...
}
```

christjt:
Thanks for reporting!

I'll try to get a patch out tomorrow to fix this.

christjt:
Hey @Strepto,

Would it be possible for you to verify this fix: https://github.com/cognitedata/reveal/pull/2239?

christjt:
@Strepto Has been merged now. Is this something you need us to release asap. or is it ok to defer?

Strepto:
No rush for us.

Feel free to close the issue 🥇

PR #2239, opened 2022-06-24
fix: mobile default settings should not include ssao, add iPad fix

christjt:
# Description

3.1.0 introduces a regression such that mobile / tablet gets ssao shading when no options are supplied. This fixes that, as well as adds an additional check for matching iPads.

For more context see #2236


Savokr (approved):
LGTM!  Only thing is that we should probably add one common way to check what platform we are running on, maybe some package is good for that, not sure.

PR #2244, opened 2022-06-27
chore: bump to Webpack 5

christjt:
# Description

NOTE THAT @haakonflatval-cognite did most of the heavy lifting here, but I simply pushed it over the goal line, and will therefore also take all the credit (just as a proper tech-lead). Had to fork his branch from an earlier commit since he started to port Reveal to be a peer-dependency (which we are not quite ready for yet).

Update to webpack 5 (includes examples and viewer).

LIMITATION / known regression:
If Reveal is built in development mode, a webpack 4 project that is also built in development mode will fail during runtime. For more info see https://github.com/webpack/webpack/issues/11277, the TLDR as far as I can see is that webpack injects some runtime glue that doesn't play well with each other across version 4 and 5 apparently.

Implications don't seem to be that bad from our side, since we also port /examples to webpack 5 here, which then circumvents the issue. Users, should of course not be affected either since we release Reveal in production mode. The documentation will however be affected by this, and will only work locally if you build Reveal in production mode. This is a regression, but hopefully something we can live with until we also port the documentation to a newer version of docusaurus (which also uses webpack 5).

So as a temporary thing we have to build Reveal for production when generating preview-docs, for it to work. Production will work fine, since it already is built for production.

Tracking for updating documentation: https://cognitedata.atlassian.net/browse/REV-403

# Checklist:


examples/package.json
```diff
@@ -4,17 +4,14 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "start": "cross-env HTTPS=true react-app-rewired start",
-    "start:e2e-server": "cross-env PORT=3000 HTTPS=false BROWSER=none react-app-rewired start",
-    "build": "react-app-rewired build",
```
pramodcog, line 9: Can we make a note that examples do not need to be build with the changes from here!

viewer/packages/sector-loader/app/index.ts
```diff
@@ -147,7 +147,7 @@ async function loadSectors(
           geometryBoundingBox: undefined,
           children: [],
           estimatedRenderCost: 0
-        },
+        } as any,
```
pramodcog, line 150: Why cast to `any` for webpack 5 and only for `sector-loader`?

pramodcog (approved):

PR #2251, opened 2022-06-29
fix: elements position in HTMLOverlay when visibility is changed from hide to show

pramodcog:
The elements position in HTML Overlay are not updated when visibility of overlay is changed from hide & show.

[REV-413](https://cognitedata.atlassian.net/browse/REV-413)


Savokr (approved):
LGTM 👍

PR #2257, opened 2022-06-30
fix: sector repository not gracefully handling sector loader errors

christjt:
# Description
If you f.ex. delete a model while sectors are still downloading, it will throw an error when the are consumed. This error was not catched and will spam the console. We still log an error, but now the user can opt out by silencing the logger.

The actual proper fix is to cancel in flight sectors, but requires a larger effort (tracking task: https://cognitedata.atlassian.net/browse/REV-414)

# Checklist:

### DIFF
--- a/viewer/packages/sector-loader/package.json
+++ b/viewer/packages/sector-loader/package.json
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "@reveal/cad-parsers": "workspace:*",
+    "@reveal/logger": "workspace:*",
     "@reveal/metrics": "workspace:*",
     "@reveal/modeldata-api": "workspace:*",
     "@reveal/rendering": "workspace:*",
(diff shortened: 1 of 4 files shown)


viewer/packages/sector-loader/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "@reveal/cad-parsers": "workspace:*",
+    "@reveal/logger": "workspace:*",
```
pramodcog, line 12: The logging is used only for unit test the changes, is possible to have an alternative avoiding logger dependencies!

viewer/packages/sector-loader/src/GltfSectorRepository.ts
```diff
@@ -55,7 +55,15 @@ export class GltfSectorRepository implements SectorRepository {
     if (this._gltfCache.has(cacheKey)) {
       return this._gltfCache.get(cacheKey);
     }
-    const consumedSector = await this._gltfSectorLoader.loadSector(sector);
+
+    const consumedSector = await this._gltfSectorLoader.loadSector(sector).catch(() => {
+      return undefined;
```
pramodcog, line 60: Do we handle the return value somewhere for the `loadSector`, wanted to understand if `return undefined` or `return null` is suitable here

christjt: undefined is returned to from the method, but from the catch lambda into the `consumedSector` variable. Which is then checked for undefined in the next line.

pramodcog (approved):
LGTM!

PR #2265, opened 2022-07-04
fix: manually loop unroll in shader to improve performance

christjt:
# Description

Whilst performance testing with EX devices, it showed a massive FPS hit from the edges visualization which turned out to be because of the loops. Manually unrolling them restored performance.

Also added an addition "Triangle count" statistic to the renderer testapp. And removed un-needed host setting in webpack config (not needed after webpack 5 migration).

### DIFF
--- a/viewer/packages/rendering/src/glsl/post-processing/edge-detect.glsl
+++ b/viewer/packages/rendering/src/glsl/post-processing/edge-detect.glsl
@@ -28,7 +28,7 @@ float edgeDetectionFilter(sampler2D baseTexture) {
 	/* fetch the 3x3 neighbourhood and use the RGB vector's length as intensity value */
   I[0][0] = length(texelFetch(baseTexture, fragCoord + ivec2(-1, -1), 0).rgb);
   I[0][1] = length(texelFetch(baseTexture, fragCoord + ivec2(-1, 0), 0).rgb);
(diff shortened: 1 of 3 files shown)


viewer/packages/rendering/src/glsl/post-processing/edge-detect.glsl
```diff
@@ -24,23 +24,48 @@ float edgeDetectionFilter(sampler2D baseTexture) {
...
-	for (int i=0; i<9; i++) {
-		float dp3 = dot(G[i][0], I[0]) + dot(G[i][1], I[1]) + dot(G[i][2], I[2]);
-		cnv[i] = dp3 * dp3;
-	}
-
-	float M = (cnv[0] + cnv[1]) + (cnv[2] + cnv[3]);
-	float S = (cnv[4] + cnv[5]) + (cnv[6] + cnv[7]) + (cnv[8] + M);
+  I[0][0] = length(texelFetch(baseTexture, fragCoord + ivec2(-1, -1), 0).rgb);
+  I[0][1] = length(texelFetch(baseTexture, fragCoord + ivec2(-1, 0), 0).rgb);
+  I[0][2] = length(texelFetch(baseTexture, fragCoord + ivec2(1, 1), 0).rgb);
```
Savokr, line 31: Should be `ivec2(-1,1),0)`.

Savokr (approved):
LGTM, only one 1 small mistake in calculations.

PR #2272, opened 2022-07-08
chore: add visual tests setup for /viewer 

christjt:
# Description

This includes a setup to run and create visual tests within the context of /viewer.

Test by running `yarn run test:visual` in /viewer.

New capabilities:

- Run visual tests with a single command `yarn run test:visual`, which both serves a development server that hosts the tests, and runs screenshot tests using puppeteer.
- All tests are run within the same context / site, so overhead of running each test is considerably lower than previous solution.
- Watch mode is partially supported. I say partially because it requires the dev server and tests to be run in different processes. To do this run `yarn run test:visual:server` and `yarn run test:visual:client --watch` in two different terminal instances. In addition to this, the test client (the one that runs jest/puppeteer) doesn't actually listen to the visual test apps, so you need to trigger a run using the cli (press ENTER key). The gain is that the feedback loop is substantially lower, and re-running a test in watch mode usually only takes a few seconds.

To register a new visual test, it has to be registered in /viewer/packages/visualTests.root.ts (though I am experimenting with webpack's ability to do dynamic loading to try to overcome this).

There are also a naming convention which must be kept when creating a new visual test which is: `[MyTest].VisualTestFixture.ts` and the class name must be `[MyTest]VisualTestFixture`.

NOTE:
LGTM and code analyser complains about remote execution logic going on, but this is intended behaviour because we want to remotely execute code that is being served on the test server from the test runner. 

# Checklist:

I have a task for this in Jira


viewer/visual-tests/VisualTests.ts
```diff
@@ -0,0 +1,73 @@
+/*!
+ * Copyright 2022 Cognite AS
+ */
+
+import { Page } from 'puppeteer';
+
+const environment = process?.env?.NODE_ENV ?? 'browser';
+
+if (environment === 'browser') {
```
larsmoa, line 9: What is this if test about? Confused

larsmoa (commented):
Looks very promising. Where will existing visual tests live? Thinking about how we can manage tests would be nice - but I'm not sure how we can make them "self registering".

viewer/visual-tests/VisualTests.ts
christjt: Was a dumb idea to have both the test runner and the server that serves the tests in the same file. And it turned out to just be confusing, so refactored to two different components.

christjt:
> Looks very promising. Where will existing visual tests live? Thinking about how we can manage tests would be nice - but I'm not sure how we can make them "self registering".

Current visual tests will be distributed to relevant packages. Most will probably belong to /rendering package.

viewer/visual-tests/VisualTest.jest.ts
```diff
@@ -0,0 +1,44 @@
+/*!
+ * Copyright 2022 Cognite AS
+ */
+
+// import assert from 'assert';
```
pramodcog, line 5:
```suggestion
```

pramodcog (approved):

PR #2332, opened 2022-08-12
fix: remove un-needed dependency (jest-environment-jsdom)

christjt:
This should not be a dependency (rather a dev-dependency, which it already is).


Savokr (approved):
LGTM

PR #2361, opened 2022-08-29
fix: incorrectly setting device pixel ratio with respect to domElement size

christjt:
# Description

Initial render size was incorrect before applying a resize. This was due to mistakenly ignoring pixel ratio on initial resize when set by user.

For more context and manifistation of the bug see:
https://cognitedata.atlassian.net/browse/REV-456.


viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1198,39 +1198,35 @@ export class Cognite3DViewer {
...
     // For the love of God someone move this to the render manager and make it dynamic based on the device.
-    const maxTextureSize = 1.4e6;
+    const maxPhysicalFrameBufferSize = 1.4e6;
 
-    const rendererSize = this.renderer.getSize(new THREE.Vector2());
-    const devicePixelRatio = this.renderer.getPixelRatio();
-    const rendererPixelWidth = rendererSize.width;
-    const rendererPixelHeight = rendererSize.height;
+    const virtualFramebufferSize = this.renderer.getSize(new THREE.Vector2());
+    const pixelRatio = window.devicePixelRatio;
```
larsmoa, line 1204: Should it always be window, or should it be possible to set using a custom `rendered`?

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1198,39 +1198,35 @@ export class Cognite3DViewer {
...
 
-    // This needs to be done such that users don't unintentionally bypass the
-    // resolution cap by setting a high device pixel ratio
-    width /= devicePixelRatio;
-    height /= devicePixelRatio;
-
-    adjustCamera(this.camera, width, height);
-    this.renderer.setSize(width, height);
+    adjustCamera(this.camera, newVirtualWidth, newVirtualHeight);
+    this.renderer.setDrawingBufferSize(newVirtualWidth, newVirtualHeight, pixelRatio);
```
pramodcog, line 1229: We are calling `renderer.setPixelRatio(window.devicePixelRatio)` in `createRenderer()`, do we need to call `renderer.setDrawingBufferSize()` which assigns pixelRatio to the renderer again, we can use `renderer.setSize()` instead which does the same as `setDrawingBufferSize()` expect for assigning `pixelRatio` for the renderer

christjt: It is exactly equivilant (both methods require framebuffers to be reinitialized). In my opinion using `setDrawingBufferSize` is clearer for readability as we know exactly what the framebuffer size will be. Not married to this though, what do you prefer?

pramodcog: If we allow the users to set `pixelRatio`, than `setDrawingBufferSize()` is relevant, having said that `setDrawingBufferSize()` does not harm us either currently.

pramodcog (approved):
Good to me!

Issue #2363, opened 2022-08-30
Fragment Interpolation causes wrong TreeIndex lookup on some GPUs


Strepto:
The title is a hypothesis, but the fix seems to work. The issue does NOT occur on certain Intel GPUs, but its reproducible on Nvidia, Mac, and iOS GPUs.

This may also have been the cause of #2176 

It seems to be a problem with fragment interpolation on some (most) GPUs that cause issues when TreeIndexes are high (+2 million), and that will introduce noise in the 3D model when filters are applied, AND when Picking based on TreeIndex.

The effect looks like this, when for instance highlighting is applied: 

![image](https://user-images.githubusercontent.com/3185998/187393466-600ea11a-100d-46fb-86b8-9dcc2274c71e.png)

Importantly this also affects picking, which may get the wrong TreeIndex (often off by 1 or up to 2) for a meshpart due to this issue.

The issue occurs frequently on TreeIndexes over approx 2_000_000 , and gets worse with higher numbers.

Our hypothesis is that the fragment interpolation causes an inaccuracy that makes the interpolation operation roundtrip give significant variance in the output fragment treeIndex. See a description of the possible algorithm here: https://stackoverflow.com/questions/24441631/how-exactly-does-opengl-do-perspectively-correct-linear-interpolation

The fix seems to be to apply `flat` to the fragment shader. See: https://www.khronos.org/opengl/wiki/Type_Qualifier_(GLSL)#Interpolation_qualifiers

christjt:
Fixed with #2364

PR #2364, opened 2022-08-30
fix: remove TreeIndex interpolation in fragment shaders

Strepto:
# Description

Related to #2363 , read that for more information about the issue.

This may be tested by adding a model where tree-indexes are surpassing 3 million, and doing some selection or similar. 

We tested this by just doing

```cs
while(treeindex < 3_000_000) treeIndex++;
```

before running any 3d model through the pipeline. And then applying a highlight, or hide NodeAppearance on click, to apply it to a specific node.

Co-authored-by: @vero-so 

# Checklist:

### DIFF
--- a/viewer/packages/rendering/src/glsl/base/determineNodeAppearance.glsl
+++ b/viewer/packages/rendering/src/glsl/base/determineNodeAppearance.glsl
@@ -1,17 +1,14 @@
 #pragma glslify: import('../math/floatBitsSubset.glsl')
 
-NodeAppearance determineNodeAppearance(sampler2D nodeAppearanceTexture, vec2 textureSize, float treeIndex) {
-  treeIndex = floor(treeIndex + 0.5);
+NodeAppearance determineNodeAppearance(sampler2D nodeAppearanceTexture, vec2 textureSize, highp float treeIndex) {
+
   float dataTextureWidth = textureSize.x;
   float dataTextureHeight = textureSize.y;
 
-  float u = mod(treeIndex, dataTextureWidth);
-  float v = floor(treeIndex / dataTextureWidth);
-  float uCoord = (u + 0.5) / dataTextureWidth;
-  float vCoord = (v + 0.5) / dataTextureHeight;
-  vec2 treeIndexUv = vec2(uCoord, vCoord);
-  
-  vec4 texel = texture(nodeAppearanceTexture, treeIndexUv);
+  int xTreeIndexTextureCoord = int(mod(treeIndex, dataTextureWidth));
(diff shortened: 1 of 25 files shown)


christjt (approved):
LGTM, thank you for contributing!

christjt:
Seems to be an issue with a visual test. Seems to be an issue on our end (we just switched systems). I am looking into it.

Strepto:
Hm, there is an issue with the changes in `viewer/packages/rendering/src/glsl/base/determineNodeAppearance.glsl` that affects the visual tests. 

I'll see if I can understand what goes wrong.

christjt:
> Hm, there is an issue with the changes in `viewer/packages/rendering/src/glsl/base/determineNodeAppearance.glsl` that affects the visual tests.
> 
> I'll see if I can understand what goes wrong.

It is due to a mock implementation being used in the visual test. Usually the colorDataTexture is the size of one color (4byte) per treeIndex in the model. For the test I just have the size of a single color (since I don't care about color overrides for this test). This previously worked fine, since texture fetches are clamped, so even if you sample out of bounds it still ends up sampling that single cell of the texture. Appearently, texelFetch doesn't work like that and it doesn't seem to clamp.

For production code everything is fine, so think I just simply have to implement the full texture in the test. I can do this, @Strepto, then you can simply rebase/merge and everything will be fine.

Strepto:
When reverting `determineNodeAppearance.glsl` the tests succeed, and its visually OK. So we could go with that. 

But the new code in `determineNodeAppearance.glsl` was copied from `determineMatrixOverride.glsl`. So there might be an issue in that code. 🤔 Not entirelly sure how to diagnose this.

Strepto:
> > Hm, there is an issue with the changes in `viewer/packages/rendering/src/glsl/base/determineNodeAppearance.glsl` that affects the visual tests.
> > I'll see if I can understand what goes wrong.
> 
> It is due to a mock implementation being used in the visual test. Usually the colorDataTexture is the size of one color (4byte) per treeIndex in the model. For the test I just have the size of a single color (since I don't care about color overrides for this test). This previously worked fine, since texture fetches are clamped, so even if you sample out of bounds it still ends up sampling that single cell of the texture. Appearently, texelFetch doesn't work like that and it doesn't seem to clamp.
> 
> For production code everything is fine, so think I just simply have to implement the full texture in the test. I can do this, @Strepto, then you can simply rebase/merge and everything will be fine.

Ok, thanks! I'll rebase when that fix is done then.

christjt:
@Strepto The fix has been merged now, so updating the base branch should hopefully work.

Strepto:
Ah, I rebased on the new changes now. Lets see if this works as well

PR #2365, opened 2022-08-30
fix: sector parser test now properly computes color data override texture

christjt:
# Description
Small refactor to properly handle color override data texture. This is needed for textureFetch logic in the shader which #2364  introduces.


Savokr (approved):
LGTM!

PR #2366, opened 2022-08-30
feat: upgrade to docusarus 2.x

haakonflatval-cognite:
# Description

This upgrades the documentation to the newest Docusaurus version.

https://cognitedata.atlassian.net/browse/REV-458

# Checklist:


documentation/docusaurus.config.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
   projectName: 'reveal',
   themeConfig: {
     algolia: {
+      appId: 'reveal-docs',
```
pramodcog, line 11: Did we add this `appId` in the `algolia` for Reveal account or do not know if it is actually needed?

documentation/src/theme/Navbar/index.tsx
```diff
@@ -1,225 +0,0 @@
-import React, {useCallback, useState, useEffect} from 'react';
```
pramodcog, line 1: Are these file auto deleted when build with new `docusarus` version

documentation/docusaurus.config.js
haakonflatval-cognite: I got an error when it wasn't included in the config.. So I just added this, and it seems to work.  I have no idea if it can be an arbitrary name, or if it should match something

haakonflatval-cognite: Well, apparently it should be set to an ID we have received from Algolia: https://docusaurus.io/docs/search But I don't think any of the current team members have an idea of what it should be

documentation/src/theme/Navbar/index.tsx
haakonflatval-cognite: No, I have deleted them manually. They seem to be made just for getting around arbitrary limitations in the Docusaurus 2 alpha which we no longer need to worry about. But I'm going to do a closer check today, to see what exactly we needed them for.

documentation/docusaurus_plugins/docusaurusWebpack5Plugin.js
```diff
@@ -0,0 +1,21 @@
...
+ * Adapted from https://gist.github.com/sibelius/24f63eef7f43b15dc73c4a0be11bbef8
+ */
+
+const webpack = require('webpack');
+
+module.exports = function (context, options) {
+  return {
+    name: 'docusaurus-webpack-plugin',
+    // eslint-disable-next-line
+    configureWebpack(config, isServer, utils) {
```
pramodcog, line 11: Is it possible to add this into webpack.config similar to viewer/examples to align, your thoughts!

haakonflatval-cognite: I don't think so, because Docusaurus has its own internal webpack config, which we are only able to change using plugins like this.

pramodcog (approved):
If someone can confirm on the [comment](https://github.com/cognitedata/reveal/pull/2366#discussion_r958638604) would be great, other than this, LGTM!

haakonflatval-cognite:
There is an issue where the space between the canvas and the following HTML content disappears. Investigation found that the styling on the outer `div` representing the `DemoWrapper` sometimes disappears, along with its id. Still no clue why.

christjt (changes requested):
Search is broken :(

haakonflatval-cognite:
As noted earlier, search is broken. We are using Algolia to index the docs and provide search for us. A former member of the team received a mail in November last year with new credentials and instructions on migrating to a newer version of Algolia, before February 2022. This was never done. Apparently, Docusaurus 2.0 uses this new Algolia API only, and we have no way of migrating now. I have sent a support ticket to Algolia to see if we can sort it out somehow.

More on the migration: https://docusaurus.io/blog/2021/11/21/algolia-docsearch-migration

PR #2383, opened 2022-09-02
fix: keyboard events dispose for input handler

pramodcog:
# Description

Fix for keyboard events which are held by `Cognite3DViewer` when user exits from viewer which cause the calling application an issue when WASD key is used.

# Checklist:

- [x ] I have performed a self-review of my own code.
- [x ] I have commented my code in hard-to-understand areas.
- [x ] I have listed any remaining work that should follow this PR in the description or jira/miro/etc.
  [REV-454](https://cognitedata.atlassian.net/browse/REV-454)


viewer/packages/camera-manager/src/Keyboard.ts
```diff
@@ -56,6 +56,11 @@ export default class Keyboard {
     return p;
   }
 
+  public dispose(): void {
```
haakonflatval-cognite, line 59: Can we just move this to the bottom of the file? Would just make it a tiny bit more tidy, in my opinion :sweat_smile:

haakonflatval-cognite (approved):
LGTM :+1:
Only a small suggestion

viewer/packages/camera-manager/src/Keyboard.ts
pramodcog: `dispose()` is a public function, all functions below this are `private`, so I placed at the end of all `public` function :). We will be refactoring the way event handling for mouse & keyboard in near future. with [REV-454](https://cognitedata.atlassian.net/browse/REV-454)

PR #2386, opened 2022-09-02
fix: remove build viewer dependency and download viewer step for unit tests

christjt:
# Description

Unit tests do not depend on viewer being built, as it will built it itself using ts-jest. So removing this dependency to increase paralellism during CI.


pramodcog (approved):
Looks good!

PR #2387, opened 2022-09-02
fix: wrong resolve of viewer symlink

christjt:
# Description

The way examples symlinks reveal is currently wrong and causes it to wrongfully resolve dependencies in the /viewers node_modules. This can cause dependencies to be resolved even though they should not.

Before this change:
image

After this change:
image


Savokr (approved):
Beautiful `false` statement! 🤩

PR #2397, opened 2022-09-06
chore: purge unused examples and all references to /internals from examples

christjt:
# Description

Removes any unused examples and utils and delete internals from development build.


examples/src/utils/renderer-debug-widget.ts
```diff
@@ -1,396 +0,0 @@
-/*!
- * Copyright 2021 Cognite AS
- */
-
```
Savokr, line 4: Could be ported to rendering visual tests because otherwise will probably be forgotten forever and reimplemented in the future. Also can be helpful in testing primitives vs meshes performance.

Savokr (approved):
Looks good, a lot of stuff removed, only one comment about renderer-debug-widget.

PR #2401, opened 2022-09-07
fix: round calculation of drawing buffer size to an integer

christjt:
# Description

Fixes an issue where drawing buffer size have a fraction.


Savokr (approved):
A really good round!

PR #2418, opened 2022-09-09
fix: remove flat-identifiers which cause rendering on iOS to be very slow

larsmoa:
# Description

It seems iOS has an issue with `flat`-"varyings", causing rendering to be very slow. This was worked around by removing flat identifiers (they were only a nicety - all values for each vertex in a triangle should have same values in our case).

In addition, precision issues with `treeIndex` has been solved by using `int` for computation of texture coordinates. A crucial step is to avoid `floor(treeIndex + 0.5)` as this breaks down for `treeIndex` around 2-4 million somewhere. After this, the result of `floor(x + 0.5)` where x is an integer, might be `x + 1`. See https://en.wikipedia.org/wiki/Single-precision_floating-point_format#Precision_limitations_on_decimal_values_(between_1_and_16777216) for details.

# Checklist:


Strepto:
Just noticed this. Was not aware of the issue, but some research shows its a WebGL + Metal compability fix with "Flat" modifiers and the WebGL spec requiring that flat shading uses the "LAST_PROVOKING_VERTEX". Metal is based on "FIRST_PROVOKING_VERTEX", and thereby requiring re-ordering of all vertices in the shader to emulate the behavior. :(

Work seems to recently have been started on an extension for WebGL2 that could solve this problem in the future.

EXT_PROVOKING_VERTEX: https://registry.khronos.org/webgl/extensions/EXT_provoking_vertex/

Unfortunately that means that a fix for the inacurracies of the fragment smoothing is not as simple as I thought.

A hack could maybe be to split the treeIndex to 2 floats, where one is the "thousands" and the other the hundreds part. And then round to int, and add together... Fun? No. Works? Yes.

```
out float _treeIndexThousands = a_treeIndex % 1000;
out float _treeIndexSubThousand = a_treeIndex // 1000;


in float _treeIndexThousands;
in float _treeIndexSubThousand;

frag() {
    _treeIndex = round(_treeIndexThousands) * 1000.0 + round(treeIndexSubThousand)
}
```

I did not however test this, and have no idea of the impact yet.

Strepto:
@larsmoa I applied the "hack/fix" above and i'm able to run good performance while avoiding noise in TreeIndexes above 10_000_000 on iPad. 

If you want I can give a link to a test env. The code is hacked into all the shaders but follows the structure in the comment above.

christjt:
@larsmoa Think we can close this? @Strepto is going to implement a work-around for this and once the provoking vertex extension lands, we can use that.

larsmoa:
Replaced with #2536

PR #2419, opened 2022-09-09
fix: set minimum point size to 1 (from 2) and add visual test for fixed size

christjt:
# Description

Setting fixed point size is off in relation to adaptive. Applying a scale to get it more consistant.


pramodcog (approved):
Looks good! Just that we have magic number here `pointSize = size * 10.0`, I guess this is just an assumption we make!

PR #2425, opened 2022-09-12
fix: issue with old frames being visible in highlighted objects

christjt:
# Description

See: https://cognitedata.slack.com/archives/G0115JQ8JM8/p1662968222863869 for context.


larsmoa (approved):
Works!

PR #2429, opened 2022-09-13
fix: downscale default resolution threshold on mobile devices by its device pixel ratio

christjt:
Lower default resolution threshold on mobile devices by its pixel ratio. This is done such that default performance is preserved (from 3.1), as well as mapping resolution to virtual pixel density (i.e. most devices will have similar resolution).


viewer/packages/api/src/public/migration/renderOptionsHelpers.ts
```diff
@@ -90,6 +90,32 @@ export function determineSsaoRenderParameters(
...
+  if (rendererResolutionThreshold) {
+    return rendererResolutionThreshold;
+  }
+
+  let resolutionThreshold = 1.4e6;
+
+  // This is an attempt at increasing performance for high-resolution mobile devices
+  // by mapping resolution to physical pixel size
+  if (device.deviceType === 'mobile' || device.deviceType === 'tablet') {
+    resolutionThreshold = Math.floor(resolutionThreshold / devicePixelRatio);
```
pramodcog, line 113:
```suggestion
    return Math.floor(resolutionThreshold / devicePixelRatio);
```

pramodcog, line 113: Can we suggest return the value here like `return Math.floor(resolutionThershold / devicePixelRatio);`

pramodcog (commented):
Test it on iOS which basically has `pixeRatio = 2`, looks good

viewer/packages/api/src/public/migration/renderOptionsHelpers.ts
christjt: Yes we can, and yes I should.

pramodcog (approved):

PR #2455, opened 2022-09-19
chore: remove support for version 8 outputs

christjt:
# Description


viewer/packages/modeldata-api/src/LocalModelMetadataProvider.ts
```diff
@@ -61,12 +61,6 @@ export class LocalModelMetadataProvider implements ModelMetadataProvider {
         return undefined;
       }
```
Savokr, line 62: V8 mention in error message of `getModelOutputs` function should be changed.

Savokr (approved):
A few tiny things here and there but overall seems really good! A lot of old stuff removed 🔥. Tested locally on M1 mac, works just fine.

PR #2508, opened 2022-09-29
fix: point cloud picking precision issue

pramodcog:
# Description

Precision error in pixel intersection in point cloud was caused due to `PickingWindowSize` was higher `20`, reduced the value to `5` pixel to bring the precision error [REV-523](https://cognitedata.atlassian.net/browse/REV-523)

While identifying the cause of the issue, I found some of the calculation done in `PointCloudOctreePickerHelper.findHit()` are not needed. We should have a detail look into the potree library if need to refactor it for our need

# Checklist:


christjt (approved):
Think this is fine for now, but down the road I would like this to be a little smarter. Having a larger window size is good in cases where users miss-click. I think a better schema is to always pick the center pixel, unless there is a pixel in the window that is a lot closer (by some relative threshold). Then we get the best of both worlds.

PR #2536, opened 2022-10-05
Remove flat modifier, while maintaining TreeIndex precision

Strepto:
# Description

This resolves an iOS Metal performance issue introduced by the fix of #2363 . The performance issue is caused by an implementation detail in the WebGL spec which is incompatible with Metal rendering, and metal emulates this detail. 

A real fix for this issue is in the works, with the EXT_PROVOKING_VERTEX extension, currently in a draft stage: https://registry.khronos.org/webgl/extensions/EXT_provoking_vertex/ . The extension is being implemented in WebKit (See https://github.com/WebKit/WebKit/pull/3701 ), but I guess we cannot expect this to be shipped soon.

As the original problem here is that "varyings" in WebGL use an algorithm that averages the three sides of the vertex, and also applies a "Perspective Correction". See "_For varyings though WebGL uses this formula_" on https://webglfundamentals.org/webgl/lessons/webgl-3d-perspective-correct-texturemapping.html 

if we do (random example proving this seems like the real issue)
```cs
float treeIndex = 3205807;
float a = treeIndex;
float b = treeIndex;
var aW = 0.9;
var bW = 0.9;
var t = 0.3;
float dividend =  (1 - t) * (float)a / aW + t * (float)b / bW;
float divisor = (1 - t) / aW + t / bW;

var result = dividend/divisor;
// Result: 3205806.5 (rounded towards nearest even: 3205806)
```

We get an offset result, but only for some values of W (and t). This varies, causing the "noise" pattern, seen in some screenshots of the "pre-fix" when tree-indices are high.

In this PR i propose a "dirty" fix, separating the TreeIndex into two different numbers in the Vertex Shader, and merging these numbers in the Fragment Shader. This removes the rounding issue given it works as the example above, as the inaccuracy does not occur for values below 2_000_000 in any of my tests. 

The current maximum supported treeIndex is 2^24 due to storing it in an RGB255 Color, so we do not yet need to handle higher values than approx 16_800_000, which the code above will handle nicely.
---- 

This PR adds another medp float to each fragment, and additional multiplications to add these together. The separation of 100ds and SubHundreds is kinda arbitrary, but medp has rounding issues when we get near 1000, so 99 as max works OK.

I might be overly defensive on rounding.

I have tested this with nodeIds up to 10_500_000, and have not spotted any issues with the approach. 

If you have any suggestion on OTHER approaches to pass these bytes through the shader I would be happy, as this feels a bit wrong... :D

# Checklist:

  - 😅

### DIFF
--- a/viewer/packages/rendering/src/glsl/sector/instancedMesh.frag
+++ b/viewer/packages/rendering/src/glsl/sector/instancedMesh.frag
@@ -13,7 +13,7 @@ uniform sampler2D matCapTexture;
 uniform vec2 treeIndexTextureSize;
 uniform int renderMode;
 
-flat in float v_treeIndex;
+in float v_treeIndex;
 in vec3 v_color;
 in vec3 v_viewPosition;
 
--- a/viewer/packages/rendering/src/glsl/sector/instancedMesh.frag
+++ b/viewer/packages/rendering/src/glsl/sector/instancedMesh.frag
@@ -13,11 +13,14 @@ uniform sampler2D matCapTexture;
 uniform vec2 treeIndexTextureSize;
 uniform int renderMode;
 
-in float v_treeIndex;
 in vec3 v_color;
 in vec3 v_viewPosition;
 
-void main() {
+in highp float v_treeIndexHundreds;
+in mediump float v_treeIndexSubHundreds;
+void main()
+{
+    highp float v_treeIndex = round(v_treeIndexHundreds) * 100.0 + round(v_treeIndexSubHundreds);
     NodeAppearance appearance = determineNodeAppearance(colorDataTexture, treeIndexTextureSize, v_treeIndex);
     if (!determineVisibility(appearance, renderMode)) {
         discard;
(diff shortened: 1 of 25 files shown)


viewer/packages/rendering/src/glsl/sector/instancedMesh.frag
```diff
@@ -13,11 +13,14 @@ uniform sampler2D matCapTexture;
 uniform vec2 treeIndexTextureSize;
 uniform int renderMode;
 
-flat in float v_treeIndex;
 in vec3 v_color;
 in vec3 v_viewPosition;
 
-void main() {
+in highp float v_treeIndexHundreds;
+in mediump float v_treeIndexSubHundreds;
```
larsmoa, line 20: Suggestion: change this to `in highp vec2 v_treeIndex` (or possibly a struct, if you want to be more explicit), and create functions `packTreeIndex` and `unpackTreeIndex` or something and add some comments to the functions. 

`mediump` most often boils down to `highp` these days, as seen on https://webglfundamentals.org/webgl/webgl-precision-check-vertex-shader.html, so there is little gain on having a mediump for subhundres.

larsmoa (approved):
Beautiful @Strepto - thank you!

larsmoa (approved):

larsmoa:
Will be automatically merge when you update your branch, @Strepto

PR #2540, opened 2022-10-06
fix: seeing old frames through in-front geometry

christjt:
# Description

For more context see: https://cognitedata.atlassian.net/browse/REV-530

Caused by multisampled rendertargets not clearing even when calling `renderer.clear`. As a work around we simply don't blend back rendertarget with in-front, when back has not been renderered.


viewer/packages/rendering/src/render-passes/PostProcessingPass.ts
```diff
@@ -86,6 +89,13 @@ export class PostProcessingPass implements RenderPass {
     inFrontBlitObject.name = 'In-front Styling blit object';
     inFrontBlitObject.renderOrder = 2;
 
+    // Removes blending with the back objects framebuffer when it is hidden i.e. not
+    // been rendered. This is a sanity check as well as a workaround for multisampled
+    // rendertargets not being able to clear properly.
```
larsmoa, line 94:
```suggestion
    // Removes blending with the back objects framebuffer when it is hidden i.e. not
    // been rendered. This is a sanity check as well as a workaround for multisampled
    // rendertargets not being able to clear properly (REV-530).
```

larsmoa (approved):
See comment

PR #2542, opened 2022-10-07
fix: remove useless Cognite3DViewer.isBrowserSupported

larsmoa:


haakonflatval-cognite (approved):
RIP `func(): true` notation :(

PR #2543, opened 2022-10-07
refac: change signature of Cognite3DViewer.models

larsmoa:
# Description

As discussed in 4.0 change document.

# Checklist:

Not filled out, too small of a change.


viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -187,8 +187,8 @@ export class Cognite3DViewer {
   /**
    * Gets a list of models currently added to the viewer.
    */
-  public get models(): CogniteModelBase[] {
-    return this._models.slice();
```
haakonflatval-cognite, line 191: Should we just change the type of `this._models` to `(Cognite3DModel | CognitePointCloudModel)[]` instead?

haakonflatval-cognite (changes requested):
See comment

christjt:
The biggest pain point I have with CogniteModelBase is that it is a base class. We should either delete it compleatly or convert it / devide it into interfaces imo.

larsmoa:
Ok, I see the comments - I'll just close this for now.

PR #2550, opened 2022-10-10
fix: point cloud custom transformation

haakonflatval-cognite:
# Description

Point clouds were not translated properly and had their bounding boxes translated wrongly when a custom transformation was applied to them. This fixes that issue, and makes sure that stylable bounding boxes also receive the appropriate transformation when traversed by the user.

A fundamental mistake in the previous code, is that it uses `updateMatrix` on the model, which computes the local matrix from `scale`, `position` and `rotation`, overwriting the matrix we just set. (See https://threejs.org/docs/#manual/en/introduction/Matrix-transformations)

# Checklist:

### DIFF
--- a/viewer/test-utilities/src/createPointCloudModel.ts
+++ b/viewer/test-utilities/src/createPointCloudModel.ts
@@ -20,6 +20,14 @@ import { LocalModelDataProvider } from '../../packages/data-providers';
 import { IPointCloudTreeGeometry } from '../../packages/pointclouds/src/potree-three-loader/geometry/IPointCloudTreeGeometry';
 
 export function createPointCloudModel(modelId: number, revisionId: number): CognitePointCloudModel {
+  const pointCloudNode = createPointCloudNode();
+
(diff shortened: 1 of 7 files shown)


viewer/test-utilities/src/createPointCloudModel.ts
```diff
@@ -41,8 +49,5 @@ export function createPointCloudModel(modelId: number, revisionId: number): Cogn
   const nodeWrapper = new PotreeNodeWrapper(pointCloudOctree, [], Symbol('dummy'), { classificationSets: [] });
 
   const pointCloudNode = new PointCloudNode(potreeGroup, nodeWrapper);
-
-  const pointCloudModel = new CognitePointCloudModel(modelId, revisionId, pointCloudNode);
-
-  return pointCloudModel;
+  return pointCloudNode;
```
christjt, line 52: You can just return the pointcloud node directly. No need to bind it to variable here.

PR #2556, opened 2022-10-11
fix: visualization artifacts due to wrongfully handling cylinder clip planes

christjt:
# Description
Fix for [REV-534](https://cognitedata.atlassian.net/browse/REV-534). Note that fixing this provoked a different issue in the backend which showed that extended centers were incorrect for inner cylinders of general cylinders with thickness. This is in almost all cases un-noticable since the segment usually doesn't stop in thin air, but also this issue only happens when the cylinder slice plane is at a slope (i.e. not the normal (0, 0, 1)). In any case, I have a fix for backend as well here: https://github.com/cognitedata/3d-optimizer/pull/1633. I updated the primitives model from this fix.


viewer/packages/sector-parser/visual-tests/GeneralCylinder.VisualTest.ts
```diff
@@ -0,0 +1,59 @@
...
+  createPrimitiveInterleavedGeometry,
+  GeneralCylinder,
+  PrimitiveName
+} from '../../../test-utilities/src/primitives';
+import { SimpleTestFixtureComponents, SimpleVisualTestFixture } from '../../../visual-tests';
+import { setConeGeometry } from '../src/reveal-glb-parser/primitiveGeometries';
+import { RevealGeometryCollectionType } from '../src/types';
+import { getMaterialsMap } from './testMaterials';
+
+export default class GeneralCylinderVisualTest extends SimpleVisualTestFixture {
```
larsmoa, line 15: First of all, this test is just crazy useful and great - something that probably should be present for all primitives. Is it an idea to generate e.g. 10 different variations of the primitive to cover more possible regressions? I'm especially thinking edge cases

viewer/packages/sector-parser/visual-tests/GeneralCylinder.VisualTest.ts
```diff
@@ -0,0 +1,59 @@
...
+    const generalCylinderMaterial = materials.get(RevealGeometryCollectionType.GeneralCylinderCollection)!;
+    const mesh = new THREE.Mesh(geometry, generalCylinderMaterial);
+    mesh.frustumCulled = false;
+    mesh.onBeforeRender = () => {
+      (generalCylinderMaterial.uniforms.inverseModelMatrix?.value as THREE.Matrix4)?.copy(mesh.matrixWorld).invert();
+      (generalCylinderMaterial.uniforms.cameraPosition?.value as THREE.Vector3)?.copy(camera.position);
+    };
+    group.add(mesh);
+
+    camera.position.set(0, 0, -3);
```
larsmoa, line 46: I notice in the snapshot image that the top of the cylinder isn't visible - I guess it might be more useful if the top was seen as well?

larsmoa (commented):
Added a few initial comments - not well tested yet

larsmoa (approved):
👏

PR #2573, opened 2022-10-17
improvement: make CadModelUpdateHandler trigger sector loading immediately after the camera stops moving

eiriklegernaes:
Previously, due to the way the RxJS observables were structured in `CadModelUpdateHandler`, the loading would not start until about 750 - 1000ms after the camera had stopped. This was only a problem when continuous sector streaming was disabled.

With this change, the sectors will be loaded immediately after the camera stops moving, and if continuous streaming is enabled, every 500ms when the camera is moving.


viewer/packages/api/src/public/RevealManager.ts
```diff
@@ -106,13 +106,14 @@ export class RevealManager {
       !this._lastCamera.position.equals(camera.position) ||
       !this._lastCamera.quaternion.equals(camera.quaternion);
 
+    this._cadManager.updateCamera(camera, hasCameraChanged);
```
christjt, line 109: This will trigger the sector culler on every frame. Any reason you moved it here?

viewer/packages/cad-geometry-loaders/src/CadManager.ts
```diff
@@ -130,9 +130,11 @@ export class CadManager {
...
   }
 
-  updateCamera(camera: THREE.PerspectiveCamera): void {
-    this._cadModelUpdateHandler.updateCamera(camera);
-    this._needsRedraw = true;
+  updateCamera(camera: THREE.PerspectiveCamera, cameraInMotion: boolean): void {
+    this._cadModelUpdateHandler.updateCamera(camera, cameraInMotion);
+    if (cameraInMotion) {
+      this._needsRedraw = true;
+    }
```
christjt, line 137: If the camera has moved then cadManager will have this flag set already. So I don't believe this is needed.

viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
```diff
@@ -69,14 +79,20 @@ export class CadModelUpdateHandler {
...
+        filter((input: CameraInput) => input.cameraInMotion),
+        mergeWith(
+          this._cameraSubject.pipe(
+            distinctUntilKeyChanged('cameraInMotion'),
+            filter((input: CameraInput) => !input.cameraInMotion)
+          )
+        )
+      ),
+      combineLatest([this._clippingPlaneSubject.pipe(startWith([]))]).pipe(map(makeClippingInput), auditTime(250)),
+      this.loadingModelObservable().pipe(auditTime(250))
```
christjt, line 95: Can you give a small TLDR on what the implications are for this change?

christjt (changes requested):
Thank you for the contribution! Overall this looks really nice. Only a few questions / clarifications that should be adressed.

viewer/packages/api/src/public/RevealManager.ts
eiriklegernaes: When the camera pose is equal to that of the previous frame, the `CadManager.updateCamera` function will be called with hasCameraChange equal to false. This first invocation will trigger the sector culling, but subsequent calls will be filtered out (due to the `distinctUntilKeyChanged` RxJS operator in `CadModelUpdateHandler`. 

One could of course do the 'is this the first frame of the camera not moving'-logic in `RevealManager`. I opted to keep things in `CadModelUpdateHandler`, but I'm more than happy to move it if you'd like.

viewer/packages/cad-geometry-loaders/src/CadManager.ts
eiriklegernaes: Ah, okey. Just assumed it was necessary since it was there before. Tested it now, and it seems to work fine without setting the flag. Removing.

viewer/packages/api/src/public/RevealManager.ts
christjt: But why should we trigger an invocation if the camera hasn't changed? If f.ex. the camera doesn't move at all we don't want to drain resources here triggering sector culling.

viewer/packages/cad-geometry-loaders/src/CadModelUpdateHandler.ts
eiriklegernaes: Sure, and I'll preface it by saying I'm no RxJS wizard, so there's probably a better way to do this. 

Two logical changes have happened in more or less the same place:
1. The "global" 250ms auditTime (old line 109) has been removed, and moved into the different observables 'triggers' above (which deals with settings, loading hint, clipping planes, models). The reason for this is that I didn't want the extra delay when the trigger was the camera.
2. The logic for the 'camera trigger' is altered. Previously, it was a combination of every 500ms, and every 250ms check whether it's more than 600ms since the last camera update (at least that's how I read it..). Now, it emits every 500ms if the camera is moving, as well as the first frame of it not moving. The `pipe(auditTime(500),  filter((input: CameraInput) => input.cameraInMotion))` fires every 500ms when the camera is moving, and the `.pipe(distinctUntilKeyChanged('cameraInMotion'), filter((input: CameraInput) => !input.cameraInMotion))` emits when once when the `cameraInMotion` property changes to false.

As before, `SectorLoader.loadSectors` will not do anything if continous streaming is disabled and the camera is moving.

viewer/packages/api/src/public/RevealManager.ts
eiriklegernaes: It gets filtered out, yes

christjt (approved):

PR #2597, opened 2022-10-22
fix: nightly publish workflow

christjt:


.github/workflows/nightly-publish-to-npm.yml
```diff
@@ -1,12 +1,29 @@
 name: Nightly Publish Package to NPM
 on:
   schedule:
-    - cron: '0 5 * * *'
+    - cron: '0 5 * * 1-6'
```
haakonflatval-cognite, line 4: So.. This means 5:00 on Monday-Saturday? I'd prefer 6:00 on Monday-Friday, if that's not too much to ask :P

.github/workflows/nightly-publish-to-npm.yml
```diff
@@ -1,12 +1,29 @@
...
-    - cron: '0 5 * * *'
+    - cron: '0 5 * * 1-6'
+
 jobs:
   deploy-to-npm:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
```
haakonflatval-cognite, line 12: I read that this makes sure the token is not stored locally. Any reason we don't want that?

haakonflatval-cognite (changes requested):
Nice! Just a small suggestion, and a question

.github/workflows/nightly-publish-to-npm.yml
christjt: I think it is due to if we wan't this workflow to do anything we have to add cogniteCICD gihub user. i.e.

git config --global user.email "cognite-cicd@users.noreply.github.com"
git config --global user.name "Cognite CICD"

So whatever credentials it initially sets would be wrong.

PR #2621, opened 2022-10-31
fix: enable depth writing for in-front pass

eiriklegernaes:
#### Type of change
Bug 

## Description :pencil:

The in-front pass doesn't appear to write to the depth buffer. This is likely because disabling depth testing in order draw on top of the back and ghost pass, also disables depth writing. The problem can be fixed by not disabling depth testing, but rather using a different depth function. 

This has not been extensively tested, but it appears to solve the problem.


christjt (approved):
LGTM!

PR #2651, opened 2022-11-04
feat: hide & show measurements

pramodcog:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-571

## Description :pencil:
It adds ability to hide/unhide all measurements in the viewer

In `examples`, under measurement option, an checkbox is add for hiding/showing the measurements added into the viewer.

Go to `examples`, add measurements, uncheck & check the `Show measurement` option.

Unit test are not added as we need to address it for Measurement Tool as documented in [Unit test for Measurement Tool](https://cognitedata.atlassian.net/browse/REV-439)

### DIFF
--- a/documentation/docs/examples/point-to-point-measurement.mdx
+++ b/documentation/docs/examples/point-to-point-measurement.mdx
@@ -53,11 +53,8 @@ measurementTool.on('added', (measurement) => {
 measurementTool.enterMeasurementMode();
 ```
 
-
-
 ### Mode of measurement
 To get the mode of measurement `MeasurementTool.isInMeasurementMode`. It returns `true` if the measurement is active else `false`.
-
 
 ## Set width, color & label units of the measuring line
 To change width, color of the line and label units, use `MeasurementTool.setLineOptions(options: MeasurementOptions)`. MeasurementOptions expects
@@ -149,10 +146,19 @@ setTimeout(() => {
 }, 6000);
 ```
 
+### Hide and show measurements
+It's possible to toggle between hide & show the measurements in the viewer.
+It can be useful to view the details of an object without measurements blocking the view.
+To do this, use `MeasurementTool.visible(enable: boolean)`.
+```jsx
+//Hide all measurements in the viewer
+measurementTool.visible(false);
+```
+
 To remove all measurement from the viewer, use `MeasurementTool.removeAllMeasurement()`
 
 ### Label visibility
-Measurement labels can be hidden/visible using `MeasurementTool.showMeasurementLabels(enable: boolean)`.
+It is also possible to hide/show only the labels of the measurement using `MeasurementTool.showMeasurementLabels(enable: boolean)`.
 ```jsx
 // Hide all measurement lables
 measurementTool.showMeasurementLabels(false);
--- a/documentation/docs/examples/point-to-point-measurement.mdx
+++ b/documentation/docs/examples/point-to-point-measurement.mdx
@@ -147,8 +147,8 @@ setTimeout(() => {
 ```
 
 ### Hide and show measurements
-It's possible to toggle between hide & show the measurements in the viewer.
-It can be useful to view the details of an object without measurements blocking the view.
+It may be useful to view the details of an object without measurements blocking the view.
+It's possible to toggle measurement visibility on and off in the viewer.
 To do this, use `MeasurementTool.visible(enable: boolean)`.
 ```jsx
 //Hide all measurements in the viewer
--- a/viewer/packages/tools/src/Measurement/MeasurementManager.ts
+++ b/viewer/packages/tools/src/Measurement/MeasurementManager.ts
@@ -116,6 +116,16 @@ export class MeasurementManager {
   }
 
   /**
+   * Hide/unhide all measurements
+   * @param enable
+   */
+  visible(enable: boolean): void {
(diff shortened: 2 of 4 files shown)


viewer/packages/tools/src/Measurement/MeasurementManager.ts
```diff
@@ -115,6 +115,16 @@ export class MeasurementManager {
     this._line.updateLineColor(color);
   }
 
+  /**
+   * Hide/unhide all measurements
+   * @param enable
+   */
+  visible(enable: boolean): void {
```
haakonflatval-cognite, line 122:
```suggestion
  showMeasurements(enable: boolean): void {
```
I was just thinking that this method should have a name similar to `showMeasurementLabels`. Or maybe it should be `setVisible(enable: boolean)` :thinking: Any opinions?

viewer/packages/tools/src/Measurement/MeasurementTool.ts
```diff
@@ -248,6 +250,7 @@ export class MeasurementTool extends Cognite3DViewerToolBase {
     this._viewer.off('click', this._handlePointerClick);
     this._events.measurementEnded.fire();
     this._measurementMode = false;
+    this._showMeasurements = true;
```
haakonflatval-cognite, line 253: Does it make sense to always show measurements when exiting the measurement mode?

viewer/packages/tools/src/Measurement/MeasurementTool.ts
```diff
@@ -283,7 +286,9 @@ export class MeasurementTool extends Cognite3DViewerToolBase {
    * @param enable
    */
   setMeasurementLabelsVisible(enable: boolean): void {
-    this._htmlOverlay.visible(enable);
+    if (this._showMeasurements) {
+      this._htmlOverlay.visible(enable);
+    }
```
haakonflatval-cognite, line 291: This seems a bit weird.. Can't the visibility of the labels be toggled even if the lines are invisible? For instance, if you hide all lines in the example, then toggle labels off, then show the lines again, the labels will also be visible, even though their checkbox is not checked.

viewer/packages/tools/src/Measurement/MeasurementTool.ts
```diff
@@ -337,6 +342,19 @@ export class MeasurementTool extends Cognite3DViewerToolBase {
...
 
+  /**
+   * Hide/unhide all measurements
+   * @param enable
+   */
+  visible(enable: boolean): void {
+    this._measurements.forEach(measurement => {
+      measurement.visible(enable);
+    });
+    this._htmlOverlay.visible(enable);
```
haakonflatval-cognite, line 353: Related to previous - do we always want to show labels whenever measurements are toggled on? (I can agree that it makes sense to hide them when the measurement line is hidden, though).

documentation/docs/examples/point-to-point-measurement.mdx
```diff
@@ -149,10 +146,19 @@ setTimeout(() => {
 }, 6000);
 ```
 
+### Hide and show measurements
+It's possible to toggle between hide & show the measurements in the viewer.
+It can be useful to view the details of an object without measurements blocking the view.
```
haakonflatval-cognite, line 151:
```suggestion
It may be useful to view the details of an object without measurements blocking the view.
It's possible to toggle measurement visibility on and off in the viewer.
```

haakonflatval-cognite (changes requested):
Nice feature - a few comments

viewer/packages/tools/src/Measurement/MeasurementManager.ts
pramodcog: We have similar function in `HTMLOverlayTool.visible()`, was following the same pattern

viewer/packages/tools/src/Measurement/MeasurementTool.ts
pramodcog: When measurement is in hide state, all objects related to measurement that is line, labels should be hidden and I think even if we try to enable labels in measurement hidden state, it should not change the labels visibility

viewer/packages/tools/src/Measurement/MeasurementTool.ts
```diff
@@ -283,7 +287,10 @@ export class MeasurementTool extends Cognite3DViewerToolBase {
    * @param enable
    */
   setMeasurementLabelsVisible(enable: boolean): void {
-    this._htmlOverlay.visible(enable);
+    if (this._showMeasurements) {
+      this._htmlOverlay.visible(enable);
+      this._showLabels = enable;
+    }
```
haakonflatval-cognite, line 293: I still think we should be able to toggle the labels on and off "behind the scenes" even though the result will not be visible before the lines are made visible. If a user has hidden both labels and lines, and then want to show both simultaneously, it will be correct only if they turn on the lines first and then the labels, and not the other way around, the way this is written now. I think both orders should work, which they would if we keep track of whether the labels are toggled on regardless of whether they are visible or not. We can do this just by moving the `showLabels` line out of the if-block.

haakonflatval-cognite (changes requested):
Looks good! But I have one last request

haakonflatval-cognite (approved):

PR #2653, opened 2022-11-04
Allow setting batchSize for SinglePropertyFilterNodeCollection execut…

j-bjorne:
#### Type of change

Bug 

#### Jira ticket :blue_book:
https://cognitedata.atlassian.net/browse/BND3D-1608

## Description :pencil:
A 1000 variables is too heavy for the database, resulting in partial table scans with **o**^**n** searches of values, resulting in time out for big models. A simple fix for this is to reduce the batch size down to 10, and at the same time set the limit.

It hasn't been tested in code, but the requests have been run with similar chunking logic in the reveal docs

You can test in docs with 1000 values for filtering doing them with batchSize 1000, and batchSize 10.

  Nope, I am not sure how I do this.
  Nope...

### DIFF
--- a/viewer/packages/cad-styling/src/SinglePropertyFilterNodeCollection.ts
+++ b/viewer/packages/cad-styling/src/SinglePropertyFilterNodeCollection.ts
@@ -59,7 +59,7 @@ export class SinglePropertyFilterNodeCollection extends CdfNodeCollectionBase {
    * @param propertyKey Node property key, e.g. `':FU'`.
    * @param propertyValues Lookup values, e.g. `["AR100APG539","AP500INF534","AP400INF553", ...]`
    */
-  executeFilter(propertyCategory: string, propertyKey: string, propertyValues: string[], batchSize: number): Promise<void> {
+  executeFilter(propertyCategory: string, propertyKey: string, propertyValues: string[], batchSize: number = 1000): Promise<void> {
     const { requestPartitions } = this._options;
 
     const outputsUrl = this.buildUrl();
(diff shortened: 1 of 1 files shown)


viewer/packages/cad-styling/src/SinglePropertyFilterNodeCollection.ts
```diff
@@ -59,7 +59,7 @@ export class SinglePropertyFilterNodeCollection extends CdfNodeCollectionBase {
    * @param propertyKey Node property key, e.g. `':FU'`.
    * @param propertyValues Lookup values, e.g. `["AR100APG539","AP500INF534","AP400INF553", ...]`
    */
-  executeFilter(propertyCategory: string, propertyKey: string, propertyValues: string[]): Promise<void> {
+  executeFilter(propertyCategory: string, propertyKey: string, propertyValues: string[], batchSize: number): Promise<void> {
```
j-bjorne, line 62:
```suggestion
  executeFilter(propertyCategory: string, propertyKey: string, propertyValues: string[], batchSize: number = 1000): Promise<void> {
```

christjt:
@j-bjorne I am very hasitant to move forward with this, as this is in my opinion clearly a backend problem. And "fixing" it here doesn't really get us anywhere as users could simply end up with exactly the same issue using the API directly or by using the SDK.  Also, I find it very unlikly that most users will bother setting this. If the API cannot handle the params that get sent it should return a 4xx, which is then documented and which we can also input validate on Reveal's side.

christjt:
Closing due to uncertanty on how to proceed with this. Lets talk this through if you believe this is a mistake.

PR #2620, opened 2022-10-28
fix: general cylinder inside rendering

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-564

## Description :pencil:

Fixes general cylinder when the camera is inside the cylinder. Also fixes the normal direction when looking at the inside of the cylinder.

## How has this been tested? :mag:

By looking

## Test instructions :information_source:

```
cd viewer && yarn && cd packages/sector-parser && yarn start
```
Have a good hard look

## Checklist :ballot_box_with_check:

- [x] I am proud of this feature.
- [x] I have performed a self-review of my own code.
- [x] I have added PR type (Feat, Bug, Chore, Test, Docs, Style, Refactor) to the PR title.
- [x] I have manually tested this for different scenarios (different models, formats, devices, browsers).
- [x] I have commented my code in hard-to-understand areas.
- [x] I have made corresponding changes to the public documentation.
- [x] I have added unit and visuals tests to prove that my feature works to the best of my ability.
- [x] I have refactored the code for readability to the best of my ability.
- [x] I have checked that my changes do not introduce regressions in the public documentation.
- [x] I have outlined any known defects / lacking capabilities in the contents of this PR.
- [x] I have listed any remaining work that should follow this PR in the description or jira/miro/etc.
- [x] I have added TSDoc to any public facing changes.
- [x] I have added "developer documentation" in any package README.md that is related / applicable to this PR.
- [x] I have noted down and am currently tracking any technical debt introduced in this PR.
- [x] I have thoroughly thought about the architecture of this implementation.
- [ ] I have asked peers to test this feature.

PR #2656, opened 2022-11-04
fix: fix low-precision problems with cylinders

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-572

## Description :pencil:

The PR https://github.com/cognitedata/reveal/pull/2620 broke cylinder rendering when viewed from far away (e.g. when approximating an orthographic camera). This seems to stem from some numeric issues. 

The main factor is apparently that the `dist` in the fragment shader was computed from the fragment in view space, and not from the view position (which would be the intuitive choice). However, fixing that in the above-mentioned PR caused the above-mentioned numeric issues. These are therefore reverted in this PR.

Furthermore, the near-plane distance computed in the vertex shader is also flaky for large distances, so this is clamped to ensure the shader doesn't suddenly decide we're inside a cylinder when we're far away. This seems to cause some trouble, as the near plane takes some time to update to adapt to the model, and it causes heavy clipping. I'll double check if this is was as bad before.

Finally, now that the distance in the fragment shader is not computed from the view position anymore, we might have to take care to check the sign of the dist when finding out whether the fragment is behind the camera.

Manually in the examples.

In examples, replace the line 
```
viewer.loadCameraFromModel(model);
```
in Viewer.tsx with
```
        cameraManager.setCameraState({position: new THREE.Vector3(10000, 0, 0), target: new THREE.Vector3(0, 0, 0)});
        cameraManager.getCamera().fov = 0.2;
        cameraManager.getCamera().updateProjectionMatrix();
```
to simulate an orthographic camera. Start the example and look at the primitives model.


christjt (approved):
LGTM

PR #2692, opened 2022-11-15
feat: Node isolation performance improvement

eiriklegernaes:
## Overview of changes

This PR improves performance when nodes are invisible, either by directly styling them with `visible = false` or because the nodes shouldn't be visible in given render mode.

* Added a `visibleNodesTreeIndices: IndexSet` in `NodeAppearanceTextureBuilder`, along side the back, ghosted and in-front sets. This keeps track of the set of tree indices whose style end up being visible. The logic is also altered to now exclude invisible tree indices from the back, ghosted and in-front sets. This was done to clean up the logic in `populateNodeSets` and to make sure the visible set is represented with as few ranges as possible. The visible set is used in `setModelRenderLayers` to disable drawing of buffers that don't contain any visible tree indices. 
* As pointed out by @christjt, additional speedup was gained by checking the appearance visibility in the vertex shader. Vertices that are deemed to be invisible are set to a hardcoded position in clip space, outside clipping boundary. For geometry types where the existing number of varyings wasn't at the limit, this lookup is only done in the vertex shader. For cone, eccentric cone, ellipsoid segment, general cylinder, general ring and spherical segment, the lookup is still also performed in the fragment shader.
  * This necessitated moving the definition of the render modes into its own glsl file
  * The `determineNodeAppearance` function was refactored into two functions, to allow parsing the appearance texel in the fragment shader without repeating the texture lookup.
  * I also had problems with mismatching precision of the `renderMode` uniform between vertex and fragment shaders, since this now is used in both. Solved by explicitly setting the precision to `lowp` in both shaders.

## Benchmarking

The benchmarking has been done by simply rendering a medium sized asset from a fixed viewpoint, with and without a handful set of nodes isolated (meaning everything else styled invisible). All timings are frametimes for a full frame (without ghosted and in-front passes), with some extra unrelated logic adding overhead. The cad budget was set high enough to never hit the fps cap. Captured on 3.3.0. Take these with a grain of salt.

```
                                            All nodes visible         Most nodes invisible
Before any changes                                62 ms                      66 ms
Using visible set                                 62 ms                      32 ms
Using visible set and shader change               62 ms                      22 ms
```

There is also noticeable improvement when some nodes are styled in-front, since this means a large part of the model will be invisible in the in-front pass, although I haven't attempted to measure this.

Although I haven't noticed any performance decreases, there are some possibilities for regressions:
* The extra check in `setModelRenderLayers` can be a problem on CPU bound setups. Since the visible set is just the union of back, ghosted and in-front sets, the set is strictly speaking not necessary, one might want to remove it. Thoughts?
* The extra texture lookup in many of the vertex shaders can lead to worse performance when this early test never rejects vertices.

### DIFF
--- a/viewer/package.json
+++ b/viewer/package.json
@@ -81,7 +81,7 @@
     "proj4": "2.8.0",
     "rxjs": "7.5.7",
     "skmeans": "0.11.3",
-    "three-stdlib": "2.17.3"
+    "three-stdlib": "2.18.1"
   },
   "devDependencies": {
     "@azure/msal-browser": "2.31.0",
--- a/viewer/yarn.lock
+++ b/viewer/yarn.lock
@@ -845,7 +845,7 @@ __metadata:
     skmeans: 0.11.3
     stats.js: 0.17.0
     three: 0.146.0
-    three-stdlib: 2.17.3
+    three-stdlib: 2.18.1
     ts-jest: 29.0.3
     ts-loader: 9.4.1
     tsc-alias: 1.7.1
@@ -11935,9 +11935,9 @@ resolve@^1.20.0:
   languageName: node
   linkType: hard
 
-"three-stdlib@npm:2.17.3":
-  version: 2.17.3
-  resolution: "three-stdlib@npm:2.17.3"
+"three-stdlib@npm:2.18.1":
+  version: 2.18.1
+  resolution: "three-stdlib@npm:2.18.1"
   dependencies:
     "@babel/runtime": ^7.16.7
     "@types/offscreencanvas": ^2019.6.4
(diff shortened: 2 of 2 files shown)


viewer/packages/rendering/src/glsl/sector/primitives/eccentricCone.frag
```diff
@@ -26,15 +25,13 @@ in vec4 v_centerB;
 in float height;
 in vec3 v_color;
 in vec3 v_normal;
+
 in highp vec2  v_treeIndexPacked;
 
 void main()
 {
     highp float v_treeIndex = unpackTreeIndex(v_treeIndexPacked);
     NodeAppearance appearance = determineNodeAppearance(colorDataTexture, treeIndexTextureSize, v_treeIndex);
```
haakonflatval-cognite, line 34: I think we should somehow document that we are doing the same lookup in the vertex and fragment shaders because we are hitting the attribute limit. I suggest just adding a comment along the lines of `// Redo appearance texture lookup from vertex shader due to limit in transferable attributes`, in this file and all the other fragment shaders where this occurs.

Open to other suggestions on how to document this.

haakonflatval-cognite (changes requested):
Thanks for the contribution!

Looks very good - it was hard to find something to nitpick on :sweat_smile: I've just added a suggestion for documenting the reason for duplicate texture lookups, otherwise looks good to me :+1:

viewer/packages/rendering/src/glsl/sector/primitives/eccentricCone.frag
eiriklegernaes: Sounds like a good idea! Just pushed that

haakonflatval-cognite (approved):
Very nice! LGTM :+1:

PR #2701, opened 2022-11-16
fix: set 360 image events limit to max (1000), vs. the default 100

christjt:
#### Type of change

Bug 

Bumps the limit on the number of 360 images that can be loaded in a single set.


haakonflatval-cognite (approved):
LGTM, but is there really that much of a difference?

PR #2727, opened 2022-11-23
fix: cad intersection model data incorrect

pramodcog:
#### Type of change

Bug 

#### Jira ticket :blue_book:
https://cognitedata.atlassian.net/browse/REV-595

## Description :pencil:
Fix for incorrect model data in `Cognite3DViewer.getIntersectionFromPixel()` when two or more CAD models are added into the viewer.

1. Load https://cognitedata.github.io/reveal-docs/docs/examples/cad-basic/
2. Run the code mentioned in the JIRA reproducible steps


viewer/packages/cad-model/src/picking/PickingHandler.ts
```diff
@@ -95,11 +95,16 @@ export class PickingHandler {
 
     try {
       for (const cadNode of cadNodes) {
+        //Make current CAD node visible & hide others to fix REV-595
```
haakonflatval-cognite, line 98: I think it's a bit unnecessary to mention the issue here in the code :sweat_smile: I expect we won't care about whether or not this was a fix, just that it's correct

Also - I think we usually have a space between the `//` and the comment? :thinking:

haakonflatval-cognite (changes requested):
Nice fix! Just some nitpick on the comments

viewer/packages/cad-model/src/picking/PickingHandler.ts
larsmoa: Better to store original visibility value and restore original value in the finally block.

pramodcog: @larsmoa do we allow user to change the visibility of a node? By yesterday discussion, I thought you mentioned, we do not expose `CADNode` as public, I know that user can still make it visible false through `Scene`, just for my knowledge to understand if we expose something to the user to make objects visibility in viewer to false!

pramodcog: I found `CogniteCadModel.visible`, so I guess storing previous state make sense

viewer/packages/cad-model/src/picking/PickingHandler.ts
```diff
@@ -92,16 +92,26 @@ export class PickingHandler {
...
+    let cadNodesVisibility: boolean[];
 
     try {
       for (const cadNode of cadNodes) {
+        // Make current CAD node visible & hide others
+        cadNodesVisibility = cadNodes.map(p => {
+          const previousVisibility = p.visible;
+          p.visible = false;
+          return previousVisibility;
+        });
```
larsmoa, line 104: Why do this for every model? Should only be done once outside the for loop. This will not work as you expect, since only the state after the last model is processed is used.

larsmoa: Should be easily unit testable.

viewer/packages/cad-model/src/picking/PickingHandler.ts
```diff
@@ -92,16 +92,24 @@ export class PickingHandler {
   ): Promise<IntersectCadNodesResult[]> {
     const results: IntersectCadNodesResult[] = [];
     const release = await this._mutex.acquire();
+    const cadNodesVisibility = cadNodes.map(p => p.visible);
```
haakonflatval-cognite, line 95: I think we can avoid this one and only use the `visibleCadNodes` array. In the end, we can just set all of the elements in that array to visible :+1:

viewer/packages/cad-model/src/picking/PickingHandler.ts
```diff
@@ -92,16 +92,24 @@ export class PickingHandler {
...
+    // Get CadNodes which are visible
+    const visibleCadNodes = cadNodes.filter(node => node.visible === true);
```
haakonflatval-cognite, line 97:
```suggestion
    const visibleCadNodes = cadNodes.filter(node => node.visible);
```

haakonflatval-cognite (changes requested):
Looks better! Think we can make it a little cleaner, though

haakonflatval-cognite (approved):
Nice! LGTM :+1:

PR #2728, opened 2022-11-23
chore: change cdf env loading for visual tests to be dynamic

christjt:
#### Type of change

Chore 

## Description :pencil:

The env file was previously tracked by git which was very annoying to work with (constantly had to revert the env I have set such that it isn't pushed).


haakonflatval-cognite (approved):
Looks good! :+1:

PR #2761, opened 2022-12-01
fix: add proper click event instead of mouse up

christjt:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-608
https://cognitedata.atlassian.net/browse/REV-611

## Description :pencil:

Adds proper click instead of "pointer up" when clicking a 360 image icon. Also returns null for intersection result if a 360 image icon is clicked.


haakonflatval-cognite (approved):
I like it

PR #2767, opened 2022-12-02
feat: add Image360Collection return type for easier extendability in the future

christjt:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-605
https://cognitedata.atlassian.net/browse/REV-621
https://cognitedata.atlassian.net/browse/REV-622

## Description :pencil:

Adds a Image360Collection type that wraps a list of 360 images such that we can add additional per-collection funcionality in the future without breaking the API.

Also fix small nits such as off the outer rim of the 360 image icon being a little bit off, as well as reducing the max size of the icon.


haakonflatval-cognite (approved):
Looks good! When the CI passes, at least

PR #2772, opened 2022-12-05
doc: Add breaking changes into the migration documentation for next

christjt:
#### Type of change

Docs 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-609

## Description :pencil:

Add breaking changes into the migration documentation for NEXT.


documentation/docs/migration-guide.md
```diff
@@ -9,6 +9,74 @@ import { DemoWrapper } from '@site/docs/components/DemoWrapper';
...
+
+## Cognite3DViewer.setBackgroundColor now alpha
+The `setBackgroundColor` now supports settings alpha value. 
+This becomes relevant when you want your 3D content to blend with your website.
+
+## Cognite3DViewer.isBrowserSupported has been removed
+This method has been left unused for a long time, and Reveal now supports all major browsers with WebGL 2.0 capabilities.
+
+## Cognite3DModel has been renamed to CogniteCadModel
+Renamed for clarity, as technically a point cloud can also be viewer as a 3D model.
```
haakonflatval-cognite, line 50:
```suggestion
Renamed for clarity, as technically a point cloud can also be viewed as a 3D model.
```

documentation/docs/migration-guide.md
```diff
@@ -9,6 +9,74 @@ import { DemoWrapper } from '@site/docs/components/DemoWrapper';
...
+
+## Cognite3DViewer.renderer has been removed
+Getter for the renderer used by Reveal has been removed.
+
+## Cognite3DViewer.getScene has been removed
+Any direct manipulation of the three.js scene which Reveal uses for rendering can cause undefined behaviour, so in an effort to limit unintentional mistakes it has now been removed.
+Please reach out if there is any missing functionality that Reveal should expose.
+
+## CameraManager now requires two new methods: activate and deactivate
+These new methods are used for transitioning to and from other `CameraManager`s and lets each implementation decide what this functionality should look like.
```
haakonflatval-cognite, line 67: Perhaps we can move the latter section about `cameraStop` event handling in the CameraManager here as well? Just so that the CameraManager-related sections are bound together

haakonflatval-cognite (changes requested):
Looks good at first glance :+1: A couple of minor suggestions

documentation/docs/migration-guide.md
christjt: Makes sense, will fix!

documentation/docs/migration-guide.md
```diff
@@ -9,6 +9,71 @@ import { DemoWrapper } from '@site/docs/components/DemoWrapper';
...
+imports will still work the same way they did with Reveal 3.x. 
+Note that this may require a bump in typescript version.
+If this is not possible for you, then you need to change all imports to be `from @cognite/reveal` vs. f.ex. `from @cognite/reveal/tools`. 
+
+## Three.js is now a peer-dependency
+For Reveal 4 we have changed threejs from being exported from Reveal and rather require applications that consume Reveal to import it.
+There are multiple reasons for this, but a major pain point has been that Reveal does not expose the three.js examples in its export, causing applications to have to install threejs anyway.
+In addition, we also see a lot of applications ending up having multiple bundles of three.js which we are trying to alliviate.
+
+## Exploded view tool has been removed
```
larsmoa, line 29: I think it's enough to have a list at the end of the document with "Removed functionality" and list the individual removed items here. They are removed because they are not used, so it's not very important to highlight them early in the doc

documentation/docs/migration-guide.md
```diff
@@ -9,6 +9,71 @@ import { DemoWrapper } from '@site/docs/components/DemoWrapper';
 
 This document is an overview of some important differences between Reveal 3.x and Reveal 4.x.
```
larsmoa, line 10: I think it would be good to rename the page to "Migrating from Reveal 3 to 4"

documentation/docs/migration-guide.md
```diff
@@ -9,6 +9,71 @@ import { DemoWrapper } from '@site/docs/components/DemoWrapper';
...
+Methods for settings and getting options has been added to support dynamically updating the state.
+
+## CogniteCadModel and PointCloudModel no longer inherit from the three.js Object3D base class
+This prevents usages that Reveal does not handle such as reparenting and other mutable operations on the three.js object.
+Methods such as setting the transformation or visibility are available
+
+## contiuousModelStreaming option has been promoted to opt-out
+This means that CAD models will now stream continuously (vs. just when camera stops moving) by default.
+This can be reverted by setting the `contiuousModelStreaming` option to false when initializing the `Cognite3DViewer`.
+
```
larsmoa, line 76: CSP requirements should be listed. Do we have a separate page that explains these requirements?

documentation/docs/migration-guide.md
```diff
@@ -9,6 +9,71 @@ import { DemoWrapper } from '@site/docs/components/DemoWrapper';
...
 ## Potree prefix removed
```
larsmoa, line 77: This is removed in version 3?

documentation/docs/migration-guide.md
```diff
@@ -9,6 +9,71 @@ import { DemoWrapper } from '@site/docs/components/DemoWrapper';
...
+For Reveal 4 we changed into using a single entry point for everything: `@cognite/reveal`.
+So extensions and tools also should be imported from this entry point.
+We have however also added support for node's module `exports` such that if your tsconfig is setup with:
+```ts
+"moduleResolution": "Node16",
+```
+
+imports will still work the same way they did with Reveal 3.x. 
+Note that this may require a bump in typescript version.
+If this is not possible for you, then you need to change all imports to be `from @cognite/reveal` vs. f.ex. `from @cognite/reveal/tools`.
```
larsmoa, line 22: It would be good to show what old tools import looks like with a code snippet, and what new import looks like here and then explain that node16 can be used to use `@cognite/reveal/tools`

larsmoa (changes requested):

PR #2801, opened 2022-12-14
fix: AssetNodeCollection.clear() doesn't clear getAreas()

larsmoa:
#### Type of change

Bug 

## Description :pencil:

`AssetNodeCollection.clear()` must also clear areas, to avoid `getAreas()` to return old collection.

Run docs with the following snippet:
```
const assetNodes = new AssetNodeCollection(sdk, model);
assetNodes.executeFilter({ assetId: 4222791488928479 }).then(() => {
  assetNodes.clear();
  alert('Is ares empty? ' + assetNodes.getAreas().isEmpty);
});
```

Before `isEmpty` would be false.


haakonflatval-cognite (approved):
LGTM :+1:

PR #2804, opened 2022-12-15
bug: Fixed controls in DefaultCameraManager not being disabled when the manager is deactivated

astrid-kg:
#### Type of change

Bug 

#### Jira ticket :blue_book:

None

## Description :pencil:

I noticed that when gizmos are used in example (Last Clicked Node -> drag and Add GLTF model with gizmo) the camera input is still handled when using the DefaultCameraManager. The manager is deactivated during drag, but the controls are not. This change fixes this.

Note: This is already implemented and working for CustomCameraManager.

## How has this been tested? :mag:

I tested in examples project. Using the two gizmos described above. Also switching between Default and Custom camera manager and going in and out of 360 images.

## Test instructions :information_source:

See above.

## Checklist :ballot_box_with_check:

- [x] I am proud of this feature.
- [x] I have performed a self-review of my own code.
- [x] I have added PR type (Feat, Bug, Chore, Test, Docs, Style, Refactor) to the PR title.
- [x] I have manually tested this for different scenarios (different models, formats, devices, browsers).
- [ ] I have commented my code in hard-to-understand areas.
- [ ] I have made corresponding changes to the public documentation.
- [ ] I have added unit and visuals tests to prove that my feature works to the best of my ability.
- [x] I have refactored the code for readability to the best of my ability.
- [x] I have checked that my changes do not introduce regressions in the public documentation.
- [ ] I have outlined any known defects / lacking capabilities in the contents of this PR.
- [ ] I have listed any remaining work that should follow this PR in the description or jira/miro/etc.
- [ ] I have added TSDoc to any public facing changes.
- [ ] I have added "developer documentation" in any package README.md that is related / applicable to this PR.
- [ ] I have noted down and am currently tracking any technical debt introduced in this PR.
- [x] I have thoroughly thought about the architecture of this implementation.
- [ ] I have asked peers to test this feature.

PR #2806, opened 2022-12-15
refactor: recalculateBoundingBox() is only executed when 3D models are added/removed/changed

astrid-kg:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-598

## Description :pencil:

 - **Note: The result from `computeBoundsOfMostGeometry` is already cached so this should not effect performance. See code from SectorScene.ts below.**

- Removed `recalculateBoundingBox` from animate.

- Renamed `recalculateBoundingBox` to `forceUpdateCameraNearAndFarBoundingBox` and made it public.

- Added calls to `forceUpdateCameraNearAndFarBoundingBox` everywhere we add or remove models to the viewer.

- Updated 'Reveal exampes' to call `forceUpdateCameraNearAndFarBoundingBox` when there are changes to models.

Cached `computeBoundsOfMostGeometry`:
```
getBoundsOfMostGeometry(): THREE.Box3 {
    this._cachedBoundsOfMostGeometry = this._cachedBoundsOfMostGeometry ?? this.computeBoundsOfMostGeometry();
    return this._cachedBoundsOfMostGeometry;
  }
```

[Finally, note that in our Example project gizmos have a huge bounding box, and since this is also included in the forceUpdateCameraNearAndFarBoundingBox the final bounding box will be too large once a gizmo is added.](https://cognitedata.atlassian.net/browse/REV-598?focusedCommentId=222338)

Using 'Reveal Examples'  I have removed, added and moved CDF models (cad&pc) and GLTF models, while checking that the bounds calculated in recalculateBoundingBox looks correct and are updated when expected.

See above. Node that you need [this change](https://github.com/cognitedata/reveal/pull/2804) for gizmos to work correctly.


viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -141,6 +141,7 @@ export class Cognite3DViewer {
   private latestRequestId: number = -1;
   private readonly clock = new THREE.Clock();
   private _clippingNeedsUpdate: boolean = false;
+  private _nearAndFarPlaneNeedsUpdate: boolean = false;
```
Savokr, line 144: I think the name for this variable should be different because near and far planes are updated almost every frame because of camera position change. So, I propose something like `_boundingBoxNeedsUpdate`.

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1252,7 +1257,10 @@ export class Cognite3DViewer {
     if (isVisible) {
       const camera = this.cameraManager.getCamera();
       TWEEN.update(time);
-      this.recalculateBoundingBox();
+      if (this._nearAndFarPlaneNeedsUpdate) {
```
Savokr, line 1260: In general this should probably work but I am not sure how we should handle cases where custom objects are not static, i.e. bounding box can change without adding/removing any objects. Because with this approach, custom objects that won't fit inside initial model bounding box could be clipped during rendering. Besides that, I guess it can apply to the model itself because it can have some animations/interactions as you've probably seen in the docs.

Savokr (changes requested):
Looks good but have one suggestion and more general question.

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
Savokr: Probably, one approach is to expose bounding box recalculation method to the user or add a flag that will enable/disable this recalculation for each frame. And the last option is to go through custom objects every frame but save time from model's bounding box.

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
astrid-kg: Roger 👍

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1220,6 +1223,34 @@ export class Cognite3DViewer {
     return this.intersectModels(offsetX, offsetY);
   }
 
+  /**
+   * Update near and far plane buffers by creating a combined bounding box from all objects in the scene.
+   * Note: this is automatically called when 3D models are added to or removed from the viewer.
+   */
```
larsmoa, line 1229: Say something about that this only should be necessary if you have custom objects whose bounds change.

larsmoa: Also, I think this is a bit dodgy - when `automaticNearFarPlane` is false, this will have no effect. As a user I would expect that this _always_ sets near/far, regardless of the `automaticNearFarPlane`-flag.

Savokr: Agree, maybe we can rename the method again to something like `forceUpdateCameraNearAndFarBoundingBox`, so that users understand that it is not about updating actual near and far values but just some data for their calculations.

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1224,6 +1227,35 @@ export class Cognite3DViewer {
     return this.intersectModels(offsetX, offsetY);
   }
 
+  /**
+   * Recalcualte the bounding box used to determine near and far planes on the camera. This should only be
+   * necessary if you have custom objects whose bounds change, as this calculation will be done automatically
+   * when models are added to or removed from the viewer.
+   */
+  public forceUpdateCameraNearAndFarBoundingBox(): void {
```
larsmoa, line 1235: Can't this simply update near and far planes, always? Even when automaticNearFarPlane is false. There is no reason anyone would call this if they do not wish to update the near/far planes - is there?

larsmoa:
Marking do-not-merge because we need to get some better numbers before we see if this fix should be added.

PR #2818, opened 2022-12-19
Avoid disposing of external renderer

eiriklegernaes:
## Description :pencil:

The viewer should not dispose of the renderer if that renderer was supplied as an option when the viewer was constructed. This PR fixes this by adding a flag which signifies whether or not the viewer owns the renderer.

Additionally, the `BasicPipelineExecutor.resize` function assumed that the camera aspect ratio was correct if the renderer's resolution matched the desired resolution (as computed from the DOM element size). This assumption failed when using an externally supplied renderer setup with the correct resolution.


haakonflatval-cognite (approved):
Nice! LGTM :+1:

PR #2825, opened 2022-12-20
Fix AxisViewTool animation

eiriklegernaes:
## Description :pencil:

The AxisViewTool currently does not keep the camera target constant during its animation. This is barely noticeable with the default field of view, but becomes a problem when using very narrow field of views. This PR alters the way the camera position and rotation is calculated, to make the effective camera target stationary during the animation. 

The `setCameraState` function with position and target arguments used at the end of the animation is inherently problematic in the "up" or "down" direction. I don't know why this seemingly worked before, but with the new calculation I had make sure the target and position were exactly on top of each other, which makes threejs use a different "up" direction when computing the camera rotation from position and target. Not sure if this could/should be done differently...


haakonflatval-cognite (approved):
Very nice! Good to see the math cleaned up as well :+1:

PR #2843, opened 2023-01-03
fix: only add exit 360 image key listener when entering 360 image

christjt:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-642

## Description :pencil:

Fixes issue when resetting camera position to (0, 0, 0) when clicking escape key when not in 360 image.

Clicking escape button in examples should not have any effect.


pramodcog (approved):
LGTM!

PR #2849, opened 2023-01-05
fix: don't enable click on invisible point cloud objects

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-626

## Description :pencil:

Don't allow points styled invisible to be clicked.

This is done by properly using the object-to-style texture with the picking material.

See Jira issue

### DIFF
--- a/viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePickerHelper.ts
+++ b/viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePickerHelper.ts
@@ -205,6 +205,7 @@ export class PointCloudOctreePickerHelper {
     pickMaterial.minSize = nodeMaterial.minSize;
     pickMaterial.maxSize = nodeMaterial.maxSize;
     pickMaterial.classification = nodeMaterial.classification;
+    pickMaterial.objectAppearanceTexture = nodeMaterial.objectAppearanceTexture;
   }
 
   public static updatePickRenderTarget(pickState: IPickState, width: number, height: number): void {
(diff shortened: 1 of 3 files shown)


viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePickerHelper.ts
```diff
@@ -205,6 +205,7 @@ export class PointCloudOctreePickerHelper {
     pickMaterial.minSize = nodeMaterial.minSize;
     pickMaterial.maxSize = nodeMaterial.maxSize;
     pickMaterial.classification = nodeMaterial.classification;
+    pickMaterial.objectAppearanceTexture = nodeMaterial.objectAppearanceTexture;
```
Savokr, line 208: Looks totally legit but found a funny bug. Texture contains not only hidden/visible information but also color, so because of that result of the picking is wrong when any of the objects has some color applied (this results in wrong `zoomToCursor` behaviour since picked point is not actually unde the cursor). See screenshot below with blue coloured pipe in `pointIndex` rendering mode that is used by picking. 

Screenshot 2023-01-05 at 15 01 09

So, the solution would be to copy only "visibility" part of the texture and the rest of it should be with default values. Not sure where to place this logic, maybe in some utility function.

Savokr (changes requested):
Looks good but found a funny bug, should be fixed before merging.

viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePickerHelper.ts
haakonflatval-cognite: Fixed by not using style color when rendering point index

Savokr (approved):
Looks good now! 👍

PR #2869, opened 2023-01-13
feat: Zoom to cursor when mouse-scrolling in 360 images

astrid-kg:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-659

## Description :pencil:

In the `StationaryCameraManager` we use FOV to zoom. I now calculate a ray to the mouse position before and after the FOV is changed, and use the arc between these vectors to rotate the camera so that the cursor stays in the same position. In order to keep the "up-vector" constant the rotation on the camera is set using `lookAt,` which will preserve the up direction.

Zoom is not reset between points, but this is something we want to do. I tried a simple implementation of this, but the alpha&position animation we do during transitions looks broken when the FOV is not animated.  I have made an additional Jira task for this.

https://cognitedata.atlassian.net/browse/REV-689

I have tested this using Reveal Examples and the MAUI_A model. I have loaded 360 images, zoomed in/out and transitioned between points. To make sure input works as intended after 360 images, I have also tested input after exiting 360 mode.

Tested in Chrome & Firefox.

See above. [Details from Slack on how to load images](https://cognitedata.slack.com/archives/G0115JQ8JM8/p1669380585627129).


viewer/packages/camera-manager/src/StationaryCameraManager.ts
```diff
@@ -176,8 +176,34 @@ export class StationaryCameraManager implements CameraManager {
 
   private readonly zoomCamera = (event: WheelEvent) => {
     const sensitivityScaler = 0.05;
+    const preCursorRay = this.getCursorRay(event).normalize();
+    const lastFov = this._camera.fov;
+
     this._camera.fov = Math.min(Math.max(this._camera.fov + event.deltaY * sensitivityScaler, 10), this._defaultFOV);
     this._camera.updateProjectionMatrix();
```
haakonflatval-cognite, line 183: Can this also be moved to after the `if`-statement below?

viewer/packages/camera-manager/src/StationaryCameraManager.ts
```diff
@@ -176,8 +176,34 @@ export class StationaryCameraManager implements CameraManager {
...
+
+    // When zooming in the camera is rotated towards the cursor position
+    const postCursorRay = this.getCursorRay(event).normalize();
+    const arcBetweenRays = new THREE.Quaternion().setFromUnitVectors(postCursorRay, preCursorRay);
+    const forwardVector = this._camera.getWorldDirection(new THREE.Vector3()).clone();
+
+    forwardVector.applyQuaternion(arcBetweenRays);
+    const targetWorldCoordinates = new THREE.Vector3().addVectors(
+      this._camera.position,
+      forwardVector.multiplyScalar(1)
```
haakonflatval-cognite, line 195:
```suggestion
      forwardVector
```

haakonflatval-cognite (approved):
A couple of comments, but looks good :+1:

viewer/packages/camera-manager/src/StationaryCameraManager.ts
astrid-kg: The preCursorRay needs to happen before fov is changed, but I moved the check for if the fov has changed further up so that we do no extra work if there is no change in fov. 👍

viewer/packages/camera-manager/src/StationaryCameraManager.ts
astrid-kg: 👍

PR #2911, opened 2023-01-26
[Issue] Picking Handler Optimization Discussion

Strepto:
#### Type of change

This change is not for merging, but used as a discussion for two possible optimizations to increase performance when raycasting with multiple models.

See the comments on the code for more information. This "replaces" an issue, as I already toyed with the code here.

This issue is created as a follow up for a discussion with @christjt 

#### Jira ticket :blue_book:

N/A

## Description :pencil:

This PR adds one file out of a few that would be changed. Its just added for a discussion point and inspiration for a possible future change.

N/A

Testing performance with multiple models, even with just a single model can highlight the performance differences between a hit and a non-hit on the bounding boxes.

DO NOT MERGE THIS. Its not for merging

### DIFF
--- a/viewer/packages/cad-model/src/picking/PickingHandler.ts
+++ b/viewer/packages/cad-model/src/picking/PickingHandler.ts
@@ -15,6 +15,7 @@ import { SceneHandler, WebGLRendererStateHelper } from '@reveal/utilities';
 import { CadNode } from '../wrappers/CadNode';
 import { readPixelsFromTargetAsync } from './readPixelsFromTargetAsync';
 import { Mutex } from 'async-mutex';
+import { Raycaster } from 'three';
 
 type PickingInput = {
   normalizedCoords: {
@@ -31,9 +32,12 @@ type TreeIndexPickingInput = PickingInput & {
   cadNode: CadNode;
 };
 
-type IntersectCadNodesResult = {
+export type IntersectResult = {
   distance: number;
   point: THREE.Vector3;
+};
+
+export type IntersectCadNodesResult = IntersectResult & {
   treeIndex: number;
   cadNode: CadNode;
   object: THREE.Object3D; // always CadNode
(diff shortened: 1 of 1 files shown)


viewer/packages/cad-model/src/picking/PickingHandler.ts
```diff
@@ -133,11 +189,9 @@ export class PickingHandler {
     if (treeIndex === undefined) {
       return undefined;
     }
-    const depth = await this.pickDepth(pickInput, async);
+    const perspectiveDepthNormalized = await this.pickDepth(pickInput, async);
```
Strepto, line 192: Below here is refactorings only to reduce code duplication for the Depth only raycast

Strepto (commented):
Comments on the two optimization approaches.

christjt:
@Strepto Ok, so looking at this a little more high level, what I could tell here is this:

1.  Raycast the the boundingboxes of the models, if the ray doesn't intersect a bounding box then we don't need to do picking on it.
2. Sort the intersecting models bbs front to back and pick in a front-to-back order. If the picked distance is closer than the next bounding box(es) closest hit, then you can prune all behind models (i.e. no need to run picking).

Anything I missed here?

Strepto:
> @Strepto Ok, so looking at this a little more high level, what I could tell here is this:
> 
> 1. Raycast the the boundingboxes of the models, if the ray doesn't intersect a bounding box then we don't need to do picking on it.

Yep.

> 2. Sort the intersecting models bbs front to back and pick in a front-to-back order. If the picked distance is closer than the next bounding box(es) closest hit, then you can prune all behind models (i.e. no need to run picking). (there is of course some edge-cases that need to be handled here as in the case where you are inside the bounding box of multiple models.)

If you are inside a bounding box the distance will be zero and the distance compared to the hit position will always be lower. But ofc needs validation. 

> Anything I missed here?

The depth only raycast avoids multiple re-renders as it does not care which model it hits, so it might be a useful addition.

Otherwise I think thats it.

christjt:
> If you are inside a bounding box the distance will be zero and the distance compared to the hit position will always be lower. But ofc needs validation.

Right, but need to have some logic that guarantees any sectors that you are inside to be rendered. But should be quite trivial.

pramodcog:
@Strepto and @christjt shall I close this PR as picking improvement discussed in here is implemented in https://github.com/cognitedata/reveal/pull/2970 & https://github.com/cognitedata/reveal/pull/2953 which are merged into the master?

Strepto:
👍

pramodcog:
Has been merged to master with https://github.com/cognitedata/reveal/pull/2970 and https://github.com/cognitedata/reveal/pull/2953

PR #2914, opened 2023-01-27
fix: entered 360 image being purged from cache

christjt:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-717

## Description :pencil:

Fixes an issue where currently opened / entered 360 images could be purged from cache.

unit test :) 
But also the image you are in should not be exited when hovering on > 10 icons of other images.


pramodcog (approved):
Looks good

PR #2969, opened 2023-02-14
fix: disallow opening multiple 360 images

christjt:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-767

## Description :pencil:

Fixes an issue where multiple 360 images can be entered / opened if you click another 360 images before the first one completes loading.

Open any 360 image set, throttle your network speed in your browser and click two 360 images. Previously this would open both (once the finish downloading). Current behaviour should be to transition into the last clicked.


haakonflatval-cognite (approved):
LGTM :+1:

PR #2973, opened 2023-02-15
chore: add metrics for 360 image add/enter/transition/exit

christjt:
#### Type of change

Chore 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-764

### DIFF
--- a/viewer/packages/api/src/public/migration/Cognite3DViewer.ts
+++ b/viewer/packages/api/src/public/migration/Cognite3DViewer.ts
@@ -736,6 +736,13 @@ export class Cognite3DViewer {
       preMultipliedRotation
     );
 
+    const numberOf360Images = image360Collection.image360Entities.length;
+
+    MetricsLogger.trackEvent('360ImageCollectionAdded', {
+      datasource,
+      numberOf360Images
+    });
+
     return image360Collection;
   }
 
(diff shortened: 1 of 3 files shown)


viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -736,6 +736,13 @@ export class Cognite3DViewer {
       preMultipliedRotation
     );
 
+    const numberOf360Images = image360Collection.image360Entities.length;
+
+    MetricsLogger.trackEvent('360ImageCollectionAdded', {
```
pramodcog, line 741: If `numberOf360Images` is `0`, should we track the event as `360ImageCollectionAdded`

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -736,6 +736,13 @@ export class Cognite3DViewer {
...
+      datasource,
+      numberOf360Images
+    });
```
pramodcog, line 744: If `numberOf360Images` is `0`, should we track the event as `360ImageCollectionAdded`

christjt: sure, why not? Gives us a good metric on how often people send in and invalid filter, which is good to know.

pramodcog: 👍

pramodcog (approved):
LGTM

PR #2980, opened 2023-02-17
feat: add average framerate user metrics  

Savokr:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-742

## Description :pencil:

Adds new Mixpanel event called `sessionEnded` that is sent together with fps value and session length in milliseconds when user ends reveal "session". By "session", I mean time from opening/"resuming" a page with reveal viewer and closing the tab/switching to other tab/navigating to a different part of the same page where reveal window is not displayed anymore.

Tested with local documentation on desktop (Chrome, Safari, Firefox) and mobile (Safari, iOS).

1. Open documentation preview link and load any example with live viewer.
2. Navigate around the viewer a little bit to "gather" fps data.
3. Open Chrome DevTools -> Network tab.
4. Switch to another tab and back or go to different documentation example preserving network log.
5. Notice that there is new request sent to mixpanel, you can also check fps value inspecting request payload.


documentation/docs/components/Cognite3DViewerDemo.tsx
```diff
@@ -23,31 +23,34 @@ export default function Cognite3DViewerDemo({
 }: DemoProps & OwnProps) {
   const canvasWrapperRef = useRef(null);
 
-  useEffect(() => {
```
haakonflatval-cognite, line 26: What is the difference here, using `useMemo` instead of `useEffect`?

viewer/packages/metrics/src/FpsLogger.ts
```diff
@@ -0,0 +1,84 @@
...
+      this._renderedFramesPerCameraMovement = 0;
+      this._wasPreviousAnimationFrameRendered = false;
+    }
+
+    if (this._wasPreviousAnimationFrameRendered) {
+      this._renderedFramesPerCameraMovement++;
+    } else {
+      this._renderedFramesStartTime = Date.now();
+      this._wasPreviousAnimationFrameRendered = true;
+    }
```
haakonflatval-cognite, line 45: I think I get the gist of the logic here - I'd just like it to be factored into more functions, to make it easier to see what happens :+1:

haakonflatval-cognite (changes requested):
Looks good - a few suggestions

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -139,7 +139,9 @@ export class Cognite3DViewer {
   private isDisposed = false;
 
   private latestRequestId: number = -1;
-  private readonly clock = new THREE.Clock();
+  private readonly fpsLogger: FpsLogger;
+
+  private readonly cameraManagerClock = new THREE.Clock();
```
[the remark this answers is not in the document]
Savokr: Sound better for me and also used specifically for `cameraManager`.

documentation/docs/components/Cognite3DViewerDemo.tsx
Savokr: It is a more proper way of storing a variable that is "persistent" between component re-renders. In current component implementation it doesn't bring much benefits but makes viewer accessible to other hooks/html in this component, not only one `useEffect`.

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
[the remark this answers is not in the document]
haakonflatval-cognite: Alright, fair enough :+1:

documentation/docs/components/Cognite3DViewerDemo.tsx
haakonflatval-cognite: Nice! :+1:

Savokr: Okay, after some tests. Previous version seems to work faster for some reason, so reverting to it.

viewer/packages/metrics/src/FpsLogger.ts
Savokr: Refactored

documentation/docs/components/Cognite3DViewerDemo.tsx
```diff
@@ -35,17 +35,15 @@ export default function Cognite3DViewerDemo({
       antiAliasingHint: 'msaa4+fxaa'
     });
 
-    async function addModel(options: AddModelOptions) {
-      const model = await viewer.addModel(options);
+    viewer.addModel({ modelId, revisionId }).then((model) => {
       viewer.loadCameraFromModel(model);
-      window.model = model;
```
haakonflatval-cognite, line 41: Don't we still need to add the model to the `window`?

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -407,6 +411,7 @@ export class Cognite3DViewer {
     }
 
     this.spinner.dispose();
+    this.sessionLogger.dispose();
```
haakonflatval-cognite, line 414: Can we be so sure this will be called at the end of every session? It seems very dependent on someone actually calling the `dispose` method on the viewer, which I'd guess doesn't always happen. But then again, I don't know what else we would do :thinking:

haakonflatval-cognite (commented):
Looks good - I'll test it tomorrow, don't have time now :see_no_evil:

documentation/docs/components/Cognite3DViewerDemo.tsx
Savokr: I think it was only for debug purposes, if it is needed, can be left there.

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
Savokr: Discussed it with Lars, specifically for that reason there is a check for canvas visibility in `SessionLogger`. So, if someone forgot to dispose the viewer, `sessionEnded` events won't be fired since canvas will be invisible.

documentation/docs/components/Cognite3DViewerDemo.tsx
Savokr: I'd prefer to remove it

Savokr: Also, you can access it through `viewer.models` easily.

haakonflatval-cognite: We use `window.model` a lot in the examples. That's how we inject the model into the example snippets: https://github.com/cognitedata/reveal/blob/10e7062654ba4260848ccd022a67c7d5f90758d6/documentation/docs/components/LiveCodeSnippet.tsx#L59 E.g. in the `Loading CAD models` page, the first example tries to hide the model, but the model stays visible.

viewer/packages/metrics/src/SessionLogger.ts
```diff
@@ -0,0 +1,115 @@
...
+    const eventMetadata: { averageFramerate?: number; sessionTime: number } = {
+      sessionTime: Date.now() - this._sessionStartTime
+    };
+    const fps = this.calculateAverageSessionFps();
+
+    if (fps) {
+      eventMetadata.averageFramerate = fps;
+    }
+
+    MetricsLogger.trackEvent('sessionEnded', eventMetadata);
```
haakonflatval-cognite, line 101: Should we reset the state of the `SessionLogger` when session end is tracked? It could be used again if e.g. the canvas is only temporarily hidden.

haakonflatval-cognite (changes requested):
Closer and closer :sweat_smile:

viewer/packages/metrics/src/SessionLogger.ts
Savokr: Not sure what you mean. Reset and similar things are handled in `tickCurrentAnimationFrame` function. Session time is reset. Or you mean resetting average fps after "resuming" reveal?

haakonflatval-cognite: Yes, I meant resetting average FPS.. But I suppose it won't change the results much anyway.

haakonflatval-cognite (approved):
LGTM :+1:

PR #2992, opened 2023-02-23
improvement: Loading 360 descriptors for large sets no longer freeze the application

astrid-kg:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-789

## Description :pencil:
I found that the problem was in Cdf360ImageEventProvider and mergeDescriptors where we go through all files (in this case 43284) for each event (7222) which is a really big loop. 

Files are now represented by a Map so that lookup can be done in mergeDescriptors. I also replaced autoPagingToArray with autoPagingEach so that the files can be inserted directly into the map as they arrive.

Using the Performance and Network tab in the debug window. Since "Performance" gave some weird results I also added timestamps in code and printed out how long different parts of the code took to finish.

Tested mostly on Windows/Chrome, but I did profile once on Android/Chrome.

1. Open any CDF model.
2. In the GUI open "360 Images" and add "Hibernia_RS2" as site_id in options.
3. Click "Add image set". (This can be done several times without refreshing the browser, but performance is affected by caching, so disable this in the Network-tab for consistent results.).
4. Move the camera around and see that there is no freezing going on 🧊 .

Optional: Use the Performance and Network tab to profile the loading.


Savokr (approved):
Looks really good 👍! Noticed funny thing though, adding big 360 image site is not blocking the thread but if you try to remove them, reveal freezes for a few seconds. But it should be a separate PR, for sure.

astrid-kg:
I noticed that remove all also did not remove everything if I had added the same dataset twice (or more). I will make a task to look into removal too then 👍

PR #2994, opened 2023-02-24
fix: Entering a clicked 360 Image is slow after hovering many icons

astrid-kg:
#### Type of change

Bug 

#### Jira ticket :blue_book:
https://cognitedata.atlassian.net/browse/REV-794

## Description :pencil:
I now throttle the amount of calls to `preload` when hovering using `debounceTime`. I am using debounce with both leading and trailing, so the first icon to be hovered will send a request right away. This makes the delay value less critical as it will not affect the download of the first image set.

I added a console.log that was printed on hover and one on preload. I then loaded Hibernia_RS2 and looked at how the functions were called. See task for movie.

Do what I did. Or just play around in Hibernia_RS2 and look for anything suspicious. 

### DIFF
--- a/viewer/packages/api/src/api-helpers/Image360ApiHelper.ts
+++ b/viewer/packages/api/src/api-helpers/Image360ApiHelper.ts
@@ -60,7 +60,7 @@ export class Image360ApiHelper {
     this._cachedCameraManager = activeCameraManager.innerCameraManager;
     this._requestRedraw = requestRedraw;
     this._preLoadObservable = new Subject();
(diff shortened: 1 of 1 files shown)


pramodcog (approved):
Looks good. Small observation, the delay of 300ms seems to work well currently, but need to have an eye for different data to make sure delay works as intended for smaller & very large data.

astrid-kg:
I updated the PR text: I am using debounce with both leading and trailing, so the first icon to be hovered will send a request right away. This makes the delay value less critical as it will not affect the download of the first image set.

PR #2996, opened 2023-02-27
fix: security issues in dependencies

christjt:


haakonflatval-cognite (approved):
Hurts a little, but okay

PR #3004, opened 2023-03-06
Astrid kg/rev 807

astrid-kg:
#### Type of change

Feat  
Bug 
Test 
Docs 
Style 
Refactor 
Chore 

#### Jira ticket :blue_book:


viewer/packages/360-images/src/icons/IconCollection.ts
```diff
@@ -85,6 +125,29 @@ export class IconCollection {
...
+    return ({ camera }) => {
+      const points =
+        this._proximityRadius < 1
+          ? this._icons
+          : octree.findPoints(camera.position, this._proximityRadius).map(pointContainer => {
+              return pointContainer.data;
+            });
+
+      let closestPoints = points.sort((a, b) => {
+        return b.position.distanceTo(camera.position) - a.position.distanceTo(camera.position);
```
christjt, line 138: Tiny optimization, but it is more efficient to check the "squared distance to", which is preferable to comparing vector lengths. See: https://threejs.org/docs/#api/en/math/Vector3.distanceToSquared

viewer/packages/360-images/src/icons/IconCollection.ts
```diff
@@ -85,6 +125,29 @@ export class IconCollection {
     };
   }
 
+  private computeProximityPoints(octree: IconOctree, iconSprites: InstancedIconSprite): BeforeSceneRenderedDelegate {
+    return ({ camera }) => {
+      const points =
+        this._proximityRadius < 1
```
christjt, line 131: This condition seems odd to me, so if proximity radius is set to 0.5 then all points will be returned? I guess this should rather be when it is smaller than some epsilon?

viewer/packages/360-images/src/icons/IconCollection.ts
```diff
@@ -85,6 +125,29 @@ export class IconCollection {
...
+          : octree.findPoints(camera.position, this._proximityRadius).map(pointContainer => {
+              return pointContainer.data;
+            });
+
+      let closestPoints = points.sort((a, b) => {
+        return b.position.distanceTo(camera.position) - a.position.distanceTo(camera.position);
+      });
+
+      if (this._proximityPointLimit > 0) {
+        closestPoints = closestPoints.slice(-(this._proximityPointLimit + 1)); //Add 1 to "skipSelf"..
```
christjt, line 142: This syntax confuses me a little. Is this taking the last this._proximityPointLimit points in the array? If so, would it not be cleaner to reverse the sorting order and then do `closestPoints.slice(this._proximityPointLimit)` (this might be missing a +1).

viewer/packages/360-images/src/icons/IconCollection.ts
```diff
@@ -17,12 +22,44 @@ export class IconCollection {
...
+        break;
+      }
+      default:
+        break;
+    }
+    this._onBeforeSceneRenderedEvent.subscribe(this._activeCullingSchemeEventHandeler);
+  }
+
+  public set360IconCullingRestrictions(radius: number, pointLimit: number): void {
+    this._proximityRadius = radius;
```
christjt, line 59: Should probably do a min clamp here. I.E. `Math.min(0, radius)`, as negative values would probably create undefined behaviour / we would not need to handle / think about it.

christjt (changes requested):
This looks great! Architecturally this seems very sound, at least in my opinion. I think the only thing I would like to see changed is having the set360IconCullingRestriction removed from the Cognite3DViewer API, and rather just be a method on the 360 image collection. Afaik. it already is, so  you would just need to expose it. I suspect that most people will not want to change the default behaviour. So I think just having it only on the collection is fine, and if we see a lot of usage, we can add a utility on the viewer to set to all collection such as you have here.

viewer/packages/360-images/src/icons/IconCollection.ts
```diff
@@ -8,6 +8,11 @@ import { Image360Icon } from './Image360Icon';
 import { InstancedIconSprite } from './InstancedIconSprite';
 import { IconOctree } from './IconOctree';
 
+export enum IconCullingScheme {
+  Clustered,
+  Proximity
+}
```
christjt, line 14: Concider changing this to `export type IconCullingScheme = 'clustered' | 'proximity';`

In any other language, using an enum here is correct. But typescript's enums are notoriously bad (here is a talk on it if you are interested in why: https://www.youtube.com/watch?v=0fTdCSH_QEU).

Also you would be able to leverage duck-typing here such that a user can do:
```ts
setCullingScheme('clustered');
```

or
```ts
const clusterScheme = 'clustered';
setCullingScheme(clusterScheme);
```
such that they don't have to import the type unless they want to.

christjt (changes requested):

viewer/packages/360-images/src/icons/IconCollection.ts
astrid-kg: Done! 🥳

viewer/packages/360-images/src/icons/IconCollection.ts
astrid-kg: I reversed it. It is now `this._proximityRadius > 0`🤸‍♀️ .

viewer/packages/360-images/src/icons/IconCollection.ts
astrid-kg: Done! 😎 Note that I now do reverse() the list before I set the icons further down in the code. As we talked about, this is to put the transparent icon in the correct order in the scene.

viewer/packages/360-images/src/icons/IconCollection.ts
astrid-kg: Done! But with max 💯

viewer/packages/360-images/src/icons/IconCollection.ts
```diff
@@ -17,12 +22,44 @@ export class IconCollection {
...
+      }
+      default:
+        break;
+    }
+    this._onBeforeSceneRenderedEvent.subscribe(this._activeCullingSchemeEventHandeler);
+  }
+
+  public set360IconCullingRestrictions(radius: number, pointLimit: number): void {
+    this._proximityRadius = radius;
+    this._proximityPointLimit = pointLimit;
```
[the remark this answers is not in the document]
astrid-kg: ..and done! 🥇

viewer/packages/360-images/src/icons/IconCollection.ts
christjt: I think you should remove the max cap, we can have point clouds that span kilometers, so in those cases if they only have a few 360 images, but they span great areas, I would still like to see them.

christjt (approved):
LGTM!

PR #3041, opened 2023-03-10
feat: Adjust the size of 360Images icons to better reflect their position

astrid-kg:
#### Type of change

Feat  

#### Jira ticket :blue_book:
https://cognitedata.atlassian.net/browse/REV-824

## Description :pencil:

I have tried to match icon sizes with what is currency used in Remote. 

I had to change the output of `computeAdaptiveScaling` to include the radiusScale, and multiply the _hoverSprite scale by 2 as Points size is measured in diameter:
https://registry.khronos.org/OpenGL-Refpages/gl2.1/xhtml/glPointSize.xml
https://stackoverflow.com/questions/25780145/gl-pointsize-corresponding-to-world-space-size
This was not needed when we were using radiusScale = 0.5, as 2 * 0.5 = 1.

With different values for:
`MIN_PIXEL_SIZE `: 8-16
`MAX_PIXEL_SIZE`: 64 - 1024 (with 0.3 radius this value seems to no have any affect after 256ish)
`_iconRadius`: 0.01 - 100.

Compared with Remote these seem to be close to the correct values:
`MIN_PIXEL_SIZE `: 16
`MAX_PIXEL_SIZE`: 256
`_iconRadius`: 0.3

However `MAX_PIXEL_SIZE` should be bigger, as the closest icons are a little smaller in Reveal than Remote. But increasing this value above ~256 does not have any effect. We suspect this is due to a threshold in Points. But the current result is good enough.

Load any 360 image. This change is visible both outside and inside images.


christjt (approved):
Looks great!

PR #3099, opened 2023-03-28
feat: Add API to list available dates, change and get selected (visible) 360 image by date (no UI)

astrid-kg:
#### Type of change

Feat  

#### Jira ticket :blue_book:
https://cognitedata.atlassian.net/browse/REV-840

## Description :pencil:

**Known defects / lacking capabilities:** 
The transform of each individual revision is currently the same as whatever uniqueEventDescriptor it is grouped on in mergeDescriptors. Atm I do not know how to match files and events to get the individual transform, (it looks like they do have this, but I can't see a way to link the event with the image sets other than station_id), but doing this at a later time should be trivial as it would mostly require changes in mergeDescriptors: https://cognitedata.atlassian.net/browse/REV-859

**Added to API:**
- (Image360) `list360ImageRevisions(): Image360Revision[];` 
- (Image360) `getActiveRevision(): Image360Revision;`
- (Image360) `changeRevision(revisionId: number): Promise<void>;`
- (Image360Revision) `readonly date: Date | undefined;`
- (Image360Collection) `targetRevisionDate: Date | undefined;`

**Larger changes:**
- When we get images from CDF these are now grouped in Historical360ImageSet. This holds a list of revisions (Image360EventDescriptors) and the metadata for the Event they are grouped by.

- I introduced `Image360Revision` and split out everything image related from `Image360Entity`, so that Image360Entity now represents a station, while Image360Revision represents all sets of images that belong to this station. A huge portion of this PR is replacing entities with revisions. _Edit: Due to breaking changes I can't move the visualization box to Revisions as planed. This is now still owned by Image360Entity. The gain from this is that we only have one mesh per entity, but we loose the ability to do fade/transition between revisions as this (today) require them to both have boxes. This can be solved in other ways later tho._ 

- ~~To be able to change the revision from the Image360Entity, I am currently passing a `reloadImage` function from 
Image360ApiHelper -> Image360Facade -> Image360CollectionFactory -> Image360Entity. It would be nice to avoid this, but this is as good as I could do without going though the Cognite3DViewer.~~ Added revision as optional parameter in enter360Image.

Using the site_id: XOM-HCU1_north_v02

I added a "Current image revision" field in the Examples UI. This only has effect when changed, and is **not** read automatically when entering a new image. I have also added some logging in Examples. This can be removed once testing is done.

For how it has been tested, see below.

1. Load the XOM-HCU1_north_v02 image set.
2. Open the debug console.
3. Navigate into any image and see that the entered revision is printed in the log and available revisions are listed.
4. Navigate to an image with more than one revision and change "Current image revision" to one of the listed ids (or one that is not listed, but nothing will happen).
7. Change "Revision date" to a really large (2000000000000) and then a really small number (2). If the image has more than one revision, the image should change to which ever is the newest (for large number) or oldest (for the small number). You can ofc test with some real times. 🕐 
8. Empty  "Revision date" and the newest image should be loaded.

Finally, test 360 images like normal:
- Enter/leave images
- Change network settings, zoom, alpha, etc.


viewer/packages/360-images/src/collection/Image360CollectionFactory.ts
```diff
@@ -28,35 +30,37 @@ export class Image360CollectionFactory<T> {
...
+    const entities = zip(historicalDescriptors, icons)
       .filter(isDefined)
       .map(([descriptor, icon]) => {
         return new Image360Entity(
           descriptor,
           this._sceneHandler,
           this._image360DataProvider,
           descriptor.transform,
-          icon!
+          icon!,
```
christjt, line 54: Don't think the `!` operator is nessesary anymore since it is type guarded by `isDefined`.

viewer/packages/360-images/src/entity/Image360.ts
christjt, line 1: We need to think about this a little. These are technically breaking changes.

1. Perhaps we could just have the transform getter here instead of on the revision, since they all share the same transform.

2. I realize this might cause some needed architectural changes, but perhaps only 1 visualization box is nessesary, and it could simply be "filled" with new texture sets as the revision changes.

If we do these two things we get to keep the current getters. And in my opinion it makes more sense that the entity is controlled with one visualization box that gets mutated based on the set revision.

viewer/packages/360-images/src/entity/Image360Entity.ts
```diff
@@ -37,111 +24,89 @@ export class Image360Entity implements Image360 {
...
-    this._transform = transform;
     this._image360Icon = icon;
-    this._image360VisualzationBox = new Image360VisualizationBox(this._transform, sceneHandler);
-    this._image360VisualzationBox.visible = false;
-    this._getFullResolutionTextures = undefined;
+    this._reloadImage = reloadImage;
+
+    this._revisions = image360Metadata.imageRevisions.map(descriptor => {
+      return new Image360RevisionEntity(imageProvider, descriptor, sceneHandler, transform);
+    });
```
christjt, line 40: Note that simply commiting this suggestion will probably have lint error.

```suggestion
    this._revisions = image360Metadata.imageRevisions.map(descriptor => new Image360RevisionEntity(imageProvider, descriptor, sceneHandler, transform));
```

viewer/packages/360-images/src/entity/Image360Entity.ts
```diff
@@ -37,111 +24,89 @@ export class Image360Entity implements Image360 {
...
+      (closest, revision) => {
+        if (revision.date) {
+          const difference = Math.abs(revision.date.getTime() - dateAsNumber);
+          if (difference < closest.difference) return { revision: revision, difference };
+        }
+        return closest;
+      },
+      { revision: this.getMostRecentRevision(), difference: Number.POSITIVE_INFINITY }
+    );
+    return closest.revision;
```
christjt, line 102: Think this can be made cleaner with something like this (note that I havn't tested this):
```ts
  public getRevisionClosestToDate(date: Date): Image360RevisionEntity {
    const dateAsNumber = date.getTime();
    const datedRevisions = this.revisions.filter(revision => revision.date !== undefined);
    const closestDatedRevision = minBy(datedRevisions, revision => Math.abs(revision.date.getTime() - dateAsNumber));
    return closestDatedRevision ?? this.getMostRecentRevision();
```

viewer/packages/360-images/src/entity/Image360Revision.ts
```diff
@@ -0,0 +1,26 @@
...
+   * of the given 360 image.
+   * @returns model-to-world transform of the 360 Image
+   */
+  readonly transform: THREE.Matrix4;
+
+  /**
+   * The object containing the unit cube with the 360 images.
+   * @returns Image360Visualization
+   */
+  readonly image360Visualization: Image360Visualization;
```
christjt, line 19: As noted in a comment earlier, lets see if we can move this back to the entity

viewer/packages/360-images/src/entity/Image360RevisionEntity.ts
```diff
@@ -0,0 +1,130 @@
...
+        () => {
+          return Promise.resolve();
+        },
+        reason => {
+          return Promise.reject(reason);
+        }
+      );
+
+    return { firstCompleted, fullResolutionCompleted };
+  }
```
christjt, line 108: `.then` is not idiomatic, try to separate these things into local functions and use await.

viewer/packages/360-images/src/entity/Image360RevisionEntity.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * Apply full resolution textures to the image360VisualzationBox. This has no effect if full resolution has already been applied.
+   */
+  public async applyFullResolution(): Promise<void> {
+    if (this._getFullResolutionTextures) {
```
christjt, line 114: invert this if

viewer/packages/360-images/src/entity/Image360Entity.ts
```diff
@@ -37,111 +24,89 @@ export class Image360Entity implements Image360 {
...
   }
 
   /**
    * @obvious
    */
   public dispose(): void {
-    this.unload360Image();
+    this._revisions.forEach(revision => revision.unload360Image());
```
christjt, line 109: NIT: concider adding a dispose method on revision and simply calling that dispose method here.

christjt (changes requested):
Great stuff, Astrid. Awesome that you were able to take lead on this, and generally it looks really good. Have some comments here and there with some opportunities for improvement. As for the passing the callback from the API helper, it is not ideal. But we can fix this later as it functionally should not matter much. Also note that documentation should be updated with these new additions, but we can do that in a separate PR (but please create a task for it).

viewer/packages/data-providers/src/image-360-data-providers/Cdf360ImageEventProvider.ts
```diff
@@ -115,22 +117,29 @@ export class Cdf360ImageEventProvider implements Image360Provider<Metadata> {
...
+                  mimeType: imageInfo.mimeType!,
+                  fileId: imageInfo.id
+                } as Image360FileDescriptor;
+              })
+            };
+          });
+        return { ...eventDescriptor, imageRevisions };
       })
-      .filter(image360Descriptor => image360Descriptor.faceDescriptors.length === 6);
+      .filter(historicalImages => historicalImages.imageRevisions.length > 0);
```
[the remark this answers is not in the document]
astrid-kg: This check is to ensure we found at least one revision. The "set of 6" check I have tried to move into the creation of revision further up:
".filter(imageSetInfo => imageSetInfo.length === 6)"

I am doing it before the map to make sure I can safely get imageSetInfo[0] when assigning timestamp. Can 'imageSetInfo.map' fail to return a descriptor for each element? If so, I will need to do the check after.

haakonflatval-cognite (changes requested):
Just some nitpick - I'm gonna go through once more today, but it looks good :+1:

viewer/packages/360-images/src/entity/Image360RevisionEntity.ts
```diff
@@ -0,0 +1,110 @@
...
+    async function awaitFullResolution(): Promise<void> {
+      await fullResolutionFaces;
+    }
+  }
+
+  /**
+   * Clear the cached textures used by this revision.
+   */
+  public clearTextures(): void {
+    this._textures = [];
```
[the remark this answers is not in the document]
astrid-kg:
```
  public clearTextures(): void {
    this._textures.forEach(t => t.texture.dispose());
    this._textures = [];
  }
```
  
  Like this?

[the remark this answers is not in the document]
haakonflatval-cognite: Yes, something like that, at least we should dispose of them somewhere :thinking:

viewer/packages/360-images/src/entity/Image360Revision.ts
```diff
@@ -0,0 +1,11 @@
+/*!
+ * Copyright 2022 Cognite AS
+ */
+
+export interface Image360Revision {
```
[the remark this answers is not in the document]
astrid-kg: Added! But I require feedback on the text:

`/**
 * Interface used to manage historical revisions of {@link Image360}.
 * One instance represents one specific image revision.
 */`

viewer/packages/360-images/src/entity/Image360RevisionEntity.ts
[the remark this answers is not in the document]
astrid-kg: Done! 🏑

astrid-kg:
Waiting for @christjt to get back so that we can make a final decision on the **reloadImage** vs **setting revision through the viewer**. Let's look at this again after easter. 🐣 ⛷️

christjt (approved):
Looks great! Ship it!!!

haakonflatval-cognite (approved):

viewer/packages/360-images/src/entity/Image360Revision.ts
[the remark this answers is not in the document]
haakonflatval-cognite: Seems good!

PR #3134, opened 2023-04-11
refac: port to Typescript 5 decorators

haakonflatval-cognite:
#### Type of change

Refactor 

## Description :pencil:

An attempt at porting from experimental Typescript decorators to the official Typescript 5 decorators. The API has changed somewhat, so this change is non-trivial.

I also decided to remove the `opacity` uniform from the point cloud material. It is unused and caused trouble, because it's declared in the base class of `PointCloudMaterial` and cannot be redeclared as an `accessor`.

**Update:** 
This seems to work well now. Biggest issue with this update, is that it requires our clients to use Typescript 4.9 or newer, because our type definitions now include the `accessor` keyword. I don't think this is a big problem, is it is only a minor version update from our last version (which I _think _ was `4.x`, but we may need to notify our users somehow.

**Update 2:**
By marking the class as `@internal`, it is no longer exposed in the type declarations, and we no longer have to worry about exposing the `accessor` key word.

**Original Rant:**
The most in-depth explanation of the new decorators (which is also referenced from the[ official Typescript 5.0 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#decorators)), seems to be [this blog](https://2ality.com/2022/10/javascript-decorators.html#class-field-decorators). However, the blog is focusing on Javascript, and I am trouble directly copying its solutions to our use case. 

The most prominent change from the experimental decorator API, is that we need to use [auto-accessors](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#auto-accessors-in-classes) to be able to overwrite the getter and setters like we do in the older implementation. This causes trouble for the `opacity` field, which is defined in the super class as a normal field, which cannot be overriden by an accessor. This is probably easily circumventable by renaming the variable.

I am also confused as to how to use the `this` variable in the decorator. It is at least not automatically recognized by the Typescript parser. It is possible that some more time is needed for the parser to reach full spec-compatibility. (But also, that might be weird, since they are both in the domain of Microsoft...)


haakonflatval-cognite:
From what I can tell, the current suggested implementation conforms to the API conventions used in https://2ality.com/2022/10/javascript-decorators.html#read-only-auto-accessors (Particularily the use of `this` and declaration as a `function`). Alas, it does not work. We may have to wait a little, until either more (correct) resources on Typescript decorators are published, or until the transpiler admits its mistake, whichever is relevant.

viewer/packages/rendering/src/pointcloud-rendering/PointCloudMaterial.ts
```diff
@@ -381,42 +374,39 @@ function getValid<T>(a: T | undefined, b: T): T {
...
-          }
+function uniform<K extends keyof IPointCloudMaterialUniforms>(uniformName: K, requireSrcUpdate: boolean = false) {
+  return (_target: any, _context: any) => ({
+    get(this: any) {
+      return this.getUniform(uniformName);
+    },
+    set(this: any, value: unknown) {
+      if (value !== this.getUniform(uniformName)) {
+        this.setUniform(uniformName, value as IPointCloudMaterialUniforms[K]['value']);
+        if (requireSrcUpdate) {
```
astrid-kg: This was only needed for opacity?

astrid-kg (approved):
LGTM ☃️

viewer/packages/rendering/src/pointcloud-rendering/PointCloudMaterial.ts
astrid-kg: This has been removed 👍

haakonflatval-cognite: Yes, that's right :+1:

PR #3135, opened 2023-04-11
fix: throw error if same image 360 collection is added twice

christjt:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-852

## Description :pencil:

Now throws if user attempts to add the same 360 image collection more than once.

In examples, attempt to add the same image360set twice, notice an error being thrown.


astrid-kg (approved):
Tested! And the code LGTM 👍

PR #3157, opened 2023-04-20
fix: remove jest polyfill

christjt:
Seems to break tests for an unknown reason. But also cannot see that it is needed anywhere.


haakonflatval-cognite (approved):
LGTM :+1:

PR #3216, opened 2023-05-02
refactor: make annotation getter async and extract from texture loading

haakonflatval-cognite:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-906

## Description :pencil:

- Factor out annotations to not go through cache. The hypothesis is that the annotations will not pose a significant memory footprint, even though they are never expelled from memory in a session
- Make annotation getter async
- Outsource default styling handling to each revision

- Start Reveal examples
- Open https://localhost:3000/?project=3d-test&env=greenfield-3d&modelId=4923307033982497&revisionId=3100284276970593
- Load 360-image set `helideck-site-2-jpeg`
- Enter lower-most 360 image and the one beside it (the latter has one annotation on the ground)
- Observe that annotations are loaded and displayed correctly
- Observe that the annotation default styling (set from the UI) works properly
- Note that clicking an annotation will set a custom appearance overriding the default style, but it should disappear when clicking another annotation


viewer/packages/360-images/src/collection/DefaultImage360Collection.ts
```diff
@@ -37,7 +37,7 @@ export class DefaultImage360Collection implements Image360Collection {
 
   private _needsRedraw: boolean = false;
 
-  private _defaultStyle: Image360AnnotationAppearance = {};
+  private readonly _defaultStyle: { appearance: Image360AnnotationAppearance } = { appearance: {} };
```
Savokr, line 40: Why this change? Is it planned to include more things?

viewer/packages/360-images/src/Image360Facade.ts
```diff
@@ -60,7 +60,14 @@ export class Image360Facade<T> {
...
 
   public preload(entity: Image360Entity, revision: Image360RevisionEntity, lockDownload?: boolean): Promise<void> {
-    return this._image360Cache.cachedPreload(entity, revision, lockDownload);
+    const annotationPromise = revision.getAnnotations();
+    const cacheLoadPromise = this._image360Cache.cachedPreload(entity, revision, lockDownload);
+
+    return awaitPromises();
+
+    async function awaitPromises(): Promise<void> {
+      await Promise.all([annotationPromise, cacheLoadPromise]);
```
Savokr, line 69: Do we care about the order of resolving these promises here?

Savokr (approved):
Looks good! One thing I noticed though is that sometimes 360 image cubemap "flickers" once when clicked for being entered. Doesn't happen every time, though. Not sure this PR introduced it but it is not present in latest reveal version in Data Exploration, so seems like something recent. Here is a video of it: 

https://user-images.githubusercontent.com/45129444/235920321-65793d0b-2af9-49aa-b8b6-030ef915d909.mov

viewer/packages/360-images/src/Image360Facade.ts
haakonflatval-cognite: Not really, no... Do you mean I should be using some other function to await them?

viewer/packages/360-images/src/collection/DefaultImage360Collection.ts
haakonflatval-cognite: No, fixed :+1:

haakonflatval-cognite:
>sometimes 360 image cubemap "flickers" once when clicked for being entered
I think this is because there is a small window between where we set the cubemap visible (line 230 in `Image360ApiHelper` currently) and where it is being incrementally made more opaque while the camera moves towards it. I think the effect appears if we render a frame in that window. Created a bug: https://cognitedata.atlassian.net/browse/REV-934

viewer/packages/360-images/src/Image360Facade.ts
Savokr: No, just curious since before this PR the order of things was defined explicitly as far as I understood.

PR #3257, opened 2023-05-12
fix: Fix a leak where 360 image blobs are never released

Strepto:
#### Type of change

Bug 

#### Jira ticket :blue_book:

N/A (External contribution)

## Description :pencil:

Fixes a leak where blob references are never released.

This may cause image blob references to never be evicted from the browsers memory until the tab is closed or refreshed.

Fixing this by releasing the url reference after the image is loaded into the texture.

The fix is inspired by a bug we had, and I scanned your code for possible similar issues.

See ie: https://developer.mozilla.org/en-US/docs/Web/API/File_API/Using_files_from_web_applications#example_using_object_urls_to_display_images

for how they also release the objectUrl after loading the image in the MDN docs.

Not reliably tested in reveal as I have not been able to see any 360 images in the reveal test scene. Hoping someone else can check if the fix has no bad effects.

You can check for blob-leaks by opening the `chrome://blob-internals/` and inspecting the growing amount of blob references when browsing 360 images. (You have to manually refresh to update the list). 

If running with the above fix the blob-internals list should not grow.


christjt (approved):
LGTM! Thanks for spotting and fixing this 💯

PR #3296, opened 2023-05-23
fix: add minimum threshold to prevent getting stuck when two icons are overlapping

christjt:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/REV-945

## Description :pencil:

Fixes an edge case where mutiple 360 images are in the same position which are not revisions.

`site_id: Akterdekk` has some 360 images that overlap exactly and are not revisions of each other. Before this fix you get stuck / cannot click anything than the other icon which overlaps.


pramodcog (approved):
Looks good

PR #3300, opened 2023-05-24
feat: add label and id to collection and fix underlying label types.

christjt:
#### Type of change

Feat  

## Description :pencil:

Add label and id to collection and fix underlying label types.

### DIFF
--- a/viewer/packages/360-images/src/collection/Image360CollectionFactory.ts
+++ b/viewer/packages/360-images/src/collection/Image360CollectionFactory.ts
@@ -2,17 +2,17 @@
  * Copyright 2023 Cognite AS
  */
 
+import assert from 'assert';
+import zip from 'lodash/zip';
+import uniqBy from 'lodash/uniqBy';
 import { Image360Provider } from '@reveal/data-providers';
 import { BeforeSceneRenderedDelegate, DeviceDescriptor, EventTrigger, SceneHandler } from '@reveal/utilities';
-import zip from 'lodash/zip';
 import { DefaultImage360Collection } from './DefaultImage360Collection';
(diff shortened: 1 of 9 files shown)


viewer/packages/360-images/src/entity/Image360.ts
```diff
@@ -33,7 +33,7 @@ export interface Image360 {
    * Get label of 360 image entity.
    * @returns Station label
    * */
-  readonly label: string;
+  readonly label: string | undefined;
```
haakonflatval-cognite, line 36: Do we want a similar field on `Image360Revision` as well?

viewer/packages/360-images/src/collection/Image360CollectionFactory.ts
```diff
@@ -74,11 +74,19 @@ export class Image360CollectionFactory<T> {
         );
       });
 
-    const collectionIds = uniq(historicalDescriptors.map(desc => desc.collectionId));
+    const uniqueCollections = uniqBy(historicalDescriptors, desc => desc.collectionId);
+
+    assert(uniqueCollections.length === 1);
 
-    assert(collectionIds.length === 1);
+    const { collectionId, collectionLabel } = uniqueCollections[0];
```
haakonflatval-cognite, line 82: Is this necessary? :sweat_smile: Aren't we already pretty sure the `collectionId` is the same for all descriptors? Can we just use `historicalDescriptors[0]`?

haakonflatval-cognite: Nevermind, I missed the previous code :sweat: I agree this sanity check should be somewhere

PR #3303, opened 2023-05-25
fix: remove syntax error in jsdoc

christjt:
#### Type of change

Bug 

## Description :pencil:

Fixes a small syntax error in `dispose` which broke some parts of the typedoc generation.


Savokr (approved):
LGTM

PR #3409, opened 2023-06-22
fix: Camera rotation target from new camera position

hugolafis:
#### Type of change

Bug 

## Description :pencil:
This PR fixes a bug when both camera position and rotation are set simultaneously. Currently, when the new target is calculated from the provided quaternion, the offset it uses to position itself in world space references the old camera position, instead of the new vector that the camera position will be set to, and results in an incorrect target position.

This fix addresses that by passing the new vector through into the utility function `CameraManagerHelper.calculateNewTargetFromRotation` and utilises that for the target offset.

I have manually tested this by utilising a locally build copy of Reveal as a dependency in another application where I first noticed the issue, and the fix appears to have solved the issue. 

If you call `setCameraState` with a provided vector and quaternion, both these values should now match the resulting `camera.quaternion` and `camera.position`. Previously, if the provided vector was different to the current position of the camera, the resulting camera quaternion would not match the input value.


haakonflatval-cognite:
Hi!, thanks for the contribution! 

I have some trouble reproducing the problem you are stating. I'm testing now by calling 
```
    viewer.cameraManager.setCameraState({
      position: new THREE.Vector3(50, 0, 51),
      rotation: new THREE.Quaternion(0.5, 0.5, 0.5, 0.5).normalize()
    });

    console.log(
      'New camera position: ',
      this.viewer.cameraManager.getCamera().position,
      'rotation: ',
      new THREE.Quaternion().setFromEuler(this.viewer.cameraManager.getCamera().rotation)
    );
```
, which prints out 
```
New camera position:  
Object { x: 49.99999999999997, y: -7.105427357601002e-15, z: 51 }

 rotation:  
Object { isQuaternion: true, _x: 0.5, _y: 0.5, _z: 0.4999999999999999, _w: 0.5000000000000001 }

```
The rotation seems to be as expected. Am I misunderstanding the problem? Do you have another specific input that produces the wrong result?

hugolafis:
Hello :)

I've just reproduced it on the master branch, using the example viewer provided in the repo. Here's the code I used:

Viewer.tsx

```
      viewer.domElement.addEventListener('keypress', (ev: KeyboardEvent) => {
        switch (ev.key) {
          case 'g':
            const vec3 = new THREE.Vector3(5, 5, 5);
            const quaternion = new THREE.Quaternion(0, 0, 0, 1);

            viewer.cameraManager.setCameraState({
              position: vec3,
              rotation: quaternion
            });

            console.log(quaternion);
            console.log(viewer.cameraManager.getCamera().quaternion);
            break;
          default:
            break;
        }
      });
```

When first pressing G, the quaternion and camera quaternion values do not match, however if you press it a second time you should see the values match (due to the camera not moving).

Interestingly I do get a matching quaternion value if I use the same value as you tried, however I'm not sure why!

haakonflatval-cognite:
@hugolafis I see - I've tried again with your value and can reproduce the error now :+1: I'm approving the PR!

I see there are some CI steps failing. The first one can be fixed by running `yarn update-api` in the `viewer` folder. 
Let me know if you need assistance with resolving the rest

haakonflatval-cognite (approved):
LGTM :+1:

PR #3412, opened 2023-06-23
fix: remove stdlib due to botched ESM transitive dependency

christjt:
#### Type of change

Bug 

## Description :pencil:

Remove three-std-lib dependency and leverage three/examples directly instead.


haakonflatval-cognite (approved):
LGTM :+1:

PR #3451, opened 2023-07-03
feat(react-components): add threed resource container

christjt:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/UX-847

## Description :pencil:

Adds a container for loading a list of 3D resources. Also wraps this in a context that can be used to listen to when all resources are loaded.


react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export const Reveal3DResources = ({ resources }: Reveal3DResourcesProps): ReactElement => {
+  const [reveal3DModels, setReveal3DModels] = useState<TypedReveal3DModel[]>([]);
+  const { setModelsAdded } = useContext(ModelsLoadingStateContext);
+  const viewer = useReveal();
+  const numModelsLoaded = useRef(0);
+
+  useEffect(() => {
+    getTypedModels(resources, viewer).then(setReveal3DModels).catch(console.error);
+  }, []);
```
Savokr, line 30: Shouldn't this recompute if we change models parameters like id/revision or just add another model? I think adding `resources` as a dependency should be correct.

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
```diff
@@ -0,0 +1,103 @@
...
+  };
+
+  return (
+    <>
+      {reveal3DModels
+        .filter(({ type }) => type === 'cad')
+        .map((addModelOption) => {
+          return (
+            <CadModelContainer
+              key={`${addModelOption.modelId}/${addModelOption.revisionId}`}
```
Savokr, line 51: In case someone for some weird reason wants to add two exactly the same models the following error will appear. 
image
I think we should add some salt/uuid/model number to the key here and in pointcloud/360 image case.

Savokr (approved):
Looks really good! A few comments/questions but nothing major.

react-components/webpack.config.js
```diff
@@ -14,7 +14,7 @@ export default (_, argv) => {
     output: {
       path: path.resolve(__dirname, 'dist'),
       filename: 'index.js',
-      clean: true,
+      clean: false,
```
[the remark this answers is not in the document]
christjt: Webpack watch clears it, and then doesn't repopulate.... :(

react-components/tsconfig.json
```diff
@@ -14,5 +14,5 @@
     "moduleResolution": "bundler",
     "allowSyntheticDefaultImports": true
   },
-  "include": ["src/*", "stories/*"]
+  "include": ["src/**/*", "stories/*"]
```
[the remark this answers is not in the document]
christjt: Since src/* only includes files directly under the src directory. Technically it didn't matter much since we use webpack to bundle, but vscode / eslint doesn't like it, so changed it for convinience.

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
christjt: I'll add it, but note that this doesn't come without down sides. F.ex. mutating the resources list / object doesn't actually cause this useeffect to run (comparrison by ref. vs. value).

PR #3468, opened 2023-07-05
feat(react-components): expose viewer options

christjt:
#### Type of change

Feat  

## Description :pencil:

Exposes some hand picked viewer options from Cognite3DViewerOptions.


Savokr (approved):
⌥

PR #3493, opened 2023-07-17
improvement: add importance to AssignStyledNodeCollection

Strepto:
#### Type of change
Feat  

#### Jira ticket :blue_book:

## Description :pencil:

This is a draft suggestion on how an api for adding importance weighting in assignStyledNodeCollection. This is very open for discussion and changes, and is an early draft of how I an prioritization system could work. Please do not hesitate to contact me or anyone else on slack or in this PR.

This PR solves a problem we are having in the Equinor use of Reveal, where we use assignStyledNodeCollection for a lot of "systems" at the same time: 

* Layer system
* Highlighting
* Selection Coloring
* Coloring Tool

All these may interact with the model at the same time, but have different importance. 

Selection is always the most important, and we always want to indicate this in a given color. 

The coloring tool overrides highlighting, so if a part is highlighted in a pink color we want to be able to color over it.

The Layer system works in the background, but has the ability to override the default colors of the model based on user input. If we want to support this for the least important layer we need to assign styled node collections very early in the bootup of reveal.

**What we do today:** 

We initialize the systems in order of "least important to most important"

1. Layers system (which has 16 NodeCollections)
2. Highlighting (Which has 1 NC)
3. Coloring tool (Which has up to 20 NC)
4. Selection coloring (Which has 2)

What is the problem with this approach?
* We need to initialize all the StyledNodeCollections in this order in startup, _even if they are never used_ by the user.
* We can not split or add styledNodeCollections at runtime for these systems without the problem of overriding more important systems.
* Restoring state etc is slowed down due to the fact that we need to restore in sequence instead of in parallel to maintain the importance order.

**What `importance` solves:**

We can assign an importance to each system, and sub-importance within.

* Layers system: -10
* Default stuff: 0 // Anything not in the systems / short lived.
* Highlighing: 10
  * Highlighted Early: 10 Color Magenta
  * Highlighted late: 10 Color LtBlue
  * Highlighted important 15 Color Magenta
* ColoringTool: 20
  * Color Blue: 20
  * Color Red: 20
* Selection Coloring: 100
* etc

N/A

I can add visual and unit tests for this if this suggested Api has any chance of progressing.

* Use the default example scene.


viewer/packages/cad-styling/src/NodeAppearanceProvider.ts
```diff
@@ -79,15 +80,18 @@ export class NodeAppearanceProvider {
...
     if (existingCollection !== undefined) {
       existingCollection.appearance = appearance;
+      existingCollection.importance = importance;
+
       this.handleNodeCollectionChanged(existingCollection);
     } else {
       const styledCollection: StyledNodeCollection = {
         nodeCollection: nodeCollection,
         appearance,
+        importance: importance,
```
haakonflatval-cognite, line 94:
```suggestion
        importance,
```

viewer/packages/cad-model/src/wrappers/CogniteCadModel.ts
```diff
@@ -156,23 +160,25 @@ export class CogniteCadModel implements CdfModelNodeCollectionDataProvider {
...
 
     const index = this._styledNodeCollections.findIndex(x => x.nodeCollection === nodeCollection);
     if (index !== -1) {
       this._styledNodeCollections[index].appearance = appearance;
     } else {
-      this._styledNodeCollections.push({ nodeCollection: nodeCollection, appearance });
+      this._styledNodeCollections.push({ nodeCollection: nodeCollection, appearance, importance });
     }
-    this.cadNode.nodeAppearanceProvider.assignStyledNodeCollection(nodeCollection, appearance);
+    this._styledNodeCollections.sort((a, b) => a.importance - b.importance);
```
haakonflatval-cognite, line 180: Javascript builtin sorting is not stable, meaning that if e.g. all collections have importance 0, we are not guaranteed that they are stored in the order they are added, which is important for our current behavior. Lodash's `sortBy` function should be stable

viewer/packages/cad-styling/src/NodeAppearanceProvider.ts
```diff
@@ -98,6 +102,9 @@ export class NodeAppearanceProvider {
       this.scheduleNotifyChanged();
     }
 
+    // Sort ascending, to set the most important styles last so they override the unimportant
+    this._styledCollections.sort((a, b) => a.importance - b.importance);
```
haakonflatval-cognite, line 106: Lodash's `sortBy` here as well

haakonflatval-cognite (changes requested):
Nice feature - a couple of comments

haakonflatval-cognite (approved):

PR #3518, opened 2023-07-27
feat: add support for node data fetching in React Components

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/UX-1136

## Description :pencil:

Implement FDM node data query and expose it.

Given a proper FDM setup, you should now be able to register a callback in `onNodeClicked` that receives the clicked node data.


Savokr (changes requested):
Looks good, a few questions but nothing major. Only thing is that I think we should add an example in the storybook in the same way as I did with styling capabilities. It will allow to easily test this functionality now and later on. So, just expanding `Reveal3DResources` story should be fine.

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
```diff
@@ -35,24 +39,28 @@ export type Reveal3DResourcesStyling = {
...
+export type Reveal3DResourcesProps<NodeType = any> = {
   resources: AddResourceOptions[];
-  fdmAssetMappingConfig?: FdmAssetMappingsConfig;
+  fdmAssetMappingConfig: FdmAssetMappingsConfig;
   styling?: Reveal3DResourcesStyling;
+  onNodeClick?: (node: NodeDataResult<NodeType>) => void;
 };
 
-export const Reveal3DResources = ({
+export const Reveal3DResources = <NodeType,>({
```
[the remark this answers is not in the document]
haakonflatval-cognite: Ah, yes.. You're right

react-components/src/components/Reveal3DResources/queryMappedData.ts
```diff
@@ -0,0 +1,131 @@
...
+          }
+        }
+      },
+      {
+        equals: {
+          property: [
+            fdmConfig.source.space,
+            `${fdmConfig.source.externalId}/${fdmConfig.source.version}`,
+            'revisionId'
+          ],
```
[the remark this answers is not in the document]
haakonflatval-cognite: What do you mean `properties` field? :thinking: The `equals` filter node only takes a `property` and `value` field as parameters.

react-components/src/components/Reveal3DResources/queryMappedData.ts
```diff
@@ -0,0 +1,131 @@
...
+      mappings.nextCursor
+    );
+
+    mappings = {
+      edges: [...mappings.edges, ...nextMappings.edges],
+      nextCursor: nextMappings.nextCursor
+    };
+  }
+
+  if (mappings.edges.length === 0) {
```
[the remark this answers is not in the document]
haakonflatval-cognite: Refactored with a `filterAllInstances` method in the FDM client.. Let me know if you want another approach.

react-components/src/components/Reveal3DResources/queryMappedData.ts
```diff
@@ -0,0 +1,131 @@
...
+    return;
+  }
```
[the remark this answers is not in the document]
haakonflatval-cognite: Yes, I think I've tested for that case :thinking: Because that would happen whenever you click on a non-mapped node, and it doesn't throw an error when that happens

react-components/src/utilities/FdmSDK.ts
```diff
@@ -26,17 +26,49 @@ export type EdgeItem<PropertiesType> = {
...
+    containers: Array<{
+      type: 'container';
+      space: string;
+      externalId: string;
+    }>;
+    views: Array<{
+      type: 'view';
+      space: string;
+      externalId: string;
+      version: string;
```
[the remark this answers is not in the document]
haakonflatval-cognite: Nice, will use that :+1:

react-components/src/utilities/FdmSDK.ts
```diff
@@ -26,17 +26,49 @@ export type EdgeItem<PropertiesType> = {
...
+  inspectionOperations: { involvedViewsAndContainers: Record<never, never> };
+  items: Array<{ instanceType: InstanceType; externalId: string; space: string }>;
+};
+
+export type InspectResult = {
+  involvedViewsAndContainers: {
+    containers: Array<{
+      type: 'container';
+      space: string;
+      externalId: string;
```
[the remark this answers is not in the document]
haakonflatval-cognite: Seems like we don't have a  type with `type: 'container'`. `Source` is hardcoded to `type: 'view'`. I'll leave this, I think

react-components/src/components/Reveal3DResources/queryMappedData.ts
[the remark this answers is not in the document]
Savokr: Assume that you have `item` object that represents retrieved edge, I mean that you want to filter by `item.properties.{dataModelSpace}.{connectionExternalId}/{version}.revisionId`. I was thinking that the api works as defining each "step" towards desired property, i.e. in that case something like `['edge', 'properties', {dataModelSpace}, ...]`. So, in your filter you don't have first two elements and I was wondering whether my understanding of the API is wrong.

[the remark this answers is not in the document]
Savokr: Ok, I guess it's just works since it is an actual `property` of the node, so you don't need to specify these two fields.

react-components/src/components/Reveal3DResources/queryMappedData.ts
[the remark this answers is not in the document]
Savokr: Maybe we can follow convention in js sdk to have a method like `autoPagingToArray`. In the sdk it is a method on the actual response but we can make it as something external if it's more convenient. Also, if it's too much work we can postpone it for later.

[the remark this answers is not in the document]
Savokr: Also, this way it will support all kinds of request that can be added in the future since the scheme for `nextCursor` is the same for all requests.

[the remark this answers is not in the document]
haakonflatval-cognite: Yes, that would be nice - but I see that the implementation of `autoPagingToArray` in the Cognite SDK is rather convoluted (https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/core/src/autoPagination.ts), and in our case we may have to support result lists with different field names, and not just `items` as they are called in the SDK. I'd rather do this at a later point

react-components/src/components/Reveal3DResources/queryMappedData.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  const dataNode = mappings.edges[0].startNode;
+
+  const inspectionResult = await inspectNode(fdmClient, dataNode);
+
+  const dataView =
+    inspectionResult.items[0]?.inspectionResults.involvedViewsAndContainers?.views[0];
+
+  const nodeData = await filterNodeData<NodeType>(fdmClient, dataNode, dataView);
```
Savokr, line 162: Why do we need these 2 requests? I think it should be nice just to return an externalId of connected equipment from `mappings` without anything else. Then the user should define what to do with that. Also, it is quite slow when having so much queries.

react-components/src/components/Reveal3DResources/queryMappedData.ts
```diff
@@ -0,0 +1,169 @@
...
+          equals: {
+            property: ['node', 'externalId'],
+            value: dataNode.externalId
+          }
+        }
+      ]
+    },
+    'node',
+    dataView
+  );
```
Savokr, line 117: Isn't it a retrieval of the node by `externalId`? Shouldn't we use [this api call](https://api-docs.cognite.com/20230101/tag/Instances/operation/byExternalIdsInstances/) for that?

react-components/stories/HighlightNode.stories.tsx
Savokr, line 1: I was thinking of adding an example usage in the same `Reveal3DResources` component story, so that all props examples belong in the same place. Also, if we want to have this as a separate story it is better to use `Reveal3DResources` component and new prop that you've added instead of doing the same thing as the prop externally. I was thinking that we don't want to expose `queryMappedData` functions, so the story is not really helping as an example usage since users won't have access to these functions.

react-components/stories/HighlightNode.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+type Story = StoryObj<typeof meta>;
+
+const token = new URLSearchParams(window.location.search).get('token') ?? '';
+const sdk = new CogniteClient({
+  appId: 'reveal.example',
+  baseUrl: 'https://greenfield.cognitedata.com',
+  project: '3d-test',
+  getToken: async () => await Promise.resolve(token)
+});
+
```
Savokr, line 47: There is a helper function `createSdkByUrlToken` that does exactly that, that is used across most stories (not sure why it was not used in toolbar and some others).

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
```diff
@@ -65,6 +73,29 @@ export const Reveal3DResources = ({
...
+          viewer,
+          client,
+          fdmSdk,
+          fdmAssetMappingConfig,
+          event
+        );
+        if (onNodeClick !== undefined && data !== undefined) {
+          onNodeClick?.(data);
+        }
+      })(event);
```
Savokr, line 89: Should consider wrapping it in a hook that uses queries but could be a different PR. Will help a little bit with performance, though.

react-components/stories/HighlightNode.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+    };
+
+    viewer.on('click', (e: PointerEventData) => {
+      queryAndSetData(e);
+    });
+
+    return (): void => {
+      viewer.off('click', queryAndSetData);
+    };
+  });
```
Savokr, line 87: Seems like infinite loop that we saw today is triggered by this hook not having any dependencies.

Savokr (changes requested):
Prop on `Reveal3DResources` works good but query feels a little too slow. I think when we will be highlighting clicked objects it will feel quite bad for the user, so we need to think how to speed it up a little bit.

react-components/src/components/Reveal3DResources/queryMappedData.ts
haakonflatval-cognite: Ah, you're right - I've just taken some shortcuts since we already had that query available

react-components/src/components/Reveal3DResources/queryMappedData.ts
haakonflatval-cognite: Right... Perhaps that is the best, yes :thinking: Will ask Deep about that

react-components/stories/HighlightNode.stories.tsx
haakonflatval-cognite: Makes sense - thanks for finding it :+1:

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
haakonflatval-cognite: Agree :+1:  https://cognitedata.atlassian.net/browse/UX-1147

react-components/stories/HighlightNode.stories.tsx
haakonflatval-cognite: As `onNodeClick` doesn't really exhibit any default behavior, I'm a bit reluctant to putting it into the Reveal3DResources story. But yes, should defiinitely use the `onNodeClick` prop here

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
```diff
@@ -35,24 +39,28 @@ export type Reveal3DResourcesStyling = {
   groups?: FdmAssetStylingGroup[];
 };
 
-export type Reveal3DResourcesProps = {
+export type Reveal3DResourcesProps<NodeType = any> = {
```
Savokr, line 42: We can change to `NodeType = { [key: string]: any }` but can be postponed for later.

Savokr (approved):
Looks good! We will probably need to do something regarding the time of all requests but can fix it in future PRs.

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
haakonflatval-cognite: Alright  :+1: I say we fix later when we have more feedback on the table

PR #3577, opened 2023-08-14
fix(react-components): mimic react.memo type signature with withSuppressRevealEvents HOC

christjt:
#### Type of change

Bug 

## Description :pencil:

Change type signature of withSuppressRevealEvents to allow wrapping components without intrisic attributes.


haakonflatval-cognite (approved):
LGTM :+1:

PR #3583, opened 2023-08-15
chore: simplify quality component

christjt:


pramodcog (commented):
Looks much simple and clean 👍. Just one minor thought on custom high fidelity option, it would be appropriate that we allow the component to have optional budget parameters like only set `maximumRenderCost` in CAD or combination of few like `maxRenderResolution` and `numberOfPoints`. Here we are mandating to add value in `QualitySettings`.

pramodcog (approved):

PR #3589, opened 2023-08-16
feat: add Image360Details component

christjt:
#### Type of change

Feat  

## Description :pencil:

Adds a wrapper around Image360HistoricalDetails that levarages the useReveal hook and adds some of the boilerplate needed for the current solution. Also adds an exit button.

Image360 story has been updated to use this new 


react-components/src/components/Image360Details/Image360Details.tsx
```diff
@@ -0,0 +1,79 @@
...
+  position: absolute;
+  bottom: ${({ isExpanded }) => (isExpanded ? '0px' : '40px')};
+  display: flex;
+  flex-direction: column;
+  height: fit-content;
+  width: fit-content;
+  max-width: 100%;
+  min-width: fill-available;
+  transition: transform 0.25s ease-in-out;
+  transform: ${({ isExpanded }) => (isExpanded ? 'translateY(0)' : 'translateY(100%)')};
```
Savokr, line 78: I feel like this animation should be handled inside `Image360HistoricalDetails` components since this div element is just a wrapper for proper positioning.

Savokr: This will also make it a little more clear about what values are used for `transform` and `bottom`. According to design `bottom` should be 16px but a different value is used because of transform.

Savokr (changes requested):
Looks good, a few comments related to design.

react-components/src/components/Image360Details/Image360Details.tsx
christjt: I agree, but this will brake classic Data Explorer, so will defer until we can combine the solutions.

react-components/src/components/Image360Details/Image360Details.tsx
```diff
@@ -0,0 +1,85 @@
...
+  );
+}
+
+const ExitButtonContainer = styled.div`
+  position: absolute;
+  right: 20px;
+  top: 20px;
+  background-color: #ffffff;
+  height: 35px;
+  width: 35px;
```
Savokr, line 69: This should be 36px as it doesn't match button's size.

react-components/src/components/Image360Details/Image360Details.tsx
```diff
@@ -0,0 +1,85 @@
...
+  border-radius: 6px;
```
Savokr, line 70: Details panel, toolbar and all other buttons in FDX has 8px radius, so it would be nice to have the same feel across all buttons. You will also need to create `StyledButton` with 8px border-radius.

Savokr (approved):
Looks good but still minor design things to fix.

react-components/src/components/Image360Details/Image360Details.tsx
Savokr: And because of that when hovered it has tiny transparent part in the corner.

Savokr (approved):
Perfect 🤌

PR #3594, opened 2023-08-18
chore: Fix miscellaneous typos

greenbech:
#### Type of change
Chore 

#### Jira ticket :blue_book:
No JIRA ticket.

## Description :pencil:
I found some TYPOs and some very long-hanging cleanups when I sat ut this repo locally.

It shouldn't be necessary to test the typos. I ran the example and believe it works the same way as previously.


README.md
```diff
@@ -39,7 +39,7 @@ that can be explored in the browser.
 
 ### Building on Macbook M1
 
-Building Reveal on Macbook M1 migth require some special care.
+Building Reveal on Macbook M1 might require some special care.
```
greenbech, line 42: Interestingly I didn't have any issues setting this up on my M2 MacBook. Probably make sense to keep this around somewhat longer, but I think this entire section can be deleted at some point.

examples/src/pages/Viewer.tsx
```diff
@@ -76,18 +75,7 @@ export function Viewer() {
...
-        client = createSDKFromToken('reveal.example.example', project, overrideToken);
-      } else if (project && environment) {
-        client = await createSDKFromEnvironment('reveal.example.example', project, environment);
-      } else {
-        client = new CogniteClient({
-          appId: 'reveal.example.example',
-          project: 'dummy',
-          getToken: async () => 'dummy'
-        });
-      }
```
greenbech, line 90: I think it is somewhat smelly first create a mutable variable and then assign it immediately below with an `if` / `else if` / `else` block. If there are just two cases, I would usually go for a ternary operator inline, otherwise, I think it's better to extract a method.

examples/src/pages/Viewer.tsx
```diff
@@ -51,7 +50,7 @@ export function Viewer() {
     // Check in order to avoid double initialization of everything, especially dat.gui.
     // See https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects for why its called twice.
     if (!canvasWrapperRef.current) {
-      return () => {};
+      return;
```
greenbech, line 53: Can revert this if you want, but I see no reason to return a function here instead of just a simple `return`

examples/src/utils/example-helpers.ts
```diff
@@ -203,3 +203,29 @@ export async function createSDKFromEnvironment(
...
+export const getCogniteClient = async ({
+  project,
+  environment,
+  overrideToken
+}: {
+  project: string | null;
+  environment: string | null;
+  overrideToken: string | null;
+}): Promise<CogniteClient> => {
+  if (project !== null) {
```
haakonflatval-cognite, line 216: Invert `if` to avoid nesting

haakonflatval-cognite (dismissed):
Looks good! Just one change request

examples/src/utils/example-helpers.ts
greenbech: This was what I did initially, but I didn't like the throw at the end. I agree about the nesting and defensive programming approach, though, so I think I like this better after all. What do you think?

```typescript
  if (project === null) {
    return new CogniteClient({
      appId: 'reveal.example.example',
      project: 'dummy',
      getToken: async () => 'dummy'
    });
  }

  if (overrideToken !== null) {
    return createSDKFromToken('reveal.example.example', project, overrideToken);
  }

  if (environment !== null) {
    return await createSDKFromEnvironment('reveal.example.example', project, environment);
  }

  throw new Error('Unsupported arguments to getCogniteClient');
```

haakonflatval-cognite: Perhaps we can make the two latter conditions into `overrideToken !== null && project !== null` and `environment !== null && project !== null`? And then move the default SDK construction to the end? A little redundant, but I think I like it better

greenbech: Make sense. Updated in: [5caf7a8](https://github.com/cognitedata/reveal/pull/3594/commits/5caf7a8c57444988a674980d45f28c0cc9e97c2f)

haakonflatval-cognite (approved):

PR #3610, opened 2023-08-22
fix: don't recompute clicked node data every render and reset state on clicking nothing

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/UX-1294

## Description :pencil:

Fix a bug where `useClickedNodeData` would return a new value every time it is called, even though no new node was clicked. This could cause unnecessary renders, and infinite render loops.

Also, make sure clicking nothing resets the clicked node data to `undefined`.

Also - don't send `undefined` as temporary result after new click.

In HighlightNode story. 

In the `HighlightNode` story, make the `useEffect` rely on `nodeData` instead of `nodeData?.fdmNode`. This will cause an infinite loop with the old implementation, but not the new.


react-components/src/hooks/useClickedNode.tsx
```diff
@@ -43,16 +44,21 @@ export const useClickedNodeData = (): ClickedNodeData | undefined => {
       cadIntersection?.treeIndex
     ).data ?? [];
 
-  if (cadIntersection === undefined || nodeData.length === 0) {
-    return undefined;
-  }
+  useEffect(() => {
+    if (cadIntersection === undefined || nodeData.length === 0) {
```
Savokr, line 48: I feel like we should remove the second condition and set `clickedNodeData` with undefined `fdmNode`, `view` and probably `cadNode` when we click on non-mapped object. And the reason for that is that with this hook we are basically creating a "react-y" API for `viewer.on('click', ...)` and if we won't add capabilities for normal non-mapped clicks the user will be forced to have both in place if they want to just access, let's say, a clicked cursor position on the model or treeIndex for styling. I think it would be really useful to have it. Since with current behaviour when you have `clickedNodeData` as a dependency in useEffect hook when you click twice on non-mapped part of the model the hook is updated only for the first time if previously some mapped node was selected.

Savokr (changes requested):
Looks and feels really good, one comment related to some change in behaviour of the hook.

react-components/src/hooks/useClickedNode.tsx
```diff
@@ -43,16 +44,28 @@ export const useClickedNodeData = (): ClickedNodeData | undefined => {
...
+    const chosenNode = nodeData[0];
 
-  const chosenNode = nodeData[0];
+    setClickedNodeData({
+      intersection: cadIntersection,
+      fdmNode: chosenNode.fdmId,
+      view: chosenNode.view,
+      cadNode: chosenNode.cadNode
+    });
+  }, [nodeData]);
```
Savokr, line 68: `cadIntersection` should be listed as a dependency here.

Savokr (approved):
Feels just perfect! 🔥

christjt (approved):

PR #3613, opened 2023-08-23
fix: changing default style does not update styling

christjt:
#### Type of change

Bug 

## Description :pencil:

Fixes an issue where changing the default styling does not affect the models.


react-components/src/hooks/useCalculateModelsStyling.tsx
```diff
@@ -38,14 +49,13 @@ export const useCalculateModelsStyling = (
...
-function useCalculateMappedStyling(models: TypedReveal3DModel[]): ModelStyleGroup[] {
-  const modelsRevisionsWithMappedEquipment = models.filter(
-    (model) => model.styling?.mapped !== undefined
-  );
-  const shouldFetchAllMappedEquipment = modelsRevisionsWithMappedEquipment.length > 0;
+function useCalculateMappedStyling(
+  models: CadModelOptions[],
+  defaultMappedNodeAppearance?: NodeAppearance
+): ModelStyleGroup[] {
+  const modelsRevisionsWithMappedEquipment = getMappedCadModelsOptions();
```
Savokr, line 56: This should be wrapped in `useMemo` since it updates the array every re-render.

Savokr (changes requested):
Looks really good, only one comment.

Savokr (dismissed):
👍

Savokr (approved):

PR #3620, opened 2023-08-26
fix: Use event.code instead of deprecated event.keyCode

greenbech:
#### Type of change
Bug 

#### Jira ticket :blue_book:
No JIRA ticket.

## Description :pencil:
This PR changes the `Keyboard` class to use the [`event.code` API](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead of the deprecated `event.keyCode`.  According to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode):

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

The practical reason I made this PR is that the `event.keyCode` doesn't work as expected on my non-standard keyboard layout (Colemak). With the use of the `event.code` API, I can now click on the physical WASDQE keys and navigate around in the 3D model.

I've also taken the liberty to refactor the code. Let's test the PR properly before merging. I did, for instance, remove one `event.preventDefault` that I didn't see a good reason to have. However, maybe I've overlooked something so please comment if you think I should add it back. 

For most people, this should work exactly as before. Open the example and navigate with WASDQE and the arrow. If you use a non-standard keyboard layout (such as Dvorak or Colemak), you should now be able to use the physical keys corresponding to the QWERTY layout.


viewer/packages/camera-manager/src/Keyboard.ts
```diff
@@ -46,51 +55,67 @@ export default class Keyboard {
...
-  private readonly onKeydown = (event: KeyboardEvent) => {
-    if (event.metaKey || event.altKey || event.ctrlKey) {
-      return;
-    }
-
-    if (event.keyCode in keyMap) {
-      if (this.keys[keyMap[event.keyCode]] === 0) {
-        this.keys[keyMap[event.keyCode]] = 2;
-      }
-      event.preventDefault();
```
haakonflatval-cognite, line 81: I think we should include this still, not sure if it's entirely necessary, but I'd prefer to break as little of the behavior as necessary. That probably means we have to check for `metaKey` etc. again too, since preventing default-behavior for those buttons disable most browser key shortcuts, i believe.

haakonflatval-cognite (changes requested):
Nice! A much-needed refactoring of this code.

A couple of comments

viewer/packages/camera-manager/src/Keyboard.ts
greenbech: Are you sure you want to keep it? I am pretty certain that removing this won't have any practical differences in behavior, and adding it back makes the code quite a lot less nice IMO. This is what ChatGPT says about it:

image

(😅 I added the word "legacy" to nudge ChatGPT to say: "That can probably be deleted")

viewer/packages/camera-manager/src/Keyboard.ts
```diff
@@ -1,30 +1,39 @@
...
+  'KeyQ',
+  'KeyS',
+  'KeyW'
+] as const;
+
+type EventCode = (typeof EVENT_CODES)[number];
 
 export default class Keyboard {
-  private keys: { [s: string]: number } = {};
+  private readonly _keys = new Set<string>();
```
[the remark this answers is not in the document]
greenbech: Yeap, makes sense. Added that functionality here: [1736c75](https://github.com/cognitedata/reveal/pull/3620/commits/1736c75cd0d0c09d48d7b89111b798b7429054ff)

viewer/packages/camera-manager/src/Keyboard.ts
haakonflatval-cognite: Okay, I'll accept it - We'll hear soon enough if this messes with someone :P

viewer/packages/camera-manager/src/Keyboard.ts
```diff
@@ -32,8 +32,12 @@ const EVENT_CODES = [
 
 type EventCode = (typeof EVENT_CODES)[number];
 
+const isEventCode = (value: string): value is EventCode => {
+  return EVENT_CODES.some(eventCode => eventCode === value);
```
haakonflatval-cognite, line 36: Can we put the events in a set as a pre-processing step to make this lookup faster?

haakonflatval-cognite (changes requested):

viewer/packages/camera-manager/src/Keyboard.ts
greenbech: Done in: [140c08c](https://github.com/cognitedata/reveal/pull/3620/commits/140c08cd96b3f1b6d81a834ed03992cdd9ccb551)

haakonflatval-cognite (approved):

PR #3641, opened 2023-08-29
feat: overwrite nodes/treeIndices if styling is equal

christjt:
#### Type of change

Feat  

## Description :pencil:

Style cache optimization when styles are equal but tree indices / node ids differ. When this happens we can simply overwrite the node collection instead of reapplying the style.


react-components/src/components/CadModelContainer/useApplyCadModelStyling.tsx
```diff
@@ -99,32 +99,54 @@ async function applyStyling(
   }
 }
 
-function isEqualStylingGroupAndCollection(
+async function tryUpdateStylingGroup(
```
haakonflatval-cognite, line 102: Still unsure about this name - could we have something that includes e.g. `checkSimilarity...`?

haakonflatval-cognite (changes requested):

haakonflatval-cognite (approved):

PR #3679, opened 2023-09-06
feat: Stop old batch loading

Strepto:
#### Type of change
Feat  

#### Jira ticket :blue_book:

## Description :pencil:

I think this is a simple fix that will potentially reduce waste in loading and then discarding sectors. I have not extensively tested this, but it seems like a easy win for improved loading performance.

I know this fix is not _as_ important with remote models as the warning `Abort signal is not supported for local models` appears in the log and I presume this would have cancelled something.

Tested with local model in chrome using dev-tools to throttle network latency to 200ms. Also tested with remote Equinor-models, with improvements when the network or device is slow. Combined with #3674 I believe this is a big difference in performance.

Lots of more possible improvements (should cancel the ongoing batch for instance) but this is a oneliner, so hard to beat.

I suggest playing around with the code to see the impact

for example this debug code is good to show the loading of old sectors still active way later.

```

    for (const batch of chunk(changedSectors, SectorLoadingBatchSize)) {
      if (currentBatchId !== this._batchId) {
        console.log('Stopped loading old batch ', currentBatchId, this._batchId);
        // Uncomment the break to see the current behavior
        break; // Stop processing this batch as a new batch has started, and will discard results from old batches.
      }

      const filteredSectors = await this.filterSectors(sectorCullerInput, batch, sectorCuller, progressHelper);
      const consumedPromises = this.startLoadingBatch(filteredSectors, cadModels);
      for await (const consumed of PromiseUtils.raceUntilAllCompleted(consumedPromises)) {
        const resolvedSector = consumed.result;
        if (currentBatchId === this._batchId) {
          if (resolvedSector !== undefined) {
            this._modelStateHandler.updateState(
              resolvedSector.modelIdentifier,
              resolvedSector.metadata.id,
              resolvedSector.levelOfDetail
            );
            yield resolvedSector;
          }
        } else {
          console.log('Discarded sector due to old batch', currentBatchId, this._batchId);
        }
        progressHelper.reportNewSectorsLoaded(1);
      }
    }
  }
```


viewer/packages/cad-geometry-loaders/src/sector/SectorLoader.ts
```diff
@@ -93,6 +93,7 @@ export class SectorLoader {
     const currentBatchId = this._batchId;
 
     for (const batch of chunk(changedSectors, SectorLoadingBatchSize)) {
+      if (currentBatchId !== this._batchId) break; // Stop processing this batch as a new batch has started, and will discard results from old batches.
```
Strepto, line 96: Line 101 discards any result, so this cancels out early before sending the request at all.

christjt (approved):

PR #3682, opened 2023-09-06
feat: add a higher order component for storing camera state to URL

christjt:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/UX-1350

## Description :pencil:

Adds a HOC which can be mounted to any component. As long as the component is mounted, camera state will be stored in the URL when the camera stop event triggers.

POC is implemented in FDM.

In the toolbar story, append:
```
&cameraPosition=%5B13.54764613001787%2C18.554701782663916%2C9.87398745289454%5D&cameraTarget=%5B12.65278161163936%2C4.1501122422409775%2C-4.756918093283632%5D
```

to the URL. Notice that the starting camera position is now defined by this state.

To test the setting of the URL open the canvas of the story in a new window (see image). This is needed because the content of the story is in an iFrame, so updates in the URL are happening in the iFrame.

image

Once opened, move the camera and confirm that the state changes.


react-components/src/hooks/useCameraNavigation.tsx
```diff
@@ -56,9 +57,14 @@ export const useCameraNavigation = (): CameraNavigationActions => {
     await fitCameraToModelNode(modelMappings.revisionId, nodeId.id);
   };
 
+  const fitCameraToState = (cameraState: CameraState): void => {
+    viewer.cameraManager.setCameraState(cameraState);
```
pramodcog, line 61: I see in the Toolbar story check for the `cameraState` validation but it make more sense to check within `fitCameraToState`

pramodcog (approved):
Works well! minor comment

react-components/src/hooks/useCameraNavigation.tsx
christjt: Why? I don't think it makes sense to accept undefined here? By that logic every argument to every function should be `T | undefined`.

react-components/src/higher-order-components/withCameraStateUrlParam.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }, []);
+
+    return <Component {...props} />;
+  };
+
+  function hasCameraStateChanged(
+    previous: CameraStateTransform,
+    current: CameraStateTransform
+  ): boolean {
+    const epsilon = 0.001;
```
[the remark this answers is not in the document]
christjt: Disagree, first of all I don't see the point in refactoring it out into a separate file since it is only used once. We can consider if it is used multiple times in different places. Having a utilities/constants.ts in the first place seems like a mistake to me. Should every constant in the entire code base be here? How does that make anything easier? In the very least it should be grouped by function. I don't f.ex. see what `DEFAULT_QUERY_STALE_TIME` has to do with `METERS_TO_FEET`; they are not related what so ever.

PR #3714, opened 2023-09-19
bug: Update Moving Resolution Factor behavior

Strepto:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:

Now takes the max resolution into account.
The MCRF would often use the max resolution when it was running on a large screen. This adjusts it to use the min of: the percentage of the max resolution and the percentage of the current estimated resolution.

Tested in client. No integration tests or unit tests.

Previous behavior:
Set the renderResolutionThreshold to a relatively low number (For instance 1360 * 720). And try to use moving camera resolution factor. On a low res screen it will work, but on a high res screen the resolution will not change.

New behavior: 
If the resolution is lower than the max, we use half that. If the current resolution is higher than max, we adjust the therenderResolutionThreshold isntead.

### DIFF
--- a/viewer/packages/rendering/src/ResizeHandler.ts
+++ b/viewer/packages/rendering/src/ResizeHandler.ts
@@ -36,10 +36,8 @@ export class ResizeHandler {
     this._cameraManager = cameraManager;
 
     this._onCameraChangeCallback = () => {
-      this._currentResolutionThreshold = Math.min(
-        this._movingResolutionFactor * getPhysicalSize(renderer),
-        this._movingResolutionFactor * this._stoppedCameraResolutionThreshold
-      );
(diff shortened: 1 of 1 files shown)


Strepto:
Hm, I see the test covering this expects the higher values of the two given resolutions to be picked I guess. But it seems wrong to me. I'll update the test if you agree that the logic is wrong.

haakonflatval-cognite:
Hi - I do see the point you're making, that it may be more intuitive that `movingCameraResolutionFactor` operates on the already-reduced resolution as opposed to the original one. 
However, the TsDoc (in `viewer/packages/api/src/public/migration/types.ts`) specifies that the factor is relative to the number of physical pixels of the canvas, so we'd rather not change it now.

PR #3756, opened 2023-09-29
fix: update revision id for all version for test model

christjt:
#### Type of change

Bug 

## Description :pencil:

Need to update the source model in docs due to regression in /byIds endpoint.


haakonflatval-cognite (approved):
LGTM :+1: :crossed_fingers:

PR #3771, opened 2023-10-05
feat(react-components): asset mapping click handling and "mapped"-styling

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/UX-1583

## Description :pencil:

Add click-handling support for nodes that are asset-mapped through classic asset mappings.

Note that this code is not optimized for _many_ asset mappings, and assumes filtering them locally is feasible for e.g. on-click operations.

Also, this breaks the API of the `useClickedNode` hook. There may be a way to better future proof this, open to suggestions.

In storybook, HighlightNode.

Run HighlightNode. Replace the model in the code by 
```
{
        modelId: 1791160622840317,
        revisionId: 502149125550840,
}
```

Click on different nodes (the asset-mapped ones should be highlighted), and see that the console logs the data from the nodes.


react-components/src/components/NodeCacheProvider/AssetMappingCache.ts
```diff
@@ -0,0 +1,87 @@
...
+import { type ModelId, type ModelRevisionKey, type RevisionId } from './types';
+import { maxBy } from 'lodash';
+import assert from 'assert';
+
+export type NodeAssetMappingResult = { node?: Node3D; mappings: AssetMapping3D[] };
+
+export class AssetMappingCache {
+  private readonly _sdk: CogniteClient;
+
+  private readonly _modelToAssetMappings = new Map<ModelRevisionKey, AssetMapping3D[]>();
```
Savokr, line 15: Would it be more efficient to use a map instead of an array for asset mappings?

react-components/src/components/NodeCacheProvider/AssetMappingCacheProvider.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const content = useContext(AssetMappingCacheContext);
+
+  if (content === undefined) {
+    throw Error('Must use useAssetMappingCache inside a AssetMappingCacheContext');
+  }
+
+  return content.cache;
+};
+
+export const useMappeNodesForRevisions = (
```
Savokr, line 36: Should we add `asset` word to the name of the hook for easier distinction between this one and fdm based?

react-components/src/hooks/useClickedNode.tsx
```diff
@@ -54,58 +59,85 @@ export const useClickedNodeData = (): ClickedNodeData | undefined => {
     cadIntersection?.treeIndex
   ).data;
 
+  const assetMappingResult = useAssetMappingForTreeIndex(
+    cadIntersection?.model.modelId,
+    cadIntersection?.model.revisionId,
+    cadIntersection?.treeIndex
+  ).data;
+
+  return useCombinedClickedNodeData(nodeDataPromises, assetMappingResult, cadIntersection);
```
Savokr, line 68: Current behaviour is that when you click, the return data for asset mapping is first undefined and then it updates, should we update the return data only when we got the data?

Savokr: Also, if you click on "non-mapped" part of the model it will still return the data two times instead of just one. Which will trigger any dependant useEffect hook.

react-components/src/components/NodeCacheProvider/AssetMappingCache.ts
```diff
@@ -0,0 +1,87 @@
...
+    ancestors: Node3D[]
+  ): Promise<NodeAssetMappingResult> {
+    if (ancestors.length === 0) {
+      return { mappings: [] };
+    }
+
+    const searchTreeIndices = new Set(ancestors.map((ancestor) => ancestor.treeIndex));
+    const allModelMappings = await this.getAssetMappingsForModel(modelId, revisionId);
+
+    const relevantMappings = [...allModelMappings.values()].filter((mapping) =>
```
Savokr, line 33: Conversion to array doesn't look very efficient here 🙂. I'd suggest testing it on at least 100-200k mappings and see whether it triggers any performance issues.

react-components/src/hooks/useClickedNode.tsx
```diff
@@ -54,58 +59,85 @@ export const useClickedNodeData = (): ClickedNodeData | undefined => {
...
+      if (views === undefined || views.length === 0) {
         return;
       }
 
-      setClickedNodeData({
-        intersection: cadIntersection,
-        fdmNode: data.fdmIds[0],
-        cadNode: data.cadNode,
+      setFdmData({
+        fdmNode: cadAndFdmNodes.fdmIds[0],
```
Savokr, line 135: Should we return all fdmIds here? The same applies to views.

Savokr (approved):
Looks and works good, a few small questions.

react-components/src/components/NodeCacheProvider/AssetMappingCache.ts
haakonflatval-cognite: Hmm, in case we need to lookup asset mappings, you mean? That's true :thinking: But I think I'll postpone that decision until it's needed

react-components/src/components/NodeCacheProvider/AssetMappingCacheProvider.tsx
haakonflatval-cognite: Yes, good call

react-components/src/hooks/useClickedNode.tsx
haakonflatval-cognite: The reason I update both when clicking, and then when getting data, is that it enables the user to interact with the `intersection` itself, without having to wait for the data.

haakonflatval-cognite: It works like this for the FDM part as well.

I did make a small change to make sure you don't get two updates when clicking an unmapped node, though :+1:

react-components/src/components/NodeCacheProvider/AssetMappingCache.ts
haakonflatval-cognite: I'm not too worried about this - Fetching the assets to begin with would be much worse performance-wise. It's a linear search through an array. But anyway, I don't know why I'm doing this `[...list.values()]`-thing at all. It's just a list to begin with :thinking: I'll simplify it.

react-components/src/hooks/useClickedNode.tsx
haakonflatval-cognite: Agreed, since we're already changing the type signature of the function :+1: Good call

react-components/src/components/NodeCacheProvider/AssetMappingCacheProvider.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const content = useContext(AssetMappingCacheContext);
+
+  if (content === undefined) {
+    throw Error('Must use useAssetMappingCache inside a AssetMappingCacheContext');
+  }
+
+  return content.cache;
+};
+
+export const useAssetMappeNodesForRevisions = (
```
Savokr, line 35:
```suggestion
export const useAssetMappedNodesForRevisions = (
```

Small typo

Savokr (approved):
Looks good, tiny rename comment.

PR #3788, opened 2023-10-13
feat: add search hooks for 360 annotation asset

pramodcog:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/UX-1536

## Description :pencil:
Created a hook for searching annotation asset in 360 image data sets. FDX or any other application will be to use it without the need to initialize `RevealContainer`.

In storybook & FDX (via `yalc`)

- `yarn storybook`
- Select `AnnotationAsset` example


react-components/src/hooks/use360ImagesFileIds.tsx
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  return queryResult;
+};
+
+const listFileIds = async (filter: FileFilterProps, sdk: CogniteClient): Promise<number[]> => {
+  const req = { filter, limit: 1000 };
+  const map = await sdk.files.list(req).autoPagingToArray({ limit: Infinity });
+
+  const fileInfo = await Promise.all(map.flat());
```
Savokr, line 45: Why do you need to await here if the request is already awaited above?

Savokr (commented):
Looks good overall but haven't tested yet, so this is more of a general feedback. A few comments and we should probably align together on the API for this search hooks.

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -0,0 +1,146 @@
...
+  return result;
+}
+
+async function get360AnnotationAssets(
+  image360Annotations: AnnotationModel[],
+  sdk: CogniteClient
+): Promise<Asset[]> {
+  const annotationMapping = image360Annotations.map((annotation) => ({
+    assetId:
+      ((annotation.data as AnnotationsBoundingVolume).assetRef?.id as number) ??
```
Savokr, line 81: What happens when assetRef is undefined? We should handle this case as well. I think type conversion to number and string shouldn't be used here.

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -0,0 +1,146 @@
...
+      ((annotation.data as AnnotationsBoundingVolume).assetRef?.externalId as string)
+  }));
+
+  const uniqueAnnotationMapping = uniqBy(annotationMapping, 'assetId');
+
+  const assets = chunk(uniqueAnnotationMapping, 1000).map(async (uniqueAssetsChunk) => {
+    const retrievedAssets = await sdk.assets.retrieve(
+      uniqueAssetsChunk.map(({ assetId }) => ({
+        [typeof assetId === 'number' ? 'id' : 'externalId']: assetId
+      })) as unknown as IdEither[],
```
Savokr, line 91: Why we have `as unknown` here?

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+  return (await Promise.all(assets)).flat();
+}
+
+async function get360ImagesFileIds(siteIds: string[], sdk: CogniteClient): Promise<number[]> {
+  const fileIdListPromises = siteIds.map(async (siteId) => {
+    const req = {
+      metadata: { site_id: siteId }
+    };
+    const fileIds = await listFileIds(req as FileFilterProps, sdk);
```
Savokr, line 105: Type conversion can be moved to just type annotation on the filter  itself

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -0,0 +1,146 @@
...
+        annotatedResourceIds: fileIdsChunk?.map((id) => ({ id })) ?? [],
+        annotatedResourceType: 'file',
+        annotationType: 'images.AssetLink'
+      };
+      const annotations = await sdk.annotations
+        .list({
+          limit: 1000,
+          filter
+        })
+        .autoPagingToArray({ limit: Infinity });
```
Savokr, line 130: Why do you need autoPaging here if you always fetch <=1000 annotations, so one page exactly? Because autoPaging will fetch all annotations, as far as I can tell, you are fetching all annotations for each chunk of initial array

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -0,0 +1,146 @@
...
+  );
+
+  return annotationArray.flat();
+}
+
+async function listFileIds(filter: FileFilterProps, sdk: CogniteClient): Promise<number[]> {
+  const req = { filter, limit: 1000 };
+  const map = await sdk.files.list(req).autoPagingToArray({ limit: Infinity });
+
+  const fileInfo = await Promise.all(map.flat());
```
Savokr, line 142: Why do you need Promise.all here if you are awaiting the result before that already?

react-components/stories/AnnotationAsset.stories.tsx
Savokr, line 1: After merge of my PR we can maybe unite the stories in one. I updated some styles in my Story, so at least we can make them the same. But it's not very important of course.

Savokr (changes requested):
Looks good but a few things that should be fixed before merging.

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
pramodcog: map.flat() returns an array of promises, and we need to wait for all of these.

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const assets = chunk(uniqueAnnotationMapping, 1000).map(async (uniqueAssetsChunk) => {
+    const retrievedAssets = await sdk.assets.retrieve(
+      uniqueAssetsChunk.map(({ assetId }) => {
+        if (typeof assetId === 'number') {
+          return { id: assetId };
+        } else if (typeof assetId === 'string') {
+          return { externalId: assetId };
+        } else {
+          throw new Error(`Invalid assetId: ${assetId}`);
```
Savokr, line 95: I think you should not throw here.  Because if you have a 360 image set where all except 1 annotation have assetRef set, then the whole set will be impossible to search through. I would suggest just initially filtering out all annotations that have undefined `assetRef`.

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -0,0 +1,149 @@
...
+  );
+
+  return annotationArray.flatMap((annotations) => annotations);
+}
+
+async function listFileIds(filter: FileFilterProps, sdk: CogniteClient): Promise<number[]> {
+  const req = { filter, limit: 1000 };
+  const map = await sdk.files.list(req).autoPagingToArray({ limit: Infinity });
+
+  const fileInfo = await Promise.all(map.flat());
```
Savokr, line 145: I don't think you need to use Promise.all here, see how we use the same method for point cloud annotation works without that part and flattening. 

https://github.com/cognitedata/reveal/blob/master/viewer/packages/data-providers/src/pointcloud-stylable-object-providers/CdfPointCloudStylableObjectProvider.ts (fetchAnnotations method)

Savokr (commented):
Looks and works really good, a few comments about one important use case.

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -0,0 +1,153 @@
...
```
Savokr, line 149: This can be removed but we can ignore it for now

Savokr (approved):
Looks good!

PR #3789, opened 2023-10-13
Bug: Picking part in multiple overlapping CADModels does not always return closest part

Strepto:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:

We need to check depth for all nodes to correctly do the sorting at the end. Currently the same depth is reused, so if we have multiple cadNodes the sorting at the end of the method may return wrong data.

Manual testing with two models overlapping

I'm not proficient enough in the testing to create a integration test with two models in Reveal. If you'd like to help me with that I might be able to add one.

Load two different slightly overlapping models. Position camera within one or both models. The intersection should be correct now, but is wrong in the old code. 


viewer/packages/cad-model/src/picking/PickingHandler.ts
```diff
@@ -123,9 +120,10 @@ export class PickingHandler {
...
         visibleCadNodes.forEach(p => (p.visible = false));
         cadNodeData.cadNode.visible = true;
-
         const treeIndex = await this.intersectCadNodeTreeIndex(cadNodeData.cadNode, input, async);
         if (treeIndex) {
+          // Assuming we have depth anywhere we hit a treeIndex
+          const depthResult = await this.intersectCadNodeDepth(depthInput, async);
           const result: IntersectCadNodesResult = {
             distance: depthResult.distance,
             point: depthResult.point,
```
Strepto, line 129: In the old code the reuse of the depthResult here will cause the `return results.sort((l, r) => l.distance - r.distance);` to just return a node, independent of real distance. If two models overlap this may not be the closest treeIndex.

haakonflatval-cognite (approved):
Nice, thanks! LGTM :+1:

PR #3883, opened 2023-11-09
fix: Camera moving resolution from scaled output size

hugolafis:
#### Type of change

Bug 

## Description :pencil:
This is a fix for how the current scaling factor works whilst the camera is moving. At the moment it either takes the minimum of either the capped stopped resolution, or the uncapped target resolution of the canvas multiplied by the `movingResolutionFactor`, which causes issues when the `_stoppedCameraResolutionThreshold` is a lower value (the movingResolutionFactor stops having an effect, and has a decreasing effect the closer these two values become).

This change adjusts the behaviour so that the calculated buffer size is always multiplied by the `movingResolutionFactor` when the camera is moving, which I believe is the intended behaviour and how I assumed it worked from the docs.

I've tested this out locally by printing out the buffer sizes when the camera is moving and paused. Beforehand, the buffer size would not change when the camera was moving if the `_stoppedCameraResolutionThreshold` was set low.

### DIFF
--- a/viewer/packages/rendering/src/ResizeHandler.ts
+++ b/viewer/packages/rendering/src/ResizeHandler.ts
@@ -26,6 +26,7 @@ export class ResizeHandler {
   private readonly _resizeObserver: ResizeObserver | undefined;
 
   private _shouldResize: boolean = false;
+  private _cameraMoving: boolean = false;
 
   constructor(renderer: WebGLRenderer, cameraManager: CameraManager, resizeOptions?: ResizeHandlerOptions) {
     this._stoppedCameraResolutionThreshold =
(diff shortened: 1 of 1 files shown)


haakonflatval-cognite:
Hi - I believe this is the same change as was proposed in https://github.com/cognitedata/reveal/pull/3714? I closed that PR, as the related TSDocs (in `viewer/packages/api/src/public/migration/types.ts`) asserts the current behavior.

But I see that this is something we could reconsider, since it is the second PR we have on this issue. I would rather wait for version 5.0 to change this though, as it will break current behavior. I am adding it to our internal list of wanted feature for the new major version :+1: 

Also, I agree that our docs are vague on how this feature actually works.

hugolafis:
Apologies, I didn't see that other PR. I'll close this one out then for the moment since the topics already been covered.

PR #3946, opened 2023-11-22
fix: Disposed geometryGroup softlocks sector load

Strepto:
#### Type of change
Bug 

#### Jira ticket :blue_book:

## Description :pencil:

This fixes an issue where a sector would load a disposed geometry group, and then never be loadable again.

The stack trace of the issue this fixes: 

```text
Uncaught Error: Already disposed/dereferenced
    at U.ensureNotDisposed (index.js:2:1391445)
    at U.reference (index.js:2:1390996)
    at bt.updateGeometry (index.js:2:1449747)
    at r (index.js:2:1911821)
    at Object.next (index.js:2:1912155)
    at e.next (index.js:2:2678779)
    at n._next (index.js:2:2678354)
    at n.next (index.js:2:2677942)
    at index.js:2:2706978
    at s._next (index.js:2:2705280)
```

I'm not actually sure why the reference imbalance happens. It seems to be easier to trigger with a small sector MemoryCache, but it may happen without any sector cache at all. It does not seem to happen if network cache is disabled. So its a probably a combination of a few factors and hard to reliably reproduce.

This PR suggests a workaround fix by not failing hard if the sector is already disposed, as this would stop all loading and the sector would never recover from the exception. Only fix for the user was a refresh of the website.

A more long term fix would be to identify the race-condition that causes a disposed sector geometry group to be loaded.

See below, I can try to add some tests if I understand what causes the issue.

(In reveal Examples with a local model): Reduce budget to (1%) and navigate around a model (such as Huldra) quickly. (Move backwards / forwards / spin / anything). The log message should appear quite reliably. In real world testing this is harder to reproduce, but still happens _especially with a low budget_ (3 - 4 million triangles budget). 


viewer/packages/cad-parsers/package.json
```diff
@@ -1,17 +1,18 @@
...
-  },
-  "dependencies": {
-    "@reveal/data-providers": "workspace:*",
-    "@reveal/metrics": "workspace:*",
-    "@reveal/model-base": "workspace:*",
-    "@reveal/sector-parser": "workspace:*",
-    "@reveal/utilities": "workspace:*"
-  }
-}
+{
```
Strepto, line 1: This file was changed to lf file endings, which I believe should be correct.

viewer/packages/cad-parsers/package.json
```diff
@@ -1,17 +1,18 @@
...
+  "private": true,
+  "main": "index.ts",
+  "scripts": {
+    "start": "yarn ws:start",
+    "test": "yarn ws:test --config ./../../jest.config.js",
+    "lint": "yarn ws:lint"
+  },
+  "dependencies": {
+    "@reveal/data-providers": "workspace:*",
+    "@reveal/logger": "workspace:*",
```
Strepto, line 12: Added a logger reference is the only change in this file except for line endings

christjt (approved):
LGTM!

PR #4081, opened 2024-01-10
Refacoring usage of getNormalizedPixelCoordinates

nilscognite:
#### Type of change

Refactor

This is just a tiny cleanup in the code. It removes some repetitive coding around

### DIFF
--- a/viewer/packages/camera-manager/src/StationaryCameraManager.ts
+++ b/viewer/packages/camera-manager/src/StationaryCameraManager.ts
@@ -2,7 +2,6 @@
  * Copyright 2022 Cognite AS
  */
 
-import { assertNever, clickOrTouchEventOffset, pixelToNormalizedDeviceCoordinates } from '@reveal/utilities';
 import * as THREE from 'three';
 import TWEEN from '@tweenjs/tween.js';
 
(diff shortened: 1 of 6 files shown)


viewer/packages/camera-manager/src/StationaryCameraManager.ts
```diff
@@ -285,7 +285,7 @@ export class StationaryCameraManager implements CameraManager {
   private getCursorRay(event: WheelEvent) {
     const { width, height } = this._domElement.getBoundingClientRect();
     const { offsetX, offsetY } = clickOrTouchEventOffset(event, this._domElement);
-    const ndcCoordinates = pixelToNormalizedDeviceCoordinates(offsetX, offsetY, width, height);
+    const ndcCoordinates = getNormalizedPixelCoordinatesBySize(offsetX, offsetY, width, height);
```
pramodcog, line 288: If `getNormalizedPixelCoordinatesBySize` is not used anywhere else within `Reveal`, we can remove it from `wotdToViewport.ts` file and use `getNormalizedPixelCoordinates` here as `this._domElement` is available.

nilscognite: It is also used by pixelToNormalizedDeviceCoordinates. So I want to keep it as it is now.

PR #4125, opened 2024-01-24
fix(react-components): don't limit search result that will be filtered

haakonflatval-cognite:
#### Type of change

Bug 

## Description :pencil:

There's a bug in the asset-search code related to limits.
The code works in two steps - it first fetches assets that satisfy the search query. It then filters these assets based on the mappings available in the 3D model(s). 
Previously, we used the same limit in both these queries, which in many cases causes the result from the first step to be trimmed down to a lower number, like 0. This tries to mitigate it somewhat by always returning the max number of results, 1000, from the first search.

Linked into fusion-shell and this PR: https://github.com/cognitedata/fusion/pull/4867

### DIFF
--- a/react-components/src/hooks/useSearchMappedEquipmentAssetMappings.tsx
+++ b/react-components/src/hooks/useSearchMappedEquipmentAssetMappings.tsx
@@ -33,9 +33,10 @@ export const useSearchMappedEquipmentAssetMappings = (
   userSdk?: CogniteClient
 ): UseQueryResult<Asset[]> => {
   const sdk = useSDK(userSdk);
+  const modelsKey = models.map((model) => [model.modelId, model.revisionId]);
 
   return useQuery(
-    ['reveal', 'react-components', 'search-mapped-asset-mappings', query, models],
+    ['reveal', 'react-components', 'search-mapped-asset-mappings', query, modelsKey],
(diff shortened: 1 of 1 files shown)


react-components/src/hooks/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -45,7 +46,7 @@ export const useSearchMappedEquipmentAssetMappings = (
         return modelsAssets.map(({ assets }) => assets).flat();
       }
 
-      const searchedAssets = await sdk.assets.search({ search: { query }, limit });
+      const searchedAssets = await sdk.assets.search({ search: { query }, limit: 1000 });
```
pramodcog, line 49: Was just wondering what would happen if search result is more than 1000 as we do not have pagination, should we consider it?

pramodcog (approved):
Nice! one minor comment

react-components/src/hooks/useSearchMappedEquipmentAssetMappings.tsx
haakonflatval-cognite: There is no pagination support for the search endpoint, so I suppose we just cross our fingers in that case...

PR #4188, opened 2024-02-13
feat(react-components): Refactor RevealContainer into RevealContext and RevealCanvas

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

## Description :pencil:

With this change, the idea is that all context resources are available separately from the canvas itself. We often want e.g. interaction UI and search to be available outside of the canvas. This was a little tedious to make work before, but with this change, the canvas and the Reveal context can be placed independently (although the canvas must still be inside the RevealContext).

This is a necessary step to make the TopBar feature.

Went through all the storybooks. Will check in Fusion too


react-components/src/components/RevealContext/RevealContext.tsx
```diff
@@ -0,0 +1,103 @@
...
+};
+
+export const RevealContext = (props: RevealContextProps): ReactNode => {
+  const viewer = useRevealFromKeepAlive(props);
+
+  const queryClient = useMemo(() => {
+    return new QueryClient();
+  }, []);
+
+  if (viewer === undefined) return <></>;
```
pramodcog, line 43: This should be moved above `useMemo()`

react-components/src/components/RevealCanvas/RevealCanvas.tsx
```diff
@@ -0,0 +1,26 @@
...
+      {mountChildren()}
+    </div>
+  );
+
+  function mountChildren(): ReactElement {
+    useEffect(() => {
+      if (parentElement.current !== null) {
+        parentElement.current.appendChild(viewer.domElement);
+      }
+    }, [parentElement.current]);
```
pramodcog, line 23: Logically, `parentElement.current` is a mutable value that will not trigger a re-render when it changes, are you sure this gets triggered when `parentElement.current`

react-components/src/components/RevealCanvas/RevealCanvas.tsx
```diff
@@ -0,0 +1,26 @@
...
+ */
+import { type ReactNode, type ReactElement, useRef, useEffect } from 'react';
+import { createPortal } from 'react-dom';
+import { useReveal } from './ViewerContext';
+
+export function RevealCanvas({ children }: { children?: ReactNode }): ReactElement {
+  const viewer = useReveal();
+  const parentElement = useRef<HTMLDivElement>(null);
+
+  return (
```
pramodcog, line 12: I feel below look more clean, what do you think?
`useEffect(() => {`
`    if (parentElement.current !== null) {`
`      parentElement.current.appendChild(viewer.domElement);`
`    }`
`  }, []);`

`  return (`
`    <div style={{ width: '100%', height: '100%', overflow: 'hidden' }} ref={parentElement}>`
`      {createPortal(children, viewer.domElement)}`
`    </div>`
`  );`

react-components/src/components/RevealContext/RevealContext.tsx
haakonflatval-cognite: Hooks cannot be called after conditional return statements. https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

react-components/src/components/RevealCanvas/RevealCanvas.tsx
haakonflatval-cognite: No, I don't think it will retrigger automatically, but if the useEffect runs again at some point, it can use `parentElement.current` to see whether the call is actually needed

haakonflatval-cognite: Maybe it's not entirely needed, but I don't think it's a problem to keep it

react-components/src/components/RevealCanvas/RevealCanvas.tsx
haakonflatval-cognite: I agree :+1: Will change

pramodcog (approved):

PR #4208, opened 2024-02-18
feat (beta): Add new class CustomObject and use it to store custom objects in the scene

nilscognite:
#### Type of change

Feat  

## Description :pencil:
This change introduce a new concept of CustomObject. Before they couldn't have any properties, so Reveal treat them equally.  The new CustomObject, there will be possible to defined the behavior of the custom object.

For the moment it will solve the problem with the groundplane. The groudplane is not part of the data itsself, only a visual effect. So the boundingBox calculation for the entire model should not include  the groundplane. I have also implemented mouse picking for these objects (default off as it was before).  This is done in  order to get correct information to the camera manager when clicking, scrolling and double clicking. Before it worked for CAD or PointClouds only , the custom objects was invisible for the camera manager and will cause misbehavior.

In the class `Cognite3DViewer` added:
```
addCustomObject(customObject: CustomObject)
removeCustomObject(customObject: CustomObject)
```
Since this is a part of the public API, I have made `CustomObject` as a part of the public API.

I have updated some other files, so it uses CustomObject.

**NOTE:**  Reveal use types for most option, but I want to use classes. So  `CustomObject` is a class. This is a more cleaner and object oriented and can be extended in the future to have methods to be overridden. 

**It be marked @beta**

- [] I have commented my code in hard-to-understand areas.


viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1631,6 +1682,42 @@ export class Cognite3DViewer {
...
 
+    const normalizedCoords = getNormalizedPixelCoordinates(this.domElement, offsetX, offsetY);
+
+    const result: CameraManagerCallbackData = {
+      intersection,
+      modelsBoundingBox: this._updateNearAndFarPlaneBuffers.combinedBbox,
+      pickedBoundingBox: undefined
+    };
+
+    let closestDistanceToCamera = result.intersection?.distanceToCamera ?? undefined;
```
haakonflatval-cognite, line 1693: You don't need to add `?? undefined`, the `result.intersection?.distanceToCamera` will already be undefined if `result.intersection` is undefined or `result.intersection.distanceToCamera` is undefined.

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1631,6 +1682,42 @@ export class Cognite3DViewer {
...
+      );
+      if (!customObjectIntersection) {
+        return;
+      }
+      closestDistanceToCamera = customObjectIntersection.distanceToCamera;
+      result.intersection = customObjectIntersection;
+      if (pickBoundingBox) {
+        result.pickedBoundingBox = customObjectIntersection.boundingBox;
+      }
+    });
```
haakonflatval-cognite, line 1719: If I understand this correctly, we iterate through each custom object and add any applicable data from the intersections to the final result? Wouldn't it make more sense that all the data comes from just one intersection? So that we only return data from the first intersection from the ray, whether that's CAD, pointcloud or a custom object?

viewer/packages/utilities/src/customObject/CustomObjectIntersection.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+import { Vector3, Box3 } from 'three';
+import { CustomObject } from './CustomObject';
+
+/**
+ * This class is used as output intersecting custom objects
+ * @beta
+ */
+export class CustomObjectIntersection {
```
haakonflatval-cognite, line 12: I wonder if this should be just a type rather than a class :thinking: If we want to combine it in the global `Intersection` type, it should be defined similarily to how `CadIntersection` and `PointCloudIntersection` are defined, in `packages/api/src/public/migration/types.ts`. I don't see the benefit of letting this be a class

haakonflatval-cognite (commented):
Looks good, although I have a few questions about conventions and behaviour I'd like to sort out before I approve :+1:

viewer/packages/utilities/src/SceneHandler.ts
```diff
@@ -56,14 +57,25 @@ export class SceneHandler {
     remove(this._cadModels, { cadNode });
   }
 
-  public addCustomObject(object: THREE.Object3D): void {
-    this._customObjects.push(object);
-    this._scene.add(object);
+  public addCustomObject(object: THREE.Object3D | CustomObject): void {
```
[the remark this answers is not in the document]
nilscognite: Fixed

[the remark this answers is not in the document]
nilscognite: Fixed

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1052,6 +1053,33 @@ export class Cognite3DViewer {
...
+   * new THREE.SphereGeometry(),
+   * new THREE.MeshBasicMaterial()
+   * );
+   * const customObject = CustomObject(sphere);
+   * customObject.isPartOfBoundingBox = false;
+   * viewer.addCustomObject(customObject);
+   * ```
+   * @beta
+   */
+  addCustomObject(customObject: CustomObject): void {
```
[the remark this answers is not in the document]
nilscognite: Left it as it was

nilscognite (commented):
Se my comments

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1629,8 +1714,30 @@ export class Cognite3DViewer {
...
-    const intersection = await this.intersectModels(offsetX, offsetY, { asyncCADIntersection: false });
+    const modelIntersection = await this.intersectModels(offsetX, offsetY, { asyncCADIntersection: false });
 
+    // Find any custom object intersection closer to the camera than the model intersection
+    const customObjectIntersection = this.getCustomObjectIntersectionIfCloser(
+      offsetX,
+      offsetY,
+      modelIntersection?.distanceToCamera
+    );
+    if (customObjectIntersection == null || modelIntersection == null) {
```
haakonflatval-cognite, line 1725:
```suggestion
    if (customObjectIntersection === undefined || modelIntersection === null) {
```
It's a little unfortunate that the two intersection functions are inconsistent in using `undefined` and `null` as the failure-value. If you have time, would be great to make them return the same thing (preferably `undefined`)

viewer/packages/utilities/src/customObject/CustomObjectIntersectInput.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Raycaster, Vector3, Vector2, PerspectiveCamera, Plane } from 'three';
+
+/**
+ * This class is used as input for intersecting custom objects
+ * @beta
+ */
+export class CustomObjectIntersectInput {
+  readonly normalizedCoords: Vector2;
+  readonly camera: PerspectiveCamera;
+  readonly clippingPlanes: Plane[] | null;
```
haakonflatval-cognite, line 14: I'd prefer using `undefined` instead of `null` here

haakonflatval-cognite (approved):
LGTM :+1: A couple of comments, but approved

viewer/packages/utilities/src/customObject/CustomObjectIntersectInput.ts
nilscognite: Fixed

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
nilscognite: Fixed

PR #4210, opened 2024-02-19
feat(react-components): Born of Rule Based coloring and bump react-version 0.33.0

danpriori:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-3708

## Description :pencil:
Main logic and backend base for the Rule Based Coloring feature. In this case it is called Rule Based outputs, as these are outputs that in the future will not necessarily only be output for colors.
It requires two main components: the RuleBasedOutputsContainer, which includes DropDown, and the component to take all instances of the RuleSets and process them within the 3D models in the scene where they are loaded in the Viewer.
The instances are from the Data Model used as a base: [data model](https://cog-3d.dev.fusion.cogniteapp.com/3d-test/data-models/rule_based_outputs_space/Rule_Based_Outputs/1/?cluster=greenfield.cognitedata.com&env=greenfield)

In other words, to use it, it is necessary to create a Data Model similar to this one and popular with instances using the type definitions/schema defined in the types file, for now. A guide will be provided on how to define the RuleSet hierarchy to be created manually via postman or via SDK/GraphiQL.

[rule-based-coloring-react-component.webm](https://github.com/cognitedata/reveal/assets/4473965/403b2c6d-57b7-4b35-8280-5c1d6b388051)


react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
```diff
@@ -0,0 +1,100 @@
...
+        placement="top-start"
+        className="rule-based-outputs-class"
+        content={
+          <Menu
+            style={{
+              maxHeight: 300,
+              overflow: 'auto',
+              marginBottom: '20px'
+            }}>
+            <Menu.Header>Select The RuleSet</Menu.Header>
```
haakonflatval-cognite, line 74: This needs to use translation - can you ask @pramodcog What the translation key will be? Then you use it by writing `const { t } = useTranslation();` above, inside this component, and then `{t(<translation-key>)}` here

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,242 @@
...
+      (output: { type: string }) => output.type === outputType
+    ) as ColorRuleOutput;
+
+    const ruleOutputAndStyleIndex: RuleAndStyleIndex = {
+      styleIndex: new TreeIndexNodeCollection(),
+      ruleOutputParams: outputSelected
+    };
+
+    if (outputSelected === undefined) {
+      console.log('No rule output found for the type requested: ', outputType, outputs);
```
haakonflatval-cognite, line 194: We can't usually expect users to see messages in the console log, so this could be removed (unless it's important, perhaps we can throw an error?)

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const getCdfCadContextualization = async ({
+  sdk,
+  modelId,
+  revisionId,
+  nodeId,
+  assetId
+}: {
+  sdk: CogniteClient;
```
haakonflatval-cognite, line 83: We already have a way of getting this - try `useAllMappedEquipmentAssetMappings`, it is also being cached

haakonflatval-cognite (changes requested):
A preliminary review

react-components/src/utilities/FdmSDK.ts
```diff
@@ -408,6 +412,47 @@ export class FdmSDK {
     throw new Error(`Failed to fetch instances. Status: ${result.status}`);
   }
 
+  public async createInstance<PropertyType>(
```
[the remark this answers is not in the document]
danpriori: Yes

react-components/src/components/RuleBasedOutputs/utils.ts
danpriori: I think it is better to just remove it because it is more a internal log when traversing the expression types

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
danpriori: Ah ok cool.
Changed here:
https://github.com/cognitedata/reveal/pull/4210/commits/d748037cc48301eb04ef215bd184c86e4158a622

react-components/src/components/RuleBasedOutputs/utils.ts
danpriori: removed here:
https://github.com/cognitedata/reveal/pull/4210/commits/d748037cc48301eb04ef215bd184c86e4158a622

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
danpriori: added here. 
https://github.com/cognitedata/reveal/pull/4210/commits/d748037cc48301eb04ef215bd184c86e4158a622

@pramodcog will add the translation item from another PR

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,242 @@
...
+      case 'numericExpression': {
+        expressionResult = checkNumericExpressionStatement(asset, expression);
+        break;
+      }
+      case 'stringExpression': {
+        expressionResult = checkStringExpressionStatement(asset, expression);
+        break;
+      }
+      default: {
+        // statements;
```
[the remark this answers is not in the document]
danpriori: ops, removed:

https://github.com/cognitedata/reveal/pull/4210/commits/d748037cc48301eb04ef215bd184c86e4158a622

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,242 @@
...
+        expressionResult = operatorResult.find((result) => !result) ?? false;
+        break;
+      }
+      case 'not': {
+        const operatorResult = traverseExpression(
+          asset,
+          [expression.expression]
+          // expression.type
+        );
+        expressionResult = !operatorResult[0]; // TODO: make the not operator
```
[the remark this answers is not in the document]
danpriori: it is working. At least it should :D 
Just the comment wasn't removed yet.
Removed here:
https://github.com/cognitedata/reveal/pull/4210/commits/0054b30448dab1198dd1110b1167a8bc89a722c9

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,242 @@
...
+  const expressionResults: boolean[] = [];
+
+  expressions.forEach((expression) => {
+    switch (expression.type) {
+      case 'or': {
+        const operatorResult = traverseExpression(
+          asset,
+          expression.expressions
+          // expression.type
+        );
```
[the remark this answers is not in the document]
danpriori: removed here:
https://github.com/cognitedata/reveal/pull/4210/commits/d748037cc48301eb04ef215bd184c86e4158a622

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,242 @@
...
+    case 'endsWith': {
+      expressionResult = asset[trigger.type]?.[trigger.key].endsWith(condition.parameter) ?? false;
+      break;
+    }
+  }
+
+  return expressionResult;
+};
+const checkNumericExpressionStatement = (asset: Asset, expression: NumericExpression): boolean => {
+  // for now it will be only for metadata
```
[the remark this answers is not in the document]
danpriori: removed here:
https://github.com/cognitedata/reveal/pull/4210/commits/d748037cc48301eb04ef215bd184c86e4158a622

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+        placement="top-start"
+        className="rule-based-outputs-class"
+        content={
+          <Menu
+            style={{
+              maxHeight: 300,
+              overflow: 'auto',
+              marginBottom: '20px'
+            }}>
+            <Menu.Header>{t('RULESET_SELECT_HEADER', 'Select The RuleSet')}</Menu.Header>
```
haakonflatval-cognite, line 77: I talked to Pramod and Per-Øyvind about terminology, I think we want this to be `Select color overlay`. Pramod will update the localization-files

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const sdk = useSDK();
+  const fdmSdk = useMemo(() => new FdmSDK(sdk), [sdk]);
+
+  const { t } = useTranslation();
+
+  useEffect(() => {
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+    const getAllInstances = async () => {
+      const versionedPropertiesKey = `${RULE_BASED_OUTPUTS_VIEW.externalId}/${RULE_BASED_OUTPUTS_VIEW.version}`;
+
```
haakonflatval-cognite, line 27: Would be nice to separate long `useEffect`s like this one into a separate function/hook

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+      const filter = {
+        in: {
+          property: [RULE_BASED_OUTPUTS_VIEW.space, versionedPropertiesKey, 'shamefulOutputTypes'],
```
haakonflatval-cognite, line 30: We better remind each other to fix this data model and field name at some point :sweat_smile:

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+export const RuleBasedOutputsContainer = (): ReactElement => {
+  const [ruleInstances, setRuleInstances] = useState<RuleAndEnabled[]>();
+  const [currentRuleSetEnabled, setCurrentRuleSetEnabled] = useState<any>();
+  const sdk = useSDK();
+  const fdmSdk = useMemo(() => new FdmSDK(sdk), [sdk]);
+
+  const { t } = useTranslation();
+
+  useEffect(() => {
```
haakonflatval-cognite, line 23: This should be a `useQuery` the only difference will be that instead of the `ruleInstances ... = useState...`, we can write `const ruleInstancesResult = useQuery(['react-components', 'color-overlay-rules', 'all'], <async-function>);` where the `<async-function>` is the same as the one you wrote here, but returns the value instead of using the `setRuleInstances`. Then, the result will be put in `ruleInstancesResult.data` and triggering a React re-render when it's ready. Let me know if you want help with this

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    if (selectedRule === undefined) return;
+    selectedRule.isEnabled = data.target.checked;
+  };
+
+  return (
+    <>
+      <Dropdown
+        placement="top-start"
+        className="rule-based-outputs-class"
```
haakonflatval-cognite, line 69: I don't see any other references to this class.. Do we need this?

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+    setRuleInstances(ruleInstances);
+    setCurrentRuleSetEnabled(selectedRule);
+
+    if (selectedRule === undefined) return;
+    selectedRule.isEnabled = data.target.checked;
+  };
+
+  return (
+    <>
+      <Dropdown
```
haakonflatval-cognite, line 67: Does this dropdown have text in it? Or is it just an icon? Since we for now will only be targetting the vertical toolbar, we will need a version with only the icon

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export type ColorOverlayProps = {
+  ruleSet: RuleOutputSet | undefined;
+};
+
+export function RuleBasedOutputs({ ruleSet }: ColorOverlayProps): ReactElement | undefined {
+  const viewer = useReveal();
+
+  // only enabled whether has loaded models
+  if (viewer.models === undefined || viewer.models.length === 0) return;
```
haakonflatval-cognite, line 20: We cannot have an early return when there are React hooks later in the component. The hooks will just have to tackle the case when `models` are undefined on their own

haakonflatval-cognite (changes requested):
A few more comments, I'll read and test a little more too

react-components/src/components/RevealTopbar/RevealTopbar.tsx
```diff
@@ -24,6 +25,7 @@ const DefaultContentWrapper = (props: CustomTopbarContent): ReactElement => {
       <LayersButton />
       <SlicerButton />
       <FitModelsButton />
+      <RuleBasedOutputsContainer />
```
haakonflatval-cognite, line 28: We should expose this - further down in this file we expose a few other buttons, we can expose this likewise

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { useAllMappedEquipmentAssetMappings, useReveal } from '../..';
+import { Color } from 'three';
+import { type RuleOutputSet } from './types';
+import { generateRuleBasedOutputs } from './utils';
+
+export type ColorOverlayProps = {
+  ruleSet: RuleOutputSet | undefined;
+};
+
+export function RuleBasedOutputs({ ruleSet }: ColorOverlayProps): ReactElement | undefined {
```
haakonflatval-cognite, line 16: Can we rename this to e.g. `RuleBasedOutputsSelector`?

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { RuleBasedOutputs } from '../RuleBasedOutputs/RuleBasedOutputs';
+import { RULE_BASED_OUTPUTS_VIEW } from '../RuleBasedOutputs/constants';
+import { FdmSDK } from '../../utilities/FdmSDK';
+import { useSDK } from '../RevealCanvas/SDKProvider';
+import { type RuleAndEnabled } from '../RuleBasedOutputs/types';
+import { useTranslation } from '../i18n/I18n';
+
+export const RuleBasedOutputsContainer = (): ReactElement => {
+  const [ruleInstances, setRuleInstances] = useState<RuleAndEnabled[]>();
+  const [currentRuleSetEnabled, setCurrentRuleSetEnabled] = useState<any>();
```
haakonflatval-cognite, line 17: Can we have better typing for this?

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
```diff
@@ -0,0 +1,103 @@
...
+        });
+      });
+
+      setRuleInstances(rulesAndEnabled);
+    };
+
+    void getAllInstances();
+  }, []);
+
+  const onChange = (data: { target: { id: any; checked: boolean } }): void => {
```
haakonflatval-cognite, line 50: Do we have a better option than using `any` here?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
```diff
@@ -0,0 +1,75 @@
...
+          .flat()
+          .map((item) => item.mappings)
+          .flat() ?? [];
+      const flatMappings = flatAssetsMappingsList.map((node) => node.items).flat();
+      const contextualizedAssetNodes =
+        assetMappings?.pages
+          .flat()
+          .map((item) => item.assets)
+          .flat() ?? [];
+
```
haakonflatval-cognite, line 64: In  `assetMappings`, each of the pages contain one asset-mapping list per model, which means that only one of the mapping-lists per page is needed to consider here

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,218 @@
...
+    const expression = rule.expression;
+
+    const outputSelected = outputs.find(
+      (output: { type: string }) => output.type === outputType
+    ) as ColorRuleOutput;
+
+    if (outputSelected === undefined) return;
+
+    const ruleOutputAndStyleIndex: RuleAndStyleIndex = {
+      styleIndex: new TreeIndexNodeCollection(),
```
haakonflatval-cognite, line 171: I think a better way to construct this node collection, is to use the `AssetNodeCollection` that is provided by `Reveal`. You send it a bunch of asset IDs and it fetches the relevant data.

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
danpriori: But should we also be prepared for multi models? To flat all in a one dimensional list to facilitate the logic?

react-components/src/components/RuleBasedOutputs/utils.ts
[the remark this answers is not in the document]
haakonflatval-cognite: Ah, okay :sweat_smile: Nice!

react-components/src/components/RuleBasedOutputs/types.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+  | {
+      type: 'within';
+      lowerBoundInclusive: number;
+      upperBoundInclusive: number;
+    }
+  | {
+      type: 'outside';
+      lowerBoundExclusive: number;
+      upperBoundExclusive: number;
```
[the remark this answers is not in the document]
danpriori: Yeah, Me too. But I think we have to keep it for now.

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  let expressionResult: boolean = false;
+
+  switch (condition.type) {
+    case 'equals': {
+      expressionResult = asset[trigger.type]?.[trigger.key] === condition.parameter;
+      break;
+    }
+    case 'notEquals': {
+      expressionResult = asset[trigger.type]?.[trigger.key] !== condition.parameter;
```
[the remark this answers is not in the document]
danpriori: The decision was more merely aesthetic, to simplify reading and maintain the pattern as was done in the traverseExpression function where it is necessary to populate an Array of expressions. But if you find it necessary not to use this for some logical reason, we'll change it, as I don't have a strong opinion on this :)

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,218 @@
...
+      break;
+    }
+    case 'contains': {
+      expressionResult = asset[trigger.type]?.[trigger.key].includes(condition.parameter) ?? false;
```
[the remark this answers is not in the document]
danpriori: Hmm yeah good point. Maybe we can send out a undefined (instead of false) and if the not-clause receives that then it keeps false. But I think we can test it more and see how it goes, or do you prefer that I change it now and see if I have room for testing before merging? I believe we can do it during the next comings more carefully and keep it like this for now, what do you think?

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,218 @@
...
+    case 'endsWith': {
+      expressionResult = asset[trigger.type]?.[trigger.key].endsWith(condition.parameter) ?? false;
+      break;
+    }
+  }
+
+  return expressionResult;
+};
+const checkNumericExpressionStatement = (asset: Asset, expression: NumericExpression): boolean => {
+  const trigger = expression.trigger as MetadataRuleTrigger;
```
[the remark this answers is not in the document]
danpriori: Hmmm... I can add a 'user-defined' type guard but what is going to happen is if it is not a metadata, it has to return undefined, I think. Do you have any suggestion on what should be the best way to keep it safer?
I can add a

```typescript
const isMetadataTrigger = (
  trigger: MetadataRuleTrigger | TimeseriesRuleTrigger
): trigger is MetadataRuleTrigger => {
  return (trigger as MetadataRuleTrigger).type !== 'metadata';
};

``` 
But I'm not sure if it is a good type guard TBH

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
danpriori: ops.. yeah.
fixed: https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
danpriori: already changed and pushed here:
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
danpriori: yep. Done
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RevealTopbar/RevealTopbar.tsx
danpriori: Ah ok.. Done here:
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,218 @@
...
+    ) as ColorRuleOutput;
+
+    if (outputSelected === undefined) return;
+
+    const ruleOutputAndStyleIndex: RuleAndStyleIndex = {
+      styleIndex: new TreeIndexNodeCollection(),
+      ruleOutputParams: outputSelected
+    };
+
+    // ======== OUTPUT - COLOR IN 3D
```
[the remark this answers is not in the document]
danpriori: Why you dont like my comments? 😢it is so perfect! Aahahhhahaahha kidding.
Done:
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,218 @@
...
+import { type AssetMapping3D, type Asset } from '@cognite/sdk';
+
+const checkStringExpressionStatement = (asset: Asset, expression: StringExpression): boolean => {
+  const { trigger, condition } = expression;
+
+  let expressionResult: boolean = false;
+
+  switch (condition.type) {
+    case 'equals': {
+      expressionResult = asset[trigger.type]?.[trigger.key] === condition.parameter;
```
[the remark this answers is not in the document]
danpriori: oh yeah :) 
Forgot that.
Done
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RuleBasedOutputs/utils.ts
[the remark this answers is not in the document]
danpriori: Did some tweaks based on my thoughts in here:
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RuleBasedOutputs/types.ts
```diff
@@ -0,0 +1,216 @@
...
+  version: number;
+  space: string;
+  externalId: string;
+  createdTime: number;
+  lastUpdatedTime: number;
+  deletedTime: number;
+  properties: FdmPropertyType<PropertyType>;
+};
+
+export type FdmPropertyType<NodeType> = Record<string, Record<string, NodeType>>;
```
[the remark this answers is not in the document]
danpriori: oh yeah. 
Using the reveal type here:
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
danpriori: fixed here

https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
danpriori: Updated with just the icon here:
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
danpriori: it was for debugging. Removed:
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
danpriori: Refactored here:
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
danpriori: yeah :D

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
danpriori: done here:
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
danpriori: Ok but should I push the english translation or Pramod will handle that as well?

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,218 @@
...
+            color: new Color(outputSelected.fill)
+          });
+        }
+        // debug
+        // console.log(' ASSET ', assetNode, finalGlobalOutputResult);
+      })
+    );
+
+    // =================================
+  });
```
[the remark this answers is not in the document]
danpriori: removed here
https://github.com/cognitedata/reveal/pull/4210/commits/e4c1637a2c6f4412f8fa79861d3d37844a646860

react-components/src/components/RuleBasedOutputs/utils.ts
danpriori: Right. But I believe it would be safer to leave it the way it is because from what I quickly saw of the AssetNodeCollection, I may have to refactor more parts throughout the code and I'm afraid of ending up breaking more things right before the release.
I think it's safer to make this specific change more calmly over the next few days and after this release. Is it okay if we do like this?

react-components/src/components/RevealTopbar/RuleBasedOutputsContainer.tsx
haakonflatval-cognite: Pramod will handle all that :+1:

react-components/src/components/RuleBasedOutputs/utils.ts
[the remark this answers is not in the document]
haakonflatval-cognite: Just worried that we'll have to change many places if we want to use another method of extracting the relevant key :sweat_smile:

react-components/src/components/RuleBasedOutputs/utils.ts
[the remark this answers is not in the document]
haakonflatval-cognite: Looks good! :+1:

react-components/src/components/RuleBasedOutputs/utils.ts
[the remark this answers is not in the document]
haakonflatval-cognite: I think the function body can be just `return trigger.type === 'metadata'`

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
haakonflatval-cognite: Isn't this happening within a loop where each iteration considers one model? So I thought we could just filter out the asset mappings relevant for the current iteration

react-components/src/components/RuleBasedOutputs/utils.ts
haakonflatval-cognite: Yes, we can do this later :+1:

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -0,0 +1,266 @@
...
+    switch (expression.type) {
+      case 'or': {
+        const operatorResult = traverseExpression(asset, expression.expressions);
+        expressionResult = operatorResult.find((result) => result) ?? false;
+        break;
+      }
+      case 'and': {
+        const operatorResult = traverseExpression(asset, expression.expressions);
+        expressionResult =
+          operatorResult.find((result) => (result !== undefined ? !result : false)) ?? false;
```
haakonflatval-cognite, line 139: Can this be just a `operatorResult.every((result) => result === true)` ?

danpriori: changed here
https://github.com/cognitedata/reveal/pull/4210/commits/03b40b181de31e95d5b302162f9a3d5d0b1ffb51

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
danpriori: If I understood your suggestion correctly (only using the first page/per model), done here:
https://github.com/cognitedata/reveal/pull/4210/commits/03b40b181de31e95d5b302162f9a3d5d0b1ffb51

react-components/src/components/RuleBasedOutputs/utils.ts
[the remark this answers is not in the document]
danpriori: changed here:
https://github.com/cognitedata/reveal/pull/4210/commits/03b40b181de31e95d5b302162f9a3d5d0b1ffb51

react-components/src/components/RuleBasedOutputs/RuleBasedOutputs.tsx
danpriori: reverted here: 
https://github.com/cognitedata/reveal/pull/4210/commits/c51354fb0dca063f92301efcff39f86fa9d9560b

PR #4231, opened 2024-02-21
fix: scene console log error

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:

There was a bug where the scene could temporarily be undefined, which in the Search code would manifest itself as an error about not finding the scene `''`. This improves the typing, so that we don't need to send in an empty string in cases where the scene space and external ID is unknown.

Includes a few more cleanups too.

### DIFF
--- a/react-components/src/hooks/useReveal3dResourcesFromScene.ts
+++ b/react-components/src/hooks/useReveal3dResourcesFromScene.ts
@@ -27,7 +27,7 @@ export const useReveal3dResourcesFromScene = (
   const [resourceOptions, setResourceOptions] = useState<AddResourceOptions[]>([]);
 
   useEffect(() => {
-    if (scene.data === undefined) {
+    if (scene.data === undefined || scene.data === null) {
       return;
     }
     const addResourceOptions: AddResourceOptions[] = [];
--- a/react-components/src/utilities/fdmViewsExist.ts
+++ b/react-components/src/utilities/fdmViewsExist.ts
@@ -1,4 +1,7 @@
-import { FdmSDK, Source } from './FdmSDK';
+/*!
+ * Copyright 2024 Cognite AS
+ */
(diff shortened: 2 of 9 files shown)


anders-hopland (approved):
LGTM

react-components/src/hooks/useReveal3dResourcesFromScene.ts
```diff
@@ -27,7 +27,7 @@ export const useReveal3dResourcesFromScene = (
   const [resourceOptions, setResourceOptions] = useState<AddResourceOptions[]>([]);
 
   useEffect(() => {
-    if (scene.data === undefined) {
+    if (scene.data === undefined || scene.data === null) {
```
danpriori, line 30: I also prefer to use the explicitly for null and undefined instead of using the equality operator ==.
Good!

haakonflatval-cognite: Yeees :vulcan_salute:

react-components/src/utilities/fdmViewsExist.ts
```diff
@@ -0,0 +1,12 @@
...
+ */
+import { type FdmSDK, type Source } from './FdmSDK';
+
+export async function fdmViewsExist(fdmSdk: FdmSDK, neededViews: Source[]): Promise<boolean> {
+  const views = await fdmSdk.getViewsByIds(neededViews);
+  if (views.items.length === neededViews.length) {
+    return true;
+  }
+  return false;
+}
```
danpriori, line 12: I know that it is not part of this refactoring and fixes but I think we should consider to not just returning boolean if some of the views doesnt exist by calling this method. We should also inform which one doesnt exist if the request is for multiple views.

PR #4261, opened 2024-02-26
fix(react-components): missing styling type for 360 annotation

pramodcog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:
 Exported missing type for 360 annotation styling


react-components/src/components/NodeCacheProvider/Image360AnnotationCache.ts
```diff
@@ -69,10 +70,30 @@ export class Image360AnnotationCache {
...
+                assetId ===
+                getAssetIdOrExternalIdFromImage360Annotation(revisionAnnotation.annotation)
+              ) {
+                const center = revisionAnnotation.getCenter().applyMatrix4(tranform);
+                if (center !== undefined) {
+                  centerPosition.push(center);
+                }
+              }
+            });
+          });
```
haakonflatval-cognite, line 88: This is not safe - we should not just assume that this async call will be done executing before the rest of this function is run

react-components/src/components/NodeCacheProvider/types.ts
```diff
@@ -59,6 +60,7 @@ export type Image360AnnotationModel = AnnotationModel & {
 export type Image360AnnotationAssetInfo = {
   asset: Asset;
   assetAnnotationImage360Info: AssetAnnotationImage360Info;
+  positions: Vector3[];
 };
```
haakonflatval-cognite, line 64: I wonder if we should just return the `Image360AnnotationObject` instead of the annotation info and the centers

haakonflatval-cognite (changes requested):
A couple of comments

react-components/src/components/NodeCacheProvider/types.ts
haakonflatval-cognite: Actually - we will need the `AssetAnnotationImage360Info` :thinking: Perhaps this is okay - but it feels like we might as well give the user the option to use `Image360Annotation` too

react-components/src/components/NodeCacheProvider/Image360AnnotationCache.ts
```diff
@@ -60,25 +66,59 @@ export class Image360AnnotationCache {
...
+        const centerPosition = new Vector3();
+
+        for (const revisionAnnotation of revisionAnnotations) {
+          if (
+            assetId !== getAssetIdOrExternalIdFromImage360Annotation(revisionAnnotation.annotation)
+          ) {
+            continue;
+          }
+          const center = revisionAnnotation.getCenter().applyMatrix4(transform);
+          console.log('assetId, center', assetId, center);
```
haakonflatval-cognite, line 104: :raised_eyebrow:

react-components/src/components/NodeCacheProvider/Image360AnnotationCache.ts
```diff
@@ -60,25 +66,59 @@ export class Image360AnnotationCache {
...
+          if (center !== undefined) {
+            centerPosition.add(center);
```
haakonflatval-cognite, line 106: Do you mean to use `add` here? Don't we only want to return the `centerPosition` of the `revisionAnnotation` that has the same ID as `annotationInfo.annotation`?

react-components/src/components/NodeCacheProvider/Image360AnnotationCache.ts
pramodcog: 🙅

haakonflatval-cognite (approved):

PR #4267, opened 2024-02-27
Swap 1 and 2 keyboard key and made the marker slightly smaller [BND3D-3866]

nilscognite:
Also move marker constants to the option class.

https://cognitedata.atlassian.net/browse/BND3D-3866


pramodcog (approved):
LGTM!

PR #4268, opened 2024-02-28
fix: solve problem with too large boundingbox when having ground plane

nilscognite:


viewer/packages/camera-manager/src/Flexible/FlexibleCameraManager.ts
```diff
@@ -191,16 +191,10 @@ export class FlexibleCameraManager implements IFlexibleCameraManager {
 
   public update(deltaTime: number, boundingBox: Box3): void {
```
haakonflatval-cognite, line 192: Can we rename the second parameter `nearFarBoundingBox`? I feel it makes this a little more clear

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1569,10 +1570,12 @@ export class Cognite3DViewer {
       const camera = this.cameraManager.getCamera();
       TWEEN.update(time);
       this.recalculateBoundingBox();
-      this._activeCameraManager.update(
-        this.cameraManagerClock.getDelta(),
-        this._updateNearAndFarPlaneBuffers.combinedBbox
-      );
+
+      if (this._activeCameraManager instanceof FlexibleCameraManager) {
+        this._activeCameraManager.updateInnerBoundingBox(this._boundingBoxes.innerBoundingbox);
```
haakonflatval-cognite, line 1575: I don't like that we depend on `FlexibleCameraManager` like this here. But I suppose you don't either :sweat_smile:

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -196,9 +196,10 @@ export class Cognite3DViewer {
   /**
    * Reusable buffers used by functions in Cognite3dViewer to avoid allocations.
    */
-  private readonly _updateNearAndFarPlaneBuffers = {
-    combinedBbox: new THREE.Box3(),
-    bbox: new THREE.Box3()
+  private readonly _boundingBoxes = {
+    nearFarPlaneBoundingbox: new THREE.Box3(),
+    innerBoundingbox: new THREE.Box3(),
```
haakonflatval-cognite, line 201: I'm not sure `innerBoundingbox` is a clear name, but not sure about what it should be either :thinking: Could it be `baseBoundingBox` or `modelBoundingBox`?

Also, we usually write `Box` in `boundingBox` with a capital `B`

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1569,10 +1570,12 @@ export class Cognite3DViewer {
       const camera = this.cameraManager.getCamera();
       TWEEN.update(time);
       this.recalculateBoundingBox();
-      this._activeCameraManager.update(
-        this.cameraManagerClock.getDelta(),
-        this._updateNearAndFarPlaneBuffers.combinedBbox
-      );
+
+      if (this._activeCameraManager instanceof FlexibleCameraManager) {
```
haakonflatval-cognite, line 1574: I don't think this would work - `this._activeCameraManager` is of type `ProxyCameraManager`. We need to check `this._activeCameraManager.innerCameraManager`.

viewer/packages/camera-manager/src/Flexible/IFlexibleCameraManager.ts
```diff
@@ -49,4 +49,10 @@ export interface IFlexibleCameraManager extends CameraManager {
    * @param callback - The callback function to be removed from the controls type change listeners.
    */
   removeControlsTypeChangeListener(callback: FlexibleControlsTypeChangeDelegate): void;
+
+  /**
+   * Updates the inner bounding box of the camera manager used for camera sensibility
+   * @param innerBoundingBox - The new inner bounding box.
+   */
+  updateInnerBoundingBox(innerBoundingBox: Box3): void;
```
haakonflatval-cognite, line 57: This does not need to be exposed on the interface. It's only intended for internal use

viewer/packages/camera-manager/src/Flexible/FlexibleCameraManager.ts
```diff
@@ -265,6 +259,16 @@ export class FlexibleCameraManager implements IFlexibleCameraManager {
     this._triggers.controlsTypeChange.subscribe(callback);
   }
 
+  public updateInnerBoundingBox(innerBoundingBox: Box3): void {
+    // If the camera haven't set the position and target before, do it now
+    if (!this.controls.isInitialized) {
```
haakonflatval-cognite, line 264: Does this mean we are dependent on the `updateInnerBoundingBox` to be called for the camera manager to be initialized correctly? Not a huge problem, since it's for internal use

haakonflatval-cognite (changes requested):
A few concrete suggestions for improvement.

But otherwise - this feels architecturally like a temporary solution (i.e. it only works specifically for `FlexibleCameraManager`). Is there any longer-term solution to this problem?

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
nilscognite: Done

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
nilscognite: Fixed

viewer/packages/camera-manager/src/Flexible/IFlexibleCameraManager.ts
nilscognite: Fixed

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
nilscognite: I think this is the best we can no with the current architecture

viewer/packages/camera-manager/src/Flexible/FlexibleCameraManager.ts
nilscognite: Fixed

haakonflatval-cognite (approved):

PR #4269, opened 2024-02-28
fix: problems with clicking in Measurement tool enables viewer click

nilscognite:
Added a isEnableClickAndDoubleClick in the FlexibleCameraManager and used it in the Measurement  tool


pramodcog (approved):
LGTM!

PR #4273, opened 2024-02-28
fix: Force to  render the axis gizmo when connecting to the viewer at the first time

nilscognite:


haakonflatval-cognite (approved):
LGTM :+1:

PR #4281, opened 2024-03-01
fix: Some adjustment in the AxisGizmo

nilscognite:
Some adjustments in the Axis Gismo

* **Fix bug in mouse move.** The mouse down in the previous version stops the propagation. This was a bug, since the camera was not able to stop the dragging when the mouse was inside the gizmo.
* Swap color on y and z axis so it is equal blender.
* Use alpha instead of blending the color with black (also equal blender) . The intended behavior with the axis fading away will then also when the background is bright. Then I could remove the mixing of colors in the code. Only a visual effect
* Make the focus circle somewhat darker


haakonflatval-cognite (approved):
LGTM :+1:

PR #4304, opened 2024-03-05
(refactor) Add customObject to the CustomObjectIntersection and export some function

nilscognite:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

## Description :pencil:
In order to get proper info of the intersection, a userData field is added. Also updated the type to be a string. I misunderstood the syntax.

I discovered this needs when updating the contextualize editor. Click on AnnotationsViedw, I need to know which annotation is clicked on. All are in the same object3D.


pramodcog (approved):
Looks good to me!

PR #4328, opened 2024-03-13
fix: Move getWheelDelta to utils and add a maximum wheel delta to 120

nilscognite:
#### Type of change
Bug 

#### Jira ticket :blue_book:

No Jira ticket. This was a tip from Henrik, that we should have a max value of the scroll wheel, since the max accelerates the wheel when used multiple times.

A also move the function out and reused it


pramodcog (approved):
LGTM

PR #4329, opened 2024-03-14
feat(react-components): expose utility functions for rule-based coloring and some internal refactoring

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

## Description :pencil:

Expose some utility functions needed for the upcoming Rule-based coloring UI in 3d-management

Also: 
- Replace `filterUndefined()` with a combination of `filter` and the new function `isDefined`, to conform better to our code convention.
- Introduce `FdmNode`, which is a `NodeItem`, but with the properties "hoisted" (accessible directly, rather than through maps with view/version). This has in reality been the type of the result from `filterInstances` all this time, we have just mislabeled it.


react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -164,41 +167,59 @@ const traverseExpression = (
...
+        return await analyzeNodesAgainstExpression({
+          model,
+          contextualizedAssetNodes,
+          assetMappings,
+          expression,
+          outputSelected
+        });
+      })
+    )
+  ).filter(isDefined);
```
danpriori, line 222: Hmm I like this approach. It is cleaner return. Nice!

danpriori (approved):
LGTM

react-components/src/hooks/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -87,18 +87,18 @@ async function get360AnnotationAssets(
...
+      const assetId = getAssetIdOrExternalIdFromImage360Annotation(annotation);
+      if (assetId === undefined) {
+        return undefined;
+      }
+      return {
+        assetId,
+        annotationId: annotation.id
+      };
+    })
+    .filter(isDefined);
```
danpriori, line 102: The PR seems like more related to exposing things for rule based coloring but also has some refactoring related to the 360s... Is it intentional?
Maybe it is better to rephrase the PR title to also mention these other changes perhaps?

haakonflatval-cognite: Good point - changed the title. (Yes, it is intentional. It is a little messy, but I want the bar for doing improvements in the code to be low, so that we don't require separate PR's for every change)

PR #4350, opened 2024-03-25
feat: add setModelTransformation capability to 360 image collection

christjt:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4011

## Description :pencil:

Adds setModelTransformation feature to 360 image collections. 

Visual regression is updated to include a transform.

- Test that setting a transformation of a 360 image collection correctly transforms the icons.
- LOD system should still work as intended
- Visualization box should be translated and rotated with the icon
- Annotations should follow the transformation, and should still be clickable

Example data set on 3d-test which has an annotation: 
- externalId: c_RC_2
- space: anders_360_image


viewer/packages/360-images/src/Image360Facade.ts
```diff
@@ -115,19 +115,29 @@ export class Image360Facade<T> {
...
+          )
+        )
+        .filter(hasIntersection)
+        .map(intersectionToCameraSpace)
+        .filter(isInFrontOfCamera)
+        .sort(byDistanceToCamera)
+        .map(selectEntity)
+    );
 
     return first(intersections);
```
danpriori, line 142: I know that is not part of this PR changes but I was wondering why using the lodash first() as usually just getting the first element from an array is way faster than using the lodash version. Unless I'm missing some context or some info here :D
Just for the sake of cleanness?

danpriori (approved):
LGTM.

I liked the implementation. Despite there being a certain amount of complexity involved in the affected parts, I think the implementation was very clean and direct, without any fuss. It made sense to me.

The only comment would be to change the use of global calls from THREE.js to its resources directly, as has already been done in some places.
In other words, I think we can make more changes in the future, changing THREE.Color, THREE.Vector 2 to Color, Vector2, etc.
Unless we have a need/pattern to keep the "THREE" walking around our code for some reason.

But we can do it in another time, it is not urgent for sure.

Approved

viewer/packages/360-images/src/Image360Facade.ts
christjt: It bounds checks under the hood, so it will be a little slower. But I have to bounds check anyway in the case where len is 0.

PR #4351, opened 2024-03-25
feat: add 360 image collection transform support to react components

christjt:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4023

## Description :pencil:

Add 360 image collection transform support to react components.

Added support in both 360 image collection component story and 3d resources components story.


react-components/stories/Image360CollectionContainer.stories.tsx
```diff
@@ -16,17 +22,47 @@ const meta = {
...
+const Image360CollectionContainerStoryContent = ({
+  collectionId,
+  transform,
+  styling
+}: CadModelContainerStoryContentProps): ReactElement => {
+  const modelsLoadedRef = useRef(0);
+  const cameraNavigationActions = useCameraNavigation();
+  const onLoad = (): void => {
+    modelsLoadedRef.current++;
+    if (modelsLoadedRef.current === 1) {
```
danpriori, line 55: Hmmm interesting. So looks like the onLoad is triggered twice or I read it wrong?
That is why it has this counter?

Maybe the add360Collection (inside the useEffect) and therefore the getOrAdd360Collection are being called twice or doesnt make no sense what I'm saying? :D :D

danpriori (approved):
in general LGTM. One comment only that probably needs some further investigation IMO if what I've commented does make sense.

react-components/stories/Image360CollectionContainer.stories.tsx
christjt: I can remove the condition here. In this case it will only be called once. If you have multiple models, then we need to make sure all are loaded before fitting the camera, which is the reason why the check is here. For one model though it doesn't really make sense. I'll remove it.

PR #4362, opened 2024-04-04
Smaller fixes on the flexible camera control

nilscognite:
#### Type of change

Bug 

## Description :pencil:
@greenbech found a few issues and suggestions:

* Bug: When camera state uses a rotation, the navigation do not work properly afterwards. This was caused by a old hack in the code, and is now removed. The hack is not longer necessary because the old AxisTool is not in used anymore. 
* Increase speed rotation x 2 with keyboard
* Increase speed rotation x 4 with keyboard + shift
* Do not need to automatically switch to first person mode when using the keyboard arrows only.


danpriori (approved):
LGTM

PR #4384, opened 2024-04-16
fix: Don't combine custom styles with DefaultAppearance definition

hugolafis:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:
This PR fixes an issue where cumulative styles would not combine as intended, as the style passed through would get merged with the `DefaultAppearance` object, overwriting unset values with the defaults provided by that object. 

This meant that styles could override previously applied styles unintentionally, e.g.:

- Style set 1 is applied, with the values: `renderGhosted: true`
- Style set 2 is applied, with the values: `color: new THREE.Color(0xFF0000)`

All nodes in the above example would render without ghosting, as the value of `renderGhosted` in the second style set would be passed through with the value of false, as a result of it being merged with the `DefaultAppearance` object.

I don't believe this change should cause any issues as the `setDefaultAppearance` method is called in the constructor of this class with the `DefaultAppearance` object, so these buffers should be populated with defaults in any subsequent styling calls, but please let me know if I'm missing something here.

I've tested this manually in the example viewer by creating two style sets, and visually checking the outcome of the combined styles. This was done with the default primitives example model.

Inside the example `Viewer` class, add the styling code from below:
```
        viewer.loadCameraFromModel(model); // line 217
        if (model instanceof CogniteCadModel) {

        ...

        // Styling testing
          const ghostedSet = new reveal.TreeIndexNodeCollection(
            new reveal.IndexSet(new reveal.NumericRange(0, model.nodeCount))
          );
          const ghostedAppearance: reveal.NodeAppearance = {
            renderGhosted: true
          };

          const customStyleSet = new reveal.TreeIndexNodeCollection(
            new reveal.IndexSet(new reveal.NumericRange(0, model.nodeCount))
          );
          const customAppearance: reveal.NodeAppearance = {
            color: new THREE.Color(0xff0000)
          };

          model.assignStyledNodeCollection(ghostedSet, ghostedAppearance);
          model.assignStyledNodeCollection(customStyleSet, customAppearance);
```

Without the fix, the nodes will not appear styled with ghosting even though the second style set doesn't set whether the node should be rendered with ghosting or not, whereas with the fix the nodes will appear ghosted.


christjt (approved):
LGTM! A little surprised this hasn't been caught before now. I am also worried that people have worked around this issue, so underlying semantics will change. But, in any case, I agree that the current behaviour is unintended. 

Thanks for contributing the fix!

PR #4403, opened 2024-04-24
(feat) Architecture on top of Reveal

nilscognite:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

## Description :pencil:

Overview of the proposed architecture her:
https://docs.google.com/presentation/d/1Y50PeqoCS5BdWyDqRNNazeISy1SYTcx4QDcjeMKIm78/edit?usp=sharing

There will also be documentation in the Readme.md file: ` src\architecture\base\README.md`  (WORK ON THIS NOW)

**Link here:**  https://github.com/cognitedata/reveal/blob/np/actitecture-in-reactcomponent/react-components/src/architecture/base/README.md

```
yarn
yarn build
yarn storybook
```

This will goto: https://localhost:3000/?modelUrl=primitives
Open Architecture/Main
**The toolbar to the right hand side contains all new functionality**


react-components/src/architecture/base/commands/BaseCommand.ts
```diff
@@ -0,0 +1,100 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+/* eslint-disable @typescript-eslint/class-literal-property-style */
```
greenbech, line 4: unnecessary — don't notice any difference when I remove this

nilscognite: No, it gives lint errors in all virtual functions that return a constant.

haakonflatval-cognite: If we need this rule disabled, we can disable it in the `eslintrc.cjs`

react-components/src/architecture/base/commands/BaseCommand.ts
```diff
@@ -0,0 +1,100 @@
...
+    return undefined;
+  }
+
+  public get tooltip(): Tooltip {
+    return { key: '', fallback: '' };
+  }
+
+  public get icon(): string {
+    return '';
+  }
```
haakonflatval-cognite, line 40: I see that some of these properties are very tied to the command being used as a button, but if I understand correctly, many commands will probably not be tied to a button? Would it make sense to make "button-commands" a separate (sub-)class?

react-components/src/architecture/base/commands/BaseCommand.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  // ==================================================
+  // VIRTUAL METHODS (To be override)
+  // =================================================
+
+  public get name(): string {
+    return this.tooltip.fallback;
+  }
+
+  public get shortCutKey(): string | undefined {
```
haakonflatval-cognite, line 30: I'm skeptical to let the tools define themselves whether they have a shortcut key. I think the application should be in absolute control of this, and not be riddled with predefined short-cuts from react-components as a library.

Or am I misunderstanding how this works?

react-components/src/architecture/base/README.md
```diff
@@ -0,0 +1,186 @@
...
+  - :
+  - Overrides from the actually base class (if any)
+  - Instance methods (If many, break it up based on what they do)
+  - Static methods (if any)
+  - Local functions, types or classes used in this file only
+
+- **Virual methods:**: Since TypeScript doesn't allow the virtual keyword, all functions are virtual. This is a weekness for the application developer. I have tried to mark function as virtual by comments. You should not override other function than this. A virtual function normally brings in uncessesary complexibility and should general be used by care.
+
+  - Function with prefex `Core` are always virtual and protected. They can be overridden, but must call the same function for the base class. A function without the `Core` prefex is the one you should call from outside and is public. Experince shows that this pattern will make the code easier to maintain and extend in the future. For instance `remove/removeCore` and `initialize/initializeCore` in the `DomainObject`.
+  - Overridden methods should alway be marked by the override keyword. You will unfortunate not get any linting error when you forget this.
```
haakonflatval-cognite, line 41: We can fix this by adding the `noImplicitOverride` option to the tsconfig file: https://www.typescriptlang.org/tsconfig/#noImplicitOverride

react-components/src/architecture/base/README.md
```diff
@@ -0,0 +1,186 @@
...
+  -
+
+- **Static methods and classes:** The linker don't like static method, epecially when the entire class have static method only. I wanted to use this more, but have used functions instead. I don't like this, because it doen't group similar function togeter. It will also be less readabble on the caller side, since you don't know what part of the code the function is coming from. In typescript we have for instance Math with act like a static class, but they use som magic to pretend it to be using interface.
```
haakonflatval-cognite, line 44: This can probably also be fixed by a lint rule

react-components/src/architecture/base/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+- **Reuse:** I try to reuse code whenever it is possible. Therefore you will find functions that are very small,
+  often one liners. This ensures the complexity of the code to be low and increase readability on the caller side. For instance:
+
+        export function clear<T>(array: T[]): void {
+            array.splice(0, array.length);
+        }
```
haakonflatval-cognite, line 51: There are a bunch of builtin utility functions like this in the `lodash` library which we can use

react-components/src/architecture/base/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+- **Single responsibility:** I try to give each class one single responsibility. Therefore I use several base classes, each class adds some complexity.
+  Examples of this is `BaseView/ThreeView/GroupThreeView` which are all base classes. If they are merge to one single class, it will be too complex.
+
+- **Interpendence:** Try to keep must classes independed of other classes. For instance the tool should not know abount the views. Utility function and classes can used whenevery you like.
+
+- **Utility function and classes:** I have added som utility function/classes in `architecture/base/utilities`.
+  - Some utility functions may be a duplicate of some function in a package we use. Please tell me.
```
haakonflatval-cognite, line 59: `lodash` has a lot of simple util functions like this, we can probably go through them

react-components/src/architecture/base/README.md
nilscognite: So why haven't you used them before? Splice is a very bad name. It means connection 2 thing in English. When I see the spice, I always wonder what it does. In this example it is actually erasing all values.......  https://www.merriam-webster.com/dictionary/splice

react-components/src/architecture/base/README.md
nilscognite: Good

react-components/src/architecture/base/README.md
nilscognite: Good, see slack message

react-components/src/architecture/base/README.md
nilscognite: Good, see slack message

react-components/src/architecture/base/commands/BaseCommand.ts
nilscognite: Not really. If you have a better solution on this, come with it. But my message from Per Øyvind was to make the apps equal. If the apps has different shortCut, we will fail.

react-components/src/architecture/base/commands/BaseCommand.ts
nilscognite: If you don not use it as buttons, you don't need to implement the tooltip and the icon. This has never been a problem actually. If you make a base class for  these, you will loose a lot of flexibility.

react-components/src/architecture/base/README.md
```diff
@@ -0,0 +1,186 @@
...
+Here is the architecture with all base classes and some utility functionality.
+
+### architecture/base/domainObject
+
+- **DomainObject:** This is the base class for the domain object. A domain object is some sort of data, that is a collection of other domainObject or the data itself.
+
+  - An important property of the domainObject is that is has a list of views that will be updated.
+  - It has a parent and a collection of children. To ensure easy use, I have added several methods to access children, descendants and ancestors in several ways.
+
+  - It has some properties like selected, expanded and active. This is not used yet, but I wanted it to be here as this is part of the pattern. They are connected to how they are visualized and behave in a tree view, but this is not used for the moment.
```
haakonflatval-cognite, line 77: I imagine we will have trouble enforcing proper usage of these properties if we can't test them in any way. Could we leave them out now until we need them?

.gitignore
```diff
@@ -61,3 +61,4 @@ examples/public/
 documentation/.docusaurus
 documentation/generated
 documentation/docs/api
+/react-components/stories/public/cadModel
```
haakonflatval-cognite, line 64: Can we add e.g. a `public/local` folder where we can put custom models like this? Just so that we don't have to update the ignore file every time we start working with a new custom model

haakonflatval-cognite: (Also, the leading `/` shouldn't be necessary)

haakonflatval-cognite (commented):
I believe this contribution will be a major cornerstone in future development, and I'm conceptually very happy about it. There are some conventions/code style choices that disrupt our current norms, but I think we should wait until we've gotten our hands wet in testing it out before addressing it and coming up with better suggestions. I'll note some of my immediate concerns here:
- The current setup treats constructions as multiple-statement processes, so that e.g. constructing a domain object often looks like this:
```
const object = new DerivedDomainObject();
object.field = value;
....
```
I'd rather put such field initializations into the constructor itself
- Users are expected to inherit from e.g. Domain objects to integrate with the architecture, and there is in general heavy emphasis on class hierarchies and inheretance throughout the architecture. Obviously, it will work, but I do wonder if we can get away with something more "Typescript-y". It also means that many fields (name, color etc.) will be present in all Domain Objects even though they may not be relevant for all (at least unless we have a 2D three-view of the DomainObject hierarchy, of which there has been some talk. I believe @christjt had some ideas around this involving composition. I'll allow him to describe his ideas himself when he's back

haakonflatval-cognite (approved):
LGTM-ish :+1:

PR #4414, opened 2024-04-29
feat(react-components): timeseries support in rule based thresholds

danpriori:
#### Type of change

Feat  

#### Jira ticket :blue_book:
[](https://cognitedata.atlassian.net/browse/BND3D-4077)

## Description :pencil:
First implementation of timeseries in Rule Based Thresholds based on the defined data model and considering the creation of the rule without UI at this time, but providing support for the implementation of the UI later. The mechanism tries to get the timeseries and asset linked from two sources: the direct link and from the relationship, which brings more complexity than it should but and we have to support that and in some cases the timeseries are not direct linked to assets (which is only 1.1) as are more assets linked to the same timeseries with this old relationship format.

The tests were carried out using lervik-industries projects, which contain the largest variety of timeseries for testing and 3d-test to perform manual regression tests.

In lervik-industries, create rules that use timeseries that contain contextualized assets or contextualize the assets in some 3d nodes to be able to visualize them.

- [X] I am proud of this feature.


react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -31,6 +37,23 @@ export function RuleBasedOutputsSelector({
...
+    .map((ruleSet) => ruleSet.rule.expression)
+    .filter(isDefined);
+  const timeseriesExternalIdsFromRule = traverseExpressionToGetTimeseries(expressions) ?? [];
+
+  const { isLoading: isLoadingAssetIdsAndTimeseriesData, data: assetIdsWithTimeseriesData } =
+    useAssetsAndTimeseriesLinkageDataQuery({
+      timeseriesExternalIdsFromRule,
+      contextualizedAssetNodes
+    });
+
```
haakonflatval-cognite, line 56: There is a `useEffect` below that should be moved up since it's associated with `useAllMappedEquipmentAssetMappings()`

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -43,45 +66,65 @@ export function RuleBasedOutputsSelector({
...
+        })
+      );
     });
-  }, [assetMappings, ruleSet]);
+  }, [isLoadingAssetIdsAndTimeseriesData, ruleSet]);
 
   return <></>;
 }
 
+const initializeRuleBasedOutputs = async ({
```
haakonflatval-cognite, line 95: Is there a reason this is a lambda-function and not a normal function?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -31,6 +37,23 @@ export function RuleBasedOutputsSelector({
...
+  const contextualizedAssetNodes =
+    assetMappings?.pages
+      .flat()
+      .flatMap((item) => item.assets)
+      .map(convertAssetMetadataKeysToLowerCase) ?? [];
+
+  const expressions = ruleSet?.rulesWithOutputs
+    .map((ruleSet) => ruleSet.rule.expression)
+    .filter(isDefined);
+  const timeseriesExternalIdsFromRule = traverseExpressionToGetTimeseries(expressions) ?? [];
```
haakonflatval-cognite, line 49: This is triggered on every re-render, so this (maybe some of the lines above too?) should be put into `useMemo`

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -64,61 +82,100 @@ const checkStringExpressionStatement = (
...
+  );
+
+  if (currentTriggerData === undefined) return;
+
+  const dataTrigger =
+    currentTriggerData.type === 'metadata' && trigger.type === 'metadata'
+      ? Number(currentTriggerData.asset[trigger.type]?.[trigger.key])
+      : currentTriggerData.type === 'timeseries' && trigger.type === 'timeseries'
+        ? getTriggerTimeseriesNumericData(currentTriggerData, trigger)
+        : undefined;
```
haakonflatval-cognite, line 101: This nested ternary expression is hard to read, can we rewrite this?

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -64,61 +82,100 @@ const checkStringExpressionStatement = (
...
+
+    const dataFound = timeseriesWithDatapoints.find(
+      (item) => item.externalId === trigger.externalId
+    );
+
+    const datapoint = dataFound?.datapoints[dataFound?.datapoints.length - 1]?.value;
+
+    return Number(datapoint);
+  }
+  return undefined;
```
haakonflatval-cognite, line 121: Can we do early return in this function?

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -289,6 +413,68 @@ const analyzeNodesAgainstExpression = async ({
...
+        case 'and': {
+          timeseriesExternalIdFound = traverseExpressionToGetTimeseries(expression.expressions);
+          break;
+        }
+        case 'not': {
+          timeseriesExternalIdFound = traverseExpressionToGetTimeseries([expression.expression]);
+          break;
+        }
+        case 'numericExpression': {
+          timeseriesExternalIdFound = getTimeseriesExternalIdFromNumericExpression(expression);
```
haakonflatval-cognite, line 467: Are we restricting ourselves to only use time series that have an `externalId`? Are we sure that won't cause us trouble?

react-components/src/components/RuleBasedOutputs/utils.ts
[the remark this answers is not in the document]
haakonflatval-cognite: I'm a little worried that it's hard to extend this setup if we add even more field types. I think the `TriggerTypeData` type is a good start, but I think we should make a better effort of creating general reusable functions for this type, so that we don't have to fill the code with conditionals on the trigger type. But again, I suppose there will be a major rewrite when we migrate the backend

react-components/src/query/useAssetsAndTimeseriesLinkageDataQuery.ts
```diff
@@ -0,0 +1,219 @@
...
+            return generateAssetAndTimeseries(
+              assetFromTimeseries,
+              timeseries,
+              assetAndTimeseriesIdsFromRelationship
+            );
+          })
+          .flat()
+          .filter(isDefined) ?? [];
+
+      const timeseriesDatapoints = await getTimeseriesLatestDatapoints(sdk, externalIds);
```
haakonflatval-cognite, line 95: I see that we do several `await`s serially in this function. Can/should we await them in parallel?

haakonflatval-cognite (changes requested):
Nice! A few comments, otherwise looks good

react-components/src/components/RuleBasedOutputs/types.ts
```diff
@@ -5,14 +5,15 @@
...
+import { type Datapoints, type Asset, type Timeseries, type ExternalId } from '@cognite/sdk/';
 
 // =========== RULE BASED OUTPUT DATA MODEL
 
 export type TriggerType = 'timeseries' | 'metadata';
 
 export type TimeseriesRuleTrigger = {
   type: 'timeseries';
-  timeseriesId: number;
+  externalId: string;
```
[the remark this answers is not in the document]
danpriori: the timeseries endpoints that we are using support externalID but not all support internalID. At least this is what I see in the documentation.
The timeseries UI will show and let the user to search the timeseries based on the timeseries names or by asset. We will only store the timeseries info by its externalId. Do you think in this case is this important to have also internalId even if the UI will control this info anyway?
Or are you thinking about when we create the Rule Based Thresholds by eval mechanism to give full freedom to the user later?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -43,45 +66,65 @@ export function RuleBasedOutputsSelector({
...
+  contextualizedAssetNodes: Asset[];
+  ruleSet: RuleOutputSet;
+  assetIdsAndTimeseries: AssetIdsAndTimeseries[];
+  timeseriesDatapoints: Datapoints[] | undefined;
+}): Promise<AssetStylingGroupAndStyleIndex[]> => {
+  const flatAssetsMappingsList = assetMappings.pages
+    .flat()
+    .map((item) => item.mappings)
+    .flat();
+  const flatMappings = flatAssetsMappingsList.map((node) => node.items).flat();
```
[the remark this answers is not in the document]
danpriori: ops.. forgot to push these changes before:

https://github.com/cognitedata/reveal/pull/4414/commits/e9274eee0c2c72c3c317bf342082e9f6db1ebcd8

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -17,27 +17,45 @@ import {
...
-  const assetTrigger = asset[trigger.type]?.[trigger.key];
+  const currentTriggerData = triggerTypeData.find(
+    (triggerType) => triggerType.type === trigger?.type
+  );
+  if (
+    trigger?.type === 'metadata' &&
+    currentTriggerData?.type === 'metadata' &&
+    currentTriggerData?.asset === undefined
+  )
+    return;
```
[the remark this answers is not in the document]
danpriori: changed here. Merging to another checker in order to have a explanatory name :

https://github.com/cognitedata/reveal/pull/4414/commits/e9274eee0c2c72c3c317bf342082e9f6db1ebcd8

react-components/src/hooks/network/filterRelationships.ts
```diff
@@ -0,0 +1,50 @@
...
+        return [] as Relationship[];
+      });
+  };
+
+  let relationships: Relationship[] = [];
+
+  do {
+    relationships = relationships.concat(await fetch());
+    // eslint-disable-next-line no-unmodified-loop-condition
+  } while (cursor !== undefined);
```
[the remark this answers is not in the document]
danpriori: changed here with autopage:
https://github.com/cognitedata/reveal/pull/4414/commits/eafa59ef52ad2ff09547b677e079f695946969a8

react-components/src/query/useTimeseriesLatestDatapointQuery.ts
```diff
@@ -0,0 +1,28 @@
...
+  const sdk = useSDK();
+
+  const timeseries = timeseriesIds.map((id) => {
+    return { id };
+  });
+  const { data: timeseriesDatapoints } = useQuery(
+    [queryKeys.timeseriesLatestDatapoint(), timeseriesIds],
+    async () => await getTimeseriesLatestDatapoints(sdk, timeseries),
+    {
+      enabled: !isLoadingAssetIdsAndTimeseries && timeseries.length > 0
```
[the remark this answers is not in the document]
danpriori: Yeah. Agree. It is my thought, but I think I have to do more changes on other components in order to support it properly. I can start to play around with that on the next coming days.

react-components/src/components/RuleBasedOutputs/utils.ts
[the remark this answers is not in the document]
danpriori: Yeah. I have the same feeling. But I think it is safer to do that on the next sprints by separating some time of refactoring to improve this place.

react-components/src/utilities/types.ts
```diff
@@ -15,3 +23,38 @@ export function isAssetInstance(instance: InstanceReference): instance is AssetI
...
+
+export type AssetAndTimeseriesIds = {
+  assetIds: Partial<ExternalId & InternalId>;
+  timeseriesIds: Partial<ExternalId & InternalId>;
+};
+
+export type AssetIdsAndTimeseriesData = {
+  assetIdsWithTimeseries: AssetIdsAndTimeseries[];
+  timeseriesDatapoints: Datapoints[];
+};
```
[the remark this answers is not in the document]
danpriori: The first idea is to let others to get info from assets + timeseries combined from the `useAssetsAndTimeseriesLinkageDataQuery `hook, eventually. Nothing really related to rule based coloring actually. Are you sure even then it is better to move this up to the rule based scope?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: You mean closer to the useAllMappedEquipmentAssetMappings ?
For design patterns/convention, Is it not better to keep all useEffects at the end of the component and right before the jsx part?
But I changed here:
https://github.com/cognitedata/reveal/pull/4414/commits/82528c8036f90d3a908a0dc63f1693254f5455bd

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: ops.. no. Changed here:
https://github.com/cognitedata/reveal/pull/4414/commits/82528c8036f90d3a908a0dc63f1693254f5455bd

react-components/src/components/RuleBasedOutputs/utils.ts
danpriori: AFAIK all timeseries have to have an externalId, isnt?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: https://github.com/cognitedata/reveal/pull/4414/commits/45046b273630ff5a219303859c636e899d25cfd8

react-components/src/components/RuleBasedOutputs/utils.ts
danpriori: https://github.com/cognitedata/reveal/pull/4414/commits/b343e596937f0a83e23dbf4967a1762fd4e7a87c

react-components/src/query/useAssetsAndTimeseriesLinkageDataQuery.ts
danpriori: Awesome! Now we're talking ahaha nice catch! Like it!
Improved here.
https://github.com/cognitedata/reveal/pull/4414/commits/0788178dd8d724640516e763f63ca14c51e0cb59

react-components/src/utilities/queryKeys.ts
```diff
@@ -0,0 +1,18 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+import { type IdEither } from '@cognite/sdk/';
+
+export const queryKeys = {
+  all: ['cdf'] as const,
+  // ASSETS
+  assets: () => [...queryKeys.all, 'assets'] as const,
+  assetsById: (ids: IdEither[]) => [...queryKeys.all, 'assets', ids] as const,
```
haakonflatval-cognite, line 10: Is `assetsById` supposed to use the `assets` query key instead of `queryKeys.all`?

react-components/src/utilities/queryKeys.ts
```diff
@@ -0,0 +1,18 @@
...
+  // TIMESERIES
+  timeseries: () => [...queryKeys.all, 'timeseries'] as const,
```
haakonflatval-cognite, line 12: I wonder if this and the `assets` query keys should be defined outside of this exported object, since they're not supposed to be used outside this file.

react-components/src/hooks/network/getTimeseriesByIds.ts
haakonflatval-cognite, line 1: it's nice to separate non-hook functions from the hook ones! Should we put the `network` folder outside the `hooks` folder?

react-components/src/utilities/queryKeys.ts
danpriori: are you suggesting to remove the assets and timeseries and just define both outside in the places where would be called?

react-components/src/hooks/network/getSourceRelationships.ts
```diff
@@ -0,0 +1,29 @@
...
+  targetExternalIds: CogniteExternalId[];
+  sourceTypes?: RelationshipResourceType[];
+} & Omit<RelationshipsFilterRequest, 'sourceExternalIds' | 'targetTypes'>;
+
+export const getSourceRelationships = async (
+  sdk: CogniteClient,
+  payload: Payload
+): Promise<ExtendedRelationship[]> => {
+  return await filterRelationships(sdk, payload).then((relationships) => {
+    return relationships.map((relationship) => ({
```
haakonflatval-cognite, line 24: Just a nitpick - I think I'd prefer just awaiting the result from `filterRelationships` and then performing the `map` normally, instead of using a `then`

haakonflatval-cognite: I see you are doing similar things in the other `network` files too

react-components/src/components/RuleBasedOutputs/utils.ts
haakonflatval-cognite: It seems like that's not the case. The API lists it as an optional field: https://api-docs.cognite.com/20230101/tag/Time-series/operation/getTimeSeries (look at the list result)

react-components/src/components/RuleBasedOutputs/types.ts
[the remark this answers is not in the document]
haakonflatval-cognite: Time-series are not guaranteed to have an external ID, but are guaranteed to have an internal one - from what I can see, all API endpoints support internal ID, while some (not all) support externalID as well. Am I missing something?

react-components/src/query/useAssetsAndTimeseriesLinkageDataQuery.ts
haakonflatval-cognite: Nice!

react-components/src/hooks/network/getTimeseriesByIds.ts
danpriori: Yeah. Agree. I dont know why the subfolder network was put in the hook folder. We can do it on another PR I think

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
haakonflatval-cognite: Aah, I see now that that's the convention you're using. I think I'd still prefer it the new way, especially since it's not clear from the names used inside the `useEffect` what is being fetched

react-components/src/utilities/types.ts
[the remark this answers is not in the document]
haakonflatval-cognite: I see - I don't have a strong opinion. I just imagine that the `utilities/types.ts` file will be pretty cluttered eventually

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: So you prefer to move closer to the associated place? If so, then It is done already in the commit above.

react-components/src/utilities/queryKeys.ts
haakonflatval-cognite: No, I just meant putting them outside of the `queryKeys` object, but in this same file, just so that they're not exported/exposed

react-components/src/hooks/network/getSourceRelationships.ts
danpriori: oh yeah.. forgot to change that. Agree

https://github.com/cognitedata/reveal/pull/4414/commits/c7f080ddaa9bff8a4c58db3dad5ae0e1492c4d86

react-components/src/utilities/queryKeys.ts
danpriori: https://github.com/cognitedata/reveal/pull/4414/commits/c9a50e52b105e8f50cceda82d6fe1e4d14b2dfe7

react-components/src/utilities/queryKeys.ts
danpriori: Ops, make sense.
https://github.com/cognitedata/reveal/pull/4414/commits/c9a50e52b105e8f50cceda82d6fe1e4d14b2dfe7

react-components/src/components/RuleBasedOutputs/types.ts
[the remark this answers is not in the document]
danpriori: as we've discussed in DM, let's skip the internalID due to the fact of relationships endpoints used to get timeseries through relationship support only externalId.

react-components/src/components/RuleBasedOutputs/utils.ts
danpriori: as we've discussed in DM, let's skip the internalID due to the fact of relationships endpoints used to get timeseries through relationship support only externalId.

react-components/src/utilities/types.ts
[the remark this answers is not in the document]
danpriori: Then we can eventually separate in categories/spaces, or breakdown in separated type groups, I think.

react-components/src/utilities/queryKeys.ts
```diff
@@ -0,0 +1,24 @@
...
+  // TIMESERIES
+  timeseriesById: (ids: IdEither[]) => [...timeseries(), ids] as const,
+  timeseriesLatestDatapoint: () => [...timeseries(), 'latest-datapoints'] as const,
+  // TIMESERIES RELATIONSHIPS WITH ASSETS
+  timeseriesLinkedToAssets: () =>
+    [...queryKeys.all, 'timeseries', 'timeseries-linked-assets'] as const
+} as const;
+
+const assets = (): string[] => {
+  return [...queryKeys.all, 'assets'];
```
haakonflatval-cognite, line 19: Nitpick - this (and the `timeseries` lambda below) could just be constants instead of functions

haakonflatval-cognite (approved):
LGTM :+1:

react-components/src/utilities/queryKeys.ts
danpriori: ok but I thought that we were using this as a pattern. I can change it

danpriori: https://github.com/cognitedata/reveal/pull/4414/commits/88550354c2c0f4e426f590f68eb38e3959679926

PR #4417, opened 2024-04-29
chore: bump dependencies in /viewer

christjt:
#### Type of change

Chore 

## Description :pencil:

Bump all deps. The only one I had to revert was ESLint, as we have a plugin(header/header) which does not, at this moment, support the next major version.


viewer/packages/utilities/src/WebGLRendererStateHelper.test.ts
```diff
@@ -118,22 +212,60 @@ describe('WebGLRendererStateHelper', () => {
...
 
     helper.setScissor(0, 0, 128, 128);
-    expect(setScissorSpy).toBeCalledWith(0, 0, 128, 128);
+    expect(setScissorSpy).toHaveBeenCalledWith(0, 0, 128, 128);
 
     helper.resetState();
-    expect(setScissorSpy).toBeCalledWith(new THREE.Vector4(0, 0, 64, 64));
+    expect(setScissorSpy).toHaveBeenCalledWith(new THREE.Vector4(0, 0, 64, 64));
+
+    function mockedWebGLRenderer(): Mock<THREE.WebGLRenderer> {
```
haakonflatval-cognite, line 234: Is it necessary with separate renderer-mocking functions in this file? Is it because this test is testing very specific parts of the renderer?

viewer/packages/cad-model/src/batching/MultiBufferBatchingManager.ts
```diff
@@ -276,17 +287,19 @@ export class MultiBufferBatchingManager implements DrawCallBatchingManager {
     const newOffset = updateRange.byteOffset / typeSize;
     const newCount = updateRange.byteCount / typeSize;
 
-    const { offset: oldOffset, count: oldCount } = attribute.data.updateRange;
-    if (oldCount === -1) {
-      attribute.data.updateRange = { offset: newOffset, count: newCount };
+    if (attribute.data.updateRanges.length === 0) {
+      attribute.data.clearUpdateRanges();
+      attribute.data.updateRanges.push({ start: newOffset, count: newCount });
```
haakonflatval-cognite, line 292: Do we need to clear here? Isn't the updateRanges-array already empty?

haakonflatval-cognite (approved):
LGTM :+1: Just a couple of questions

viewer/packages/cad-model/src/batching/MultiBufferBatchingManager.ts
christjt: Guess we don't was just scared if there is some other metadata or something that it tracks internally.

PR #4499, opened 2024-05-15
fix: contexts not being cleaned up properly

christjt:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4232
(maybe this) https://cognitedata.atlassian.net/browse/BND3D-4175

## Description :pencil:

Removes missing and / or wrong cleanup causing contexts not being cleaned up properly.

Create an instance of Cognite3DViewer and instantly dispose it. Doing a memory snapshot should not show any missing webgl / rendering contexts (that reveal creates at least).


haakonflatval-cognite (approved):
LGTM :+1:

PR #4501, opened 2024-05-15
fix(react-components): unknown space `cog_3d_data` error when models are added into different space

pramodcog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4100

## Description :pencil:
If 3D models are not added into `cog_3d_data` space than on node clicking it would result in console error, this PR fixes the issue by fetching and maintain space the 3D model belong to.
 
In `Search`

- `yarn && yarn build && yalc publish` in `react-components`
- `yalc add @cognite/reveal-react-components && yarn && nx serve flexible-data-explorer` in `fusion` repo
- Load `omv` project and browse to 3D page
- Load `Maui Production Station` site
- Click on any 3D node, console.error should not popup


react-components/src/components/CacheProvider/RevisionFdmNodeCache.ts
```diff
@@ -247,9 +251,13 @@ export class RevisionFdmNodeCache {
 
       return edgesAndNodes.map((edge) => edge.edge);
     }
+    const modelInstances = await this._model;
```
danpriori, line 254: Can we have getters and setters instead of accessing internal/private parameters? Even internally I think it would be better to use it, unless we dont follow this OOP pattern or it has some consequences that I'm not aware of in our case.

pramodcog: getter and setter is useful only when it is used outside this class and we access it else where

danpriori: Actually, in some cases getter and setter are also good to be used internally, mostly for future code, single point of truth, future simplicity, etc.. is a common OOP pattern.
Like if we have to introduce some extra check before return/set that object then we have the opportunity to do that in one single place (inside the getter or setter) without replicating the same code all over other places that needs to have that treatment. Then we're ending up creating a function for that, which behaves like a getter/setter in the end of the day.

But ofc, I dont have a strong opinion in our case, as we're using this pattern already.

danpriori (approved):
LGTM

PR #4521, opened 2024-05-22
fix(react-components): make sure model matrices are compared properly before removal and bump to react-components 0.44.4

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:

Fix a bug where models would be removed if transform was not set (?)

### DIFF
--- a/react-components/src/utilities/isSameModel.ts
+++ b/react-components/src/utilities/isSameModel.ts
@@ -56,7 +56,7 @@ export function isSamePointCloudModel(
 
 const identity = new Matrix4();
 
-function isSameTransform(m0: Matrix4 | undefined, m1: Matrix4 | undefined) {
+function isSameTransform(m0: Matrix4 | undefined, m1: Matrix4 | undefined): boolean {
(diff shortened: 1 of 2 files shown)


react-components/src/utilities/isSameModel.ts
```diff
@@ -52,13 +50,24 @@ export function isSamePointCloudModel(
...
+    isSameTransform(model0.transform, model1.transform)
   );
 }
 
+const identity = new Matrix4();
+
+function isSameTransform(m0: Matrix4 | undefined, m1: Matrix4 | undefined): boolean {
+  if ((m0 === undefined || m0.equals(identity)) && (m1 === undefined || m1.equals(identity))) {
+    return true;
+  }
```
danpriori, line 62: m0 being undefined and m1 being a matrix4 identity is not really the same transform, for instance. Or something I've forgot?

danpriori: Ok so based on our discussion, the idea is to either being one undefined and another one is identity, it will be treated as with the same transform later on, so it is intentional. Ok.

PR #4550, opened 2024-05-29
fix: renderlayers get updated with stale data from material manager

christjt:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4264

## Description :pencil:

Cad node is updating the renderlayers based on when the node appearance provider triggers a change, but the material manager throttles (yields), and may have stale data when the render layers are computed. So correct behaviour is to rather listen for when the material manager updates.

Can be reproduced with the follow visual test setup (Default.VisualTest.ts)

```ts
export default class DefaultVisualTest extends ViewerVisualTestFixture {
  public async setup({ models, viewer }: ViewerTestFixtureComponents): Promise<void> {
    const model = models[0] as CogniteCadModel;
    const a = new TreeIndexNodeCollection([1]);
    const b = new TreeIndexNodeCollection([2]);
    const appearance: NodeAppearance = { renderInFront: true, color: new Color('red') };
    model.assignStyledNodeCollection(b, appearance);
    model.unassignStyledNodeCollection(b);
    await new Promise(resolve => setTimeout(resolve, 0));
    model.assignStyledNodeCollection(a, appearance);
  }
}
```

### DIFF
--- a/viewer/packages/cad-model/src/wrappers/CadNode.ts
+++ b/viewer/packages/cad-model/src/wrappers/CadNode.ts
@@ -91,7 +91,7 @@ export class CadNode extends Object3D {
     this._sourceTransform = new Matrix4().copy(model.modelMatrix);
     this._customTransform = new Matrix4();
 
-    this.nodeAppearanceProvider.on('changed', this._setModelRenderLayers);
+    this.materialManager.on('materialsChanged', this._setModelRenderLayers);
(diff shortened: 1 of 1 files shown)


pramodcog (approved):
Nice!
Appreciate the help.

PR #4567, opened 2024-06-04
(fix) Fix bounding box problems

nilscognite:
The bounding box was not updated when new models was added. Could probably done it other places as well, since reveal doesn't care about it's bounding box.  To wait for animate() to be run, it is too late


haakonflatval-cognite (approved):
LGTM :+1:

PR #4569, opened 2024-06-05
feat: Fixes for the architecture

nilscognite:
#### Type of change

Feat  

## Description :pencil:
* Added 2 function in ICustomObject and implemented in CustomObject 
* Refactored in some intersection to use useDepthTest  and added a predicate to filter intersections
* beforeRender  is needed to speed up the rendering for the architecture
* useDepthTest   will fix a bug in the architecture
* All are tested on the architecture  and works as intended

### DIFF
--- a/viewer/packages/api/src/public/migration/Cognite3DViewer.ts
+++ b/viewer/packages/api/src/public/migration/Cognite3DViewer.ts
@@ -1566,24 +1566,30 @@ export class Cognite3DViewer {
   /**
    * Raycasting model(s) for finding where the ray intersects with all models, including custom objects.
    * @param pixelCoords Pixel coordinate in pixels (relative to the domElement).
-   * @param predicate Check whether a CustomObject should be intersected.
    * @param options
    * @param options.stopOnHitting360Icon
+   * @param options.predicate Check whether a CustomObject should be intersected.
    * @returns A promise that if there was an intersection then return the intersection object - otherwise it
    * returns `null` if there were no intersections.
    * @beta
    */
   public async getAnyIntersectionFromPixel(
     pixelCoords: THREE.Vector2,
-    predicate?: (event: ICustomObject) => boolean,
-    options?: { stopOnHitting360Icon?: boolean }
+    options?: {
+      stopOnHitting360Icon?: boolean;
+      predicate?: (customObject: ICustomObject) => boolean;
+    }
   ): Promise<AnyIntersection | undefined> {
     if ((options?.stopOnHitting360Icon ?? true) && this.isIntersecting360Icon(pixelCoords)) {
       return undefined;
     }
 
+    const predicate = options?.predicate;
     let intersection: AnyIntersection | undefined = undefined;
-    intersection = this.getCustomObjectIntersectionIfCloser(pixelCoords, false, undefined, predicate);
+    intersection = this.getCustomObjectIntersectionIfCloser(pixelCoords, {
+      useDepthTest: false,
+      predicate
+    });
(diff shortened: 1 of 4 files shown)


viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1685,7 +1694,9 @@ export class Cognite3DViewer {
     const start = Date.now();
 
     this._events.beforeSceneRendered.fire({ frameNumber, renderer: this.renderer, camera });
-
+    this._sceneHandler.customObjects.forEach(customObject => {
+      customObject.beforeRender(camera);
+    });
```
haakonflatval-cognite, line 1699: Should the `beforeRender`-funcion of each `customObject` be added to the `this._events.beforeSceneRendered` trigger instead of manually getting called here?

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1772,14 +1783,22 @@ export class Cognite3DViewer {
 
   private getCustomObjectIntersectionIfCloser(
     pixelCoords: THREE.Vector2,
-    closestDistanceToCamera: number | undefined
+    useDepthTest: boolean,
+    closestDistanceToCamera: number | undefined,
+    predicate?: (event: ICustomObject) => boolean
   ): CustomObjectIntersection | undefined {
```
haakonflatval-cognite, line 1789: I think this would be cleaner if we made the second argument an `options`-object that contained all of these options

haakonflatval-cognite (changes requested):
Looks good, just a few questions

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1574,28 +1575,36 @@ export class Cognite3DViewer {
...
     options?: { stopOnHitting360Icon?: boolean }
   ): Promise<AnyIntersection | undefined> {
     if ((options?.stopOnHitting360Icon ?? true) && this.isIntersecting360Icon(pixelCoords)) {
       return undefined;
     }
 
-    const modelIntersection = await this.intersectModels(pixelCoords.x, pixelCoords.y, { asyncCADIntersection: false });
-
+    let intersection: AnyIntersection | undefined = undefined;
+    intersection = this.getCustomObjectIntersectionIfCloser(pixelCoords, false, undefined, predicate);
```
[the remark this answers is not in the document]
nilscognite: Yes.  I want the intersection to be a AnyIntersection. Otherwise you will get an compiler error

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
nilscognite: It was like that it worked before on the architeture side. But the loop was too compicated

viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1574,28 +1575,36 @@ export class Cognite3DViewer {
    */
   public async getAnyIntersectionFromPixel(
     pixelCoords: THREE.Vector2,
+    predicate?: (event: ICustomObject) => boolean,
     options?: { stopOnHitting360Icon?: boolean }
   ): Promise<AnyIntersection | undefined> {
```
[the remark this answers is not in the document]
nilscognite: Fixed

haakonflatval-cognite (approved):
LGTM :+1:

PR #4585, opened 2024-06-10
refactor: Use Vector3Pool  instead of ReusableVector3s and extra argument to IFlexibleCameraManager .onWheel

nilscognite:
#### Type of change

Refactor 

## Description :pencil:

- The class Vector3Pool is moved from react-components
- Removed ReusableVector3s
- Use Vector3Pool  instead of ReusableVector3s
- Expose Vector3Pool out of Reveal. Change a little bit on the usage.
- Expose getWheelEventDelta
- Updated onWheel in IFlexibleCameraManager to have an extra argument. The reason for this is to synchronize what delta means, since it may depend of the browser. I then calculated it once and used everywhere.


haakonflatval-cognite (approved):
LGTM :+1:

PR #4618, opened 2024-06-19
feat: Make options for  FlexibleControls available outside Reveal + Add speed option

nilscognite:
#### Type of change

Feat  

## Description :pencil:

* Expose the options out of Reveal
* Made a new speed option. Default this is 0

Can not test this yet. Another PR in components

### DIFF
--- a/viewer/packages/camera-manager/src/Flexible/FlexibleControlsOptions.ts
+++ b/viewer/packages/camera-manager/src/Flexible/FlexibleControlsOptions.ts
@@ -19,10 +19,10 @@ export class FlexibleControlsOptions {
   // INSTANCE FIELDS
   //================================================
 
-  // Main behaivor
+  // Main behavior
   public controlsType = FlexibleControlsType.Orbit;
 
-  // Mouse click, double click and wheel behaivor
+  // Mouse click, double click and wheel behavior
   public mouseWheelAction = FlexibleWheelZoomType.Auto;
   public mouseClickType = FlexibleMouseActionType.SetTarget;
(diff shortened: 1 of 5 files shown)


viewer/packages/camera-manager/src/Flexible/FlexibleControlsOptions.ts
```diff
@@ -117,6 +118,13 @@ export class FlexibleControlsOptions {
     }
   }
 
+  public getKeyboardSpeed(shift: boolean): number {
+    if (!shift) {
```
danpriori, line 122: If I understood this correctly, the keyboard speed will be enabled only when the user is pressing shift?
It was requested? I thought that the user only needs to set the speed from the dropdown and just use it. If the user has to keep pressing the shift while wanting to keep the movement faster, is it not be annoying?

danpriori (approved):
Just a comment.
I think it is good if the "shift" is intentional and we can approve. Otherwise I think we have to change it.

viewer/packages/camera-manager/src/Flexible/FlexibleControlsOptions.ts
nilscognite: No. Look at the code.  The control I have made changes this.keyboardSpeed. This used when moving by keyboard. So the user is able to change that. 

If shift is pressed, the speed will be multiplied by  this.keyboardFastMoveFactor.

danpriori: Ah, yeah, right. So we already have a move factor when pressing shift and it is already there from a long time, way before this keyboard speed... you're right.
All good!

PR #4619, opened 2024-06-19
feat (react-components): Add general option and use it to make camera speed settings

nilscognite:
#### Type of change
Feat  

#### Jira ticket :blue_book:
https://cognitedata.atlassian.net/browse/BND3D-4365

## Description :pencil:

* Generalize possibility to add option to the general toolbar. Any thing that can have a icon or sytring can be a part of the options.
* Use it to make a SetSpeedOptionCommand, which sets the speed of the camera keyboard.

Use storybook, architecture:. You se the option there.


react-components/src/architecture/concrete/config/StoryBookConfig.ts
```diff
@@ -19,6 +19,7 @@ import { type BaseTool } from '../../base/commands/BaseTool';
 import { ToggleMetricUnitsCommand } from '../../base/concreteCommands/ToggleMetricUnitsCommand';
 import { MeasurementTool } from '../measurements/MeasurementTool';
 import { ClipTool } from '../clipping/ClipTool';
+import { SetSpeedOptionCommand } from '../../base/concreteCommands/SetKeyboardSpeedCommand';
```
danpriori, line 22: `SetKeyboardSpeedCommand`  instead of `SetSpeedOptionCommand` ?

Also, one comment about naming: is it not better to not use "Sets" and "Gets" as a prefix of classes? Maybe only "KeyboardSpeedCommand" or something like this.

As in OOP we use getters and setters for methods to manage parameter values, it might be confusing IMO.

react-components/src/components/Architecture/OptionButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [uniqueId, setUniqueId] = useState<number>(0);
+  const [icon, setIcon] = useState<IconType | undefined>(undefined);
+
+  useEffect(() => {
+    function update(command: BaseCommand): void {
+      setEnabled(command.isEnabled);
+      setVisible(command.isVisible);
+      setUniqueId(command.uniqueId);
+      setIcon(getIcon(command));
+    }
```
danpriori, line 38: Maybe is better to move this update function out of the useEffect and use the hook useCallback

react-components/src/components/Architecture/Utilities.tsx
```diff
@@ -0,0 +1,47 @@
...
+    return undefined;
+  }
+  return command.icon as IconType;
+}
+
+export function getTooltipPlacement(isHorizontal: boolean): 'top' | 'right' {
+  return isHorizontal ? 'top' : 'right';
+}
+
+export function getButtonType(command: BaseCommand): 'ghost' | 'ghost-destructive' | 'primary' {
```
danpriori, line 21: Maybe is better to define a type for this.
Perhaps:
```typescript
type ButtonType = 'ghost' | 'ghost-destructive' | 'primary';
```

react-components/src/components/Architecture/Utilities.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export function getIcon(command: BaseCommand): IconType | undefined {
+  if (command.icon === undefined) {
+    return undefined;
+  }
+  return command.icon as IconType;
+}
+
+export function getTooltipPlacement(isHorizontal: boolean): 'top' | 'right' {
+  return isHorizontal ? 'top' : 'right';
```
danpriori, line 18: Maybe is better to define a type for this.
Perhaps:

```typescript
type TooltipPlacementType = 'top' | 'right';
``` 

And also create/use a type file to store all these new type definitions there

react-components/src/architecture/base/concreteCommands/SetKeyboardSpeedCommand.ts
```diff
@@ -0,0 +1,52 @@
...
+    super();
+    this._value = value;
+  }
+
+  public override get tooltip(): TranslateKey {
+    return { fallback: this._value.toString() + 'x' };
+  }
+
+  public override get isChecked(): boolean {
+    return this._value === this.renderTarget.flexibleCameraManager.options.keyboardSpeed;
```
danpriori, line 45: Do we have the options exposed from the flexibleCameraManager ?

danpriori (changes requested):
A few comments, but the rest looks good!

react-components/src/architecture/concrete/config/StoryBookConfig.ts
nilscognite: I agree. Thanks for comment!

react-components/src/components/Architecture/OptionButton.tsx
nilscognite: I don't know why or why not. I just do things that is working. Can you suggest by code it?

react-components/src/components/Architecture/Utilities.tsx
nilscognite: Fix

react-components/src/components/Architecture/Utilities.tsx
nilscognite: Fixed

react-components/src/architecture/base/concreteCommands/SetKeyboardSpeedCommand.ts
nilscognite: Yes. The other PR

nilscognite:
The fixes are here: 836d8d5cb7910a4a03fdb0382dd39ad7e56abdb5

react-components/src/architecture/base/concreteCommands/SetKeyboardSpeedCommand.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public override invokeCore(): boolean {
+    this.renderTarget.flexibleCameraManager.options.keyboardSpeed = this._value;
```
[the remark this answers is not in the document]
nilscognite: Yes. But it require a new version of Reveal. On my side it works perfectly.

react-components/src/components/Architecture/OptionButton.tsx
danpriori: if we define a function inside the `useEffect`, then the compiler will re create the same function every time when that `useEffect `is triggered and, in this case, with no reason.
So one way to avoid that is to or define this "regular" function outside of the `useEffect `or define it as a `useCallback `hook (outside the `useEffect `as well). `useCallback `will use the memoization mechanism for the `update `function.
So maybe something like this:

```typescript
   const update = useCallback((command: BaseCommand) => {
    setEnabled(command.isEnabled);
    setVisible(command.isVisible);
    setUniqueId(command.uniqueId);
    setIcon(getIcon(command));
  }, [setEnabled, setVisible, setUniqueId, setIcon]);
  

useEffect(() => {
    update(newCommand);
    newCommand.addEventListener(update);
    return () => {
      newCommand.removeEventListener(update);
    };
 }, [newCommand]);
```

nilscognite: Ok, I see. This is now fixed

danpriori (approved):
Super! Nice work man!
LGTM

PR #4631, opened 2024-06-21
feat(react-components): initial support for Observations

haakonflatval-cognite:
#### Type of change

Feat  

## Description :pencil:

Initial support for observations. Includes a couple of other changes.

- Add ObservationsTool which currently is used for visualizing ObservationDomainObjects, representing Observations stored in the backend. These concepts are all work in progress.
- Add generic argument to the `BaseView` class, to make it explicit what DomainObject it concerns.
- Change the CommandButton form to follow React conventions
-  Use `useMemo` to avoid recreating command objects repeatedly in React command buttons
- Make sure `filter` in the `instanceFilter` query is made undefined if empty (otherwise it's an invalid request), as was done with the `search` endpoint earlier.

In storybook (Toolbar) and linked into Fusion


react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
```diff
@@ -0,0 +1,118 @@
...
+import { type FdmSDK, type InstanceFilter } from '../../../utilities/FdmSDK';
+import { OBSERVATION_SOURCE, type Observation, type ObservationProperties } from './models';
+import { ObservationsDomainObject } from './ObservationsDomainObject';
+import { Color, Vector3 } from 'three';
+import { CDF_TO_VIEWER_TRANSFORMATION, type Overlay3D, Overlay3DCollection } from '@cognite/reveal';
+import { NavigationTool } from '../../base/concreteCommands/NavigationTool';
+import { type RevealRenderTarget } from '../../base/renderTarget/RevealRenderTarget';
+
+const DEFAULT_OVERLAY_COLOR = new Color('lightblue');
+const SELECTED_OVERLAY_COLOR = new Color('red');
```
danpriori, line 15: just a cosmetic question: is it not better to move out these const to a `constant` file?

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export class ObservationsTool extends NavigationTool {
+  // private _domainObjects: ObservationsDomainObject[] = [];
```
danpriori, line 18: I think it is better to clean this up if wont being use it.

react-components/src/architecture/concrete/observationsDomainObject/ObservationsView.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright 2024 Cognite AS
+ */
+import { Box3 } from 'three';
+import { type ObservationsDomainObject } from './ObservationsDomainObject';
+import { GroupThreeView } from '../../base/views/GroupThreeView';
+import { type Overlay3DCollection } from '@cognite/reveal';
+import { type Observation } from './models';
+
+export class ObservationsView extends GroupThreeView<ObservationsDomainObject> {
+  private readonly _collection: Overlay3DCollection<Observation>;
```
nilscognite, line 11: Please follow the example code. You should not store the data here, but get it by the DomainObject.

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
haakonflatval-cognite: Sure, I don't have a strong opinion :sweat_smile:

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
```diff
@@ -0,0 +1,118 @@
...
+  private readonly _domainObjectPromise: Promise<ObservationsDomainObject>;
+
+  private _selectedOverlay: Overlay3D<Observation> | undefined;
+
+  constructor(fdmSdk: FdmSDK) {
```
nilscognite, line 23: You should have an empty constructor here. The data should be fetched by the domain object. Se the example code how it is done. The  _domainObjectPromise should be in the domain object itself.

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
```diff
@@ -0,0 +1,118 @@
...
+    void this._domainObjectPromise.then((domainObject) => {
+      domainObject.setVisibleInteractive(false, this.renderTarget);
+    });
+  }
+
+  public override async onClick(event: PointerEvent): Promise<void> {
+    const domainObject = await this._domainObjectPromise;
+    this._selectedOverlay?.setColor(DEFAULT_OVERLAY_COLOR);
+
+    const normalizedCoord = this.getNormalizedPixelCoordinates(event);
```
nilscognite, line 86: You should follow the example code here. This code will not work properly since you don't see any other object.

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
```diff
@@ -0,0 +1,118 @@
...
+    const collection = new Overlay3DCollection<Observation>(observationOverlays, {
+      defaultOverlayColor: DEFAULT_OVERLAY_COLOR
+    });
+
+    const observationDomainObject = new ObservationsDomainObject(collection);
+
+    return observationDomainObject;
+  }
+
+  public override get icon(): IconType {
```
nilscognite, line 55: Please move all override to the top. It will be easier to read

nilscognite (changes requested):
In this PR you do to much outside of the architecture. Our observation should be in the domain object itself and added to the root at startup.

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
danpriori: @nilscognite  and also move the `initializeDomainObjects` method to the domainObject?

react-components/src/components/Architecture/RevealButtons.tsx
```diff
@@ -4,24 +4,53 @@
...
+import { useFdmSdk } from '../RevealCanvas/SDKProvider';
+import { ObservationsTool } from '../../architecture/concrete/observationsDomainObject/ObservationsTool';
 
 export class RevealButtons {
-  static FitView = (): ReactElement => CreateCommandButton(new FitViewCommand());
-  static NavigationTool = (): ReactElement => CreateCommandButton(new NavigationTool());
-  static SetAxisVisible = (): ReactElement => CreateCommandButton(new SetAxisVisibleCommand());
-  static Measurement = (): ReactElement => CreateCommandButton(new MeasurementTool());
-  static Clip = (): ReactElement => CreateCommandButton(new ClipTool());
+  static FitView = (): ReactElement => (
```
nilscognite, line 18: Please revert this. file. The commands doesn't need a common constructor argument. Beside this the coide complexity has increased a lot in this file.

nilscognite (changes requested):

react-components/src/components/Architecture/RevealButtons.tsx
haakonflatval-cognite: The constructor argument is necessary to avoid having the commands/tools reconstructed on every React re-render. But I can change back to using functions, but in that case we must be clear that they are not React-components, just regular functions that return React-components

haakonflatval-cognite: So I will change the casing to camel-case

haakonflatval-cognite: The result did not end up as clean as I had hoped, but I'm not sure how to do it better. If we can guarantee that we never want the tool that is attached to a button, to be reconstructed, we can make it a little simpler

haakonflatval-cognite: Went with that last idea. Ready for re-review

react-components/src/architecture/concrete/observationsDomainObject/ObservationsDomainObject.ts
```diff
@@ -0,0 +1,68 @@
...
+    return { fallback: ObservationsDomainObject.name };
+  }
+
+  constructor(fdmSdk: FdmSDK) {
+    super();
+    void fetchObservations(fdmSdk).then((observations) => {
+      this.initializeCollection(observations);
+      this.notify(Changes.geometry);
+    });
+  }
```
danpriori, line 30: Btw, on classes, in general for easy reading, we usually define the constructor right after the parameters and as Nils mentioned, the override functions on the top (but IMO it should be right below the constructor). At the bottom we keep the rest of the methods (public and then private/protected at the end).
But TBH I dont have a strong opinion on that. As long as we all agree on a unique way to organize the classes, I'm in :)

react-components/src/architecture/concrete/exampleDomainObject/ExampleView.ts
```diff
@@ -16,13 +16,13 @@ import {
...
 
-export class ExampleView extends GroupThreeView {
+export class ExampleView extends GroupThreeView<ExampleDomainObject> {
   // ==================================================
   // INSTANCE PROPERTIES
   // ==================================================
 
   public override get domainObject(): ExampleDomainObject {
-    return super.domainObject as ExampleDomainObject;
+    return super.domainObject;
```
danpriori, line 25: Is it not better to call the domainObject as his inheritance instead of calling direct from the super class? Or doesnt work in this case for some reason that I'm not following?

react-components/src/architecture/concrete/primitives/box/BoxView.ts
```diff
@@ -68,7 +68,7 @@ export class BoxView extends GroupThreeView {
   // ==================================================
 
   public override get domainObject(): BoxDomainObject {
-    return super.domainObject as BoxDomainObject;
+    return super.domainObject;
```
danpriori, line 71: Is it not better to call the domainObject as his inheritance instead of calling direct from the super class? Or doesnt work in this case for some reason that I'm not following?

react-components/src/architecture/concrete/primitives/line/LineView.ts
```diff
@@ -41,13 +41,13 @@ import { BoxView } from '../box/BoxView';
...
 
-export class LineView extends GroupThreeView {
+export class LineView extends GroupThreeView<LineDomainObject> {
   // ==================================================
   // INSTANCE PROPERTIES
   // ==================================================
 
   public override get domainObject(): LineDomainObject {
-    return super.domainObject as LineDomainObject;
+    return super.domainObject;
```
danpriori, line 50: same question here :D

react-components/src/architecture/concrete/primitives/plane/PlaneView.ts
```diff
@@ -45,7 +45,7 @@ export class PlaneView extends GroupThreeView {
   // ==================================================
 
   public override get domainObject(): PlaneDomainObject {
-    return super.domainObject as PlaneDomainObject;
+    return super.domainObject;
```
danpriori, line 48: and here :D

react-components/src/architecture/concrete/terrainDomainObject/TerrainThreeView.ts
```diff
@@ -35,13 +35,13 @@ import { type RegularGrid2 } from './geometry/RegularGrid2';
...
 
-export class TerrainThreeView extends GroupThreeView {
+export class TerrainThreeView extends GroupThreeView<TerrainDomainObject> {
   // ==================================================
   // INSTANCE PROPERTIES
   // ==================================================
 
   public override get domainObject(): TerrainDomainObject {
-    return super.domainObject as TerrainDomainObject;
+    return super.domainObject;
```
danpriori, line 44: same question here (about the super usage) :D

react-components/src/components/Architecture/CommandButton.tsx
```diff
@@ -9,6 +9,14 @@ import { useTranslation } from '../i18n/I18n';
 import { type BaseCommand } from '../../architecture/base/commands/BaseCommand';
 import { getButtonType, getDefaultCommand, getIcon, getTooltipPlacement } from './utilities';
 
+export const createCommandButton = (
+  commandConstructor: () => BaseCommand,
+  isHorizontal = false
+): ReactElement => {
+  const command = useMemo(commandConstructor, []);
```
danpriori, line 16: Hmmm interesting usage of the useMemo with the command constructor function! I like it.

danpriori (approved):
LGTM. Some few comments but they are not blockers. Good job!
Note: Let's wait Nils reviews as he can see some pitfalls that I'm not seeing since he knows this architecture better atm.

react-components/src/architecture/concrete/observationsDomainObject/ObservationsDomainObject.ts
nilscognite: Yes. Please keep the order as I have started with. Otherwise, the navigation is harder!

react-components/src/architecture/base/commands/BaseTool.ts
```diff
@@ -150,7 +149,7 @@ export abstract class BaseTool extends RenderTargetCommand {
       if (!(customObject instanceof ThreeView)) {
         return false;
       }
-      return domainObjectPredicate(customObject.domainObject);
+      return domainObjectPredicate(customObject.domainObject as DomainObject);
```
nilscognite, line 152: ?? I get a lint error on this. Why this cast?

react-components/src/architecture/concrete/exampleDomainObject/ExampleView.ts
nilscognite: All override get domainObject() is not necessary anymore when using template in views!

react-components/src/architecture/concrete/observationsDomainObject/ObservationsDomainObject.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  public override get typeName(): TranslateKey {
+    return { fallback: ObservationsDomainObject.name };
+  }
+
+  constructor(fdmSdk: FdmSDK) {
+    super();
+    void fetchObservations(fdmSdk).then((observations) => {
+      this.initializeCollection(observations);
+      this.notify(Changes.geometry);
```
nilscognite, line 28: Skip notify here. Should not be called in the constructor. Unnecessary, since constructor change everything.

react-components/src/architecture/concrete/observationsDomainObject/ObservationsDomainObject.ts
```diff
@@ -0,0 +1,68 @@
...
+  public get overlayCollection(): Overlay3DCollection<Observation> {
+    return this._collection;
+  }
+
+  protected override createThreeView(): ThreeView<ObservationsDomainObject> | undefined {
+    return new ObservationsView();
+  }
+}
+
+const observationsFilter: InstanceFilter = {};
```
nilscognite, line 58: Strange syntax. Is this a field? Should be at top of the class.

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
```diff
@@ -0,0 +1,76 @@
...
+import { type Observation } from './models';
+import { ObservationsDomainObject } from './ObservationsDomainObject';
+import { type Overlay3D } from '@cognite/reveal';
+import { DEFAULT_OVERLAY_COLOR, SELECTED_OVERLAY_COLOR } from './constants';
+import { BaseEditTool } from '../../base/commands/BaseEditTool';
+import { type VisualDomainObject } from '../../base/domainObjects/VisualDomainObject';
+
+export class ObservationsTool extends BaseEditTool {
+  private _selectedOverlay: Overlay3D<Observation> | undefined;
+  private _observationsDomainObject: ObservationsDomainObject | undefined;
```
nilscognite, line 15: Never use direct references like this. Use can easily get it by the root, as I told you on the course. Use `this.rootDomainObject.getDescendantByType(ObservationsDomainObject)` each time use should get it. You can make a property that do this.

```
  public get observationsDomainObject(): ObservationsDomainObject | undefined {
    return this.rootDomainObject.getDescendantByType(ObservationsDomainObject);
  }


```

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
```diff
@@ -0,0 +1,76 @@
...
+import { type TranslateKey } from '../../base/utilities/TranslateKey';
+import { type Observation } from './models';
+import { ObservationsDomainObject } from './ObservationsDomainObject';
+import { type Overlay3D } from '@cognite/reveal';
+import { DEFAULT_OVERLAY_COLOR, SELECTED_OVERLAY_COLOR } from './constants';
+import { BaseEditTool } from '../../base/commands/BaseEditTool';
+import { type VisualDomainObject } from '../../base/domainObjects/VisualDomainObject';
+
+export class ObservationsTool extends BaseEditTool {
+  private _selectedOverlay: Overlay3D<Observation> | undefined;
```
nilscognite, line 14: _selectedOverlay. This should be in the view. Remember that there should be no visualization code in the tools.

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
```diff
@@ -0,0 +1,76 @@
...
+    const normalizedCoords = this.getNormalizedPixelCoordinates(event);
+
+    const intersectedIcon = domainObject.overlayCollection.intersectOverlays(
+      normalizedCoords,
+      this.renderTarget.camera
+    );
+
+    this.handleIntersectedOverlay(intersectedIcon);
+
+    this.renderTarget.viewer.requestRedraw();
```
nilscognite, line 58: No redrawing in the tools either. You should call observationsDomainObject.notify(Changed.selected ???)

react-components/src/architecture/concrete/observationsDomainObject/ObservationsTool.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  protected override canBeSelected(domainObject: VisualDomainObject): boolean {
+    return domainObject instanceof ObservationsDomainObject;
+  }
+
+  private handleIntersectedOverlay(overlay: Overlay3D<Observation> | undefined): void {
```
nilscognite, line 65: View code here also. Move to the view. You should handle selection by the same pattern as the example.

react-components/src/components/Architecture/CommandButton.tsx
```diff
@@ -9,6 +9,14 @@ import { useTranslation } from '../i18n/I18n';
 import { type BaseCommand } from '../../architecture/base/commands/BaseCommand';
 import { getButtonType, getDefaultCommand, getIcon, getTooltipPlacement } from './utilities';
 
+export const createCommandButton = (
+  commandConstructor: () => BaseCommand,
```
nilscognite, line 13: Please remove the commandConstructor. The architecture doesn't need this feature.

react-components/src/components/Architecture/RevealButtons.tsx
```diff
@@ -10,20 +10,31 @@ import { SetFlexibleControlsTypeCommand } from '../../architecture/base/concrete
 import { SetAxisVisibleCommand } from '../../architecture/concrete/axis/SetAxisVisibleCommand';
```
nilscognite, line 10: Can not see the purpose of this way of doing it anymore. The way it was done before your changes is a lot cleaner.

react-components/src/components/RevealContext/RevealContext.tsx
```diff
@@ -106,7 +106,7 @@ const useRevealFromKeepAlive = ({
         useFlexibleCameraManager: true,
         hasEventListeners: false
       });
-      renderTarget = new RevealRenderTarget(viewer);
+      renderTarget = new RevealRenderTarget(viewer, sdk);
```
nilscognite, line 109: Good!

react-components/src/components/RuleBasedOutputs/types.ts
```diff
@@ -220,11 +220,6 @@ export type SimpleSource = {
   version: string;
 } & DmsUniqueIdentifier;
 
-export type DmsUniqueIdentifier = {
```
nilscognite, line 223: Dms? What is that? I prefer proper names.

nilscognite (changes requested):
Better now. But here is things you still have misunderstood. Having a hard reference to a DomainObject directly in a tool wrong. I said in the course that number of references to a object should be minimum and controlled by the architecture. What happens for instance when another command delete it, but the reference is still in the tool? The architecture will guarantee that the domain objects hold by the root (as a hierarchy)  is "alive".

react-components/src/architecture/concrete/observationsDomainObject/ObservationsDomainObject.ts
haakonflatval-cognite: It's not called directly in the constructor, it's called as a reaction to a promise (in the `.then` function). I want to notify when `fetchObservations` is done, which is an async function

react-components/src/architecture/concrete/observationsDomainObject/ObservationsDomainObject.ts
haakonflatval-cognite: No, it's not a field, just a constant. I'll put it into a function, I think :thinking:

react-components/src/architecture/concrete/exampleDomainObject/ExampleView.ts
haakonflatval-cognite: You're right - removed them :+1:

react-components/src/architecture/base/commands/BaseTool.ts
haakonflatval-cognite: You get a Lint error? Weird. It's because Typescript is unable to understand that `domainObject` is always of the generic type DomainObjectType, which always extends DomainObject, and thus is always a DomainObject. So, annoyingly, I had to add this cast

react-components/src/components/RuleBasedOutputs/types.ts
haakonflatval-cognite: It's not super-relevant to the PR, it's just a small cleanup. But DMS is the low-level API to the FDM-backend. Really not sure how to name this differently without being very verbose (also, I don't remember what the letters stand for)

nilscognite (approved):
Better now

PR #4651, opened 2024-07-01
fix(react-components): adopt Cogs chips and fix 360 image details panel appearance

haakonflatval-cognite:
#### Type of change

Chore 

#### Jira ticket :blue_book:

## Description :pencil:

~~Migrate components according to Cogs.js adoption plan.~~

This grew a bit out of it's original plans, but the change seems sort-of necessary: The Image360 panel would not always show up when entering 360 panels because of some React re-rendering shenanigans - the 360 image container is temporarily unmounted after getting added, and we did not properly react to when it was added anew. This PR fixes the second problem - the 360 is still mounted twice, but we now correctly react to it (by using the `use360Collections` hook, and making sure the resource count includes 360 images). 

Aside from that, I've simplified two occurrences of cogs.js chips as part of the cogs-adaption.

### DIFF
--- a/react-components/src/components/CadModelContainer/CadModelContainer.tsx
+++ b/react-components/src/components/CadModelContainer/CadModelContainer.tsx
@@ -44,7 +44,7 @@ export function CadModelContainer({
     addModel(addModelOptions, transform)
       .then((model) => {
         onLoad?.(model);
-        setRevealResourcesCount(viewer.models.length);
+        setRevealResourcesCount(viewer.models.length + viewer.get360ImageCollections().length);
       })
       .catch((error) => {
         const errorReportFunction = onLoadError ?? defaultLoadErrorHandler;
@@ -98,7 +98,7 @@ export function CadModelContainer({
       return;
 
     viewer.removeModel(model);
-    setRevealResourcesCount(viewer.models.length);
+    setRevealResourcesCount(viewer.models.length + viewer.get360ImageCollections().length);
     setModel(undefined);
   }
 }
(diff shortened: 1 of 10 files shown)


react-components/src/components/CadModelContainer/CadModelContainer.tsx
```diff
@@ -44,7 +44,7 @@ export function CadModelContainer({
     addModel(addModelOptions, transform)
       .then((model) => {
         onLoad?.(model);
-        setRevealResourcesCount(viewer.models.length);
+        setRevealResourcesCount(viewer.models.length + viewer.get360ImageCollections().length);
```
danpriori, line 47: I would prefer to use a const variable to set this summed counter, but it is more a personal taste, TBH. We can keep this as it is, if you prefer :)

danpriori (approved):
one comment but LGTM

react-components/src/components/CadModelContainer/CadModelContainer.tsx
haakonflatval-cognite: I agree that this should be more standardized across different model types... I added a `getViewerResourceCount` function that is used on all locations

react-components/src/components/CadModelContainer/CadModelContainer.tsx
```diff
@@ -44,7 +45,7 @@ export function CadModelContainer({
     addModel(addModelOptions, transform)
       .then((model) => {
         onLoad?.(model);
-        setRevealResourcesCount(viewer.models.length);
+        setRevealResourcesCount(getViewerResourceCount(viewer));
```
danpriori, line 48: Super! Way better IMO :)

PR #4652, opened 2024-07-01
fix(react-components): improve asset mapping cache mechanism and refactoring to stabilize rule threshold styling and switching

danpriori:
#### Type of change

Bug 

#### Jira ticket :blue_book:
[](https://cognitedata.atlassian.net/browse/BND3D-4199)

https://cognitedata.atlassian.net/browse/BND3D-4199

## Description :pencil:
This fixes the rule threshold for large rule-based + 3d model content + asset mappings by changing the asset mapping cache system in react components and some refactoring to speed up the rule threshold applying onto the 3d models, including a loading indicator when the rule is processing, loading, and applying the styling.
Before this, the asset mapping cache was based on a list of asset IDs or node IDs that were requested at once. However, the asset IDs of the requests are not always the same, forcing the cache mechanism to request the mappings even for those cached already in a different list. 
So now the cache key is based on the asset against all its mappings rather than a summed key based on a chunk of mappings.
This way it is possible to have more freedom and flexibility to use the cache without relying on a static list.

1 - Open a 3D scene in Search UI that contains contextualized assets and rule based thresholds (recommend to test in a scene that contains a huge amount of geometries and asset mappings to see the changes like Celanese, for instance).
2 - Select an Rule Based Thresholds and switch to another rule.
3. Observe the loading indicator replacing the radio button while the rule is loading and the performance improvements compared with the previous version (not crashing after loaded, more stable).

Note: it is connected to the Fusion PR https://github.com/cognitedata/fusion/pull/7149 for working with the Rule Based loading indicator

- [X] I am proud of this feature.


react-components/src/components/RevealToolbar/RuleBasedOutputsButton.tsx
```diff
@@ -43,6 +49,10 @@ export const RuleBasedOutputsButton = ({
     setRuleInstances(ruleInstancesResult.data);
   }, [ruleInstancesResult]);
 
+  useEffect(() => {
+    setCurrentRuleSetEnabled(newRuleSetEnabled);
+  }, [newRuleSetEnabled]);
```
haakonflatval-cognite, line 54: Do we need this? Could we just use `setCurrentRuleSetEnabled` instead of `setNewRuleSetEnabled` below?

react-components/src/components/RevealToolbar/RuleBasedOutputsButton.tsx
```diff
@@ -43,6 +49,10 @@ export const RuleBasedOutputsButton = ({
...
+
   const onChange = useCallback(
     (data: string | undefined): void => {
       ruleInstances?.forEach((item) => {
```
haakonflatval-cognite, line 58: This is a big function - can we factor it out?

react-components/src/components/Reveal3DResources/types.ts
```diff
@@ -108,4 +108,5 @@ export type CommonResourceContainerProps = {
   image360Settings?: CommonImage360Settings;
   onResourcesAdded?: () => void;
   onResourceLoadError?: (failedResource: AddResourceOptions, error: any) => void;
+  onCallback?: (data?: any) => void;
```
haakonflatval-cognite, line 111: Can this have a more descriptive name?

react-components/src/components/CacheProvider/types.ts
```diff
@@ -64,3 +65,13 @@ export type Image360AnnotationAssetInfo = {
...
+
+export type ChunkInCacheTypes = {
+  chunkInCache: Array<Required<AssetMapping3D>>;
+  chunkNotInCache: number[];
+};
+
+export type ChunkInCacheTypesNode3D = {
+  chunkInCache: Node3D[];
+  chunkNotInCache: number[];
+};
```
haakonflatval-cognite, line 77: Don't have a very strong opinion, but I think I'd prefer having one type that is generic here, e.g. 
```
export type CachedAndUncachedObjects<ObjectType> {
    cachedObjects: ObjectType[],
    uncachedIds: number[]
}
```

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
```diff
@@ -62,6 +69,11 @@ export const Reveal3DResources = ({
     [reveal3DModels]
   );
 
+  const { data: assetMappings } = useAssetMappedNodesForRevisions(cadModelOptions);
+
+  void useGenerateAssetMappingCachePerItemFromModelCache(cadModelOptions, assetMappings);
+  void useGenerateNode3DCache(cadModelOptions, assetMappings);
```
haakonflatval-cognite, line 75: @pramodcog did some work recently where we fetch all asset mappings when the scene is loaded, so that we can eventually color the contextualized nodes. I assume there's some overlap between the data you fetch here, and the data we're fetching for that. Could we make sure that those two data fetching workloads are happening together?

react-components/src/components/CacheProvider/AssetMappingCacheProvider.tsx
```diff
@@ -38,6 +38,53 @@ const useAssetMappingCache = (): AssetMappingCache => {
   return content.cache;
 };
 
+export const useGenerateNode3DCache = async (
+  cadModelOptions: CadModelOptions[],
+  assetMappings: ModelWithAssetMappings[] | undefined
+): Promise<void> => {
+  const assetMappingCache = useAssetMappingCache();
+
+  useMemo(() => {
```
haakonflatval-cognite, line 47: Hooks should not be `async` you cannot `await` inside a React component. So this `useMemo` can then be a `useEffect`

react-components/src/components/CacheProvider/AssetMappingCacheProvider.tsx
```diff
@@ -38,6 +38,53 @@ const useAssetMappingCache = (): AssetMappingCache => {
...
+    });
+  }, [cadModelOptions, assetMappings]);
+};
+
+export const useGenerateAssetMappingCachePerItemFromModelCache = async (
+  cadModelOptions: CadModelOptions[],
+  assetMappings: ModelWithAssetMappings[] | undefined
+): Promise<void> => {
+  const assetMappingCache = useAssetMappingCache();
+  useMemo(() => {
```
haakonflatval-cognite, line 72: Same for this

haakonflatval-cognite (changes requested):
Some initial comments - I'll have to look a little closer on the changes in `AssetMappingCache`, but I think I'll wait until you've had a chance to make changes based on the comments

react-components/src/components/Reveal3DResources/types.ts
danpriori: I was thinking to use a more generic name for other purposes any other than rule base callback. 
Because it can be used to anything else if we isolate this component into some other context.

react-components/src/components/RuleBasedOutputs/components/RuleBasedSelectionItem.tsx
```diff
@@ -3,34 +3,47 @@
...
           {label}
         </Flex>
-        <Radio name={id ?? ''} value={id} checked={checked} />
-      </Flex>
+        {!isEmptyRuleItem && isLoading && checked === true ? (
+          <Icon type="Loader" />
+        ) : (
+          <Radio name={id ?? ''} value={id} checked={checked} />
+        )}
+      </StyledFlex>
```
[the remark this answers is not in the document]
danpriori: You mean the rule set name while loading? But it is showing while still loading. Hmmmm did you face something odd ?

react-components/src/components/RuleBasedOutputs/utils.ts
```diff
@@ -494,9 +466,11 @@ const applyNodeStyles = (
   };
 
   const nodeIndexSet = ruleOutputAndStyleIndex.styleIndex.getIndexSet();
-  treeNodes.forEach((node) => {
-    nodeIndexSet.addRange(node.subtreeRange);
+  nodeIndexSet.clear();
+  treeNodes?.forEach((node) => {
+    nodeIndexSet.add(node.treeIndex);
   });
```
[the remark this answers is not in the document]
danpriori: I was talking to @pramodcog and he recommended me to use the single tree index in this case as we will coloring the parent of the descendant nodes which is the actual node that is linked to the asset and it will not require an extra fetch from the `model.getSubtreeTreeIndices` which gets 429 mostly and slow down the process.
Do you think we should reconsider that?

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
danpriori: Yeah.. I'm gonna check with him

react-components/src/components/RevealToolbar/RuleBasedOutputsButton.tsx
danpriori: Nice observation. But this is to let the spinner start render one event cycle before triggering the process of setting the current rule set enabled on the next cycles.
If we set the flag IsRuleLoading on the same event cycle of setting the CurrentRuleSetEnabled, then the spinner will only shows up a bit later in the middle of the rule processing/when the event flow is free to trigger a new render.
Do you have any other suggestion to do that?

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
pramodcog: `useAssetMappedNodesForRevisions` is a cached mappings when scene is loaded, so it is re-used is several places, which I think is used the right way here.

react-components/src/hooks/useCalculateModelsStyling.tsx
```diff
@@ -38,6 +38,16 @@ type ModelStyleGroup = {
...
+type ModelStyleGroupWithMappingsFetched = {
+  combinedMappedStyleGroups: ModelStyleGroup[];
+  modelMappingsIsFetched: boolean;
+};
+
+type StyledModelWithMappingsFetched = {
+  styledModels: StyledModel[];
+  modelMappingsIsFetched: boolean;
+};
+
```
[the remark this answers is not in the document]
danpriori: Hmmm several hooks are involved here and only one or two are returning query result type... so I think it wont be a good fit for this case :(
Unless you want to give me an example how to use in this case.

react-components/src/components/CacheProvider/AssetMappingCacheProvider.tsx
danpriori: https://github.com/cognitedata/reveal/pull/4652/commits/01d9e4192802dde46298958dce3453c03f561cce

react-components/src/components/CacheProvider/AssetMappingCacheProvider.tsx
danpriori: https://github.com/cognitedata/reveal/pull/4652/commits/01d9e4192802dde46298958dce3453c03f561cce

react-components/src/components/CacheProvider/types.ts
danpriori: I agree... make sense.

Changed here:
https://github.com/cognitedata/reveal/pull/4652/commits/01d9e4192802dde46298958dce3453c03f561cce

react-components/src/components/Reveal3DResources/types.ts
danpriori: changed here
https://github.com/cognitedata/reveal/pull/4652/commits/01d9e4192802dde46298958dce3453c03f561cce

react-components/src/components/RuleBasedOutputs/utils.ts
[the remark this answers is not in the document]
danpriori: moved to addRange again with a way better way now :D Thanks!

https://github.com/cognitedata/reveal/pull/4652/commits/01d9e4192802dde46298958dce3453c03f561cce

react-components/src/components/CacheProvider/AssetMappingCache.ts
```diff
@@ -81,18 +86,29 @@ export class AssetMappingCache {
...
+    const assetIdsList = Array.from(relevantAssetIds);
+    const chunkSize = Math.round(assetIdsList.length / amountOfChunks);
+    const listChunks = chunk(assetIdsList, chunkSize);
+
+    const allAssetMappingsReturned = listChunks.map(async (itemChunk) => {
+      const assetMappings = await this.getAssetMappingsForAssetIds(modelId, revisionId, itemChunk);
+      return assetMappings;
+    });
+
+    const allAssetMappings = await Promise.all(allAssetMappingsReturned);
```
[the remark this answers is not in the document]
danpriori: if you see the `_amountOfAssetIdsChunks` which is 1 then it will run only one chunk in total.
If we set the _amountOfAssetIdsChunks to more than 1, then it will split the total in smaller chunks and run in parallel.

react-components/src/components/CacheProvider/AssetMappingCache.ts
```diff
@@ -13,30 +13,35 @@ import {
...
+
+  private readonly _nodeAssetIdsToNode3D = new Map<ModelNodeIdKey, Promise<Node3D>>();
 
   constructor(sdk: CogniteClient) {
     this._sdk = sdk;
   }
 
+  // PUBLIC METHODS
+
   public async getAssetMappingsForLowestAncestor(
```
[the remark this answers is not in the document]
danpriori: Hmmm yes. I like to have comments to drive the understanding ofc. But we have to agree on keeping up to date, which is not the case sometimes in most of the code bases ahahaha.
But if you prefer I can remove these comments until we decide together :).
What do you prefer?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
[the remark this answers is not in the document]
danpriori: done here

https://github.com/cognitedata/reveal/pull/4652/commits/fa47fce31bc5089a629ac5cc150a509d9659c0bd

react-components/src/components/CacheProvider/AssetMappingCache.ts
```diff
@@ -13,30 +13,35 @@ import {
...
-
 export class AssetMappingCache {
   private readonly _sdk: CogniteClient;
 
   private readonly _modelToAssetMappings = new Map<ModelRevisionKey, Promise<AssetMapping[]>>();
-  private readonly _nodeAssetIdsToAssetMappings = new Map<
-    ModelNodeIdKey,
-    Promise<AssetMapping[]>
-  >();
+  private readonly _assetIdsToAssetMappings = new Map<ModelNodeIdKey, Promise<AssetMapping[]>>();
```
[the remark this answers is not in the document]
danpriori: ops... fixed here:
https://github.com/cognitedata/reveal/pull/4652/commits/511c4959cd5a1621b15fa1b9727f7d970fce56ed

react-components/src/components/CacheProvider/AssetMappingCache.ts
```diff
@@ -13,30 +13,35 @@ import {
...
+
+  private readonly _nodeIdsToAssetMappings = new Map<ModelNodeIdKey, Promise<AssetMapping[]>>();
+
+  private readonly _nodeAssetIdsToNode3D = new Map<ModelNodeIdKey, Promise<Node3D>>();
```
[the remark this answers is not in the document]
danpriori: This cache is for node ids against node 3d. Fixed the name here:
https://github.com/cognitedata/reveal/pull/4652/commits/511c4959cd5a1621b15fa1b9727f7d970fce56ed

react-components/src/components/CacheProvider/AssetMappingCache.ts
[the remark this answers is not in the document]
haakonflatval-cognite: Perhaps we can remove them for now? I know Nils Petter has his own convention altogether, so I'd like to minimize conflicts by not introducing something new yet :sweat_smile:

react-components/src/components/CacheProvider/AssetMappingCache.ts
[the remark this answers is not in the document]
haakonflatval-cognite: But from what I can see, there is nothing stopping the code from calling `this.getAssetMappingsForAssetIds` in parallel - when using `await` inside the `async` function in `map`, I think it will go ahead and pick the next chunk in the list and call `getAssetMappingsForAssetIds` without waiting for the first one to succeed.

react-components/src/components/RuleBasedOutputs/components/RuleBasedSelectionItem.tsx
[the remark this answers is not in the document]
haakonflatval-cognite: No, I just read the code, I might be wrong :thinking:

react-components/src/components/CacheProvider/AssetMappingCache.ts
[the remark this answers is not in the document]
danpriori: Yes. But only if we change the _amountOfAssetIdsChunks  value. For now with the value being 1, it will request just once.
My thought was to provide this flexibility to change in the future when we have a more robust backend to support more parallel requests. Do you think it is better to have a fixed unique request without this logic that splits in chunks and only change to this current approach when we have sure that the backend is improved?

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
```diff
@@ -62,6 +69,11 @@ export const Reveal3DResources = ({
     [reveal3DModels]
   );
 
+  const { data: assetMappings } = useAssetMappedNodesForRevisions(cadModelOptions);
+
+  useGenerateAssetMappingCachePerItemFromModelCache(cadModelOptions, assetMappings);
+  useGenerateNode3DCache(cadModelOptions, assetMappings);
+
```
haakonflatval-cognite, line 76: I don't think we should do this here. I think updating the cache should happen from within the cache itself, and not relying on these hooks getting called with the output from the cache. Sorry for not catching this before :see_no_evil:

haakonflatval-cognite:
I notice that clicking a rule that is already loading will put it in the "loading"-state, but it won't reset it back, so it's forever spinning :sweat_smile:

danpriori:
> I notice that clicking a rule that is already loading will put it in the "loading"-state, but it won't reset it back, so it's forever spinning 😅

Yeah. I saw this. I'm working on this with the Provider approach

react-components/src/components/CacheProvider/AssetMappingCache.ts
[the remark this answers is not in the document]
danpriori: Hmmm yeah. I can do that :)

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
danpriori: Hmmm if I understood your thoughts correctly, actually these two hooks are here to cache the asset mappings per item based on asset ids and node ids. The `assetMappings` as the output from the hook `const { data: assetMappings } = useAssetMappedNodesForRevisions(cadModelOptions);` above those two hooks are from the model cache which is used in different places in the app. 
The cache per item is used by rule based and any other components that doesnt need all the asset mappings per model at once and in the same order, which is a custom list of asset mappings everytime.

react-components/src/components/CacheProvider/AssetMappingCache.ts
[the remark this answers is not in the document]
danpriori: some improvements here:

https://github.com/cognitedata/reveal/pull/4652/commits/f0282f28241e5ec2563ad8aeb7bf49082e5be6cb

ofc we can improve it more

react-components/src/components/CacheProvider/AssetMappingAndNode3DCache.ts
haakonflatval-cognite, line 1: Nice! Looks a lot better when it's refactored! I was wondering if we could also abstract away the Asset-Mapping fetching to another class, like you already did for the node3D-fetching?

react-components/src/components/CacheProvider/Node3DPerNodeIdCache.ts
```diff
@@ -0,0 +1,81 @@
...
+    nodeIds: number[] | undefined
+  ): Promise<void> {
+    const node3Ds = await this.getNodesForNodeIds(modelId, revisionId, nodeIds ?? []);
+    node3Ds.forEach((node) => {
+      const key = modelRevisionNodesAssetsToKey(modelId, revisionId, [node.id]);
+      this.setNodeIdToNode3DCacheItem(key, Promise.resolve(node));
+    });
+  }
+
+  public async getNodesForNodeIds(
```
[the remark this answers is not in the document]
danpriori: This one is also used by the `getNodesForAssetIds` in here:

```typescript
const nodes = await this.nodeIdsToNode3DCache.getNodesForNodeIds(
      modelId,
      revisionId,
      relevantAssetMappings.map((assetMapping) => assetMapping.nodeId)
    );
``` 
Which was obtained from the asset mappings

react-components/src/components/CacheProvider/Node3DPerNodeIdCache.ts
```diff
@@ -0,0 +1,81 @@
...
+export class Node3DPerNodeIdCache {
+  private readonly _sdk: CogniteClient;
+
+  private readonly _nodeIdsToNode3D = new Map<ModelNodeIdKey, Promise<Node3D>>();
+
+  constructor(sdk: CogniteClient) {
+    this._sdk = sdk;
+  }
+
+  public async splitChunkInCacheNode3D(
```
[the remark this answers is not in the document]
danpriori: ops :D 
https://github.com/cognitedata/reveal/pull/4652/commits/a123bf92ee2143c026d521f271187d2bc5d0f472

react-components/src/components/CacheProvider/AssetMappingAndNode3DCache.ts
danpriori: Hmmm... some fetch methods are for both: asset and node id caches. 
For example:
`fetchAndCacheMappingsForIds`
`fetchMappingsInQueue`
`fetchAssetMappingsRequest`

So if we abstract these fetch methods away, then we probably will overcomplicate these classes... we're forced to create another abstraction layer in the main class and ending up having a duplicated code in both sub classes instead of reusing the same methods for both sub classes within the main class. 

Or do you have a suggestion to architect this differently?

danpriori: I included some more refactoring here:

https://github.com/cognitedata/reveal/pull/4652/commits/c46f3cbc984f32c736b18fa4e905de241a92c154

react-components/src/components/Reveal3DResources/Reveal3DResources.tsx
danpriori: After testing moving those two hooks into the main hook `useAssetMappedNodesForRevisions` to fetch the asset mappings, the Search UI start to crash and I think it is because the same main hook is called in many other components and it looks like they are trying to fetch the same items in parallel as they are not cached by the useMemo yet. So I'm thinking some alternatives but for now I believe it is safer to keep these two hooks to cache the asset mappings and nodes per item in the Reveal3DResource only or remove it entirely and them we can rethink carefully a better way to accomodate these cache hooks into the same `useAssetMappedNodesForRevisions`

haakonflatval-cognite (approved):
LGTM :+1:

react-components/src/hooks/useConvertAssetMetadatasToLowerCase.tsx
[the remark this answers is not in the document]
danpriori: yes. It has a sub folder hook inside the rule based outputs. I'm gonna put it there. Good point

react-components/src/components/CacheProvider/AssetMappingAndNode3DCache.ts
```diff
@@ -0,0 +1,363 @@
...
+    assetMapping3D.forEach(async (item) => {
+      const key: ModelNodeIdKey = modelRevisionNodesAssetsToKey(modelId, revisionId, [
+        item.assetId
+      ]);
+      if (filterType === 'nodeIds') {
+        await this.nodeIdsToAssetMappingCache.setAssetMappingsCacheItem(key, item);
+      } else {
+        await this.assetIdsToAssetMappingCache.setAssetMappingsCacheItem(key, item);
+      }
+    });
```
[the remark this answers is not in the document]
danpriori: https://github.com/cognitedata/reveal/pull/4652/commits/2e9ec23a6792163272406b31b114111990b11ac2

react-components/src/hooks/useConvertAssetMetadatasToLowerCase.tsx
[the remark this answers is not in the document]
danpriori: https://github.com/cognitedata/reveal/pull/4652/commits/58a592e277e99b52a7894b334f7da652b7e65f8d

PR #4654, opened 2024-07-02
chore(react-components): cogs v10

haakonflatval-cognite:
#### Type of change

Chore 

#### Jira ticket :blue_book:

## Description :pencil:

To build locally - add
```
  "resolutions": {
    "@mui/base": "5.0.0-beta.58"
  }
```
at the end of the object in `react-components/package.json`. This fixes a bug inherent in an earlier `@mui/base` beta version. I've reached out to the cogs-team to fix it https://cognitedata.slack.com/archives/C011E10CW2F/p1727771787512549.

List of biggest changes:
- Use `Menu` component from `cogs-lab` instead of the old `Dropdown` + `Menu` component combination from `cogs.js`. The syntax is somewhat similar, but now the `Menu` is the wrapping component, and the `renderTrigger` contains the button that opens the menu.
- Rewrite the layer handling code to use `SelectPanel`
- Introduce `IconName` as an alias to internal "known" icons. External users can add their own icon names through the `IconComponentMapper`, which knows which icon name corresponds to which icon. However, any actual external use of that probably lies a bit into the future.

#### The following is a worklog I used for venting throughout the work (not relevant anymore)

Migration to Cogs v10. 

Biggest change is that Icons are now provided as separate icons, instead of one icon taking the name as input. I've chosen to solve this by having a function translating from `IconName` (new type) to icon component. One advantage of doing this instead of e.g. returning the icon directly from each domain object / tool, is that we remain framework-agnostic in the architecture.

Will mark as "ready" (not draft) when cogs releases a version that correctly includes the `@tippy/react` and `core-js` dependencies, so we don't have to put them in our `package.json`

A lot of the functionality is working fine now. Still waiting for cogs' PR.

There are some janky behavior associated with some of the older buttons. In particular `SettingsButton` and `HelpButton` (somewhat). We should have a plan for these before merging this

Also, OptionButton is not properly migrated. May not be necessary before we merge this.

Update - Behavior in `HelpButton` is pretty good, although button itself still shows as "toggled" even when closing menu.
`SettingsButton` works pretty well, still closing panel on toggling options, though
`SlicerButton` does not close on clicking the button again after activation.

A good variety of trouble, in other words

--- 

30.08.24

We're mostly good. Remaining work (as far as I've been able to test today):
- Help content does not stop expanding, ending up too wide
- Rule-based coloring not working as it should (items not reacting to click?) and does not show color palette icon
- Filter list does not show color square alongside items

Good luck, future Håkon


christjt (approved):

react-components/src/architecture/base/utilities/IconName.ts
```diff
@@ -0,0 +1,39 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+
+export type IconName =
```
nilscognite, line 5: This don't scale. Assume you defined a command outside react-component, and you are stuck

nilscognite (changes requested):
We must find a better solution for the icon

react-components/src/components/Architecture/DomainObjectPanel.tsx
```diff
@@ -89,14 +91,17 @@ export const DomainObjectPanel = (): ReactElement => {
   );
 
   function addTextWithNumber(item: NumberPanelItem, unitSystem: UnitSystem): ReactElement {
-    const icon = item.icon as IconType;
+    const icon = item.icon;
     const { quantity, value } = item;
     const text = item?.getText(t);
+    if (iconName === undefined) {
```
pramodcog, line 97: if condition should be above other statement here as it return empty component

react-components/src/components/Architecture/FilterButton.tsx
```diff
@@ -101,61 +93,73 @@ export const FilterButton = ({
...
+      renderTrigger={(props: any) => (
+        <CogsTooltip
+          content={<LabelWithShortcut label={label} command={command} />}
+          disabled={usedInSettings || label === undefined}
+          appendTo={document.body}
+          placement={placement}>
+          <Button
+            type={usedInSettings ? 'tertiary' : getButtonType(command)}
+            icon={
+              usedInSettings ? (
```
pramodcog, line 120: would like this to be declaration outside as a function

react-components/src/components/Architecture/FilterButton.tsx
```diff
@@ -101,61 +93,73 @@ export const FilterButton = ({
...
-              overflow: 'auto',
-              flexDirection
+              minWidth: usedInSettings ? OPTION_MIN_WIDTH : undefined,
+              padding: usedInSettings ? DEFAULT_PADDING : undefined
+            }}
+            {...props}
+            onClick={(event: MouseEvent<HTMLElement>) => {
+              props.onClick(event);
+              event.stopPropagation();
+              event.preventDefault();
```
pramodcog, line 143: Ideally all event propagation prevent should happen at the top, that before` props.onClick() `is called

react-components/src/components/Architecture/useClickOutside.tsx
```diff
@@ -1,24 +0,0 @@
-/*!
```
pramodcog, line 1: Is click outside now handled by cogs library?

haakonflatval-cognite: Yes, it is :+1:

react-components/src/components/Architecture/FilterButton.tsx
haakonflatval-cognite: Hmm, okay :thinking: Is there a reason for that?

pramodcog (approved):
LGTM!

PR #4655, opened 2024-07-02
fix(react-components):  asset search hook to support more than 1000 results

pramodcog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4459

## Description :pencil:
Existing search hook support 1000 asset as the end point `asset.search` supports only 1000, in this PR I have changed the way searching of asset work. I have added support for infinite asset search using `asset.list` utilizing advanced filter option.

In storybook and fusion

- Load `SearchHooks` storybook example
- Play around with `All asset mappings`, `Asset search hook` and `Load more`


react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -27,53 +28,92 @@ export type ModelMappingsWithAssets = ModelMappings & {
...
+      }
+      const assetsResponse = await getAssetsList(sdk, {
+        query,
         limit,
-        searchedAssets.map((asset) => asset.id)
-      );
+        cursor: pageParam
+      });
+
+      const assets = assetsResponse.items.filter((asset) => asset !== undefined);
```
haakonflatval-cognite, line 84: Can the asset response contain undefined items? In any case, can we use `filter(isDefined)` here?

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -27,53 +28,92 @@ export type ModelMappingsWithAssets = ModelMappings & {
...
 
-      const assetMappingsSet = createAssetMappingsSet(assetMappingsWithSearch);
-      const filteredSearchedAssets = searchedAssets.filter((asset) =>
-        assetMappingsSet.has(asset.id)
-      );
+      const assetMappingsWithSearch = assetMappingList.flatMap((mapping) => {
+        return mapping.assetMappings.filter((assetMapping) =>
+          assets.map((asset) => asset.id).includes(assetMapping.assetId)
```
haakonflatval-cognite, line 88: I think you can do `assets.some(asset => asset.id === assetMapping.assetId)`

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -27,53 +28,92 @@ export type ModelMappingsWithAssets = ModelMappings & {
...
   userSdk?: CogniteClient
-): UseQueryResult<Asset[]> => {
+): UseInfiniteQueryResult<InfiniteData<AssetPage[]>, Error> => {
   const sdk = useSDK(userSdk);
-  const modelsKey = models.map((model) => [model.modelId, model.revisionId]);
-  const { data: assetMappings, isFetched } = useAllMappedEquipmentAssetMappings(models, sdk);
+  const { data: assetMappingList, isFetched } = useAssetMappedNodesForRevisions(
+    models.map((model) => ({ ...model, type: 'cad' }))
+  );
+  const allAssetMappings = useAllMappedEquipmentAssetMappings(models, sdk);
```
haakonflatval-cognite, line 51: What is the different between these two calls? It looks like they fetch the same data?

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -27,53 +28,92 @@ export type ModelMappingsWithAssets = ModelMappings & {
...
-      );
+      const assetMappingsWithSearch = assetMappingList.flatMap((mapping) => {
+        return mapping.assetMappings.filter((assetMapping) =>
+          assets.map((asset) => asset.id).includes(assetMapping.assetId)
+        );
+      });
 
-      return filteredSearchedAssets;
+      const assetMappingsSet = new Set(assetMappingsWithSearch.map((mapping) => mapping.assetId));
+      const filteredSearchedAssets = assets.filter((asset) => assetMappingsSet.has(asset.id));
```
haakonflatval-cognite, line 93: I don't understand what's happening here :see_no_evil: We first get all asset mappings for the revisions (at the top of this hook), and then we find the overlap between the fetched asset mappings and the fetched assets (with an O(n^2) algorithm), and then we construct a set from that overlap and finally filter the fetched assets based on that set. It seems that this set could be created from the `assetMappingList` up-front, and the assets can be filtered against it on one line here?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -106,19 +118,16 @@ async function initializeRuleBasedOutputs({
   timeseriesDatapoints
 }: {
   model: CogniteCadModel;
-  assetMappings: InfiniteData<ModelMappingsWithAssets[]>;
+  assetMappings: AssetMapping3D[];
```
haakonflatval-cognite, line 121: Just to be sure, I'd like @danpriori to approve that this is okay :sweat_smile:

haakonflatval-cognite (dismissed):
Looks nice, will spend some time testing after you've had a chance to go through the review :+1:

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: Hmmm I think all of these changes will affect the changes I've made in the PR

Among a couple of changes, I already removed the request for asset mappings from the hook 
`useAllMappedEquipmentAssetMappings`. I changed to the `useAssetMappedNodesForRevisions` and with that a couple of other changes were made and affected by this and other improvements for caching with rule based. And it is not using the pagination as is getting from that hook now.

Although some of these changes in this PR are aligned with the changes I made in the Rule Based PR.

See it here: https://github.com/cognitedata/reveal/pull/4652/files#

So what you guys think is it better to do then?

pramodcog: This is not affecting any changes to rule based coloring

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
pramodcog: kind of true, but with a catch 😄. `allAssetMappings` will contain only first 1000 mappings which will be displayed/listed which will be quick to display to the user, but when search query key is passed than `assetMappingList` (asset cache data) value will be used.

react-components/src/utilities/buildFilter.ts
```diff
@@ -0,0 +1,33 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+
+export const buildFilter = (query: string): any => {
```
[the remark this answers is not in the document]
pramodcog: Could not find a type in SDK

react-components/src/hooks/network/getAssetsList.ts
```diff
@@ -0,0 +1,44 @@
...
+import { type CogniteClient, type CursorResponse, type Asset } from '@cognite/sdk';
+import { buildFilter } from '../../utilities/buildFilter';
+
+export const getAssetsList = async (
+  sdk: CogniteClient,
+  {
+    query,
+    cursor,
+    limit = 1000,
+    sort = [{ property: ['_score_'] }],
```
danpriori, line 13: I dont have a strong opinion on this, but is it possible to put it as a const and a type?

react-components/src/hooks/network/getAssetsList.ts
```diff
@@ -0,0 +1,44 @@
...
+    cursor?: string;
+    limit?: number;
+    sort?: Array<{ property: string[] }>;
+    aggregatedProperties?: string[];
+  }
+): Promise<{ items: Asset[]; nextCursor: string | undefined }> => {
+  const advancedFilter = buildFilter(query);
+
+  return await sdk
+    .post<CursorResponse<Asset[]>>(`/api/v1/projects/${sdk.project}/assets/list`, {
```
danpriori, line 26: Can we use the sdk version of it?
`sdk.assets.list()`

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
danpriori: Maybe then is it not better then to change the name of the const to be more understandable based on your description?
Perhaps instead of `const allAssetMappings` then we could have something like
`const initialAssetMappings` if it does make sense, ofc :)

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -100,21 +138,46 @@ export const useAllMappedEquipmentAssetMappings = (
...
+      });
     },
     initialPageParam: models.map((model) => ({ cursor: 'start', model })),
     staleTime: Infinity,
-    getNextPageParam
+    getNextPageParam: (lastPage: {
+      modelsAssets: ModelMappingsWithAssets[];
+      nextCursors: string[];
+    }): Array<{ cursor: string | undefined; model: AddModelOptions }> | undefined => {
+      const nextCursors = lastPage.nextCursors
```
danpriori, line 167: I would prefer to be more explicit in var names. But this is more a personal taste :) 

For example: as this const is holding not only the cursor but also the model props, maybe the name could be something like: `nextCursorAndModelOptions`
Also, the returning I think could be a type as well:
`NextCursorAndModelOptions[]  | undefined`

react-components/src/utilities/queryKeys.ts
```diff
@@ -7,6 +7,7 @@ export const queryKeys = {
   all: ['cdf'] as const,
   // ASSETS
   assetsById: (ids: IdEither[]) => [...assets, ids] as const,
+  assetsList: (filter?: any) => [...queryKeys.all, 'assets', 'list', filter],
```
danpriori, line 10: Are we using this queryKey?
If not, do we need to add this now for future usage?

react-components/src/hooks/network/getAssetsList.ts
pramodcog: we are using advanced filter which I could not find the support with sdk.assets.list()

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
pramodcog: return `type` should match the `query` return type, so there is not much we can do to change it.

react-components/src/utilities/queryKeys.ts
pramodcog: Overhead code 😄

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
danpriori: Hmmm yeah. I see

react-components/src/hooks/network/getAssetsList.ts
danpriori: Oh yeah.. right

danpriori (approved):
LGTM

PR #4661, opened 2024-07-11
refactor: simplify isClipped

haakonflatval-cognite:
#### Type of change

Refactor 

## Description :pencil:

Make use of the general `isClipped` method in the point cloud rendering too. 

This was done in an effort to support clipping in 3D overlays, but it seems to crash into some further complications...


viewer/packages/rendering/src/glsl/base/isClipped.glsl
```diff
@@ -2,7 +2,7 @@
 uniform vec4 clippingPlanes[NUM_CLIPPING_PLANES];
 #endif
 
-bool isClipped(NodeAppearance nodeAppearance, vec3 point) {
+bool isClipped(vec3 point) {
 #if NUM_CLIPPING_PLANES > 0
   vec3 pointFlipped = -point;
   vec4 plane;
```
nilscognite, line 8: The lines above sucks in my eyes! Why not simplify with early return and remove clipped? Cannot understand why this code has been approved.

nilscognite (changes requested):
Had a comment in the unchanged code. Otherwise good.

viewer/packages/rendering/src/glsl/base/isClipped.glsl
haakonflatval-cognite: Hehe, not sure why.. It could be that they were afraid that shader threads running in parallel would diverge (one returns early, another one doesn't), which could cause some unfortunate performance loss, but not sure if that's a real concern or not :sweat_smile:

nilscognite (approved):

PR #4664, opened 2024-07-12
bug: fix unsafe usage of static variables

nabati:
## Description :pencil:

TLDR: We're using unsafe static variables to keep state. The issue is that if there are multiple instances of the viewer, then they will end up overwriting each other and we get bugs. 

Threads: 
https://cognitedata.slack.com/archives/C8A0QGMLN/p1720636595411859
https://cognitedata.slack.com/archives/C8A0QGMLN/p1720612784051089

I guess this would be a breaking change FYI. 

Locally nothing breaks. Could use some help with validating this though.


viewer/packages/camera-manager/src/CameraManagerHelper.ts
```diff
@@ -76,9 +55,16 @@ export class CameraManagerHelper {
...
    */
-  static updateCameraNearAndFar(camera: PerspectiveCamera, boundingBox: Box3): void {
-    const { cameraPosition, cameraDirection, corners } = this._updateNearAndFarPlaneBuffers;
+  static updateCameraNearAndFar(
+    camera: PerspectiveCamera,
+    boundingBox: Box3,
+    nearAndFarPlaneBuffers: NearAndFarPlaneBuffers,
+    cameraFarBuffers: CameraFarBuffers
+  ): void {
+    const { cameraPosition, cameraDirection, corners } = nearAndFarPlaneBuffers;
```
haakonflatval-cognite, line 67: This update breaks the API... :thinking: Is there some other way we can do this?

viewer/packages/camera-manager/src/types.ts
```diff
@@ -106,3 +107,14 @@ export const CAMERA_MANAGER_EVENT_TYPE_LIST = ['cameraChange', 'cameraStop'] as
  * Union type of the supported camera manager event types
  */
 export type CameraManagerEventType = (typeof CAMERA_MANAGER_EVENT_TYPE_LIST)[number];
+
+export type NearAndFarPlaneBuffers = {
+  corners: Vector3[];
+  cameraPosition: Vector3;
+  cameraDirection: Vector3;
+};
```
haakonflatval-cognite, line 115: I don't think `cameraPosition` and `cameraDirection` need to be part of this type - can't they be derived from the input camera in `CameraManagerHelper`?

haakonflatval-cognite: The reason I'm pedantic on this, is that this type (and the one below) needs to be added to Reveal's API, which I don't want to bloat with many small types

viewer/packages/camera-manager/src/CameraManagerHelper.ts
nabati: "Getting away" with it doesn't make sense IMHO, since this function would then be a noop since it's writing the results to the input parameters.

viewer/packages/camera-manager/src/types.ts
nabati: Not sure what change you want to see here. Could you clarify? 

Would you prefer to see the parameters spelled out? 

aka

```
static updateCameraNearAndFar(
    camera: PerspectiveCamera,
    boundingBox: Box3,
    cameraPosition: Vector3,
    cameraDirection: Vector3,
    corners: Vector3[]
    ...
  )
```

viewer/packages/camera-manager/src/CameraManagerHelper.ts
haakonflatval-cognite: My suggestion would be to let the function write to the static fields as before if the buffers were not supplied

haakonflatval-cognite: Actually - what if we make a non-static version of this method, with corresponding non-static fields inside CameraManagerHelper? I know it's probably intended to be a static class, but I think it's relatively safe to repurpose it a little. We would still need to keep the static fields/method though, but that would keep the usage simple, without need for extra book-keeping by every camera manager using it

viewer/packages/camera-manager/src/types.ts
haakonflatval-cognite, line 115: I think this type and the one below can be moved into `CameraManagerHelper.ts`, since they're only being used there

viewer/packages/camera-manager/src/CameraManagerHelper.ts
```diff
@@ -76,9 +103,16 @@ export class CameraManagerHelper {
...
    */
-  static updateCameraNearAndFar(camera: PerspectiveCamera, boundingBox: Box3): void {
-    const { cameraPosition, cameraDirection, corners } = this._updateNearAndFarPlaneBuffers;
+  static updateCameraNearAndFar(
+    camera: PerspectiveCamera,
+    boundingBox: Box3,
+    nearAndFarPlaneBuffers: NearAndFarPlaneBuffers = CameraManagerHelper._updateNearAndFarPlaneBuffers,
+    cameraFarBuffers: CameraFarBuffers = CameraManagerHelper._calculateCameraFarBuffers
+  ): void {
+    const { cameraPosition, cameraDirection, corners } = nearAndFarPlaneBuffers;
```
haakonflatval-cognite, line 115: I think we can remove these extra parameters, and deprecate this function to suggest that users always use the instance-methods instead.

examples/src/utils/CustomCameraManager.ts
```diff
@@ -20,6 +21,29 @@ export class CustomCameraManager implements CameraManager {
...
+      new Vector3(),
+      new Vector3(),
+      new Vector3()
+    )
+  };
+
+  private readonly _calculateCameraFarBuffers = {
+    nearPlaneCoplanarPoint: new Vector3(),
+    nearPlane: new Plane()
+  };
```
haakonflatval-cognite, line 45: Can we use a CameraManagerHelper instance here as well, instead of these big field declarations?

haakonflatval-cognite (changes requested):
Looks good! A few comments

examples/src/utils/CustomCameraManager.ts
nabati: My bad, fixed.

PR #4667, opened 2024-07-17
fix(react-components): Remove fitView when crop the slicing planes + Rename 2 directories

nilscognite:
* Rename the directories so they are aligned to the rest (dropping the domainObject suffix)
* Remove fitView when crop the slicing planes as @haakonflatval-cognite pointed out
* Remove unnecessary key in LabelWithShortcut (was done inside LabelWithShortcut anywhy)


haakonflatval-cognite (approved):
LGTM :+1:

PR #4695, opened 2024-08-06
fix(react-components): Improve Scene home camera button and bump to 0.55.2

anders-hopland:
#### Type of change

Feat  

## Description :pencil:
Improves Scene home camera position.
* If default camera position is not provided, use a default if no models are present, of not, fit to models
* Fixes bug where rotation was not calculated correctly when rotating camera instead of providing camera target

Have built the package locally and tested it in Fusion

Build locally and use in Fusion. Verify that Scene home camera is correct

### DIFF
--- a/react-components/src/hooks/useSceneDefaultCamera.tsx
+++ b/react-components/src/hooks/useSceneDefaultCamera.tsx
@@ -47,6 +47,8 @@ export const useSceneDefaultCamera = (
       return {
         fitCameraToSceneDefault: () => {
           if (viewer.models.length === 0) {
+            // If no models are loaded, set a default camera position
+            // This is the same default position that have been used in SceneBuilder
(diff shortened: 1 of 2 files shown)


react-components/src/hooks/useSceneDefaultCamera.tsx
```diff
@@ -32,13 +32,42 @@ export const useSceneDefaultCamera = (
       };
     }
 
+    const cameraNotSet =
```
greenbech, line 35: Boolean values should be prefix with `is`, `should` or `has` in my opinion

greenbech: Usually, I would also prefer them to be positive. In this case, I would go for `isCameraSet`

greenbech (approved):
This seems to work!

PR #4702, opened 2024-08-13
fix(react-components): various small fixes and bump react-components to 0.55.4

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:

Fix a couple of small bugs in the code:
- `RevisionFdmNodeCache.ts`: remove `assert` and instead ignore invalid values to make code more resilient (I've seen this `assert` trigger a couple of times in the wild)
- Make sure mapped-equipment-query has `undefined` filter, if the incoming filter is empty
- Don't send "asset ID Search" request if there are no assets. Saves a query which would have an empty response

In the Fusion UI

### DIFF
--- a/react-components/src/components/CacheProvider/RevisionFdmNodeCache.ts
+++ b/react-components/src/components/CacheProvider/RevisionFdmNodeCache.ts
@@ -237,16 +237,18 @@ export class RevisionFdmNodeCache {
     mappingConnections: FdmCadConnection[],
     nodes: Node3D[]
   ): Array<{ connection: FdmCadConnection; treeIndex: TreeIndex }> {
-    return mappingConnections.map((connection) => {
+    return mappingConnections.reduce((acc, connection) => {
       const nodeInConnection = nodes.find((node) => node.id === connection.nodeId);
 
-      assert(nodeInConnection !== undefined);
+      if (nodeInConnection !== undefined) {
+        acc.push({
+          connection,
+          treeIndex: nodeInConnection.treeIndex
+        });
+      }
 
-      return {
-        connection,
-        treeIndex: nodeInConnection.treeIndex
-      };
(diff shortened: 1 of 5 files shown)


react-components/src/components/CacheProvider/RevisionFdmNodeCache.ts
```diff
@@ -237,16 +237,18 @@ export class RevisionFdmNodeCache {
...
+        });
+      }
 
-      return {
-        connection,
-        treeIndex: nodeInConnection.treeIndex
-      };
-    });
+      return acc;
+    }, new Array<{ connection: FdmCadConnection; treeIndex: TreeIndex }>());
```
danpriori, line 251: I Like this change 👍

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -208,7 +208,7 @@ export const useMappingsForAssetIds = (
             nextCursor.model.revisionId === model.revisionId
         )?.cursor;
 
-        if (nextCursor === undefined) {
+        if (nextCursor === undefined || assetIds.length === 0) {
```
danpriori, line 211: useInfiniteQuery has no "enabled" option, right? Otherwise we could use it if the asset Ids is empty.
But I guess it doesnt have it, right?

haakonflatval-cognite: I think the code here is a better way to handle the situation, rather than disabling the query altogether. If the query is disabled, I think the result would be `undefined`, which we would then have to deal with as a special case somewhere else. It feels nicer to just return an empty list here as the query result, in my opinion

PR #4703, opened 2024-08-16
fix(react-components): Fix PointCloudFilterCommand because it didn't show up in Search

nilscognite:
#### Type of change

Bug 

## Description :pencil:

I have changed the way to get the point cloud only.  The new way is more secure for not having bugs.

### DIFF
--- a/react-components/src/architecture/base/concreteCommands/PointCloudFilterCommand.ts
+++ b/react-components/src/architecture/base/concreteCommands/PointCloudFilterCommand.ts
@@ -10,6 +10,11 @@ import { CommandsUpdater } from '../reactUpdaters/CommandsUpdater';
 import { type RevealRenderTarget } from '../renderTarget/RevealRenderTarget';
 
 export class PointCloudFilterCommand extends BaseFilterCommand {
+  // ==================================================
+  // INSTANCE FIELDS
+  // ==================================================
+
+  private _modelId: number | undefined = undefined;
   private _revisionId: number | undefined = undefined;
 
   // ==================================================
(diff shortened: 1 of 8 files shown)


react-components/src/architecture/base/concreteCommands/PointCloudFilterCommand.ts
```diff
@@ -82,19 +85,39 @@ export class PointCloudFilterCommand extends BaseFilterCommand {
...
+  private getPointCloud(): CognitePointCloudModel | undefined {
+    if (this._modelId === undefined || this._revisionId === undefined) {
+      return undefined;
+    }
+    for (const pointCloud of this.renderTarget.getPointClouds()) {
+      if (this._modelId === pointCloud.modelId && this._revisionId === pointCloud.revisionId) {
+        return pointCloud;
+      }
+    }
+    return undefined;
```
haakonflatval-cognite, line 102: This function could be rewritten as `return this.renderTarget.getPointClouds().find(pointCloud => pointCloud.modelId === this._modelId && pointCloud.revisionId === this._revisionId)`. It returns `undefined` if no applicable result is found

react-components/src/architecture/base/concreteCommands/PointCloudFilterCommand.ts
```diff
@@ -128,14 +151,27 @@ class FilterItemCommand extends BaseFilterItemCommand {
...
+  // INSTANCE METHODS
+  // ==================================================
+
+  private getPointCloud(): CognitePointCloudModel | undefined {
+    for (const pointCloud of this.renderTarget.getPointClouds()) {
+      if (this._modelId === pointCloud.modelId && this._revisionId === pointCloud.revisionId) {
+        return pointCloud;
+      }
+    }
+    return undefined;
```
haakonflatval-cognite, line 173: This could also be written with `find`

haakonflatval-cognite (approved):
LGTM :+1: Even though I'm not sure what exactly is making this work contra the previous solution :sweat_smile: Also, a couple of suggestions

PR #4708, opened 2024-08-19
feat(react-components): CoreDM queries

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4541

## Description :pencil:

Now - this was originally just a PR to implement the CoreDM-part of the Fdm3dDataProvider. However, a few cascading effects occured, making this monstrosity of a PR. I'll go through the relevant changes:

- Introduce typed queries. This is adapted directly from @christjt 's implementation in fusion: https://github.com/cognitedata/fusion/blob/master/libs/3d/src/api/datamodeling/sdk-extensions/queryNodesAndEdges.ts
  - This of course had a series of effects, due to requiring more explicitly typed calls. This especially lead to a long list of necessary changes in the Scenes-implementation, where I ended up doing a major refactoring to make the code more maintainable and usable type-wise.
- The `FdmCadConnection` type no longer uses nodeIds, but treeIndex. This type is used as a common grounds for an "asset mapping" for the legacy FDM-API and the new CoreDM. Since CoreDM does not really have a concept of nodeId, I decided to change it to `treeIndex` for future proofing. This may have performance implications for the legacy code, as I was forced to make another query translating between nodeIds and treeIndices. But all the better for CoreDM.
- Along with this, the caches now use treeIndex to as cache key, instead of nodeId, which also lead to a few changes.

I'll do my best to delegate the review across multiple team members

Remaining work: 
- ~~Implement remaining functions (3-4 of them?)~~
- ~~Have proper data~~
- ~~Test it all~~

In Fusion using this PR: https://github.com/cognitedata/fusion/pull/7854

- Open in Fusion using PR mentioned above, enter project 3d-dev-aws
- In ThreeDSplitView.tsx (in fusion) and in ThreeDWidget, make sure that the `RevealContext` has the `useCoreDm` prop set to `true`
- See that all 3D functionality (listing, search, clicking in list, clicking in 3D, opening instance page, show mappings etc.) works


react-components/src/architecture/concrete/observations/network.ts
```diff
@@ -5,18 +5,15 @@ import { chunk } from 'lodash';
...
 import { type ObservationFdmNode, OBSERVATION_SOURCE, type ObservationProperties } from './models';
 
 import { v4 as uuid } from 'uuid';
 
 export async function fetchObservations(fdmSdk: FdmSDK): Promise<ObservationFdmNode[]> {
-  const observationsFilter: InstanceFilter = {};
-
   const observationResult = await fdmSdk.filterAllInstances<ObservationProperties>(
-    observationsFilter,
+    undefined,
```
danpriori, line 16: I was wondering if we can later change the argument list to an object to avoid "forced" undefineds as an argument.
But this is more my opinion on cosmetic changes :D

react-components/src/components/CacheProvider/AssetMappingPerNodeIdCache.ts
```diff
@@ -2,26 +2,29 @@
...
+import { type ModelTreeIndexKey } from './types';
 import { type AssetMapping } from './AssetMappingAndNode3DCache';
 
 export class AssetMappingPerNodeIdCache {
-  private readonly _nodeIdsToAssetMappings = new Map<ModelNodeIdKey, Promise<AssetMapping[]>>();
+  private readonly _nodeIdsToAssetMappings = new Map<ModelTreeIndexKey, Promise<AssetMapping[]>>();
 
   public setNodeIdsToAssetMappingCacheItem(
-    key: ModelNodeIdKey,
+    key: ModelTreeIndexKey,
```
danpriori, line 12: I like that we are moving to treeindex and abandoning nodeId going forward!

react-components/src/components/CacheProvider/FdmNodeCache.ts
```diff
@@ -339,20 +337,20 @@ async function createRevisionToConnectionsMap(
 
 function createFdmConnectionWithNode(
   modelRevisionId: ModelRevisionId,
-  modelNodeIdToNodeMap: Map<ModelNodeIdKey, Node3D>,
+  modelTreeIndexToNodeMap: Map<ModelTreeIndexKey, Node3D>,
   connection: FdmCadConnection,
-  view?: Source
+  views?: Source[]
```
danpriori, line 342: Is there any reason to pass the views here even not really used inside the function? Couldnt be just integrated into the object outside where is called?

react-components/src/components/CacheProvider/RevisionFdmNodeCache.ts
```diff
@@ -245,7 +243,7 @@ export class RevisionFdmNodeCache {
     nodes: Node3D[]
   ): Array<{ connection: FdmCadConnection; treeIndex: TreeIndex }> {
     return mappingConnections.reduce((acc, connection) => {
-      const nodeInConnection = nodes.find((node) => node.id === connection.nodeId);
+      const nodeInConnection = nodes.find((node) => node.treeIndex === connection.treeIndex);
```
danpriori, line 246: As this mappingConnection is potentially large and we are more considering performance changes going forward, I would suggest to replace the `reduce ` to some regular for loop: default for loop, for-of, optimized-for, forEach or while.

react-components/src/components/CacheProvider/requests.ts
```diff
@@ -54,28 +55,59 @@ export async function inspectNodes(
...
+
+  return nodeIds;
+}
+
+export async function nodeIdsToTreeIndices(
+  modelId: ModelId,
+  revisionId: RevisionId,
+  nodeIds: NodeId[],
+  cogniteClient: CogniteClient
+): Promise<NodeId[]> {
```
danpriori, line 93: As we are returning treeindices, can we use the TreeIndex type instead?

react-components/src/data-providers/FdmSDK.ts
```diff
@@ -529,16 +530,32 @@ export class FdmSDK {
...
-    throw new Error(`Failed to fetch instances. Status: ${result.status}`);
+
+    return { items };
+  }
+
+  public async queryNodesAndEdges<
+    TQueryRequest extends QueryRequest,
+    TypedSelectSources extends SelectSourceWithParams = SelectSourceWithParams
+  >(query: TQueryRequest): Promise<QueryResult<TQueryRequest, TypedSelectSources>> {
+    return await queryNodesAndEdges<TQueryRequest, TypedSelectSources>(query, this._sdk);
```
danpriori, line 552: Although I dont see anything smelling with the queryNodesAndEdges and the code related, I think it would be good if Anders could pay some extra attention in here, including the QueryResult type definition inside the queryNodesAndEdges.ts

react-components/src/data-providers/core-dm-provider/CoreDm3dDataProvider.ts
```diff
@@ -0,0 +1,210 @@
...
+  async getFdmConnectionsForNodes(
+    models: DmsUniqueIdentifier[],
+    revisionId: number,
+    nodes: Node3D[]
+  ): Promise<FdmCadConnection[]> {
+    if (models.length !== 1) {
+      throw new Error(`Expected 1 model, got ${models.length}`);
+    }
+
+    const model = models[0];
```
danpriori, line 128: I think this way of having an argument as an array (in this case: `models`) , suggesting multiple instances but in the function description it ends up forcing the first one to always be taken and having to be unique, is a bit contradictory in terms of semantic coding in relation to what the function requests and what it does.
Perhaps if we could make it explicit in the method name that the argument is an array with a single instance, I feel it would be less bad. But that is just my opinion and not something crucial.

react-components/src/data-providers/core-dm-provider/cogniteAssetSourceWithProperties.ts
```diff
@@ -0,0 +1,36 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+import { type SourceSelectorV3 } from '@cognite/sdk/dist/src';
```
danpriori, line 4: can we import without dist/src here?

react-components/src/data-providers/core-dm-provider/cogniteCadNodeSourceWithProperties.ts
```diff
@@ -0,0 +1,23 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+import { type SourceSelectorV3 } from '@cognite/sdk/dist/src';
```
danpriori, line 4: can we import without dist/src here?

react-components/src/data-providers/core-dm-provider/cogniteCadRevisionSourceWithProperties.ts
```diff
@@ -0,0 +1,12 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+import { type SourceSelectorV3 } from '@cognite/sdk/dist/src';
```
danpriori, line 4: can we import without dist/src here?

react-components/src/data-providers/core-dm-provider/cogniteObject3dSourceWithProperties.ts
```diff
@@ -0,0 +1,27 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+import { type SourceSelectorV3 } from '@cognite/sdk/dist/src';
```
danpriori, line 4: can we import without dist/src here?

react-components/src/data-providers/core-dm-provider/filterNodesByMappedTo3d.ts
```diff
@@ -0,0 +1,251 @@
...
+    direct_nodes_point_cloud_volumes: { sources: pointCloudVolumeSourceWithProperties },
+    indirect_nodes_object_3ds: { sources: cogniteObject3dSourceWithProperties },
+    indirect_nodes_cad_nodes: {
+      sources: cadNodeSourceWithProperties
+    },
+    indirect_nodes_point_cloud_volumes: {
+      sources: pointCloudVolumeSourceWithProperties
+    }
+  }
+} as const satisfies Omit<QueryRequest, 'parameters' | 'cursor'>;
```
danpriori, line 207: Is it possible (and if does make sense) to move out the consts and the type definitions? Or at least organize into a separated place inside the file. I think moving out we can reduce the file size.

react-components/src/data-providers/core-dm-provider/getCadConnectionsForRevisions.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  return returnResult;
+}
+
+function createObject3dToAssetMap<T extends NodeItem<CogniteAssetProperties>>(
+  assets: T[]
+): Map<FdmKey, T> {
+  return new Map(
+    assets.map((asset) => [
+      createFdmKey(asset.properties[CORE_DM_SPACE]?.['CogniteAsset/v1'].object3D),
```
danpriori, line 69: Does make sense to move the key string 'CogniteAsset/v1' into a const?

react-components/src/data-providers/core-dm-provider/getCadConnectionsForRevisions.ts
```diff
@@ -0,0 +1,194 @@
...
+        direction: 'inwards'
+      },
+      limit: 10000
+    }
+  },
+  select: {
+    cad_nodes: { sources: cogniteCadNodeSourceWithProperties },
+    assets: { sources: cogniteAssetSourceWithProperties }
+  }
+} as const satisfies Omit<QueryRequest, 'cursor' | 'parameters'>;
```
danpriori, line 194: Is it possible (and if does make sense) to move out the consts and the type definitions? Or at least organize into a separated place inside the file. I think moving out we can reduce the file size.

react-components/src/data-providers/core-dm-provider/getCadModelsForInstance.ts
```diff
@@ -0,0 +1,86 @@
...
+      nodes: {
+        from: 'object_3ds',
+        through: { view: COGNITE_CAD_NODE_SOURCE, identifier: 'object3D' }
+      }
+    }
+  },
+  select: {
+    cad_nodes: { sources: cogniteCadNodeSourceWithProperties }
+  }
+} as const satisfies Omit<QueryRequest, 'parameters' | 'cursor'>;
```
danpriori, line 86: Is it possible (and if does make sense) to move out the consts and the type definitions? Or at least organize into a separated place inside the file. I think moving out we can reduce the file size.

react-components/src/data-providers/core-dm-provider/getDMSRevision.ts
```diff
@@ -0,0 +1,80 @@
...
+    revision: {
+      sources: [
+        {
+          source: COGNITE_CAD_REVISION_SOURCE,
+          properties: ['status', 'published', 'type', 'model3D', 'revisionId']
+        }
+      ]
+    }
+  }
+} as const satisfies Omit<QueryRequest, 'cursors' | 'parameters'>;
```
danpriori, line 80: Is it possible (and if does make sense) to move out the consts and the type definitions? Or at least organize into a separated place inside the file. I think moving out we can reduce the file size.

danpriori (approved):
Nice work! Some few comments (which doesnt block it) but the rest LGTM at least. 
However, of course Anders or someone else that has experience with the CDM needs to have a look on this.

react-components/src/data-providers/core-dm-provider/cogniteCadNodeSourceWithProperties.ts
```diff
@@ -0,0 +1,23 @@
...
+import { COGNITE_CAD_NODE_SOURCE } from './dataModels';
+
+export const cogniteCadNodeSourceWithProperties = [
+  {
+    source: COGNITE_CAD_NODE_SOURCE,
+    properties: [
```
anders-hopland, line 10: Do you need the properties from Describable like name, description, tags and aliases? If so, it would be nice to compose this property list of Describable properties and Cad Node Properties

react-components/src/data-providers/core-dm-provider/cogniteCadRevisionSourceWithProperties.ts
```diff
@@ -0,0 +1,12 @@
...
+import { COGNITE_CAD_REVISION_SOURCE } from './dataModels';
+
+export const cogniteCadNodeSourceWithPRoperties = [
```
anders-hopland, line 7: `cogniteCadNodeSourceWithPRoperties`  -> `cogniteCadNodeSourceWithProperties`

react-components/src/architecture/concrete/observations/network.ts
haakonflatval-cognite: Right :thinking: Good point - I don't have much against it, but I'd rather do it in a subsequent PR

react-components/src/components/CacheProvider/FdmNodeCache.ts
haakonflatval-cognite: I do see your point.. But I actually kind-of like that the value is created immutably here, and the connection isn't added later. I think I want to keep it :see_no_evil:

react-components/src/components/CacheProvider/RevisionFdmNodeCache.ts
haakonflatval-cognite: Right... Do you think that would reduce the time much? In any case, this function in particular is used only by the ancestor-mapping-finding function, and we shouldn't have more than about 20 results for that query normally

react-components/src/components/CacheProvider/requests.ts
haakonflatval-cognite: Definitely, good catch :+1:

react-components/src/data-providers/FdmSDK.ts
haakonflatval-cognite: This is taken directly from @christjt - but I'll ask Anders to look as well :sweat_smile:

react-components/src/data-providers/core-dm-provider/CoreDm3dDataProvider.ts
haakonflatval-cognite: Right... This is to conform to the interface - where it could happen in the legacy FDM model, that there were multiple 3D model FDM instances for one model-ID. But I think maybe that would also count as a bug :thinking: In any case, I think I'll rather change this to a warning if there is more than one model

react-components/src/data-providers/core-dm-provider/filterNodesByMappedTo3d.ts
```diff
@@ -0,0 +1,251 @@
...
+    .concat(connectionData.items.direct_nodes_cad_nodes)
+    .concat(connectionData.items.indirect_nodes_cad_nodes)
+    .map((node) => createFdmKey(node.properties.cdf_cdm['CogniteCADNode/v1'].object3D));
+
+  const pointCloudObject3dList = [...connectionData.items.initial_nodes_point_cloud_volumes]
+    .concat(connectionData.items.direct_nodes_point_cloud_volumes)
+    .concat(connectionData.items.indirect_nodes_point_cloud_volumes)
+    .map((pointCloudVolume) =>
+      createFdmKey(pointCloudVolume.properties.cdf_cdm['CognitePointCloudVolume/v1'].object3D)
+    );
```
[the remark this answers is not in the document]
haakonflatval-cognite: Right, I'll do it :+1: I didn't think about it, because it's auto-completed as I access the properties, but I agree that it's easier to maintain if we factor them out as constants :+1:

react-components/src/data-providers/core-dm-provider/filterNodesByMappedTo3d.ts
haakonflatval-cognite: I agree, the file was becoming a bit too big

react-components/src/data-providers/core-dm-provider/getCadConnectionsForRevisions.ts
haakonflatval-cognite: I did it :+1:

react-components/src/data-providers/core-dm-provider/getCadConnectionsForRevisions.ts
haakonflatval-cognite: Agreed, moved it out :+1:

react-components/src/data-providers/core-dm-provider/cogniteCadNodeSourceWithProperties.ts
haakonflatval-cognite: Good point, I'll do that

react-components/src/components/CacheProvider/RevisionFdmNodeCache.ts
danpriori: Hmmm ok. I thought we could have a large array here. 
In general, using regular for-loop, for-of, forEach reduce considerably if the array is large (the "reduce" method is faster than for-in usually, but it is slower than the others mostly). You can make some benchmark to compare the average.

PR #4711, opened 2024-08-20
Finalize the string for translation

nilscognite:
For those strings that should not appear in production, the key is undefined.


pramodcog (approved):
LGTM

PR #4713, opened 2024-08-20
fix(react-components): Just make a base class for SettingsCommand and add ALT shortcut in the architecture

nilscognite:
#### Type of change
Refactor 

This PR is just a refactor. No change in the applications.

Made a base class for SettingsCommand, called BaseSettingsCommand. This enable the possibility to make several settings buttons with different types of settings.

Also added possible Alt shortcut, which can be combined with Ctrl or Shift.

### DIFF
--- a/react-components/src/architecture/base/concreteCommands/SettingsCommand.ts
+++ b/react-components/src/architecture/base/concreteCommands/SettingsCommand.ts
@@ -1,0 +1,39 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+
+import { BaseSettingsCommand } from '../commands/BaseSettingsCommand';
+import { SetQualityCommand } from './SetQualityCommand';
+import { SetPointSizeCommand } from './SetPointSizeCommand';
+import { SetPointColorTypeCommand } from './SetPointColorTypeCommand';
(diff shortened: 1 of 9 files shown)


react-components/src/architecture/base/concreteCommands/SettingsCommand.ts
```diff
@@ -0,0 +1,39 @@
...
+  // ==================================================
+
+  public constructor() {
+    super();
+
+    this.add(new SetQualityCommand());
+    this.add(new SetPointSizeCommand());
+    this.add(new SetPointColorTypeCommand());
+    this.add(new SetPointShapeCommand());
+    this.add(new PointCloudFilterCommand());
```
pramodcog, line 25: If I understand correctly, this will add all point cloud operation (Point Size, Point color, point shape, etc) by default to the `Setting` button, assume if there is no point cloud model added into the `3D viewer`, as a user these option should populate only when point cloud model is in `viewer`.

nilscognite: This is taken care of. There commands will not show up when no point cloud in the viewer.

pramodcog (approved):

PR #4714, opened 2024-08-21
feat (react-components): Export the base classes and some utility from the new architecture

nilscognite:
#### Type of change

Feat  

## Description :pencil:

* 3D management need to use some of the basics in the architecture, so I exported all base classes and some of the utility classes/functions used by the actitecture.

* Also updated the listeners for commands. The reason for this is that you can listen to if a tool is activated or deactivated.


haakonflatval-cognite (approved):
LGTM :+1: But I really want to revisit this at a later date to determine if there's some other way we can expose the functionality we want

PR #4715, opened 2024-08-21
(fix): Update recalculateBoundingBox  in Cognite3DViewer

nilscognite:
#### Type of change

Bug 

## Description :pencil:

the function `fitCameraToModels `only take all the models, even invisible and doesn't use the `customobject `at all. In order not to break the code, I made an alternative which is `fitCameraToSceneBoundingBox`.  This uses the getSceneBoundingBox.

I also change the `recalculateBoundingBox `to use only visible object. See comment in that function. I believe that was the intension of the original code. It might break the code.... An alternative would be to make another function for doing this.

I saw this bug during a demo today. The user had many models, but only one was visible. The navigation was pretty hard and the fit all button didn't work as expected.

### DIFF
--- a/viewer/packages/api/src/public/migration/Cognite3DViewer.ts
+++ b/viewer/packages/api/src/public/migration/Cognite3DViewer.ts
@@ -1323,7 +1323,7 @@ export class Cognite3DViewer {
   }
 
   /**
-   * Move camera to a place where a all objects in the scene are visible
+   * Move camera to a place where a all objects in the scene are visible.
    * @param duration The duration of the animation moving the camera. Set this to 0 (zero) to disable animation.
    */
   fitCameraToVisualSceneBoundingBox(duration?: number): void {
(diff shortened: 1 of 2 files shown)


viewer/packages/api/src/public/migration/Cognite3DViewer.ts
```diff
@@ -1880,34 +1893,45 @@ export class Cognite3DViewer {
...
-        return;
+        // The getModelBoundingBox is using restrictToMostGeometry = true
+        model.getModelBoundingBox(temporaryBox, true);
+        if (temporaryBox.isEmpty()) {
+          return;
+        }
+        sceneBoundingBox.union(temporaryBox);
+      });
+      this._sceneHandler.customObjects.forEach(customObject => {
+        if (!customObject.object.visible) {
```
haakonflatval-cognite, line 1918: Do we want to have the `pass === 0`-check here as well?

haakonflatval-cognite (approved):
LGTM :+1: Just one question

PR #4718, opened 2024-08-21
fix(react-components): Smaller adjustment in clipping and measurement after meeting with user

nilscognite:
#### Type of change

Bug 

## Description :pencil:

This is based on feedback on a meeting today:

* Increase selection radius for lines. Before it was hard to pick them.
* Increate transparence of the clipping planes.
* Change color of the clipping planes to be less saturated. Also switch the side colors.
* When new plane is made, make sur that it is clipped in front of you.


danpriori (approved):
LGTM

PR #4725, opened 2024-08-27
fix(react-component): Tiny fix to visualize the point cloud filter in the settings drop down

nilscognite:
* Fix to show the  point cloud filter in the settings drop down
* Add a class in the index file 8forgot this class)
* Expose RevealButtons.PointCloudFilter so it can be reused


danpriori (approved):
LGTM

PR #4730, opened 2024-08-29
fix(react-components): Fix filter menus and som styling of the other autogenerated options

nilscognite:
#### Type of change

Bug 

## Description :pencil:

Renamed some virtual functions to` ...Core` and made them protected.  This is because I wanted update to be centralized in the according non virtual function. This force the updating when `setChecked `and `toggleAllChecked `are called.

In addition I fixed the width to be width to prevent shifting between 1 and 2 lines.

Open storybook/architecture. Click in bug icon on the right toolbar

### DIFF
--- a/react-components/src/architecture/base/commands/mocks/MockFilterCommand.ts
+++ b/react-components/src/architecture/base/commands/mocks/MockFilterCommand.ts
@@ -5,7 +5,6 @@
 import { Color } from 'three';
 import { BaseFilterCommand, BaseFilterItemCommand } from '../BaseFilterCommand';
 import { type TranslateKey } from '../../utilities/TranslateKey';
-import { CommandsUpdater } from '../../reactUpdaters/CommandsUpdater';
 
 export class MockFilterCommand extends BaseFilterCommand {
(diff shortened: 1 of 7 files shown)


react-components/src/architecture/base/commands/mocks/MockFilterCommand.ts
```diff
@@ -98,11 +97,11 @@ class FilterItemCommand extends BaseFilterItemCommand {
...
 
-  public override setChecked(value: boolean): void {
+  protected override setCheckedCore(value: boolean): boolean {
     if (this._use === value) {
-      return;
+      return false;
     }
     this._use = value;
-    CommandsUpdater.update(this._renderTarget);
+    return true;
```
danpriori, line 105: In OOP classic concept, we usually have the setter method to set a value in a parameter or some logic around updating a state of some object property. In this change, you are only returning a boolean based on another value. Is it not better than change the method label? Like: getCheckedCore or something more "getter" style?

danpriori (approved):
LGTM

PR #4731, opened 2024-08-29
feat(react-components): Move m/ft button to the domain object card

nilscognite:
#### Type of change

Feat  

## Description :pencil:

See this conversation: https://cognitedata.slack.com/archives/G0115JQ8JM8/p1724831633376919

* I also did some refactoring in the `ObservationsDomainObject`, because the `Quantity `was missing
* Fix missing "Copy to clipboard" button.

Open storybook/architecture
Make a crop box or a measurement.

### DIFF
--- a/react-components/src/components/Architecture/DomainObjectPanel.tsx
+++ b/react-components/src/components/Architecture/DomainObjectPanel.tsx
@@ -37,16 +37,16 @@ export const DomainObjectPanel = (): ReactElement => {
   useEffect(() => {
     DomainObjectPanelUpdater.setDomainObjectDelegate(setCurrentDomainObjectInfo);
 
-    if (commands === undefined || info === undefined) {
-      return;
-    }
-
     // Set in the get string on the copy command if any
+  }, [setCurrentDomainObjectInfo, commands]);
+
+  // Fore the getString to be updated
+  if (commands !== undefined && info !== undefined) {
     for (const command of commands) {
       if (command instanceof CopyToClipboardCommand)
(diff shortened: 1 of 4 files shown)


react-components/src/components/Architecture/DomainObjectPanel.tsx
```diff
@@ -37,16 +37,16 @@ export const DomainObjectPanel = (): ReactElement => {
...
+  }, [setCurrentDomainObjectInfo, commands]);
+
+  // Fore the getString to be updated
+  if (commands !== undefined && info !== undefined) {
     for (const command of commands) {
       if (command instanceof CopyToClipboardCommand)
         command.getString = () => toString(info, t, unitSystem);
     }
-  }, [setCurrentDomainObjectInfo, commands]);
+  }
```
haakonflatval-cognite, line 50: What is this change for? I think the for-loop should still be inside the useEffect, as it seems like something that's unnecessary to do on every render. But Perhaps `unitSystem` should be in the dependency array?

haakonflatval-cognite (changes requested):
Looks good, just a comment

react-components/src/components/Architecture/DomainObjectPanel.tsx
nilscognite: The promblem vas that the getString wasn't apply and the button disabled. I dont know why, but when I changed the code, it worked.

haakonflatval-cognite (approved):
LGTM :+1:

PR #4745, opened 2024-09-06
feat(react-components): Move global camera controls to Reveal context

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

## Description :pencil:

This moves the global camera controls to props in RevealContext. Earlier, they would be given to a component that had been created from the `withCameraStateControl`. This makes it very abritrary where in the code this camera control happens, and can be controlled from multiple places in the same page, which may cause unintended behavior. Now, the controls (a getter and setter) of the camera state are given as props directly to RevealContext, making it more obvious where this global state is controlled. Of course, the user may still mutate the state otherwise (either through the useNavigation hook, direct access to the Reveal camera manager, or just moving the camera with mouse or keyboard), and this will be handled gracefully by the RevealContext (the setter sent to RevealContext will be called in all these cases)


react-components/src/components/RevealCanvas/ViewerContext.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const ViewerContextProvider = ({
+  cameraState,
+  setCameraState,
+  value,
+  children
+}: {
+  cameraState?: CameraStateParameters;
+  setCameraState?: (cameraState?: CameraStateParameters) => void;
+  value: RevealRenderTarget | null;
+  children: ReactNode;
```
pramodcog, line 20: Can be moved to `props`.

react-components/src/components/RevealCanvas/ViewerContext.tsx
```diff
@@ -0,0 +1,52 @@
...
+  setCameraState
+}: {
+  cameraState?: CameraStateParameters;
+  setCameraState?: (cameraState?: CameraStateParameters) => void;
+}): ReactNode => {
+  useCameraStateControl(cameraState, setCameraState);
+  return null;
+};
+
+export const useReveal = (): Cognite3DViewer => {
```
pramodcog, line 41: This should be in `hooks` folder?

react-components/src/components/RevealCanvas/hooks/useCameraStateControl.ts
```diff
@@ -0,0 +1,98 @@
+/*!
+ * Copyright 2023 Cognite AS
+ */
+
+import { type MutableRefObject, useEffect, useRef } from 'react';
+import { useReveal } from '../../..';
```
pramodcog, line 6: Proper path!

pramodcog (approved):
Looks good, nice works with the tests!
Minor comments

react-components/src/components/RevealCanvas/ViewerContext.tsx
haakonflatval-cognite: I don't think so, the convention we have been using so far, is that hooks that are sort-of exposing values from a Context, are put together with that Context. That's how this hook was placed before this PR as well, in ViewerContext.ts

PR #4754, opened 2024-09-12
fix(react-components): Fix a small bug in compare equalsIgnoreCaseAndSpace and use correct undo icon

nilscognite:
1. Fix a small bug in compare equalsIgnoreCaseAndSpace
2. Use correct undo icon


danpriori (approved):
LGTM

PR #4768, opened 2024-09-20
feat: Make the camera manager marker better

nilscognite:
#### Type of change

Feat  

## Description :pencil:

Now, the marker will be render with a weaker color when it is behind something in the scene.

image

image

Go to reveal/examples
Run.

### DIFF
--- a/viewer/packages/camera-manager/src/Flexible/FlexibleCameraMarkers.ts
+++ b/viewer/packages/camera-manager/src/Flexible/FlexibleCameraMarkers.ts
@@ -5,10 +5,12 @@
 import { Scene, Object3D, Vector3, Sprite, SpriteMaterial, CanvasTexture } from 'three';
 import { FlexibleCameraManager } from './FlexibleCameraManager';
 import { FlexibleControlsType } from './FlexibleControlsType';
+import { FlexibleControlsOptions } from './FlexibleControlsOptions';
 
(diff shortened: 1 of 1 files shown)


viewer/packages/camera-manager/src/Flexible/FlexibleCameraMarkers.ts
```diff
@@ -75,9 +91,13 @@ function setPosition(object3D: Object3D, position: Vector3, manager: FlexibleCam
   object3D.updateMatrixWorld();
 }
 
-function createSprite(outerColor: string, innerColor: string): Sprite {
-  const texture = createTexture(25, 3, outerColor, innerColor);
-  const material = new SpriteMaterial({ map: texture, depthTest: false });
+function createSprite(options: FlexibleControlsOptions, depthTest: boolean): Sprite {
+  const texture = createTexture(25, 3, options.outerMarkerColor, options.innerMarkerColor);
```
danpriori, line 95: I would suggest to move these magic numbers to a class constants instead

nilscognite: Yes, I move it to constants like this:

```
const TEXTURE_SIZE = 25;
const LINE_WIDTH = 3;
const NO_DEPTH_TEST_OPACITY = 0.5;

```

PR #4801, opened 2024-10-09
feat: Optionally remove the event mouse functions from the 360 images

nilscognite:
#### Type of change

Feat  

## Description :pencil:

Optionally remove the event mouse functions from the 360 images. This follows the same pattern as when I removed the event functions from the camera manager. I had to changed some return values. If returning true something is changed. 

I also added 2 function on the viewer. This will be called from the architecture to do the same. Then we have more control over it.

Use examples with 360 images. It shouldn't be different than it was before.

Sorry, I'm not proud of this PR. I don't feel comfortable with this. The 360 image code is already very messy and I added even more......


viewer/packages/360-images/src/Image360Facade.ts
```diff
@@ -37,8 +37,14 @@ export class Image360Facade<T> {
     this.getCollectionContainingEntity(entity).setSelectedVisibility(visible);
   }
 
-  hideAllHoverIcons(): void {
-    this._image360Collections.forEach(collection => collection.setSelectedVisibility(false));
+  hideAllHoverIcons(): boolean {
+    let changed = false;
+    for (const collection of this._image360Collections) {
```
pramodcog, line 42: We can use `Array.some()` method instead of `for` here like below, it will also remove temp variable as well
```
hideAllHoverIcons(): boolean {
  return this._image360Collections.some(collection => collection.setSelectedVisibility(false));
}
```

nilscognite: You cant'. I have to call setSelectedVisibility() for all anyway

pramodcog: I get that you have to set visibility for all objects in collection than, we can use Array.reduce() get the required results as below
```
hideAllHoverIcons(): boolean {
  return this._image360Collections.reduce((changed, collection) => {
    return collection.setSelectedVisibility(false) || changed;
  }, false);
}
```

pramodcog (approved):
Nice!

PR #4815, opened 2024-10-22
feat(react-components): Add "expected resource count" hook

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

## Description :pencil:

Expose a hook that gives the expected number of loaded models. It gives the number of currently mounted `*ModelContainer` components, even though the models associated with those containers may not yet have been loaded.

The need for this arose in Search, where we would show an empty search panel if no model was loaded in the viewer, which would be problematic when the current scene didn't have any models in it, and it would make it impossible for the user to switch scenes. 

It also has the added benefit that we no longer need to reload the 3D search results panel every time a model finishes loading, but can instead wait until all models that are expected to load, loads.

In Fusion, Search


danpriori (approved):
LGTM

PR #4821, opened 2024-10-22
fix(react-components): Fixing updating of commands using the achitecture

nilscognite:
#### Type of change
Bug 

## Description :pencil:
Updating vas not correct in the menu components. I use the same pattern as for the other command component.
This pattern in marked in the code as `@update-ui-component-pattern`

So the updating of the toggle (for instance Show 360 images)  in the menu should be done when changing it  other places in the UI.

Other things on this PR:

* Little refactoring regarding the short cut buttons.
* Marked `@update-ui-component-pattern` in the code to recognize the same pattern. Good for bug fixing.

I tested it by creating some Mocks. Simulated that the toggle can be turn off several places. On in the settings button. Another in the toolbar. This is now working both ways.

### DIFF
--- a/react-components/src/components/Architecture/CommandButton.tsx
+++ b/react-components/src/components/Architecture/CommandButton.tsx
@@ -23,6 +23,7 @@ export const CommandButton = ({
   const { t } = useTranslation();
   const command = useMemo<BaseCommand>(() => getDefaultCommand(inputCommand, renderTarget), []);
 
+  // @update-ui-component-pattern
   const [isChecked, setChecked] = useState<boolean>(false);
   const [isEnabled, setEnabled] = useState<boolean>(true);
   const [isVisible, setVisible] = useState<boolean>(true);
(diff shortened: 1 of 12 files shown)


react-components/src/components/Architecture/CommandButton.tsx
```diff
@@ -23,6 +23,7 @@ export const CommandButton = ({
   const { t } = useTranslation();
   const command = useMemo<BaseCommand>(() => getDefaultCommand(inputCommand, renderTarget), []);
 
+  // @update-ui-component-pattern
```
pramodcog, line 26: I think, we can move update-ui-component-pattern as a separate file as it is used in several components, your thoughts?

nilscognite: I want, but do not know how in the react framework. Any suggestions? I have to use all the states in the rest of the code.

nilscognite: Anyway, this is a fix. I rather do this on another PR. Understand?

pramodcog (approved):

PR #4825, opened 2024-10-24
fix(react-components): Fix layout of filter drop down menu

nilscognite:
Test it by this:

image

Before it looked like this:

image


pramodcog (approved):
LGTM!

PR #4835, opened 2024-10-30
feat(react-components) Fix tooltip placement and make it possible to set it outside react components

nilscognite:
#### Type of change
Feat  

## Description :pencil:

The placement variable is now used instead of isHorizontal. It means the placement of the toolbar the button is in.
This will automatically set the tooltip at correct position.

Default is left, so in fusion we have to set this property at the top toolbar only. That what it was before, so no real changes.

The toolbar that will be generated automatically based on position will be correct now.

Also set tooltip delay to 500ms, which is normal.

Open storybook/architecture
Check the tooltips.
Open Measurement
Check also the tooltip there

### DIFF
--- a/react-components/src/architecture/base/domainObjectsHelpers/PopupStyle.ts
+++ b/react-components/src/architecture/base/domainObjectsHelpers/PopupStyle.ts
@@ -13,19 +13,19 @@ type PopupProps = {
 };
 
 export class PopupStyle {
-  private readonly _left?: number = undefined;
-  private readonly _right?: number = undefined;
-  private readonly _top?: number = undefined;
-  private readonly _bottom?: number = undefined;
+  public readonly left?: number = undefined;
+  public readonly right?: number = undefined;
+  public readonly top?: number = undefined;
+  public readonly bottom?: number = undefined;
   private readonly _margin: number = 8; // margin outside the popup
   private readonly _padding: number = 6; // margin inside the popup
   private readonly _horizontal: boolean = true; // Used for toolbars only
 
   public constructor(props: PopupProps) {
-    this._left = props.left;
-    this._right = props.right;
-    this._top = props.top;
-    this._bottom = props.bottom;
+    this.left = props.left;
+    this.right = props.right;
+    this.top = props.top;
+    this.bottom = props.bottom;
     if (props.margin !== undefined) {
       this._margin = props.margin;
     }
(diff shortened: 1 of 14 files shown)


react-components/src/components/Architecture/utilities.ts
```diff
@@ -15,12 +15,27 @@ export function getIcon(command: BaseCommand): IconName | undefined {
...
+      return 'bottom';
+    case 'bottom':
+      return 'top';
+    case 'left':
+      return 'right';
+    case 'right':
+      return 'left';
+    default:
+      return 'top';
+  }
```
danpriori, line 38: Maybe a way to help this function be self-explanatory would be to change the name of the argument to something like "parentPlacement" or something like that to understand that it is getting the placement of a source and returning where the placement of the correlated tooltip would be. Because when you look at the definition of this function and what it checks and what it returns, it seems a bit ambiguous, IMO. So maybe it would be good to make it more explicit or create a comment explaining the reason for this switch and the argument.

danpriori (approved):
LGTM. A few comments but nothing that blocks. However, it will be cool if you can have a look :)

react-components/src/architecture/base/domainObjectsHelpers/PopupStyle.ts
```diff
@@ -13,19 +13,19 @@ type PopupProps = {
...
 
 export class PopupStyle {
-  private readonly _left?: number = undefined;
-  private readonly _right?: number = undefined;
-  private readonly _top?: number = undefined;
-  private readonly _bottom?: number = undefined;
+  public readonly left?: number = undefined;
+  public readonly right?: number = undefined;
+  public readonly top?: number = undefined;
+  public readonly bottom?: number = undefined;
```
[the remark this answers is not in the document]
nilscognite: Need only setters. And they are read only. Getters don't give any more value.

react-components/src/components/Architecture/utilities.ts
nilscognite:
```
export function getTooltipPlacement(toolbarPlacement: PlacementType): PlacementType {
  // Try to keep the tooltip on the opposite side of the toolbar
  switch (toolbarPlacement) {
    case 'top':
      return 'bottom';
    case 'bottom':
      return 'top';
    case 'left':
      return 'right';
    case 'right':
      return 'left';
    default:
      return 'top';
  }
}

```

PR #4838, opened 2024-10-31
fix(react-components):  Upgrade to 0.64.0, remove unused code and fix one tooltip placement for one button used at the top toolbar

nilscognite:
#### Type of change

Bug 

## Description

* Fix one tooltip placement
* Remove unused code. This is moved to the architecture for some time ago
* Upgrade to new version since I have to update Fusion

Test in search/3d. But the removed buttons are not in use in Fusion.


haakonflatval-cognite (approved):
LGTM :+1:

PR #4840, opened 2024-11-01
fix(react-components): Remove the exit button. Now it is in the top toolbar instead

nilscognite:
#### Type of change

Bug 

image


haakonflatval-cognite (approved):
LGTM :+1:

PR #4861, opened 2024-11-12
fix: Fix wrong event type in onHover

nilscognite:
Change from PointEvent to MouseEvent everywhere when hover

And update to "version": "4.21.1

### DIFF
--- a/viewer/packages/utilities/src/events/PointerEvents.ts
+++ b/viewer/packages/utilities/src/events/PointerEvents.ts
@@ -7,7 +7,7 @@
  * @beta
  */
 export class PointerEvents {
-  public onHover(_event: PointerEvent): void {}
+  public onHover(_event: MouseEvent): void {}
 
   public async onClick(_event: PointerEvent): Promise<void> {
     return Promise.resolve();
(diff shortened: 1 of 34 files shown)


viewer/packages/utilities/src/events/PointerEvents.ts
```diff
@@ -7,7 +7,7 @@
  * @beta
  */
 export class PointerEvents {
-  public onHover(_event: PointerEvent): void {}
```
pramodcog, line 10: Why Hover event is changed to `MouseEvent`?
`PointerEvent` basically inherits `MouseEvent`, if `MouseEvent` is used than Touch/Pen will not work

nilscognite: I know. But hover is never used on Touch. The position of your finders are unknown anyway. So I wanted to synchronize it. sometimes it uses Mouse and sometime Pointer.

nilscognite: I will update the API doc when you have approved.

pramodcog (approved):
LGTM!
Can you make sure API docs are updated as well

PR #4934, opened 2024-12-04
feat(react-components): distinguish whether clicked node data is pending or empty, bump 0.73.6

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

## Description :pencil:

Up to now, it's been impossible to know if the fields of the `useClickedNodeData` return value have been pending, or if there was no result, as the fields would be `undefined` in both cases.

This PR encances useClickedNodeData to return `null` for fields that evaluates to no result.

This is used for a usability improvement in Fusion, where we wait with closing the side panel on click until we're sure whether the new click gave a new instance for the panel.

### DIFF
--- a/react-components/package.json
+++ b/react-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cognite/reveal-react-components",
-  "version": "0.73.5",
+  "version": "0.73.6",
   "exports": {
     ".": {
       "import": "./dist/index.js",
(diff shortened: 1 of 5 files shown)


haakonflatval-cognite:
> LGTM. We have to just keeping in mind that as now we are having one extra "state" (the null) we have to always be careful when checking this value in if conditions/operators. We need to check the values explicitly in order to distinguish undefined and null and not use the shorter version that combines both undefined and null at once / nullish coalescing operator / nullish coalescing assignment operator...

Yes, good point :thinking: Is there a better way of doing this? Ideally, I'd want something like a `Promise<Option<Data>>`, but since we're in React, we don't have a Promise, and instead the return value is optional...

PR #4946, opened 2024-12-18
fix(react-components): avoid fetching classic asset mappings when in coredm only and recache all revisionIds when requesting the connections with views

danpriori:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:
This fixes the use of Rule Based in projects with CoreDM Only, avoiding the need to fetch classic asset mappings to avoid triggering an error and making the Rule Based process unfeasible.

It also redoes the fdm cache when the cache request is made together with the views, because even when requesting the cache with views, the system was returning without the views because the revision IDs were already cached, causing problems when the user activates/deactivates the button to view all asset mappings of the model and then tries to activate some Rule Based Coloring, which needs the Views and uses this same cache to process the rule criteria.

Tested locally with the quickstart project in 3d test

1. Login into Search with 3d-dev-quickstart project
2. Go to Browse 3D and select Hakon Scene
3. Enable any Rules from Color overlay dropdown before and after toggling the asset mappings button

### DIFF
--- a/react-components/src/hooks/useIsCoreDmOnly.ts
+++ b/react-components/src/hooks/useIsCoreDmOnly.ts
@@ -4,5 +4,7 @@
 import { useRenderTarget } from '../components';
 
 export function useIsCoreDmOnly(): boolean {
-  return useRenderTarget().cdfCaches.coreDmOnly;
+  const renderTarget = useRenderTarget();
+  const isCoreDmOnly = renderTarget.cdfCaches.coreDmOnly;
+  return isCoreDmOnly ?? false;
 }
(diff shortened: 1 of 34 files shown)


react-components/src/hooks/useIsCoreDmOnly.ts
```diff
@@ -0,0 +1,8 @@
+/*!
+ * Copyright 2025 Cognite AS
+ */
+import { useRenderTarget } from '../components';
+
+export function useIsCoreDmOnly(): boolean {
+  return useRenderTarget().cdfCaches.coreDmOnly;
```
pramod-cog, line 7: It is better to have a variable for `useRenderTarget()` and also can you make sure `cdfCaches.coreDmOnl`y is not `undefined`/exists

danpriori: here
https://github.com/cognitedata/reveal/pull/4946/commits/6ca995277599d4d7ab1f076ccbfba4b5a1a19636

pramod-cog (approved):
LGTM

PR #4958, opened 2025-01-15
fix(react-components): improved performance of asset search slowness

pramod-cog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5057

## Description :pencil:
When there are `100000` or more assets in the project and user `search` for a specific `asset` with string `query` in search inputs in 3D page of `Search` application, it is taking long duration to filter the `assets`, this PR improves the filtering of searched query assets in the search hook

In Search application

- `yarn build` react-components
- update `package.json` file in `fusion` repo with  `"@cognite/reveal-react-components": "file:../reveal/react-components"`
- `run pnpm install && pnpm nx serve flexible-data-explorer`
- Open 3D page in `Search` application with more than `40000` assets in project ( like `cog-3d`)
- search for a string in 3D search, results should be listed much quicker than production build

### DIFF
--- a/react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
+++ b/react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
@@ -18,6 +18,7 @@ import { getAssetsList } from '../hooks/network/getAssetsList';
 import { isDefined } from '../utilities/isDefined';
 import { uniqBy } from 'lodash';
 import { useAssetMappedNodesForRevisions } from '../hooks/cad';
+import { useMemo, useState } from 'react';
 
 export type ModelMappings = {
   model: AddModelOptions;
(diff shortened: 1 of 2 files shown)


react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -48,51 +49,69 @@ export const useSearchMappedEquipmentAssetMappings = (
...
     ],
-    queryFn: async ({ pageParam }: { pageParam: string | undefined }) => {
-      if (query === '' || assetMappingList === undefined) {
-        return { assets: [], nextCursor: undefined };
-      }
+    [query, models]
+  );
+
+  const fetchAssets = async (
+    cursor: string | undefined,
```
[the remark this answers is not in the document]
haakonflatval-cognite: I see that this is a recursive function.. Maybe we need to create a new function that is created from a useCallback, which wraps this one, in order to keep it recursive and inside a useCallback

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -48,51 +49,69 @@ export const useSearchMappedEquipmentAssetMappings = (
...
-        if (uniqueAssets.length >= limit || assetsResponse.nextCursor === undefined) {
-          return { assets: uniqueAssets, nextCursor: assetsResponse.nextCursor };
-        }
+    const newUnmappedAssetIds = new Set(unmappedAssetIds);
+    fetchedAssets.forEach((asset) => {
+      if (!assetIds.has(asset.id)) {
+        newUnmappedAssetIds.add(asset.id);
+      }
+    });
+    setUnmappedAssetIds(newUnmappedAssetIds);
```
haakonflatval-cognite, line 104: Do we need to create a new Set in every iteration? We could just keep the same Set and add into it? 
That would also mean we don't need a `useState`, but could use a `useRef` instead. Unless we do the `useCallback` suggestion above, then it can just be a local variable

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -48,51 +49,69 @@ export const useSearchMappedEquipmentAssetMappings = (
...
-            .filter((assetMapping) =>
-              fetchedAssets.some((asset) => asset.id === assetMapping.assetId)
-            )
-            .map((assetMapping) => fetchedAssets.find((asset) => asset.id === assetMapping.assetId))
-            .filter(isDefined);
-        });
+    const uniqueAssets = uniqBy(
+      [...accumulatedAssets, ...filteredSearchedAssets],
+      (asset) => asset.id
+    );
```
haakonflatval-cognite, line 92: I don't think we need to worry about the assets not being unique - it should be the responsibility of the Asset endpoint to ensure that.

haakonflatval-cognite: Also - isn't this what we're trying to solve with the `unmappedAssetIds`-Set? I think that part might also be unnecessary

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -48,51 +49,69 @@ export const useSearchMappedEquipmentAssetMappings = (
...
-      };
+    return await fetchAssets(assetsResponse.nextCursor, uniqueAssets);
+  };
+
+  return useInfiniteQuery({
+    queryKey,
+    queryFn: async ({ pageParam }: { pageParam: string | undefined }) => {
+      if (query === '' || assetMappingList === undefined) {
+        return { assets: [], nextCursor: undefined };
+      }
```
haakonflatval-cognite, line 114: Not really part of this PR - but do you know why we return `nextCursor` here? I can't see that the user can use it in any way?

pramod-cog: `nextCursor` is needed as input while fetching next page

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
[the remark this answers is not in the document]
pramod-cog: `useCallback` will be useful if we are triggering the function outside the loop

haakonflatval-cognite (approved):
LGTM :+1:

PR #4977, opened 2025-02-03
refactor(react-component): Add label on domain objects 

nilscognite:
#### Type of change

Refactor 

Label is the UI version of the name and should be used in UI. It will be translated.

## Description :pencil:

**The new field is label:**
* Generates a label for the domain object using the provided translation delegate.
* If the name is provided, the label is set to the name, else the label is constructed based on the object's type name, index, and label extension.
* The label is used in UI

There is 2 new methods

`getLabel:` Gets the translated label.
`updateLabel`: Update the label field. This will be used for the tree view since the tree view itself can translate the labels, it has to be done upfront.

**hasIndexOnLabel**

This is a new virtual method that tells if an index should be added to the label. It is useful for making unit labels among siblings.

**PanelInfo**

The label is now used as header. This is the intended behavior, but was create by the type name (Since the name could be translated before)

** Domain Object changes **

I removed some unused methods that used the name, since it make the code more complicated.


react-components/src/architecture/base/domainObjects/RootDomainObject.ts
```diff
@@ -51,8 +50,12 @@ export class RootDomainObject extends DomainObject {
   // OVERRIDES
   // ==================================================
 
+  public override get hasIndexOnLabel(): boolean {
+    return false;
+  }
+
   public override get typeName(): TranslationInput {
-    return { untranslated: 'Root' };
+    return { key: 'SCENE' };
```
pramod-cog, line 58: Not sure if we want root domain object to be have `SCENE` as the key, I prefer to have `ROOT`

pramod-cog (approved):
LGTM, one minor comment, let me know what you think!

PR #4978, opened 2025-02-04
fix: Fix bug in intersection between ray and 360 image icon [BND3D-5179]

nilscognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:
](https://cognitedata.atlassian.net/browse/BND3D-5179)

## Description :pencil:

The intersection point was calculated wrong (wrong coordinate system and also substracted the camera position)

Use example app in Reveal and scroll mouse when mouse hover over the 360 icons.

### DIFF
--- a/viewer/package.json
+++ b/viewer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cognite/reveal",
-  "version": "4.23.2",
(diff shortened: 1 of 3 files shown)


christjt (approved):
👏👏👏

PR #4991, opened 2025-02-10
fix(react-components): filter undefined asset ids in rule based coloring when extracting unique asset ids

danpriori:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:
With recent updates to the SDK, it may have affected the processing of IDs and you may be receiving undefined IDs when retrieving asset IDs from mapped. This small fix creates a filter to not return undefined IDs and therefore not request empty IDs from the getByIds endpoint.


react-components/src/components/RuleBasedOutputs/hooks/useExtractUniqueAssetIdsFromMapped.tsx
```diff
@@ -6,17 +6,20 @@ import { type InternalId } from '@cognite/sdk';
 import { type ModelWithAssetMappings } from '../../../hooks/cad/ModelWithAssetMappings';
 import { useMemo } from 'react';
 import { uniqBy } from 'lodash';
+import { isDefined } from '../../../utilities/isDefined';
 
-export const useExtractUniqueAssetIdsFromMapped = (
+export const useExtractUniqueAssetIdsFromMapped: (
   assetMappings: ModelWithAssetMappings[] | undefined
-): InternalId[] => {
+) => InternalId[] = (assetMappings: ModelWithAssetMappings[] | undefined) => {
```
haakonflatval-cognite, line 13: Is it necessary to rewrite the function signature like this? It seems like it's equivalent with the previous signature, but more verbose

haakonflatval-cognite (commented):
Looks good - but I'm wondering if this should perhaps not be a hook (I don't see why it should be a hook :sweat_smile: ), and then we could add a test to it, following current trend

react-components/src/components/RuleBasedOutputs/hooks/useExtractUniqueAssetIdsFromMapped.tsx
danpriori: It was required by the linter now. But I changed the signature.
https://github.com/cognitedata/reveal/pull/4991/commits/f92343494da34de8f7bc76fdfaf9cf7c3ae1553c

danpriori:
> Looks good - but I'm wondering if this should perhaps not be a hook (I don't see why it should be a hook 😅 ), and then we could add a test to it, following current trend

Hmmm I think as this hook is calling some react hooks like useCallbacks and useMemos then it should be inside a hook, or a FC (ofc). 

About the tests, I was thinking to do the rule based coloring tests altogether in a unique cycle of writing tests as following the QA jira tickets for that. But I can write tests specific for this element to reduce the test workload later.

haakonflatval-cognite (approved):
LGTM :+1:

PR #4997, opened 2025-02-13
refactor(react-components): Use signals for updating the Active Toolbar and the Selected Domainobject Panel

nilscognite:
#### Type of change

Refactor 

## Description :pencil:

This use **signals** for updating the Active toolbar and the selected domain object panel.
Before we use some "react hacks"  to make them reactive. With signals it is a lot cleaner.

1. Open Architecture in storybook
2. Select Measurement tool
3. Check if the toolbar open
4. Make some measurements, for instance boxes.
5. Check if the panel updated when changing selection or resizing
2. Deselect Measurement tool
3. Check if the toolbar closes


react-components/package.json
```diff
@@ -48,6 +49,7 @@
     "@cognite/cogs.js": "^10.36.0",
     "@cognite/reveal": "^4.23.2",
     "@cognite/sdk": "^9.13.0",
+    "@cognite/signals": "^0.0.4",
```
christjt, line 52: Remove this. It is already defined in "dependencies". 

The difference is that devDependencies are only needed for building/testing or other things that is not the actual source code.

react-components/src/components/Architecture/DomainObjectPanel.tsx
```diff
@@ -20,44 +17,33 @@ import { type TranslateDelegate } from '../../architecture/base/utilities/Transl
...
 const TEXT_SIZE = 'x-small';
 const HEADER_SIZE = 'medium';
 
 export const DomainObjectPanel = (): ReactElement => {
-  const [currentDomainObjectInfo, setCurrentDomainObjectInfo] = useState<
-    DomainObjectInfo | undefined
-  >();
-
-  const domainObject = currentDomainObjectInfo?.domainObject;
+  useSignalValue(DomainObjectPanelUpdater.update);
```
christjt, line 26: Is this something we will need in the future? I would expect the UI to update if any of the things it uses changes without having to listen to a generic update signal.

react-components/src/architecture/base/renderTarget/CommandsController.ts
```diff
@@ -35,6 +36,10 @@ export class CommandsController extends PointerEvents {
   // ==================================================
 
   public get activeTool(): BaseTool | undefined {
+    return this._activeTool();
+  }
+
+  public get activeToolSignal(): Signal<BaseTool | undefined> {
     return this._activeTool;
   }
```
christjt, line 44: This is so nice! I guess we don't technically need the `activeTool` then? Can't we just only use the active tool with the signal?

christjt (approved):
I think this is really nice improvement. I would like to discuss which of these changes needed intermidiatly, and where you want to end up eventually. I think in general there should not be any need for generic "update" signals. Why can't we just have the things that the UI depends on be signals, and let it update whenever it needs to. If the reason is that it is needed before everything is using the signals pattern, then I am of course absolutly ok with that.

PR #5005, opened 2025-02-19
chore: add simple unit test for Reveal3DResources

christjt:
#### Type of change

Test 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5419

## Description :pencil:

Add simple unit test for Reveal3DResources. There are still a bunch of business logic in the hooks which are untested, but as we are moving logic to domain objects, I'll defer until that path is more carved out.


haakonflatval-cognite (approved):
LGTM, just a question

PR #5012, opened 2025-02-21
chore: minor improvements and fixes

christjt:
#### Type of change

Chore 

## Description :pencil:

1. Optimize tests by turning off test isolation
2. Fix package.json by correctly defining dependencies
3. Recover proper JSX syntax when passing as props
4. Fix small syntax error in styled component


react-components/package.json
```diff
@@ -32,17 +32,22 @@
   "peerDependencies": {
     "@cognite/cogs-lab": "^9.0.0-alpha.153",
     "@cognite/reveal": "4.23.2",
+    "@cognite/sdk": "^9.13.0",
```
haakonflatval-cognite, line 35: Won't this cause trouble when linking to projects that have another SDK version? The newest SDK version is 10 (and it is used in Fusion, I believe)

christjt: I mean currently we don't define the dependency at all... By declaring it as a peer dependency, fusion will at least be able to provide its version (which is the current behaviour).

PR #5020, opened 2025-02-28
chore: expose viewer on window object

haakonflatval-cognite:
Great for debugging

#### Type of change

Chore 

#### Jira ticket :blue_book:

## Description :pencil:

Expose the viewer on the `window` object, enabling debugging in action.

Implementation of global types follows https://stackoverflow.com/questions/59459312/using-globalthis-in-typescript


react-components/src/global.d.ts
christjt, line 1: I think you should able to do something like: 

```ts
declare global { 
  interface Window { ... } 
}
```

This would then prevent the lint issues as well as being explicit about this being part of the window instead of any global scope.

react-components/src/components/RevealCanvas/ViewerContext.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+      </ViewerContextProvider>
+    );
+
+    expect((window as any).renderTarget).toBe(renderTargetMock);
+    expect((window as any).viewer).toBe(renderTargetMock.viewer);
+
+    cleanup();
+
+    expect((window as any).renderTarget).toBeUndefined();
+    expect((window as any).viewer).toBeUndefined();
```
christjt, line 72: Any reason why the global d.ts isn't respected here?

christjt (changes requested):
Overall, it looks great. Nice that you added tests for it. We need to be a little cautious about this solution with respect to multiple instances of Reveal, though. F.ex. Industry Canvas will need multiple instances of Reveal, which makes this solution a little ambiguous. In any case, we don't "really" support that now anyway, but this is an additional thing we need to watch out for.

react-components/src/components/RevealCanvas/ViewerContext.test.tsx
haakonflatval-cognite: Yes, forgetfulness 👍

haakonflatval-cognite: Forgetfulness :+1:

react-components/src/global.d.ts
haakonflatval-cognite: Ah, nice - my sources didn't mention that one :sweat_smile:

christjt (approved):

larshagencognite (approved):

PR #5023, opened 2025-03-04
fix: separate build and test typescript configs

christjt:
#### Type of change

Chore 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5457

## Description :pencil:

Separate type checking and setup for build and tests. This fixes the issues of index.d.ts being emited the wrong place.


react-components/tsconfig.test.json
```diff
@@ -0,0 +1,26 @@
...
+    "src/**/*.test.tsx",
+    "src/**/*.spec.tsx",
+    "src/**/*.stories.ts",
+    "src/**/*.test.ts",
+    "src/**/*.spec.ts",
+    "stories/**/*",
+    "tests/**/*",
+    "vite.config.ts",
+    "playwright.config.ts"
+  ]
```
haakonflatval-cognite, line 25: I see the previous tsconfig.json had an `exclude`-block that excluded node_modules, dist and storybook-static. Should we have the same here?

haakonflatval-cognite (commented):
Had a comment, but otherwise looks good! Also fixes the issue of the type outputs :raised_hands:

haakonflatval-cognite (approved):

react-components/tsconfig.test.json
christjt: According to https://www.typescriptlang.org/tsconfig/#exclude, exclude only excludes things that are included. So f.ex. storybook-static and node_modules are not in the include.

BugGambit:
I will let @cemreyavuz review this. But @christjt, the PR title is wrong. This is a `chore:` not a `fix:` as this doesn't fix anything for the user.

christjt:
> I will let @cemreyavuz review this. But @christjt, the PR title is wrong. This is a `chore:` not a `fix:` as this doesn't fix anything for the user.

I think you are missing some context here. Currently the d.ts. files get emited in the wrong directory, due to rootDir being inferred wrong due to types and source being evaluated together. This causes issues for users of this lib, as exported types are not where they are supposed to. So in my opinion this is a fix. Happy to change though.

cemreyavuz (commented):
One question with the eslint config change

react-components/.eslintrc.json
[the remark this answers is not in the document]
cemreyavuz: had a chat with @christjt in slack, it is intentional (leaving a comment for future reference)

cemreyavuz (approved):
Verified the build locally as well, seems like the only difference is the directory structure for types (which fixes the issue mentioned) 👍🏻

BugGambit (approved):
:unicorn:

PR #5024, opened 2025-03-04
fix: add circular dependency checking

christjt:
#### Type of change

Chore 

## Description :pencil:

Add CI step for checking circular dependencies.


anders-hopland (approved):
LGTM

pramod-cog:
@christjt It would be nice to add eslint config rules to show circular dependency, something like
```
rules: {
      'import/no-cycle': ['error', { maxDepth: Infinity }],
    },
```
But I will leave it to you, if you think to add it in new PR or in here!

christjt:
> @christjt It would be nice to add eslint config rules to show circular dependency, something like
> 
> ```
> rules: {
>       'import/no-cycle': ['error', { maxDepth: Infinity }],
>     },
> ```
> 
> But I will leave it to you, if you think to add it in new PR or in here!

I would prefer not to have 2 systems that does dependency checking (which may or may not have the same rules), so in any case I will defer. We can evaluate if we should let eslint deal with this on its own later.

react-components/package.json
```diff
@@ -24,6 +24,7 @@
     "test:coverage": "vitest run --dom --coverage",
     "test:playwright": "playwright test",
     "test:watch": "vitest watch --dom",
+    "check-circular-deps": "madge --circular --extensions ts,tsx --exclude 'setClippingPlanes.ts$|CommandButtons.tsx$' ./src",
```
cemreyavuz, line 27: Why do we exclude these two files?

cemreyavuz (commented):
Just a small question on excluded files

react-components/package.json
christjt: The reason is that these two files still contain circular dependencies. We have another PR working on resolving those. But I want to add this check now, as I want to protect against any new circular dependencies getting added in the mean time, which happend a couple days ago...

cemreyavuz: Sounds good! Just wanted to understand 👍🏻

cemreyavuz (approved):

BugGambit (approved):
:unicorn:

anders-hopland (approved):
LGTM

anders-hopland (approved):
LGTM

BugGambit:
risk review ok

PR #5033, opened 2025-03-10
test(react-components): Add test for LineDomainObject

nilscognite:
#### Type of change

Test 

## Description :pencil:

This PR test
* LineDomainObject
* LineCreator
* LineView

**TODO 1:** Was not able to test the labels in the view. Because `Canvas.getContext() `return null. I have to figure out how to do this properly in another PR.

**TODO 2:** The Solid part of the view is not tested. It is not used my the `MeasureLineDomainObject`. It will be tested when making unit tests for the 2D annotation.


react-components/src/architecture/concrete/primitives/line/LineCreator.ts
```diff
@@ -49,18 +49,20 @@ export class LineCreator extends BaseCreator {
   }
 
   public override get minimumPointCount(): number {
-    return 2;
+    switch (this._domainObject.primitiveType) {
```
nilscognite, line 52: Tiny bug in the code discovered by the tests

react-components/src/architecture/concrete/primitives/line/LineDomainObject.ts
```diff
@@ -48,7 +48,7 @@ export abstract class LineDomainObject extends VisualDomainObject {
   }
 
   public get lineSegmentCount(): number {
-    return this.primitiveType === PrimitiveType.Polygon ? this.pointCount + 1 : this.pointCount;
+    return this.primitiveType === PrimitiveType.Polygon ? this.pointCount : this.pointCount - 1;
```
nilscognite, line 51: Bug in the code discovered by the tests, but the bug was justified (+1)  when it was used. So no tooth bugs for the user was discovered.

react-components/src/architecture/concrete/primitives/line/LineDomainObject.ts
```diff
@@ -205,7 +205,7 @@ export abstract class LineDomainObject extends VisualDomainObject {
 
   public getAverageLength(): number {
     const { lineSegmentCount } = this;
-    if (lineSegmentCount <= 1) {
+    if (lineSegmentCount <= 0) {
```
nilscognite, line 208: Inconsequent, fixed

react-components/src/architecture/concrete/primitives/line/LineView.ts
```diff
@@ -138,19 +139,30 @@ export class LineView extends GroupThreeView<LineDomainObject> {
...
+          return undefined;
+        }
+      }
+      if (!closestFinder.isClosest(intersection)) {
         continue;
       }
       const objectIntersection: DomainObjectIntersection = {
         type: 'customObject',
-        point: intersection,
+        point: intersection.clone(),
```
nilscognite, line 160: Bug in the code discovered by the tests. No real bug, since the point was never used.

react-components/src/architecture/concrete/primitives/line/LineView.ts
```diff
@@ -218,16 +230,16 @@ export class LineView extends GroupThreeView<LineDomainObject> {
     if (pointCount < 3) {
       return undefined;
     }
+    const indices = domainObject.getTriangleIndexes();
```
nilscognite, line 233: Optimalization

react-components/src/architecture/concrete/primitives/line/LineView.ts
```diff
@@ -272,39 +284,33 @@ export class LineView extends GroupThreeView<LineDomainObject> {
...
-      return;
-    }
-    const spriteHeight = this.getTextHeight(style.relativeTextSize);
+    let spriteHeight = this.getTextHeight(style.relativeTextSize);
     if (spriteHeight <= 0) {
-      return;
+      spriteHeight = 1;
     }
-    const segmentCount = domainObject.lineSegmentCount - 1;
+    const unitSystem = getRoot(domainObject)?.unitSystem ?? new UnitSystem();
```
nilscognite, line 291: Make it more robust, so the domainObject doesn't need to be in the hierarchy. It is easier to test it

react-components/src/architecture/concrete/primitives/line/LineView.ts
```diff
@@ -272,39 +284,33 @@ export class LineView extends GroupThreeView<LineDomainObject> {
...
-      const distance = point1.distanceTo(point2);
-
-      center.copy(point1).add(point2).divideScalar(2);
-      center.applyMatrix4(CDF_TO_VIEWER_TRANSFORMATION);
-
-      const text = root.unitSystem.toStringWithUnit(distance, Quantity.Length);
-      const sprite = BoxView.createSprite(text, style, spriteHeight);
-      if (sprite === undefined) {
-        continue;
+    let prevPoint = domainObject.isClosed
```
nilscognite, line 293: Refactoring. Should have the same result

react-components/src/architecture/concrete/primitives/line/LineCreator.test.ts
christjt, line 1: To me this is more an integration test of MeasureLineDomainObject and LineCreator, since you are using real implementations of both. There is absolutly no problem with this, but this is also technically testing the MeasureLineDomainObject. So perhaps we should think about if this test should be placed somewhere else.

react-components/src/architecture/concrete/primitives/line/LineDomainObject.test.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    case PrimitiveType.Polyline:
+    case PrimitiveType.Polygon:
+      domainObject.points.push(new Vector3(0, 0, 0));
+      domainObject.points.push(new Vector3(1, 0, 0));
+      domainObject.points.push(new Vector3(1, 1, 0));
+      domainObject.points.push(new Vector3(0, 1, 0));
+      break;
+  }
+  // Translate the points
```
christjt, line 134: This comment is obvious from the code in my opinion.

christjt (approved):
Overall looks good to me. As mentioned in one of my comments, I would argue that you are integration testing here rather than unit testing, as a unit test would only be testing one thing in isolation. There is of course no problem with this, but we may want to concider changing the name of the test and perhaps putting it somewhere else. We have not really defined where are integration tests should be, so I am ok with deferring it.

react-components/src/architecture/concrete/primitives/line/LineDomainObject.test.ts
nilscognite: I remove this in the next PR

react-components/src/architecture/concrete/primitives/line/LineCreator.test.ts
nilscognite: In the original code, I used a mock for the `LineDomainObject`. But since I move most functionality down to the base classes, it was only exactly the same implementation of the `MeasureLineDomainObject  `left. I didn't see any point of duplicate this in my testes.

simenl (approved):
🦄

PR #5035, opened 2025-03-12
test(react-components): Various tests

nilscognite:
#### Type of change

Test 

#### Jira ticket :blue_book:

Just for getting up the code coverage


react-components/src/architecture/base/utilities/extensions/stringExtensions.ts
```diff
@@ -29,10 +30,14 @@ export function equalsIgnoreCaseAndSpace(value1: string, value2: string): boolea
       const lowerChar2 = char2.toLowerCase();
       if (lowerChar2 === lowerChar1) {
         j++;
+        found = true;
         break;
       }
       return false;
     }
+    if (!found) {
```
nilscognite, line 38: Found a bug during the test

react-components/src/architecture/concrete/primitives/box/BoxView.test.ts
```diff
@@ -0,0 +1,133 @@
...
+
+describe('BoxView', () => {
+  let domainObject: BoxDomainObject;
+  let view: BoxView;
+
+  beforeEach(() => {
+    domainObject = createBoxDomainObject();
+    view = new BoxView();
+    addView(domainObject, view);
+    expectChildrenLength(view, 2);
```
pramod-cog, line 33: Any reason we validate the children length here, it run for every test case and it would not give any results. Why is it here?

react-components/src/architecture/base/utilities/extensions/arrayExtensions.ts
```diff
@@ -2,40 +2,131 @@
...
+}
+
+/**
+ * Returns true if the two arrays contain the same elements (ignoring order).
+ *
+ * @param array1
+ * @param array2
+ * @returns
+ */
+export function containsTheSameSet<T>(array1: readonly T[], array2: readonly T[]): boolean {
```
pramod-cog, line 126: `return lodash.isEqual(lodash.sortBy(array1), lodash.sortBy(array2));`

react-components/src/architecture/base/utilities/extensions/arrayExtensions.ts
```diff
@@ -2,40 +2,131 @@
...
   }
 }
 
+/**
+ * Removes an element from an array. Return true if removed, otherwise false
+ *
+ * @param array
+ * @param element
+ */
 export function remove<T>(array: T[], element: T): boolean {
```
pramod-cog, line 61: `lodash.remove(array, (n) => n === elementToRemove);`

react-components/src/architecture/base/utilities/extensions/arrayExtensions.ts
```diff
@@ -2,40 +2,131 @@
...
   array.splice(0, array.length);
 }
 
+/**
+ * Copy array from another array.
+ *
+ * @param array
+ * @param from
+ */
 export function copy<T>(array: T[], from: T[]): void {
```
pramod-cog, line 27: `lodash.cloneDeep()`

react-components/src/architecture/concrete/primitives/box/BoxView.test.ts
nilscognite: No. I just move it now 33f2e30f2d87288c36f095b575f8474faa30ecf2

react-components/src/architecture/base/utilities/extensions/arrayExtensions.ts
```diff
@@ -2,40 +2,131 @@
...
   return true;
 }
+
+/**
+ * Removes an element by the index from an array.
+ *
+ * @param array
+ * @param index
+ */
+export function removeAt<T>(array: T[], index: number): void {
```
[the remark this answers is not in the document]
nilscognite: Using lodash here will not get down the complexity, so I leave it as it is. The code it partly taken by Pulse, and we agree in this before

react-components/src/architecture/base/utilities/extensions/arrayExtensions.ts
nilscognite: Using lodash here will not get down the complexity, so I leave it as it is. The code it partly taken by Pulse, and we agree in this before

react-components/src/architecture/base/utilities/extensions/arrayExtensions.ts
nilscognite: No. This is an in place copy. Deep clone is doing a very different thing

react-components/src/architecture/base/utilities/extensions/arrayExtensions.ts
```diff
@@ -2,40 +2,131 @@
...
  * Use there function in order top increase the readability of your code
  */
 
+/**
+ * Clears an array and keeps the array reference (i.e. does not create a new array).
+ *
+ * @param array
+ */
 export function clear<T>(array: T[]): void {
   array.splice(0, array.length);
```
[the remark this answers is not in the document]
nilscognite: Using lodash here will not get down the complexity, so I leave it as it is. The code it partly taken by Pulse, and we agree in this before

nilscognite:
> @nilscognite Regarding arrayExtension file, there are lodash function which can be used directly for the array, have listed them which I am aware of. I would recommend to use it rather than writing our own functions

I will have them. I have talk about in this code course and the reason way.

pramod-cog (dismissed):

react-components/src/architecture/base/utilities/extensions/stringExtensions.ts
simenl: I think the function will incorrectly return true if `value1` is a prefix (after ignoring case and space) of `value2`
```
equalsIgnoreCaseAndSpace('', 'abc')
equalsIgnoreCaseAndSpace('a', 'abc')
equalsIgnoreCaseAndSpace('A', 'abc')
equalsIgnoreCaseAndSpace('a b', 'abc')
```

To address this the remaining characters (from `j` and onward) in `value2` could be processed. If a non-space character is found, it should return false.

nilscognite: Nice catch! Fixed here: 43d33da1a9d9555e2af4244766bb17f81199ea51

pramod-cog (approved):

simenl (approved):
:unicorn:

PR #5041, opened 2025-03-12
test(react-components): Add test for the mathExtensions

nilscognite:
#### Type of change

Test 

## Description :pencil:

Just add a unit test

### DIFF
--- a/react-components/src/architecture/base/utilities/extensions/mathExtensions.test.ts
+++ b/react-components/src/architecture/base/utilities/extensions/mathExtensions.test.ts
@@ -1,0 +1,235 @@
+import { describe, expect, test } from 'vitest';
+import {
+  ceil,
+  compare,
+  floor,
+  forceAngleAround0,
+  forceBetween0AndPi,
+  forceBetween0AndTwoPi,
+  getRandomGaussian,
+  getRandomInt,
+  getRandomIntByMax,
+  isAbsEqual,
+  isBetween,
+  isEqual,
+  isEven,
+  isIncrement,
+  isInteger,
+  isOdd,
+  isZero,
+  round,
+  roundIncrement,
+  square
+} from './mathExtensions';
+
+describe('MathExtensions', () => {
+  describe('isZero', () => {
+    test('should be zero', () => {
+      expect(isZero(0)).toBe(true);
+    });
+    test('should not be zero', () => {
+      expect(isZero(0.0001)).toBe(false);
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


react-components/src/architecture/base/utilities/extensions/mathExtensions.ts
```diff
@@ -21,9 +21,9 @@ export function isEqual(x: number, y: number): boolean {
 }
 
 export function isAbsEqual(x: number, y: number, tolerance: number): boolean {
-  const error = x - y;
+  let error = x - y;
```
pramod-cog, line 24: Is there a reason not to use `Math.abs(x -y) < tolerance`?

nilscognite: By version is a little bit faster. But I have hanged it

pramod-cog (approved):
LGTM!

react-components/src/architecture/base/utilities/extensions/mathExtensions.test.ts
```diff
@@ -0,0 +1,235 @@
...
+  isIncrement,
+  isInteger,
+  isOdd,
+  isZero,
+  round,
+  roundIncrement,
+  square
+} from './mathExtensions';
+
+describe('MathExtensions', () => {
```
rbtcollins, line 25: I think these tests would be a great fit for https://fast-check.dev/

react-components/src/architecture/base/utilities/extensions/mathExtensions.ts
```diff
@@ -46,9 +42,16 @@ export function isInteger(value: number): boolean {
...
   return isInteger(value / increment);
 }
 
+export function isOdd(value: number): boolean {
+  return value % 2 !== 0;
+}
+
 export function isEven(value: number): boolean {
   return value % 2 === 0;
 }
```
rbtcollins, line 57: should these be throwing a domain error when given a non integer?

rbtcollins:
:unicorn: thank you for increasing the test coverage.

One note: your summary "test(react-components): Add test for the mathExtensions" is misleading: your patch adds isOdd and alters the impplementation of some extensions

react-components/src/architecture/base/utilities/extensions/mathExtensions.ts
nilscognite: Maybe - maybe not. This is the standard impl.

PR #5043, opened 2025-03-14
fix: layers drop down mounting outside of layers button parent

christjt:
#### Type of change

Bug 

## Description :pencil:

Fixes an issue where the drop down spawned from the Layers Button mounts outside of the dom hierachy of the component. In Unified-3d, this causes it to mount outside of the style scope which breaks styles.
Is needed for [BND3D-5620](https://cognitedata.atlassian.net/browse/BND3D-5620)

Added unit test.

[BND3D-5620]: https://cognitedata.atlassian.net/browse/BND3D-5620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react-components/src/components/RevealToolbar/LayersButton/LayersButton.test.tsx
+++ b/react-components/src/components/RevealToolbar/LayersButton/LayersButton.test.tsx
@@ -1,7 +1,7 @@
 /*!
  * Copyright 2025 Cognite AS
  */
-import { render } from '@testing-library/react';
+import { act, render, screen } from '@testing-library/react';
 import { describe, expect, test, vi } from 'vitest';
 import type { ReactElement, ReactNode } from 'react';
 import { LayersButton } from './LayersButton';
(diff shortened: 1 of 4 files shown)


react-components/vite.config.ts
```diff
@@ -30,6 +30,7 @@ export default defineConfig(({ command }) => {
     test: {
       include: ['src/**/*.{test,spec}.?(c|m)[jt]s?(x)'],
       environment: 'happy-dom',
+      globals: true,
```
christjt, line 33: Enabling this fixes an issue where the dom persists between specific test cases within a suite. This causes issues as tests assume clean dom on running.

react-components/src/components/RevealToolbar/LayersButton/LayersButton.tsx
```diff
@@ -31,11 +31,7 @@ export const LayersButton = ({
 
   return (
     <>
-      <SelectPanel
-        placement="right"
-        appendTo={'parent'}
```
christjt, line 36: There is most likely a bug here in cogs. Ironically removing the appendTo parent actually causes it to append to parent (which seems to be the default behaviour). @shubhsk88 Has been notified, though we are struggling to reproduce it outside of the reveal react components lib.

react-components/src/components/RevealToolbar/LayersButton/LayersButton.test.tsx
```diff
@@ -66,6 +66,26 @@ describe(LayersButton.name, () => {
...
+      <div data-test-id="layers-button">
+        <LayersButton {...defaultProps} />
+      </div>,
+      {
+        wrapper
+      }
+    );
+
+    const button = screen.getByRole('button', { name: 'Filter 3D resource layers' });
+    await act(async () => {
```
shubhsk88, line 80: You can utilize `userEvent` here instead of using `act`

christjt: Ah! Thanks. Why is that better?

shubhsk88: userEvent.click() works with async updates without requiring act() and the other reason `userEvent ` will mimic the user behavior like `button.click` doesn't handle focus while the `userEvent` will add the focus  as it should

react-components/vite.config.ts
pramod-cog: Nice!

anders-hopland (approved):
LGTM

anders-hopland (dismissed):
LGTM

anders-hopland (dismissed):

anders-hopland (approved):

psalaberria002 (approved):
:unicorn:

PR #5044, opened 2025-03-14
chore: add imports subpath for test utilities

christjt:
#### Type of change

Chore 

## Description :pencil:

Add imports subpath such that you don't need to use a relative import such as `../../../../tests/tests-utilities/...` but rather just `#test-utilities/...`


nilscognite (approved):
LGTM

larshagencognite (approved):
risk review ok

PR #5052, opened 2025-03-19
refactor(react-components): more test util import path replacements

haakonflatval-cognite:
#### Type of change

Refactor 

## Description :pencil:

Use the new test-utils import alias in some places that were missed in #5044 

In addition, I had to introduce a second alias, as the first one only covers `.ts`-files, but there is a `.tsx`-files in the utils that we also want to export. If there's a better way to support this than adding a second import, I'd be interested to know.

By running the tests


danpriori (approved):
LGTM

cemreyavuz (approved):
risk review ok

PR #5074, opened 2025-04-02
refactor(react-components): rewrite create*-functions as actual React components (BND3D-5504 part 2)

haakonflatval-cognite:
#### Type of change

Refactor 
#### Jira ticket :blue_book:

Depends on https://github.com/cognitedata/reveal/pull/5070

## Description :pencil:

A small refactoring of function calls creating components, to actual React components. This ensures that the components behave consistently e.g. with respect to hooks.

Note that this depends on #5070 , but it's not a tightly related change - however, they will generate some convoluted merge conflicts, which is why I put it as the base of this branch. 


react-components/src/architecture/concrete/terrain/geometry/RegularGrid2.ts
```diff
@@ -18,7 +18,7 @@ export class RegularGrid2 extends Grid2 {
   public readonly origin: Vector2;
   public readonly increment: Vector2;
 
-  private _buffer: Float32Array; // NaN value in this array means undefined node
+  private _buffer: Float32Array<ArrayBuffer>; // NaN value in this array means undefined node
```
haakonflatval-cognite, line 21: This broke in type checking for me locally. I'm suspecting that the type checker in CI doesn't catch this properly, but I include it anyway. It's an issue we should definitely check promptly

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -101,73 +101,80 @@ export const SettingsButton = ({
...
-    return createDivider(command);
+    return <DividerComponent key={command.uniqueId} command={command} />;
   }
   if (command instanceof SectionCommand) {
-    return createSection(command, t);
+    return <SectionComponent key={command.uniqueId} command={command} t={t} />;
   }
-  return createButton(command, t);
+
+  return <ButtonComponent key={command.uniqueId} command={command} t={t} />;
```
danpriori, line 122: I believe the linter will complain about not using the key as a argument/props of the ReactNodes as "key" doesnt exist on the component arguments type?
Related to all of these components above

danpriori (changes requested):
One comment only. The rest looks good 👍

react-components/src/components/Architecture/SettingsButton.tsx
haakonflatval-cognite: No, it doesn't - `key` is a special property, and React complains in runtime if it's not able to distinguish between generated components... I see I'm not explaining it very well - there are some docs here: https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

danpriori: Ah yeah, you're right. Forgot that. I saw my linter complaining as well in a web workspace but you're right. Nvmd.

danpriori (dismissed):
LGTM

danpriori (approved):
LGTM

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -177,30 +186,33 @@ function createToggle(command: BaseCommand, t: TranslateDelegate): ReactNode {
...
-      }}
-      trailingContent={<Switch checked={isChecked} disabled={!isEnabled} />}
-    />
+        if (isEnabled) {
+          command.invoke();
+        }
+      }}>
+      <Switch checked={isChecked} disabled={!isEnabled} />
+      <TextLabel text={label} />
+    </StyledToggleContainer>
```
frxstrem, line 197: What's the reason behind this change? It seems somewhat unrelated to the rest of the PR?

frxstrem:
🦄

PR #5091, opened 2025-04-10
test(react-components): Add unit test for Index2 and TrianglesBuffers

nilscognite:
#### Type of change
Test 

### Description:

- `Index2`: Unit test and change of 2 function names
- `TrianglesBuffers`: Unit test and change in one function (reused one function)
- `Vector3ArrayUtils`: Unit test and added handling of empty list of points.
- `PrimitiveTestUtil`: Here I added 2 more utility functions
- The **PointsUtils** is renamed to **Vector3ArrayUtils**


react-components/src/architecture/base/utilities/geometry/TrianglesBuffers.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      const normals = geometry.getAttribute('normal');
+      const uvs = geometry.getAttribute('uv');
+
+      expect(positions.count).toBe(2);
+      expect(normals.count).toBe(2);
+      expect(uvs).toBeUndefined();
+
+      expect(positions.array[5]).toBe(point2.z);
+      expect(normals.array[5]).toBe(normal2.z);
+    }
```
haakonflatval-cognite, line 54: Would be nice to check the other coordinates as well, and not just `z` (same in the other tests here)

react-components/src/architecture/base/utilities/geometry/TrianglesBuffers.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    buffer.addTriangle(0, 1, 2);
+    buffer.addTriangle(3, 4, 5);
+    buffer.addTriangle(6, 7, 8);
+
+    const geometry = buffer.createBufferGeometry();
+    const index = geometry.getIndex();
+    expect(index).toBeDefined();
+    if (index === null) {
+      return;
+    }
```
haakonflatval-cognite, line 93: Shouldn't the test fail if `index` is `null`?

react-components/src/architecture/base/utilities/geometry/TrianglesBuffers.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    const uvs = geometry.getAttribute('uv');
+    const index = geometry.getIndex();
+
+    expect(positions.count).toBe(4);
+    expect(normals.count).toBe(4);
+    expect(uvs.count).toBe(4);
+    expect(index).toBeDefined();
+    if (index === null) {
+      return;
+    }
```
haakonflatval-cognite, line 121: Same here?

haakonflatval-cognite (commented):
Looks good! Just a few comments

react-components/src/architecture/base/utilities/geometry/TrianglesBuffers.test.ts
nilscognite: I don't like to test entire three.js, therefor the last value only.

react-components/src/architecture/base/utilities/geometry/TrianglesBuffers.test.ts
nilscognite: Fixed

react-components/src/architecture/base/utilities/geometry/TrianglesBuffers.test.ts
nilscognite: Fixed

haakonflatval-cognite (dismissed):
LGTM :+1:

react-components/tests/tests-utilities/primitives/primitiveTestUtil.ts
```diff
@@ -72,6 +88,22 @@ export function equalsEuler(a: Euler, b: Euler, epsilon = EPSILON): boolean {
   return true;
 }
 
+export function equalsRange1(a: Range1, b: Range1, epsilon = EPSILON): boolean {
+  if (!a.isEmpty && !b.isEmpty) {
+    return true;
+  }
```
thorkildcognite, line 94: Doesn't this read as "if both a and be are not empty, return true"? That sounds unexpected? Shouldn't this be:
```
if (a.isEmpty && b.isEmpty) {
   return true
}
```
To say that two empty ranges are the same?

thorkildcognite:
Left a question. Apart from that it looked good.

haakonflatval-cognite (dismissed):
LGTM again :+1:

react-components/tests/tests-utilities/primitives/primitiveTestUtil.ts
nilscognite: Also updated an import statement when resolving merge conflict

haakonflatval-cognite (approved):
LGTM Again again :+1: :+1:

thorkildcognite:
risk review ok

PR #5098, opened 2025-04-14
refactor: rewrite create as components (BND3D-5504 part 2, second take)

haakonflatval-cognite:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

## Description :pencil:

A small refactoring of function calls creating components, to actual React components. This ensures that the components behave consistently e.g. with respect to hooks.

Note that this depends on https://github.com/cognitedata/reveal/pull/5070 , but it's not a tightly related change - however, they will generate some convoluted merge conflicts, which is why I put it as the base of this branch.

This is a recreation of https://github.com/cognitedata/reveal/pull/5074, because it got stuck in CI for unknown reasons (it seems to believe the PR was created or modified externally, which prohibits the use of the CI secret for fetching some of our dependencies, like Cogs.js). 


danpriori (approved):
LGTM

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -266,74 +281,78 @@ function createSlider(command: BaseSliderCommand, t: TranslateDelegate): ReactNo
...
+  color: var(--text-icon-muted, rgba(0, 0, 0, 0.55));
+  font-feature-settings: 'cv05' on;
+
+  /* Label/XSmall */
+  font-family: Inter;
+  font-size: 12px;
+  font-style: normal;
+  font-weight: 500;
+  line-height: 20px; /* 166.667% */
+  letter-spacing: -0.036px;
```
cemreyavuz, line 345: I believe this is the styles for the `Body` component from cogs. I believe we have it as a dependency for this package - can we use that instead?

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -266,74 +281,78 @@ function createSlider(command: BaseSliderCommand, t: TranslateDelegate): ReactNo
...
+`;
+
+const StyledToggleContainer = styled(Flex).attrs({
+  direction: 'row'
+})`
+  width: 100%;
+  justify-content: start;
+  gap: 8px;
+`;
```
cemreyavuz, line 354:
```suggestion
const StyledToggleContainer = styled.div`
  display: flex;
  width: 100%;
  justify-content: start;
  gap: 8px;
`;
```
`row` is default value for the `flex-direction` and it's not inherited as far as I know. We can simply define this as a `div`. It's a bit of a gray area, but if we're gonna override styles for cogs' `Flex` component, we can might as well define it as a `div`

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -266,74 +281,78 @@ function createSlider(command: BaseSliderCommand, t: TranslateDelegate): ReactNo
...
-  padding: 4px 8px;
+  gap: 4px;
   font-size: 14px;
 `;
 
 const StyledSlider = styled(Slider)`
   offset-anchor: right top;
   float: center;
   display: flex;
+
```
cemreyavuz, line 331:
```suggestion
```

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -266,74 +281,78 @@ function createSlider(command: BaseSliderCommand, t: TranslateDelegate): ReactNo
...
   }
-  return (
-    <FilterButton
-      key={uniqueId}
-      inputCommand={command}
-      placement={'bottom'}
-      usedInSettings={true}
-    />
-  );
+  return <FilterButton inputCommand={command} placement={'bottom'} usedInSettings={true} />;
```
cemreyavuz, line 314:
```suggestion
  return <FilterButton inputCommand={command} placement="bottom" usedInSettings />;
```
(extreme nitpick alert)
I'm not sure about the codestyle in this repo, but in general I think we often do things like these: 
- `"bottom"` instead of `{'bottom'}`
- `usedInSettings` instead of `usedInSettings={true}`

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -266,74 +281,78 @@ function createSlider(command: BaseSliderCommand, t: TranslateDelegate): ReactNo
...
 
-  return (
-    <DropdownButton
-      key={uniqueId}
-      inputCommand={command}
-      placement={'bottom'}
-      usedInSettings={true}
-    />
-  );
+  return <DropdownButton inputCommand={command} placement={'bottom'} usedInSettings={true} />;
```
cemreyavuz, line 297:
```suggestion
  return <DropdownButton inputCommand={command} placement="bottom" usedInSettings />;
```
(made a similar comment below)

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -266,74 +281,78 @@ function createSlider(command: BaseSliderCommand, t: TranslateDelegate): ReactNo
...
   });
   // @end
 
   if (!isVisible) {
     return null;
   }
 
-  return (
-    <DropdownButton
-      key={uniqueId}
```
cemreyavuz, line 286: What happened to the `key`s now? Are we passing these while using `DropdownButtonComponent` now?

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -177,30 +186,33 @@ function createToggle(command: BaseCommand, t: TranslateDelegate): ReactNode {
...
-      }}
-      trailingContent={<Switch checked={isChecked} disabled={!isEnabled} />}
-    />
+        if (isEnabled) {
+          command.invoke();
+        }
+      }}>
+      <Switch checked={isChecked} disabled={!isEnabled} />
+      <TextLabel text={label} />
+    </StyledToggleContainer>
```
cemreyavuz, line 197: This seem to change a lot. We don't seem to do the `setChecked(command.isChecked)` call anymore. And I didn't see any tests around this as well 🤔 What do we do here really?

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -101,73 +103,80 @@ export const SettingsButton = ({
...
-    setUniqueId(command.uniqueId);
   });
   // @end
 
   if (!isVisible) {
     return null;
   }
   const label = command.getLabel(t);
-  return <Menu.Section key={uniqueId} label={label} />;
+  return <StyledSectionHeader>{label} </StyledSectionHeader>;
```
cemreyavuz, line 161: We switch from `Menu` subcomponents to normal divs, are we sure about these changes? If they were used inside a `Menu`, this will mean a change in behaviour as well.

react-components/src/architecture/concrete/terrain/geometry/RegularGrid2.ts
```diff
@@ -18,7 +18,7 @@ export class RegularGrid2 extends Grid2 {
   public readonly origin: Vector2;
   public readonly increment: Vector2;
 
-  private _buffer: Float32Array; // NaN value in this array means undefined node
+  private _buffer: Float32Array<ArrayBuffer>; // NaN value in this array means undefined node
```
cemreyavuz, line 21: I didn't understand how this change is related to the PR, can you elaborate a bit?

react-components/src/components/Architecture/FilterButton.tsx
cemreyavuz, line 1: (I think) This change also doesn't match the PR title/description. Why do we remove the `uniqueId` here?

cemreyavuz (commented):
I had a few questions around how certain changes relate to the PR. It would also be nice to see tests related to the behaviour changes for `SectionComponent` and `ToggleComponent`.

react-components/src/architecture/concrete/terrain/geometry/RegularGrid2.ts
haakonflatval-cognite: Yes, I briefly mentioned this in the previous PR: https://github.com/cognitedata/reveal/pull/5074#discussion_r2024961805

react-components/src/components/Architecture/SettingsButton.tsx
haakonflatval-cognite: Right, Fredrik caught a similar issue in his risk review on the other, identical PR: https://github.com/cognitedata/reveal/pull/5074#discussion_r2037466261. This is largely a mistake on my end, this change should have been part of the previous PR in this chain, where we switched out the parent component from `Menu` to a custom component. Therefore we no longer want to use `Menu` subcomponents here

react-components/src/architecture/concrete/terrain/geometry/RegularGrid2.ts
cemreyavuz: Ah I see! I think it would be nice to create a task for fixing the typecheck for this package then - it's quite risky to continue working like this.

PR #5103, opened 2025-04-21
feat(react-components): add hybrid cad mappings support - step 6

danpriori:
#### Type of change

Feat  

This is a stacked PR:

* #5107
* #5105 
* #5106 
* #5101
* #5102 
* #5103  (**This PR**) 
* #5100
* #5104 
* #5099

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4989
https://cognitedata.atlassian.net/browse/BND3D-5576

https://cognitedata.atlassian.net/browse/

## Description :pencil:
- Break down other styling group hooks and add hybrid fdm instance style group
- Add calculate instance styling hook

## Checklist :ballot_box_with_check:

- [X] I am happy with this implementation.
- [X] I have performed a self-review of my own code.
- [X] I have added unit and visuals tests to prove that my feature works to the best of my ability.
- [ ] I have added documentation to new and changed elements; both public and internally shared ones
- [X] I have refactored the code for testability, readability and extendibility to the best of my ability.
- [X] I have listed the JIRA tasks covering remaining work or tech debt related to this PR in the description.

PR #5102, opened 2025-04-21
feat(react-components): add hybrid cad mappings support - step 5

danpriori:
#### Type of change

Feat  

This is a stacked PR:

* #5107
* #5105 
* #5106 
* #5101
* #5102  (**This PR**) 
* #5103 
* #5100
* #5104 
* #5099

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4989
https://cognitedata.atlassian.net/browse/BND3D-5576

https://cognitedata.atlassian.net/browse/

## Description :pencil:

- Add reveal 3d resource types for stylings, style translation and custom viewer for unit testing.
- Add hybrid cad asset mappings hook to be used by useCalculateInstanceStyling hook

## Checklist :ballot_box_with_check:

- [X] I am happy with this implementation.
- [X] I have performed a self-review of my own code.
- [X] I have added unit and visuals tests to prove that my feature works to the best of my ability.
- [ ] I have added documentation to new and changed elements; both public and internally shared ones
- [X] I have refactored the code for testability, readability and extendibility to the best of my ability.
- [X] I have listed the JIRA tasks covering remaining work or tech debt related to this PR in the description.

PR #5107, opened 2025-04-21
feat(react-components): add hybrid cad mappings support - step 1

danpriori:
#### Type of change

Feat  

This is a stacked PR:

* #5107 (**This PR**)
* #5105 
* #5106 
* #5101
* #5102 
* #5103 
* #5100
* #5104 
* #5099

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-4989
https://cognitedata.atlassian.net/browse/BND3D-5576

## Description :pencil:
- Add base types, typeguards, keys and combined asset fetching supports

- [X] I am happy with this implementation.
- [X] I have listed the JIRA tasks covering remaining work or tech debt related to this PR in the description.

### DIFF
--- a/react-components/src/architecture/concrete/pointsOfInterest/ads/createUpsertRequestFromPois.ts
+++ b/react-components/src/architecture/concrete/pointsOfInterest/ads/createUpsertRequestFromPois.ts
@@ -6,6 +6,7 @@ import { type ExternalId } from '../../../../data-providers/FdmSDK';
 import {
   type InstanceReference,
   isDmsInstance,
+  isHybridAssetCoreDmsInstance,
   isInternalId
 } from '../../../../utilities/instanceIds';
 import { type PointsOfInterestInstance, type PoiVisibility, type SceneState } from '../models';
@@ -54,6 +55,11 @@ function instanceReferenceToPoiExternalInstanceReference(
     return instanceRef;
   } else if (isDmsInstance(instanceRef)) {
     return { externalId: instanceRef.externalId, instanceSpace: instanceRef.space };
+  } else if (isHybridAssetCoreDmsInstance(instanceRef)) {
+    return {
+      externalId: instanceRef.assetInstanceId.externalId,
+      instanceSpace: instanceRef.assetInstanceId.space
+    };
   }
 
   return undefined;
--- a/react-components/src/architecture/concrete/pointsOfInterest/ads/createUpsertRequestFromPois.ts
+++ b/react-components/src/architecture/concrete/pointsOfInterest/ads/createUpsertRequestFromPois.ts
@@ -6,7 +6,6 @@ import { type ExternalId } from '../../../../data-providers/FdmSDK';
 import {
   type InstanceReference,
   isDmsInstance,
-  isHybridAssetMappingsInstance,
   isInternalId
 } from '../../../../utilities/instanceIds';
 import { type PointsOfInterestInstance, type PoiVisibility, type SceneState } from '../models';
--- a/react-components/src/architecture/concrete/pointsOfInterest/ads/createUpsertRequestFromPois.ts
+++ b/react-components/src/architecture/concrete/pointsOfInterest/ads/createUpsertRequestFromPois.ts
@@ -6,7 +6,7 @@ import { type ExternalId } from '../../../../data-providers/FdmSDK';
 import {
   type InstanceReference,
   isDmsInstance,
-  isHybridAssetCoreDmsInstance,
+  isHybridAssetMappingsInstance,
   isInternalId
 } from '../../../../utilities/instanceIds';
 import { type PointsOfInterestInstance, type PoiVisibility, type SceneState } from '../models';
@@ -55,11 +55,6 @@ function instanceReferenceToPoiExternalInstanceReference(
     return instanceRef;
   } else if (isDmsInstance(instanceRef)) {
     return { externalId: instanceRef.externalId, instanceSpace: instanceRef.space };
-  } else if (isHybridAssetCoreDmsInstance(instanceRef)) {
-    return {
-      externalId: instanceRef.assetInstanceId.externalId,
-      instanceSpace: instanceRef.assetInstanceId.space
-    };
   }
 
   return undefined;
--- a/react-components/src/components/CacheProvider/AnnotationModelUtils.test.ts
+++ b/react-components/src/components/CacheProvider/AnnotationModelUtils.test.ts
@@ -60,9 +60,13 @@ describe('AnnotationModelUtils', () => {
     test('should fetch assets for given asset references', async () => {
       const dmAssets: ExternalIdsResultList<AssetProperties> =
         createDMAssetMock('asset-external-id1');
+      const hybridAssets: ExternalIdsResultList<AssetProperties> =
+        createDMAssetMock('asset-hybrid-external-id1');
       const assetReferences: InstanceReference[] = [
         { id: 1 },
-        { externalId: 'asset-external-id1', space: 'asset-space' }
+        { externalId: 'asset-external-id1', space: 'asset-space' },
+        { assetInstanceId: { externalId: 'asset-hybrid-external-id1', space: 'asset-space' } },
+        { assetId: 1 }
       ];
       retrieveMock.mockResolvedValueOnce([assets[0]]);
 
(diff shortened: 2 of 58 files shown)


react-components/src/architecture/concrete/pointsOfInterest/ads/createUpsertRequestFromPois.ts
```diff
@@ -54,6 +55,11 @@ function instanceReferenceToPoiExternalInstanceReference(
     return instanceRef;
   } else if (isDmsInstance(instanceRef)) {
     return { externalId: instanceRef.externalId, instanceSpace: instanceRef.space };
+  } else if (isHybridAssetCoreDmsInstance(instanceRef)) {
```
anders-hopland, line 58: Points of Interest only points to assets, not asset mappings. We do thus not need to check for hybrid here

react-components/src/utilities/instanceIds/typeGuards.ts
```diff
@@ -32,3 +36,15 @@ export function isDmsInstance(instance: unknown): instance is DmsUniqueIdentifie
     'space' in instance
   );
 }
+
+export function isHybridAssetCoreDmsInstance(
```
anders-hopland, line 40: Can we rename this to reflect that this is a hybrid asset mapping and not a hybrid asset, as the asset itself is either classic or CDM

anders-hopland (changes requested):
Looks mostly good, only minor comments. Good work 💯

react-components/src/utilities/instanceIds/typeGuards.ts
danpriori: how about `isHybridAssetMappingsInstance` ?

anders-hopland: I like that one 💯

danpriori: done https://github.com/cognitedata/reveal/pull/5107/commits/731a34c4f2510c4f6a6808dc795e1024378352a1

react-components/src/architecture/concrete/pointsOfInterest/ads/createUpsertRequestFromPois.ts
danpriori: done https://github.com/cognitedata/reveal/pull/5107/commits/731a34c4f2510c4f6a6808dc795e1024378352a1

react-components/src/utilities/instanceIds/types.ts
```diff
@@ -5,5 +5,10 @@ import { type IdEither } from '@cognite/sdk';
...
 export type FdmInstanceWithView = DmsUniqueIdentifier & { view: Source };
-
-export type InstanceReference = IdEither | DmsUniqueIdentifier;
+export type AssetInstanceReference = { assetId: number };
+export type AssetHybridInstanceReference = { assetInstanceId: DmsUniqueIdentifier };
+export type InstanceReference =
+  | IdEither
+  | DmsUniqueIdentifier
+  | AssetHybridInstanceReference
+  | AssetInstanceReference;
```
haakonflatval-cognite, line 14: Do we really need the `AssetInstanceReference` here? :thinking: The `IdEither` type already has a `{ id: number; }` variant which is intended for classic asset ids.

I also wonder whether we need the `AssetHybridInstanceReference` here as well, since we have the `DMSUniqueIdentifier` type already? Is there a reason we need to distinguish between those two?

danpriori: I believe that the most that can be done here is to remove `AssetHybridInstanceReference ` and `AssetInstanceReference ` from the `InstanceReference` types and directly use the types in places where hybrid and classic references are involved for mapping, because we need to have this distinction in all subsequent steps that go from `react-components` to the App (Search UI) that uses the mappings and when it returns to `react-components` to calculate the cad styling, we check again if the References (Classic, DMS, Hybrid) have mapping and if we need to get it from the cache. If I remove this distinction at the beginning of the cycle, all subsequent codes that request the mapping again will think that the mapping is via DMS and will not find it, because the mapping is via 3d api classic. 

You can see in the file:

https://github.com/cognitedata/reveal/blob/fc9d20b65ea01e29eb8ba597e4f4bbe4fe593341/react-components/src/components/Reveal3DResources/hooks/useCalculateInstanceStyling.ts

at the end when you have already retrieved the mapping via hybrid and from then on you will only show the results in the app, then the code combines all the mappings into a single type and no longer needs to have this differentiation.

react-components/src/components/CacheProvider/AnnotationModelUtils.test.ts
```diff
@@ -71,13 +75,54 @@ describe('AnnotationModelUtils', () => {
...
+        ...item,
+        properties: item.properties[CORE_DM_SPACE][COGNITE_ASSET_VIEW_VERSION_KEY]
+      }));
+      const expectedAssets = [...dmExpectedAssets];
+
+      expect(retrieveMock).not.toHaveBeenCalledWith([assetReferences[0]], {
+        ignoreUnknownIds: true
+      });
+      expect(result).toEqual(expectedAssets);
+    });
```
haakonflatval-cognite, line 125: Should we assert that we call the intended endpoint here as well? (And also - the first `expect` should probably be `expect(retrieveMock).not.toHaveBeenCalled()`.

haakonflatval-cognite (changes requested):
Some more comments for the journey

react-components/src/components/CacheProvider/AnnotationModelUtils.test.ts
danpriori: I think we shouldnt care if some specific endpoint is called in this case. The important is if the assets were returned as expected, isnt?
Also, if I assert the dms endpoint, then or I have to use the postMock for the sdk.post (which is too deeper for the unit test purpose I think) or I have to remove the fdmSDK instantiation from inside the fetchAssetsForDmsIds and fetchAssetsForAssetReferences to be a dependency injection to be able to mock it more easily and adapt the places that calls the fetchAssetsForAssetReferences

react-components/src/components/CacheProvider/AnnotationModelUtils.ts
```diff
@@ -56,10 +62,20 @@ export async function fetchAssetsForAssetReferences(
   assetIds: InstanceReference[],
   sdk: CogniteClient
 ): Promise<AssetInstance[]> {
-  const [classicIds, dmIds] = partition(assetIds, isIdEither);
+  const [classicIds] = partition(assetIds, isIdEither);
+  const [dmIds] = partition(assetIds, isDmsInstance);
+  const [hybridIds] = partition(assetIds, isHybridAssetMappingsInstance);
+  const [assetIdReferences] = partition(assetIds, isAssetInstanceReference);
```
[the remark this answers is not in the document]
danpriori: It was my first implementation but I thought that using partition would be also elegant as it was used here originally 🙈 .
but yeah, returned back using `filter`:
https://github.com/cognitedata/reveal/pull/5107/commits/544f0d18bb0ed295d4d52de816e13a6d4fcc492c

react-components/src/components/CacheProvider/types.ts
```diff
@@ -6,13 +6,18 @@ import {
...
+  type AssetMapping3D,
 } from '@cognite/sdk';
 import { type Source, type DmsUniqueIdentifier } from '../../data-providers/FdmSDK';
 import { type AssetAnnotationImage360Info, type DataSourceType } from '@cognite/reveal';
 import { type Vector3 } from 'three';
 import { type AssetInstance } from '../../utilities/instances';
 
+export type NodeAssetMappingResult = { node?: Node3D; mappings: CdfAssetMapping[] };
+
+export type AssetMapping = AssetMapping3D;
```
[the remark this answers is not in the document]
danpriori: Nop. It is an old definition before CdfAssetMapping. But we have to convert it right after fetching from the APIs. Removed and adapted.
https://github.com/cognitedata/reveal/pull/5107/commits/544f0d18bb0ed295d4d52de816e13a6d4fcc492c

react-components/src/components/CacheProvider/types.ts
```diff
@@ -58,6 +63,8 @@ export type FdmKey = `${string}/${string}`;
 export type ModelTreeIndexKey = `${ModelId}/${RevisionId}/${TreeIndex}`;
 export type ModelAssetIdKey = `${ModelId}/${RevisionId}/${AssetId}`;
 
+export type ModelDMSUniqueInstanceKey = `${ModelId}/${RevisionId}/${string}/${string}`;
+
```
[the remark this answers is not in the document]
danpriori: It is model + instance unique reference, as the api request is per model/revision.
https://github.com/cognitedata/reveal/pull/5107/commits/544f0d18bb0ed295d4d52de816e13a6d4fcc492c

PR #5121, opened 2025-04-29
fix: lint setup

christjt:
#### Type of change

Chore 

## Description :pencil:

Lint setup has been broken for a little while which was not caught. This PR recovers the lint command as well as fixing all lint errors and warnings that accumulated while it was broken.

For review it is recommended to look at each commit in separation to gain ordering of the fix:

- Nr 1: fix the lint command
- Nr 2: apply eslint auto-fix
- Nr 3: manual intervention
- Nr 4: disabling of eslint for certain edge cases related to const inferring types which is extremly cumbersome to type out.
- Nr 5: set max allowed lint warnings to 0.

### DIFF
--- a/react-components/.eslintrc.json
+++ b/react-components/.eslintrc.json
@@ -77,7 +77,8 @@
     },
     {
       "files": [
-        "vite.config.ts"
+        "vite.config.ts",
+        "playwright.config.ts"
       ],
       "env": {
         "node": true
--- a/react-components/package.json
+++ b/react-components/package.json
@@ -25,7 +25,7 @@
     "tsc": "tsc",
     "typecheck:source": "tsc --noEmit --project tsconfig.build.json",
     "typecheck:test": "tsc --noEmit --project tsconfig.test.json",
-    "lint": "eslint --cache --ext=.tsx,.ts .",
+    "lint": "eslint --cache --max-warnings 0 --ext=.tsx,.ts .",
     "test": "vitest run --dom --silent",
     "test:coverage": "vitest run --dom --coverage --silent",
     "test:playwright": "playwright test",
--- a/react-components/package.json
+++ b/react-components/package.json
@@ -25,7 +25,7 @@
     "tsc": "tsc",
     "typecheck:source": "tsc --noEmit --project tsconfig.build.json",
     "typecheck:test": "tsc --noEmit --project tsconfig.test.json",
-    "lint": "eslint . --cache",
+    "lint": "eslint --cache --ext=.tsx,.ts .",
     "test": "vitest run --dom --silent",
     "test:coverage": "vitest run --dom --coverage --silent",
     "test:playwright": "playwright test",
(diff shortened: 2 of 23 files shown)


react-components/.eslintrc.json
```diff
@@ -77,7 +77,8 @@
     },
     {
       "files": [
-        "vite.config.ts"
+        "vite.config.ts",
```
nilscognite, line 80: This file is just extremely cryptical for me. What does files means here? Files not using lint at all?

react-components/package.json
```diff
@@ -25,7 +25,7 @@
     "tsc": "tsc",
     "typecheck:source": "tsc --noEmit --project tsconfig.build.json",
     "typecheck:test": "tsc --noEmit --project tsconfig.test.json",
-    "lint": "eslint . --cache",
+    "lint": "eslint --cache --max-warnings 0 --ext=.tsx,.ts .",
```
nilscognite, line 28: How did the **ext** argument disappear in the first place?

react-components/src/data-providers/core-dm-provider/cadAndPointCloudAndImage360AssetQuery.ts
```diff
@@ -176,6 +176,7 @@ export function cadAndPointCloudAndImage36AssetQuery(
   } as const satisfies Omit<QueryRequest, 'parameters' | 'cursors'>;
 }
 
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
nilscognite, line 179: Do we really need this? Is it possible to add this the global rule file instead?

nilscognite (dismissed):
I approved, but had some questions. By the way, very nice of you to fix this so fast!

react-components/.eslintrc.json
christjt: If you want more detail I recommend the eslint documentation. TLDR; is that files here have an override. These two files have a lint overrides because they are configuration files which are not run in the browser, but they run on your machine and so I tell the linter that they are in a local enviroment instead of the browser (which gives it other access such as access to files on your machine f.ex.).

react-components/package.json
christjt: I have no idea. Maybe I made a mistake some time in the past...

react-components/src/data-providers/core-dm-provider/cadAndPointCloudAndImage360AssetQuery.ts
christjt: Well the eslint rule is very good rule so I don't want to disable it globally. I think we should actually add the return type here, but the type is very large and I just want to get linting up and running again. The original author should fix this. I believe that is @anders-hopland.

nilscognite (approved):

psalaberria002 (approved):
:unicorn:

PR #5126, opened 2025-05-04
fix(react-component): Fix various issues with the DropdownButton not used in settings (standalone in a toolbar) [BND3D-5561]

nilscognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5561

## Description :pencil:

This bug was caused by some other refactoring. It is only in the user interface code, not in the business  layer.

I have refactor the `DropdownButton` component so it is clear by moving some code to inner component. IsVisible is used in the `DropdownButton`, and removed from the inner components.

Now the option closes when selected, as it normally should do for option menus of this sort.

Since it wasn't under unit test, I also added unit test for this part. I have used some time to understand making test for react-components, and I think the test should cover the feature in this component. The test is for usage when **usedInSettings = false**, the opposite is already make a unit test for.

image

1. `yarn storybook`
2. Open **Architecture/Main**
3. Check the top toolbar as image above shows
4. The bug was that that the option was not close after selection

### DIFF
--- a/react-components/src/components/Architecture/DropdownButton.tsx
+++ b/react-components/src/components/Architecture/DropdownButton.tsx
@@ -78,6 +78,7 @@ const DropdownElement = ({
 
   const label = command.getLabel(t);
   const selectedLabel = command.selectedChild?.getLabel(t);
+  const isDisabled = label === undefined || isOpen;
   const OpenButtonIcon = isOpen ? ChevronUpIcon : ChevronDownIcon;
   return (
     <Menu
@@ -90,7 +91,7 @@ const DropdownElement = ({
       renderTrigger={(props: any) => (
         <CogsTooltip
           content={<LabelWithShortcut label={label} command={command} />}
-          disabled={label === undefined || isOpen}
+          disabled={isDisabled}
           enterDelay={TOOLTIP_DELAY}
           placement={getTooltipPlacement(placement)}>
           <Button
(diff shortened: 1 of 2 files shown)


react-components/src/components/Architecture/DropdownButton.tsx
```diff
@@ -35,74 +35,62 @@ export const DropdownButton = ({
...
         setOpen(open);
       }}
+      open={isOpen}
       placement={'bottom-start'}
       disableCloseOnClickInside
       renderTrigger={(props: any) => (
         <CogsTooltip
           content={<LabelWithShortcut label={label} command={command} />}
-          disabled={label === undefined}
+          disabled={label === undefined || isOpen}
```
danpriori, line 93: Maybe is it better to use a boolean const that check both label and isOpen together which is defined above this jsx? For readability, I mean.

Perhaps:
```typescript
const isDisabled = label === undefined || isOpen;
```

danpriori (approved):
LGTM. Just a suggestion. The rest looks good.

react-components/src/components/Architecture/DropdownButton.tsx
nilscognite: Fixed here a49540d5a42e01811e085b446101c7770ef5b402

PR #5129, opened 2025-05-05
test(react-components): Unit test for some commands

nilscognite:
#### Type of change

Test

## Description :pencil:

**Unit test only.**

* `CopyToClipboardCommand`
* `DeleteDomainObjectCommand`
* `ToggleMetricUnitsCommand` (update only since I had this from before)
* `SettingsCommand`
* `UndoCommand`

**Other things:**

* Made `GeneratorUtils `as a separete file and made unit test
* Move `EventChangeTester `to test folder, and fix a bug in `EventChangeTester `(used in testing only)
* Change so `ToggleMetricUnitsCommand `also notify the root.


react-components/src/architecture/base/concreteCommands/ToggleMetricUnitsCommand.test.ts
```diff
@@ -33,4 +36,17 @@ describe(ToggleMetricUnitsCommand.name, () => {
     expect(command.isChecked).toBe(false);
     expect(renderTarget.rootDomainObject.unitSystem.isMetric).toBe(false);
   });
+
+  test('should switch from metric unit and imperial and back and check if domain objects are notified twice', async () => {
```
pramod-cog, line 40:
```suggestion
  test('should toggle between metric and imperial units and notify domain objects twice', async () => {
```

react-components/src/architecture/base/utilities/extensions/generatorUtils.test.ts
```diff
@@ -0,0 +1,25 @@
+import { describe, expect, test } from 'vitest';
+import { count, first, last } from './generatorUtils';
+
+describe('generatorUtils', () => {
+  test('should count', () => {
+    expect(count(getPositiveNumbers(0))).toBe(0);
+    expect(count(getPositiveNumbers(5))).toBe(5);
+  });
+
+  test('should get first', () => {
```
pramod-cog, line 10:
```suggestion
  test('should return the first item from a generator', () => {
```

react-components/src/architecture/base/utilities/extensions/generatorUtils.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    expect(first(getPositiveNumbers(0))).toBeUndefined();
+    expect(first(getPositiveNumbers(5))).toBe(0);
+  });
+
+  test('should get last', () => {
```
pramod-cog, line 15:
```suggestion
  test('should return the last item from a generator', () => {
```

react-components/src/architecture/base/utilities/extensions/generatorUtils.ts
```diff
@@ -0,0 +1,23 @@
...
+export function last<T>(iterable: Generator<T>): T | undefined {
+  let lastItem: T | undefined;
+  for (const item of iterable) {
+    lastItem = item;
+  }
+  return lastItem;
+}
+
+export function first<T>(iterable: Generator<T>): T | undefined {
+  return iterable.next().value;
```
pramod-cog, line 22: I have question on the `Generator` here, would it be right to check for done to ensure this is handling empty generator like below?
```
const { value, done } = iterable.next();
return done ? undefined : value;
```

nilscognite: The done variable is sometimes undefined. I don't understand why. But this code is working and I have used this several other places.

pramod-cog (approved):
LGTM!

react-components/src/architecture/base/concreteCommands/SettingsCommand.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    const command = new SettingsCommand();
+    expect(isEmpty(command.tooltip)).toBe(false);
+    expect(command.icon).toBe('Settings');
+    expect(command.isEnabled).toBe(true);
+  });
+
+  test('should have all children', async () => {
+    const command = new SettingsCommand(true, true);
+    expect(command.hasChildren).toBe(true);
+    expect(command.children.length).toBe(19);
```
anders-hopland, line 17: I would prefer to not have this hardcoded to 19 as it migt change if we add more commands. Would it be possible to calculate this value somehow?

react-components/src/architecture/base/concreteCommands/ToggleMetricUnitsCommand.test.ts
```diff
@@ -1,27 +1,30 @@
...
 
   beforeEach(() => {
+    renderTarget = createFullRenderTargetMock();
     command = new ToggleMetricUnitsCommand();
     command.attach(renderTarget);
   });
 
-  test('Should have have icons, tooltip and toggle', () => {
-    expect(command.icon).not.toBe('');
+  test('should have following default behavior', async () => {
```
anders-hopland, line 19: Can we keep the old description for the test or add a more descriptive one? E.g. "Should by default have Ruler icon, tooltip and toggle" or something like that?
I would prefer this for other tests having `should have following default behavior` as well

anders-hopland (commented):
A few comments, other than that it looks good

react-components/src/architecture/base/concreteCommands/SettingsCommand.test.ts
nilscognite: No. The only way is to look in the code

react-components/src/architecture/base/concreteCommands/ToggleMetricUnitsCommand.test.ts
nilscognite: My personal view of this is that is doesn't add anything, since the code is self explaining what it test. So comments about this is redundant and will get outdated for each change.

nilscognite: I did this, but for me my code will end up to be inconsistent, since I have used this comment other places.

pramod-cog (approved):

anders-hopland:
🦄

PR #5120, opened 2025-04-29
test(react-components): Make unit test for example domain object

nilscognite:
#### Type of change
Test 

## Description :pencil:

Add unit test for:

* `ExampleDomainObject`
* `ExampleRenderStyle`
* `ExampleView`

The example code is not in production. It is used for training and a good starting point for new domain objects.

Also fix some other similar tests and use a new function `PanelInfo.getItemsByQuantity` to avoid code duplications

PR #5131, opened 2025-05-07
fix(react-components): disable PoI-comment button on empty content

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5422

## Description :pencil:

Disable the PoI comment post-button when the comment content is empty. 

In Fusion, main 3D page in Search

- Build this and link it to Fusion
- Go to the main 3D page in Search
- Open any scene
- Create a new PoI (or select one already created)
- In the right side panel, notice that the comment text area is empty and the "Send"-button underneath is disabled
- Type in the comment text area
- Notice that the "Send"-button becomes enabled once the text area has text in it

### DIFF
--- a/react-components/src/architecture/base/commands/BaseInputCommand.test.ts
+++ b/react-components/src/architecture/base/commands/BaseInputCommand.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, test, vi } from 'vitest';
+import { BaseInputCommand } from './BaseInputCommand';
+import { TestInputCommand } from '#test-utils/architecture/commands/TestInputCommand';
+
+describe(BaseInputCommand.name, () => {
+  test('getters/setters work as expected', () => {
+    const mockContent = 'mock-content';
+    const mockOnCancel = vi.fn();
+    const mockCommandUpdateListener = vi.fn();
+
+    const command = new TestInputCommand();
+
+    command.addEventListener(mockCommandUpdateListener);
+
+    expect(command.content).toBe('');
+    expect(command.onCancel).toBe(undefined);
+
+    command.content = mockContent;
+    expect(mockCommandUpdateListener).toHaveBeenCalled();
+    expect(command.content).toBe(mockContent);
+
+    command.onCancel = mockOnCancel;
+    expect(command.onCancel).toBe(mockOnCancel);
+  });
+
+  test('post button should always be enabled', () => {
+    const command = new TestInputCommand();
+
+    expect(command.isPostButtonEnabled).toBeTruthy();
+
+    command.content = 'some-content';
+
+    expect(command.isPostButtonEnabled).toBeTruthy();
+
+    command.content = '';
+
+    expect(command.isPostButtonEnabled).toBeTruthy();
+  });
+});
--- a/react-components/src/architecture/base/commands/BaseInputCommand.ts
+++ b/react-components/src/architecture/base/commands/BaseInputCommand.ts
@@ -5,11 +5,7 @@ import { type TranslationInput } from '../utilities/TranslateInput';
 import { RenderTargetCommand } from './RenderTargetCommand';
 
 export abstract class BaseInputCommand extends RenderTargetCommand {
-  protected _placeholder?: TranslationInput;
-  protected _content?: string;
-  protected _okButtonLabel?: TranslationInput;
-
-  protected _onFinish?: () => void;
+  protected _content: string = '';
   protected _onCancel?: () => void;
 
   public getCancelButtonLabel(): TranslationInput | undefined {
@@ -17,30 +13,26 @@ export abstract class BaseInputCommand extends RenderTargetCommand {
   }
 
   public abstract getPostButtonLabel(): TranslationInput | undefined;
-
   public abstract getPlaceholder(): TranslationInput | undefined;
 
-  public get onFinish(): (() => void) | undefined {
-    return this._onFinish;
-  }
-
-  public set onFinish(onFinish: () => void) {
-    this._onFinish = onFinish;
+  public get onCancel(): (() => void) | undefined {
+    return this._onCancel;
   }
 
   public set onCancel(onCancel: () => void) {
     this._onCancel = onCancel;
   }
 
-  public get onCancel(): (() => void) | undefined {
-    return this._onCancel;
+  public get content(): string {
+    return this._content;
   }
 
-  invokeWithContent(content: string): boolean {
+  public set content(content: string) {
     this._content = content;
+    this.update();
+  }
 
-    const invokeResult = this.invoke();
-    this._content = '';
-    return invokeResult;
+  public get isPostButtonEnabled(): boolean {
+    return true;
   }
 }
--- a/react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
+++ b/react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
@@ -1,0 +1,93 @@
+import { beforeAll, describe, expect, test, vi } from 'vitest';
+import { CreatePoiCommentCommand } from './CreatePoiCommentCommand';
+import { getTranslationKeyOrString } from '#test-utils/architecture/getTranslationKeyOrString';
+
+import { type PointOfInterest, PointsOfInterestStatus } from './types';
+import { PointsOfInterestDomainObject } from './PointsOfInterestDomainObject';
+import { Mock } from 'moq.ts';
+import { type RevealRenderTarget } from '../../base/renderTarget/RevealRenderTarget';
+import { type RootDomainObject } from '../../base/domainObjects/RootDomainObject';
+import { type CommandsController } from '../../base/renderTarget/CommandsController';
+
+const TEST_POINT_OF_INTEREST: PointOfInterest<string> = {
+  properties: {
+    name: 'poi_name',
+    positionX: 1,
+    positionY: 2,
+    positionZ: 3,
+    scene: {
+      externalId: 'scene_external_id',
+      space: 'scene_space'
+    },
+    sceneState: {}
+  },
+  id: 'poi_id',
+  status: PointsOfInterestStatus.Default
+} as const;
+
+describe(CreatePoiCommentCommand.name, () => {
+  beforeAll(() => {
+    vi.clearAllMocks();
+  });
+
+  test('has correct initial content', () => {
+    const poiCommentCommand = new CreatePoiCommentCommand(TEST_POINT_OF_INTEREST);
+
+    expect(getTranslationKeyOrString(poiCommentCommand.getCancelButtonLabel())).toBe('CANCEL');
+    expect(getTranslationKeyOrString(poiCommentCommand.getPostButtonLabel())).toBe('SEND');
+    expect(getTranslationKeyOrString(poiCommentCommand.getPlaceholder())).toBe(
+      'COMMENT_PLACEHOLDER'
+    );
+  });
+
+  test('post button is disabled when content is empty', () => {
+    const poiCommentCommand = new CreatePoiCommentCommand(TEST_POINT_OF_INTEREST);
+
+    expect(poiCommentCommand.content).toBe('');
+    expect(poiCommentCommand.isPostButtonEnabled).toBeFalsy();
+
+    poiCommentCommand.content = 'non-empty-content';
+
+    expect(poiCommentCommand.isPostButtonEnabled).toBeTruthy();
+  });
+
+  test('invoking command posts comment on Poi', () => {
+    const mockPostComment = vi.fn().mockReturnValue(Promise.resolve());
+    const mockDomainObjectNotify = vi.fn();
+
+    const mockPointsOfInterestDomainObject = new Mock<PointsOfInterestDomainObject<string>>()
+      .setup((p) => p.postCommentForPoi)
+      .returns(mockPostComment)
+      .setup((p) => p.notify)
+      .returns(mockDomainObjectNotify)
+      .object();
+
+    const mockRenderTarget = new Mock<RevealRenderTarget>()
+      .setup((p) => p.rootDomainObject)
+      .returns(
+        new Mock<RootDomainObject>()
+          .setup((p) => p.getDescendantByType(PointsOfInterestDomainObject))
+          .returns(mockPointsOfInterestDomainObject)
+          .object()
+      )
+      .setup((p) => p.commandsController)
+      .returns(
+        new Mock<CommandsController>()
+          .setup((p) => p.update)
+          .returns(vi.fn())
+          .object()
+      )
+      .object();
+
+    const commentContent = 'comment-content';
+
+    const poiCommentCommand = new CreatePoiCommentCommand(TEST_POINT_OF_INTEREST);
+
+    poiCommentCommand.attach(mockRenderTarget);
+
+    poiCommentCommand.content = commentContent;
+    poiCommentCommand.invoke();
+
+    expect(mockPostComment).toHaveBeenCalledWith(TEST_POINT_OF_INTEREST, commentContent);
+  });
+});
--- a/react-components/src/components/Architecture/InputField.test.tsx
+++ b/react-components/src/components/Architecture/InputField.test.tsx
@@ -1,0 +1,144 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { InputField } from './InputField';
+import { TestInputCommand } from '#test-utils/architecture/commands/TestInputCommand';
+import { act, render } from '@testing-library/react';
+import { RevealRenderTarget } from '../../architecture';
+import { type PropsWithChildren, type ReactElement } from 'react';
+import { viewerMock } from '#test-utils/fixtures/viewer';
+import { sdkMock } from '#test-utils/fixtures/sdk';
+import { ViewerContextProvider } from '../RevealCanvas/ViewerContext';
+
+import userEvent from '@testing-library/user-event';
+import assert from 'assert';
+
+const ARBITRARY_PLACEMENT = 'right';
+
+describe(InputField.name, () => {
+  let renderTargetMock: RevealRenderTarget;
+  let wrapper: (props: PropsWithChildren) => ReactElement;
+
+  beforeEach(() => {
+    renderTargetMock = new RevealRenderTarget(viewerMock, sdkMock);
+
+    wrapper = ({ children }: PropsWithChildren): ReactElement => (
+      <ViewerContextProvider value={renderTargetMock}>{children}</ViewerContextProvider>
+    );
+  });
+
+  test('should render an empty text field with a post-button by default ', () => {
+    const postButtonLabel = 'post-label-text';
+    const command = new TestInputCommand({ postButtonLabel });
+
+    const { container } = render(
+      <InputField inputCommand={command} placement={ARBITRARY_PLACEMENT} />,
+      { wrapper }
+    );
+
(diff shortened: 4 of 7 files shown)


react-components/src/architecture/base/commands/BaseInputCommand.ts
```diff
@@ -5,42 +5,34 @@ import { type TranslationInput } from '../utilities/TranslateInput';
 import { RenderTargetCommand } from './RenderTargetCommand';
 
 export abstract class BaseInputCommand extends RenderTargetCommand {
-  protected _placeholder?: TranslationInput;
-  protected _content?: string;
-  protected _okButtonLabel?: TranslationInput;
```
haakonflatval-cognite, line 10: This field turned out to be unused

react-components/src/architecture/base/commands/BaseInputCommand.ts
```diff
@@ -5,42 +5,34 @@ import { type TranslationInput } from '../utilities/TranslateInput';
...
-
-  protected _onFinish?: () => void;
```
haakonflatval-cognite, line 12: The fields `placeholder`, `okButtonLabel` and `onFinish` turned out to be unused/unneeded in this context, so they have been removed altogether

react-components/src/architecture/base/commands/BaseInputCommand.test.ts
```diff
@@ -0,0 +1,39 @@
+import { describe, expect, test, vi } from 'vitest';
+import { BaseInputCommand } from './BaseInputCommand';
+import { TestInputCommand } from '#test-utils/architecture/commands/TestInputCommand';
+
+describe(BaseInputCommand.name, () => {
+  test('getters/setters work as expected', () => {
+    const mockContent = 'mock-content';
+    const mockOnCancel = vi.fn();
+    const mockCommandUpdateListener = vi.fn();
```
nilscognite, line 9: Remove this part (mockCommandUpdateListener) of the test. It is tested in the general function \(I have a PR on this now)

react-components/src/architecture/base/commands/BaseInputCommand.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    expect(command.content).toBe(mockContent);
+
+    command.onCancel = mockOnCancel;
+    expect(command.onCancel).toBe(mockOnCancel);
+  });
+
+  test('post button should always be enabled', () => {
+    const command = new TestInputCommand();
+
+    expect(command.isPostButtonEnabled).toBeTruthy();
```
nilscognite, line 29: This test is just wrong. isPostButtonEnabled has nothing to do with content anyway. Use setPostButtonEnabled instead and check if is enabled/disabled.

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
```diff
@@ -0,0 +1,92 @@
...
+import { getTranslationKeyOrString } from '#test-utils/architecture/getTranslationKeyOrString';
+
+import { type PointOfInterest, PointsOfInterestStatus } from './types';
+import { PointsOfInterestDomainObject } from './PointsOfInterestDomainObject';
+import { Mock } from 'moq.ts';
+import { type RevealRenderTarget } from '../../base/renderTarget/RevealRenderTarget';
+import { type RootDomainObject } from '../../base/domainObjects/RootDomainObject';
+import { type CommandsController } from '../../base/renderTarget/CommandsController';
+
+const TEST_POINT_OF_INTEREST: PointOfInterest<string> = {
```
nilscognite, line 12: Move this to end of the file. The test should be first since it is the most important

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    expect(getTranslationKeyOrString(poiCommentCommand.getCancelButtonLabel())).toBe('CANCEL');
+    expect(getTranslationKeyOrString(poiCommentCommand.getPostButtonLabel())).toBe('SEND');
+    expect(getTranslationKeyOrString(poiCommentCommand.getPlaceholder())).toBe(
+      'COMMENT_PLACEHOLDER'
+    );
+  });
+
+  test('post button is disabled when content is empty', () => {
+    const poiCommentCommand = new CreatePoiCommentCommand(TEST_POINT_OF_INTEREST);
```
nilscognite, line 44: Rename to command. That is is poiCommentCommand is given from the context

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    const mockPostComment = vi.fn().mockReturnValue(Promise.resolve());
+
+    const mockPointsOfInterestDomainObject = new Mock<PointsOfInterestDomainObject<string>>()
+      .setup((p) => p.postCommentForPoi)
+      .returns(mockPostComment)
+      .setup((p) => p.notify)
+      .returns(vi.fn())
+      .object();
+
+    const mockRenderTarget = new Mock<RevealRenderTarget>()
```
nilscognite, line 64: Can you use createFullRenderTargetMock instead of all this code. Seems verbose and complicated

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        new Mock<CommandsController>()
+          .setup((p) => p.update)
+          .returns(vi.fn())
+          .object()
+      )
+      .object();
+
+    const commentContent = 'comment-content';
+
+    const poiCommentCommand = new CreatePoiCommentCommand(TEST_POINT_OF_INTEREST);
```
nilscognite, line 83: Rename to command. More readable

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    poiCommentCommand.content = 'non-empty-content';
+
+    expect(poiCommentCommand.isPostButtonEnabled).toBeTruthy();
+  });
+
+  test('invoking command posts comment on Poi', () => {
+    const mockPostComment = vi.fn().mockReturnValue(Promise.resolve());
+
+    const mockPointsOfInterestDomainObject = new Mock<PointsOfInterestDomainObject<string>>()
```
nilscognite, line 57: Whay not use `domainObject `here. `mockPointsOfInterest `is gived by the context

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    const poiCommentCommand = new CreatePoiCommentCommand(TEST_POINT_OF_INTEREST);
+
+    poiCommentCommand.attach(mockRenderTarget);
+
+    poiCommentCommand.content = commentContent;
+    poiCommentCommand.invoke();
+
+    expect(mockPostComment).toHaveBeenCalledWith(TEST_POINT_OF_INTEREST, commentContent);
+  });
+});
```
nilscognite, line 92: Just check out: Can you ensure all methods of CreatePoiCommentCommand to be tested?

react-components/src/components/Architecture/InputField.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+      { wrapper }
+    );
+
+    const textArea = container.querySelector('textarea');
+
+    assert(textArea !== null);
+
+    expect(textArea.value).toBe(textContent);
+
+    await act(() => (command.content = ''));
```
nilscognite, line 91: One question. What is act doing? It is necessary?

react-components/src/components/Architecture/InputField.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+      <InputField inputCommand={command} placement={ARBITRARY_PLACEMENT} />,
+      { wrapper }
+    );
+
+    const postButton = container.querySelector('button');
+
+    assert(postButton !== null);
+
+    await act(async () => {
+      await userEvent.click(postButton);
```
nilscognite, line 117: What is difference of fireEvent.click(element); and 

    await act(async () => {
      await userEvent.click(postButton);
      
 Need explanation that you do this in a more complicated way

react-components/src/components/Architecture/InputField.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+      { wrapper }
+    );
+
+    const postButton = container.querySelector('button');
+
+    assert(postButton !== null);
+
+    expect(postButton.getAttribute('aria-disabled')).toBe('false');
+
+    act(() => {
```
nilscognite, line 138: Same here. I have done a very similar thing in another PR. Wonder if it is necessary  for me also. I didn't use act, but got some trouble with the updating

react-components/src/components/Architecture/InputField.test.tsx
haakonflatval-cognite: It ensures that all effects on rendering that is triggered by the thing in the callback have been finished before continuing, so that I can be sure that my assertions on the next lines operate on up-to-date data

react-components/src/components/Architecture/InputField.test.tsx
haakonflatval-cognite: As Christopher mentioned in the thread with us earlier, `userEvent` is recommended over `fireEvent` in most cases (although I admit I haven't taken the time to understand the full difference). I believe I got a warning from the test that I needed to put it inside an `act` too, to be sure changes to DOM where propagated correctly before continuing

react-components/src/components/Architecture/InputField.test.tsx
haakonflatval-cognite: Yes, I have some places where I should have used this earlier too :sweat_smile:

react-components/src/components/Architecture/InputField.test.tsx
pramod-cog: @haakonflatval-cognite @nilscognite `React` does things asynchronously by queuing up as micro-tasks, it would be not be possible for `testing-library` to catch these, when `act` is used then `React` will put those tasks in a special queue that is owned by `act` and this will execute those tasks before returning.

For example, we set some `state` from a `async` call, then `testing-library` will not know when this async operation is completed and update/sets the state of the variable, so when `act` is used, it will notify the `testing-library` that async operation is done and state is set. 

You can also use `waitFor` with `await` in `execute` check for asynchronous calls

react-components/src/components/Architecture/InputField.test.tsx
pramod-cog: @haakonflatval-cognite @nilscognite `userEvent` sends events like user interactions with the component like user clicking, it simulates real user behavior by triggering multiple events within a single interaction which helps to catch bugs, `fireEvent` sends events which user tells even if these events does not exists in the component, it might not catch bugs.

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
haakonflatval-cognite: It worked :+1: Only caveat is that the test now has to care that the provider is called by the domain object with some specific arguments (i.e. POI ID and text content) - but I suppose it's not really that bad altogether

react-components/src/architecture/base/commands/BaseInputCommand.ts
```diff
@@ -5,42 +5,34 @@ import { type TranslationInput } from '../utilities/TranslateInput';
...
-
   public abstract getPlaceholder(): TranslationInput | undefined;
 
-  public get onFinish(): (() => void) | undefined {
-    return this._onFinish;
-  }
-
-  public set onFinish(onFinish: () => void) {
-    this._onFinish = onFinish;
+  public get onCancel(): (() => void) | undefined {
```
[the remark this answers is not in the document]
nilscognite: VoidFunction  is a lot better

react-components/src/architecture/base/commands/BaseInputCommand.test.ts
nilscognite: Yes. You are correct

react-components/src/architecture/base/commands/BaseInputCommand.test.ts
nilscognite: These is in the mock.  

public setPostButtonEnabled(enable: boolean): void {
    this._isPostButtonEnabled = enable;
    this.update();
  }

The point is that is just testing the mock. I cannot se any value in this unit test then.

react-components/src/architecture/base/commands/BaseInputCommand.ts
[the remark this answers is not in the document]
haakonflatval-cognite: Fixed it now

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
nilscognite: Yes, maybe. But in all normal development the following rules are applied: **the most important first**

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
nilscognite: Good. I make the test more realistic

react-components/src/architecture/base/commands/BaseInputCommand.ts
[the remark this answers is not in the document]
nilscognite: Your code is not symmetric. Get and set should have the same type.
image

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    expect(poiCommentCommand.isPostButtonEnabled).toBeTruthy();
+  });
+
+  test('invoking command posts comment on Poi and calls onFinish', async () => {
+    const mockOnFinish = vi.fn();
+    const mockPostComment = vi.fn().mockReturnValue(Promise.resolve());
+
+    const mockRenderTarget = createRenderTargetMock();
+
+    const mockPointOfInterestProvider: PointsOfInterestProvider<string> = {
```
nilscognite, line 63: Move this to a separate function. createPointsOfInterestDomainObjectMock or something

nilscognite (changes requested):

react-components/src/components/Architecture/InputField.test.tsx
haakonflatval-cognite: From what I've understood, yes

nilscognite: Ok. Then I will change it on my PR as well

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
nilscognite: Yes.  You can still have it first. Ok for me. But you can define the creation as a function in the end and do:

`const TEST_POINT_OF_INTEREST = Create....();`

Then it is easier to read, at least for me.

haakonflatval-cognite: Alright, I can do that :+1:

nilscognite (dismissed):
Nice

nilscognite (approved):
LGTM

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    );
+
+    await waitFor(() => {
+      expect(mockOnFinish).toHaveBeenCalled();
+    });
+  });
+});
+
+function createPointsOfInterestProviderMock(): PointsOfInterestProvider<string> {
+  const mockPostComment = vi.fn().mockReturnValue(Promise.resolve());
```
larsmoa, line 75: Avoid `vi.fn().mockReturn<something>` as its not type safe. A better alternative is either
```
vi.fn(() => <return something>)
```
or using `vi.mocked(...)`

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      positionZ: 3,
+      scene: {
+        externalId: 'scene_external_id',
+        space: 'scene_space'
+      },
+      sceneState: {}
+    },
+    id: 'poi_id',
+    status: PointsOfInterestStatus.Default
+  } as const;
```
larsmoa, line 101: Nit: Since you are returning `PointsOfInterest` I dont treally think this has an effect?

react-components/tests/tests-utilities/architecture/commands/TestInputCommand.ts
```diff
@@ -0,0 +1,65 @@
+import { type TranslationInput } from '../../../../src/architecture';
```
larsmoa, line 1: Same comment as here:

https://github.com/cognitedata/reveal/pull/5120#discussion_r2095420087

(This being the second time this pops up today I do think you should look into it)

larsmoa (commented):
See code comment. Ping me on slack when I should have another look (also; might be that my comment can/should be fixed in a follow up and not this PR)

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
haakonflatval-cognite: You're right - removing

react-components/src/architecture/concrete/pointsOfInterest/CreatePoiCommentCommand.test.ts
haakonflatval-cognite: :+1:

react-components/tests/tests-utilities/architecture/commands/TestInputCommand.ts
haakonflatval-cognite: It seems to me that the `vite.config.ts` in the project only looks for things referenced by the `src/index.ts`. This file is also put in the `tests-utilities` folder, which is only intended for test-specific code. But are you thinking we should have some way of explicitly excluding this in the build output?

larsmoa: tsconfig.*.js(on)

haakonflatval-cognite: It was suggested that we instead do this in the `vite` config. The `exclude`-field in the `tsconfig` doesn't guarantee exclusion like we want. In any case, we will do this in a separate PR

nilscognite (approved):

nilscognite (approved):

larsmoa (commented):
🦄

PR #5133, opened 2025-05-08
fix: click depth on CAD models with newer ThreeJS versions

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5631

## Description :pencil:

Starting from ThreeJS r167, clicking on CAD nodes would set the camera pivot point at a wrong depth - usually much closer to the camera than the clicked surface. 

The error can be traced back to PR https://github.com/mrdoob/three.js/pull/28789 in ThreeJS, which reverses the significance of the components in the `vec4` when packing the depth as an RGBA vector in the shader. The unpacking function was changed accordingly, but we have our own implementation because we do this unpacking on the CPU-side. 

It's a little unconventional of us to actually pack the depth like this in the shader at all - it seems like we should be using a depth-only framebuffer / render target for this purpose. But I imagine getting it to work with the current render pipeline can be quite involved.

In `/examples`

- Build Reveal
- (Remove intersection handling code starting with `if (intersection !== undefined) {`... in `Viewer.tsx`, to avoid annoying errors when clicking
- Start examples and open the default Primitives model
- Click around and move the camera
- Note that the Pivot point appears at the exact point of click.

### DIFF
--- a/viewer/packages/cad-model/src/picking/PickingHandler.ts
+++ b/viewer/packages/cad-model/src/picking/PickingHandler.ts
@@ -49,11 +49,16 @@ export class PickingHandler {
   };
 
   private readonly _rgbaVector = new THREE.Vector4();
+
+  /**
+   * These factors are used for undoing the `packDepthToRGBA` GLSL operation defined by ThreeJS.
+   * They are taken from https://github.com/WestLangley/three.js/blob/bc58fecba18150103b95fbde5aaa3cc7cddf95a7/src/renderers/shaders/ShaderChunk/packing.glsl.js#L19
+   */
   private readonly _unpackFactors = new THREE.Vector4(
-    255 / 256 / (256 * 256 * 256),
-    255 / 256 / (256 * 256),
+    255 / 256,
     255 / 256 / 256,
-    255 / 256
+    255 / 256 / (256 * 256),
+    1 / (256 * 256 * 256)
   );
   private readonly _pipelineExecutor: BasicPipelineExecutor;
   private readonly _depthRenderPipeline: CadGeometryRenderModePipelineProvider;


pramod-cog (approved):
Tested in the examples, the issue is fixed, Nice!!
LGTM

**_Note to risk-reviewer: It would not be possible to add test to this fix as proper rendering of object is necessary to test the fix where we do not have any mechanism today to achieve it in test._**

viewer/packages/cad-model/src/picking/PickingHandler.ts
```diff
@@ -49,11 +49,16 @@ export class PickingHandler {
...
+   * They are taken from https://github.com/WestLangley/three.js/blob/bc58fecba18150103b95fbde5aaa3cc7cddf95a7/src/renderers/shaders/ShaderChunk/packing.glsl.js#L19
+   */
   private readonly _unpackFactors = new THREE.Vector4(
-    255 / 256 / (256 * 256 * 256),
-    255 / 256 / (256 * 256),
+    255 / 256,
     255 / 256 / 256,
-    255 / 256
+    255 / 256 / (256 * 256),
+    1 / (256 * 256 * 256)
```
larsmoa, line 61: How hard would writing a test for this bug be?

larsmoa (commented):
Would love to see a test for this. Not sure how hard it would be, but the PickingHandler.test.ts seems a bit thin, not testing intersection at all.

viewer/packages/cad-model/src/picking/PickingHandler.ts
pramod-cog: @larsmoa To add test for this bug, we would we need to render real objects which jest has no access!

larsmoa:
🦄

PR #5134, opened 2025-05-12
test(React-components): Unit test for command controller

nilscognite:
#### Type of change
Test 

## Description :pencil:

This is a unit test only.

I have added some doc in the top of the `CommandsController `class.

**Coverage:**

image


react-components/src/architecture/base/renderTarget/CommandsController.ts
```diff
@@ -9,6 +9,19 @@ import { type BaseCommand } from '../commands/BaseCommand';
 import { type Class, isInstanceOf } from '../domainObjectsHelpers/Class';
 import { type Signal, signal } from '@cognite/signals';
 
+/**
+ * The main purpose of the command controller is to give the correct event to the correct command,
+ * and maintain a list of all available commands in the system for updating and other actions.
+ * Note that the tool is only a specific type command, where only one can be active at the time.
+ * The active tool will receive all mouse events. It also maintain the default tool, which is
+ * activates at startup or if the current active tool is disabled.
```
haakonflatval-cognite, line 17:
```suggestion
 * activated at startup or if the current active tool is disabled.
```

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+    return true;
+  }
+
+  protected override get shortCutKeyOnShift(): boolean {
+    return true;
+  }
+}
+
+export class UpdateTester {
+  public CalledTimes = 0;
```
haakonflatval-cognite, line 438:
```suggestion
  public calledTimes = 0;
```

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+
+  // Set isCalled to true if the change is detected
+  public constructor(command: BaseCommand, change?: symbol) {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const self = this;
+    function listener(_command: BaseCommand, inputChange?: symbol): void {
+      if (change === inputChange) {
+        self.CalledTimes++;
+      }
```
haakonflatval-cognite, line 447: I think you would get rid of this alias if you declare `listener` as `const listener = (...): void => { ... }`, if you still want to keep this class

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+    });
+
+    test('Should set active tool and check if it is updating', () => {
+      const mockTool1 = createMockTool1();
+      const mockTool2 = createMockTool2();
+
+      const activateTester1 = new UpdateTester(mockTool1, CommandChanges.active);
+      const deactivateTester1 = new UpdateTester(mockTool1, CommandChanges.deactive);
+      const activateTester2 = new UpdateTester(mockTool2, CommandChanges.active);
+      const deactivateTester2 = new UpdateTester(mockTool2, CommandChanges.deactive);
```
haakonflatval-cognite, line 78: Not sure if it would be better, but have you considered using `vi.fn()` here? Then these lines would become e.g. 
```
const mockToolListener1 = vi.fn();
const mockToolListener2 = vi.fn();
mockTool1.addEventListener(mockToolListener1);
mockTool2.addEventListener(mockToolListener2);
```
Then you could check e.g. `expect(mockToolListener1).toHaveBeenCalledWith(expect.anything(), CommandChanges.active);`, or `expect(mockToolListener1.mock.calls[0]).toEqual([expect.anything(), CommandChanges.active]);`

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+
+    test('Should get command by type', () => {
+      const mockTool = createMockTool1();
+      const mockCommand = new MockCommand();
+
+      expect(controller.getCommandByTypeRecursive(MockTool1)).toBeUndefined();
+      expect(controller.getCommandByTypeRecursive(MockCommand)).toBeUndefined();
+
+      controller.add(mockTool);
+      controller.add(mockCommand);
```
haakonflatval-cognite, line 155: Just curious - since this is called "Recursive", do we intend that e.g. the `mockCommand` should be a child of the `mockTool`?

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+      mockTool.onFocusChanged = onFocusChangedMock;
+
+      domElement.dispatchEvent(new FocusEvent('focus'));
+      domElement.dispatchEvent(new FocusEvent('blur'));
+
+      expect(onFocusChangedMock).toHaveBeenNthCalledWith(1, true);
+      expect(onFocusChangedMock).toHaveBeenNthCalledWith(2, false);
+    });
+
+    test('Should call onWhen when WheelEvent', () => {
```
haakonflatval-cognite, line 293:
```suggestion
    test('Should call onWheel when WheelEvent', () => {
```

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+      controller.setActiveTool(mockTool);
+
+      const onWheelMock = vi.fn();
+      mockTool.onWheel = onWheelMock;
+
+      domElement.dispatchEvent(new WheelEvent('wheel'));
+      expect(onWheelMock).toHaveBeenCalledTimes(1);
+    });
+
+    test('Should call onWhen when WheelEvent', () => {
```
haakonflatval-cognite, line 304: This seems to have the same name as the test above - should it be called something else?

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+      const mockTool = createMockTool1();
+      controller.setActiveTool(mockTool);
+
+      const onWheelMock = vi.fn();
+      mockTool.onHover = onWheelMock;
+
+      domElement.dispatchEvent(createMouseHoverEvent());
```
haakonflatval-cognite, line 311: This mock function may also be misnamed, from what I can see

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+      expect(onWheelMock).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('Test some event with the event function at the controller itself', () => {
+    test('Should call correction function on the correct tool on hover event', async () => {
```
haakonflatval-cognite, line 317:
```suggestion
    test('Should call correct function on the correct tool on hover event', async () => {
```
The same goes for the rest of the tests in this section

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+      controller.setDefaultTool(defaultTool);
+      controller.setActiveTool(activeTool);
+
+      for (const isLeftButton of [false, true]) {
+        const mock = vi.fn();
+        if (isLeftButton) {
+          activeTool.onLeftPointerDown = mock;
+        } else {
+          defaultTool.onRightPointerDown = mock;
+        }
```
haakonflatval-cognite, line 359: Is it intentional that the first one is registered on `activeTool` and the second on `defaultTool`?

haakonflatval-cognite: (Seems like it's a recurring pattern in the other tests in this section, but I must admit I still don't understand it completely :sweat_smile: )

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -0,0 +1,498 @@
...
+  function createMockTool2(): MockTool2 {
+    const tool = new MockTool2();
+    tool.attach(renderTarget);
+    return tool;
+  }
+});
+
+class MockTool1 extends BaseTool {
+  // Mock any methods or properties as needed
+}
```
haakonflatval-cognite, line 416: Do we need this comment?

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
haakonflatval-cognite: Ok, think I got it :+1:

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
nilscognite: Commands are Recursive in general. Here not

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
nilscognite: Fixed

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
nilscognite: Fixed

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
nilscognite: This will give more verbose code!

nilscognite:
Here is changed according to Review 7a10dff904a923372c691fef34aaf9617292bacf

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
nilscognite: Left mouse events goes to the active tool
Right mouse events goes to the default tool, which is normally the navigation.
This ensure max flexibility

PR #5137, opened 2025-05-14
refactor(react-components): Cleaning up misuse of override method

nilscognite:
#### Type of change

Refactor 

## Description :pencil:

This PR removes the `override getLabel`. This function was never intended to be virtual and should not be overridden. Instead the test should come from tooltip. The base class use the tooltip to get the label.

**Be careful when overriding a function which is not overridden before.  Look in the existing code. A quick search could guide you easily. for instance** `override getLabel`. Also the `getLabel `function is not in the `VIRTUAL METHODS (To be overridden) `section of the `BaseCommand` file.


haakonflatval-cognite (approved):
LGTM :+1:

larsmoa (changes requested):
🦄

larsmoa (approved):

larsmoa:
🦄

pramod-cog (approved):

PR #5140, opened 2025-05-20
refactor(react-components): Add count function for both array and generators and use them around in the code

nilscognite:
#### Type of change
Refactor 

## Description :pencil:

This tries to reuse code around.


react-components/src/architecture/base/utilities/extensions/generatorUtils.test.ts
```diff
@@ -16,6 +21,16 @@ describe('generatorUtils', () => {
...
   });
+
+  describe('count', () => {
+    test('should count high numbers and find none', () => {
+      expect(count(getPositiveNumbers(6), (a) => a > 100)).toBe(0);
+    });
+
+    test('should count odd numbers', () => {
+      expect(count(getPositiveNumbers(6), isOdd)).toBe(3);
+    });
```
haakonflatval-cognite, line 32: There seems to be two tests that check that it counts odd numbers correctly - is that intentional?

react-components/src/architecture/base/utilities/extensions/arrayExtensions.ts
```diff
@@ -130,3 +130,20 @@ export function containsTheSameSet<T>(array1: readonly T[], array2: readonly T[]
...
+/**
+ * Counts the number of elements in the given array that satisfy the specified predicate.
+ *
+ * @param array - The array to search through.
+ * @param predicate - A function that tests each element for a condition.
+ * @returns The number of elements in the array that satisfy the predicate.
+ */
+export function count<T>(array: readonly T[], predicate: (t: T) => boolean): number {
+  let count = 0;
+  for (const item of array) {
```
haakonflatval-cognite, line 143: Should we write this to be defined in terms of the generatorsUtil-count-function (or write the generatorsUtil-count-function in terms of this function)?

haakonflatval-cognite (commented):
Two comments, otherwise looks good

react-components/src/architecture/base/utilities/extensions/generatorUtils.test.ts
nilscognite: One on the generator and one on the array.

larsmoa (commented):
🦄

PR #5143, opened 2025-05-21
test(react-components): Unit test for BaseFilterCommand 

nilscognite:
#### Type of change
Test 

## Description :pencil:

* This is a unit testfor `BaseFilterCommand `and `BaseFilterItemCommand`.
* Also added a new method to  `BaseFilterCommand `: `isSomeChecked()`

I use the mock `MockFilterCommand `to test it.


react-components/src/architecture/base/commands/BaseFilterCommand.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      option.invoke();
+    }
+  });
+});
+
+function* getOptions(command: BaseFilterCommand): Generator<BaseFilterItemCommand> {
+  if (command.children === undefined) {
+    return;
+  }
+  for (let i = 0; i < command.children.length; i++) {
```
pramod-cog, line 110: We can use `for(const option of command.children)`, you are using similar `for` loop structure above

nilscognite: Remove the function

pramod-cog (dismissed):
LGTM!

react-components/src/architecture/base/commands/BaseFilterCommand.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      expect(command.toggleAllChecked()).toBe(true);
+      expect(command.isAllChecked).toBe(true);
+    });
+
+    test('should have label and color at most options', () => {
+      assert(command.children !== undefined);
+      let childIndex = 0;
+      for (const option of command.children) {
+        expect(option.label).not.toBe('');
+        if (childIndex < command.children.length - 1) {
```
anders-hopland, line 91: why do we test if every option except the last option has a color? Why should all but the last option have colors?

react-components/src/architecture/base/commands/BaseFilterCommand.test.ts
```diff
@@ -0,0 +1,106 @@
...
+          expect(option.color).toBeDefined();
+        }
+        childIndex++;
+      }
+    });
+    test('should checked when invoke', () => {
```
anders-hopland:
```suggestion
    test('should be checked after invoke', () => {
```

react-components/src/architecture/base/commands/BaseFilterCommand.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      assert(command.children !== undefined);
+      for (const option of command.children) {
+        option.setChecked(false);
+        expect(option.isChecked).toBe(false);
+        option.invoke();
```
anders-hopland, line 102: Missing assertion after `option.invoke()`

react-components/src/architecture/base/commands/BaseFilterCommand.test.ts
nilscognite: It is a mock. So I think it is good to have some with color and some without.

react-components/src/architecture/base/commands/BaseFilterCommand.test.ts
nilscognite: Fixed

PR #5144, opened 2025-05-21
chore(react-component): Make sure the app language is set in the new translation system

nilscognite:
#### Type of change
Chore 

## Description :pencil:

This make sure that the app language is set at the new translation system from outside(i.e. Fusion)

### DIFF
--- a/react-components/src/components/i18n/I18n.tsx
+++ b/react-components/src/components/i18n/I18n.tsx
@@ -12,6 +12,7 @@ import {
 } from '../../architecture/base/utilities/TranslateInput';
 
 import english from '../../common/i18n/en/reveal-react-components.json';
+import { setCurrentLanguage, translate } from '../../architecture/base/utilities/translateUtils';
 
 const I18nContext = createContext<I18nContent | null>(null);
 
@@ -98,6 +99,7 @@ export const useTranslation = (fallbackLanguage?: string | undefined): I18nConte
 };
 
 export const I18nContextProvider = ({ appLanguage, children }: I18nProps): ReactElement => {
+  void setCurrentLanguage(appLanguage);
   const i18nContent = useTranslationContent(appLanguage);
   return <I18nContext.Provider value={i18nContent}>{children}</I18nContext.Provider>;
 };
--- a/react-components/src/components/i18n/I18n.tsx
+++ b/react-components/src/components/i18n/I18n.tsx
@@ -99,7 +99,9 @@ export const useTranslation = (fallbackLanguage?: string | undefined): I18nConte
 };
 
 export const I18nContextProvider = ({ appLanguage, children }: I18nProps): ReactElement => {
-  void setCurrentLanguage(appLanguage);
+  useEffect(() => {
+    void setCurrentLanguage(appLanguage);
+  }, [appLanguage]);
   const i18nContent = useTranslationContent(appLanguage);
   return <I18nContext.Provider value={i18nContent}>{children}</I18nContext.Provider>;
 };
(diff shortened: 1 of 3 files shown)


haakonflatval-cognite (approved):
LGTM :+1:

react-components/src/components/i18n/I18n.tsx
```diff
@@ -98,6 +99,9 @@ export const useTranslation = (fallbackLanguage?: string | undefined): I18nConte
 };
 
 export const I18nContextProvider = ({ appLanguage, children }: I18nProps): ReactElement => {
+  useEffect(() => {
+    void setCurrentLanguage(appLanguage);
```
cemreyavuz, line 103: Do we need the `void` keyword here? 🤔

cemreyavuz (commented):
Minor question, otherwise looks good to me

react-components/src/components/i18n/I18n.tsx
nilscognite: `setCurrentLanguage()` is async. Any other suggestions for calling async inside an effect() ?

cemreyavuz: I wasn't aware this function was async. Had a chat with @haakonflatval-cognite on slack briefly - the question I have now is: do we handle the race condition between two consecutive function calls?

pramod-cog: @nilscognite should we have `await` here if it is `async`?

cemreyavuz: > Any other suggestions for calling async inside an effect() ?

Btw, i still think that we don't need the `void` here. You can call an async function inside a sync function - or am i missing something?

nilscognite: image

cemreyavuz: I see - so it's an eslint warning. Not suggesting to remove the `void` keyword since this rule is clearly a team preference. Thanks for the explanation 👍🏻 

This question still stands though:
> do we handle the race condition between two consecutive function calls?

nilscognite: > @nilscognite shouldn't we have `await` here if it is `async`?

I can't use await anything inside an effect. Or do you know how @pramod-cog

nilscognite: @pramod-cog Since I'm using your hook. Can you answer @cemreyavuz question?

haakonflatval-cognite: @cemreyavuz I think it will be overengineering to guard against race conditions here - the language is not expected to change that frequently, and the async call is almost instantaneous (we dynamically import a file in the original source code, so it will already be bundled in the package).

pramod-cog: I think some thing like below should handle race condition, but I must say its a bit overkill, I remember using it in styling for something 🤔. Not sure if we need to have `ref` for `AbortController`.
```
useEffect(() => {
    const controller = new AbortController();
    const { signal } = controller;

    const updateLanguage = async (): Promise<void> => {
      if (signal.aborted) return;
      await setCurrentLanguage(appLanguage);
    };

    void updateLanguage();

    return () => {
      controller.abort();
    };
  }, [appLanguage]);
```

cemreyavuz: > @cemreyavuz I think it will be overengineering to guard against race conditions here - the language is not expected to change that frequently, and the async call is almost instantaneous (we dynamically import a file in the original source code, so it will already be bundled in the package).

This is not a blocker for me, I was trying to understand the impact. I don't have a trivial suggestion to make this better, but I believe we should go over the `setCurrentLanguage` logic separately at some point.

cemreyavuz (commented):
risk review ok

react-components/src/components/i18n/I18n.tsx
nilscognite: setCurrentLanguage and the other functions there are under unit test.

PR #5148, opened 2025-05-26
refactor(react-components) : Remove translation delegate in getLabel and use the function translate instead

nilscognite:
#### Type of change

Refactor 

## Description :pencil:

This get rid of the `useTranslation `hook in some of the components. The goal is to get rid of it totally and delete it.

The reason for that is to be independent of React hooks, to make simpler test and use the translation code in the business logic, not only in the React layer. It also get the complexity down.

**Changes in the code :**

* `BaseCommand.getLabel() `is now without an argument. So it is renamed to `get label()`.
* Same with `DomainObject.getLabel()`.

## How to test

1. yarn storybook
2. open Architecture/Main
3. Check the strings and the tooltips. Are they readable?

### Unit test coverage

Added missing unit test for `DomainObjectPanel`.
Other missing unit tests are in other PR's that is under risk review.

### I have tested this in Fusion (here Swedish):

image


react-components/src/components/Architecture/DomainObjectPanel.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+  type TranslationInput
+} from '../../architecture';
+import { act, type PropsWithChildren, type ReactElement } from 'react';
+import { ViewerContextProvider } from '../RevealCanvas/ViewerContext';
+import { DomainObjectPanel } from './DomainObjectPanel';
+import { DomainObjectPanelUpdater } from '../../architecture/base/reactUpdaters/DomainObjectPanelUpdater';
+import { type IconName } from '../../architecture/base/utilities/IconName';
+import { createFullRenderTargetMock } from '../../../tests/tests-utilities/fixtures/createFullRenderTargetMock';
+
+// Help page here:  https://bogr.dev/blog/react-testing-intro/
```
pramod-cog, line 22:
```suggestion
```

react-components/src/components/Architecture/DomainObjectPanel.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+      expect(screen.getByText(subString)).toBeDefined();
+    }
+  });
+
+  test('should copy info text to clipboard when clicking at the copy button', async () => {
+    const domainObject = new MockDomainObject();
+    renderDomainObjectPanel(domainObject);
+    const buttons = screen.getAllByRole('button');
+    await act(async () => {
+      await userEvent.click(buttons[0]); // Copy button
```
anders-hopland, line 57: non blocking
Would in general recommend get button by id as getting all buttons and choosing the first one has an implicit dependency on the order of the buttons

anders-hopland (commented):
risk review ok

react-components/src/components/Architecture/DomainObjectPanel.test.tsx
nilscognite: Ok. Fix this later

PR #5150, opened 2025-05-28
feat(react-components): Add hook returning function for filtering on classic assets in scene

anders-hopland:
#### Type of change

Feat  

## Description :pencil:
Adds hook returning function for filtering on classic assets in scene

### DIFF
--- a/react-components/src/query/FilterOnClassicAssetsInScene.context.ts
+++ b/react-components/src/query/FilterOnClassicAssetsInScene.context.ts
@@ -1,0 +1,21 @@
+import { createContext } from 'react';
+import { useReveal3dResourcesFromScene } from '../hooks';
+import { useAllAssetsMapped360Annotations } from './useSearchAssetsMapped360Annotations';
+import { useAllAssetsMappedPointCloudAnnotations } from './useSearchAssetsMappedPointCloudAnnotations';
+import { useAllMappedEquipmentAssetMappings } from './useSearchMappedEquipmentAssetMappings';
+
+export type FilterOnClassicAssetsInSceneDependencies = {
+  useReveal3dResourcesFromScene: typeof useReveal3dResourcesFromScene;
+  useAllMappedEquipmentAssetMappings: typeof useAllMappedEquipmentAssetMappings;
+  useAllAssetsMapped360Annotations: typeof useAllAssetsMapped360Annotations;
+  useAllAssetsMappedPointCloudAnnotations: typeof useAllAssetsMappedPointCloudAnnotations;
+};
+
+// Plese create a cntext for the FilterOnClassicAssetsInScene
+export const FilterOnClassicAssetsInSceneContext =
+  createContext<FilterOnClassicAssetsInSceneDependencies>({
+    useReveal3dResourcesFromScene,
+    useAllMappedEquipmentAssetMappings,
+    useAllAssetsMapped360Annotations,
+    useAllAssetsMappedPointCloudAnnotations
+  });
--- a/react-components/src/query/useFilterOnClassicAssetsInScene.test.tsx
+++ b/react-components/src/query/useFilterOnClassicAssetsInScene.test.tsx
@@ -1,0 +1,80 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { renderHook } from '@testing-library/react';
+import React from 'react';
+import { type Asset, type CogniteClient, type DirectRelationReference } from '@cognite/sdk';
+import {
+  FilterOnClassicAssetsInSceneContext,
+  type FilterOnClassicAssetsInSceneDependencies
+} from './FilterOnClassicAssetsInScene.context';
+import { useFilterOnClassicAssetsInScene } from './useFilterOnClassicAssetsInScene';
+import { Mock } from 'moq.ts';
+
+describe(useFilterOnClassicAssetsInScene.name, () => {
+  let sdk: CogniteClient;
+  let scene: DirectRelationReference;
+  let mockDependencies: FilterOnClassicAssetsInSceneDependencies;
+
+  const wrapper = ({ children }: { children: React.ReactNode }): React.ReactElement => (
+    <FilterOnClassicAssetsInSceneContext.Provider value={mockDependencies}>
+      {children}
+    </FilterOnClassicAssetsInSceneContext.Provider>
+  );
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+    sdk = new Mock<CogniteClient>().object();
+    scene = { externalId: 'scene1', space: 'space1' };
+
+    mockDependencies = {
+      useReveal3dResourcesFromScene: vi.fn().mockReturnValue([]),
+      useAllMappedEquipmentAssetMappings: vi.fn().mockReturnValue({
+        data: { pages: [{ assets: [createMockAsset(1), createMockAsset(2)] }] },
+        isFetching: false,
+        hasNextPage: false,
+        fetchNextPage: vi.fn()
+      }),
+      useAllAssetsMapped360Annotations: vi.fn().mockReturnValue({
+        data: [{ asset: createMockAsset(3) }],
+        isLoading: false
+      }),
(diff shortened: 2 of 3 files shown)


react-components/src/query/FilterOnClassicAssetsInScene.context.ts
```diff
@@ -0,0 +1,21 @@
...
+import { useAllMappedEquipmentAssetMappings } from './useSearchMappedEquipmentAssetMappings';
+
+export type FilterOnClassicAssetsInSceneDependencies = {
+  useReveal3dResourcesFromScene: typeof useReveal3dResourcesFromScene;
+  useAllMappedEquipmentAssetMappings: typeof useAllMappedEquipmentAssetMappings;
+  useAllAssetsMapped360Annotations: typeof useAllAssetsMapped360Annotations;
+  useAllAssetsMappedPointCloudAnnotations: typeof useAllAssetsMappedPointCloudAnnotations;
+};
+
+// Plese create a cntext for the FilterOnClassicAssetsInScene
```
anders-hopland:
```suggestion
```

react-components/src/query/useFilterOnClassicAssetsInScene.ts
```diff
@@ -0,0 +1,107 @@
...
+    (resource): resource is AddCadResourceOptions | AddPointCloudResourceOptions =>
+      !is360ImageAddOptions(resource)
+  );
+
+  const {
+    data: pagedCadAssetMappings,
+    isFetching: isFetchingCadAssetMappings,
+    hasNextPage: hasNextCadAssetMappings,
+    fetchNextPage: fetchNextCadAssetMappingsPage
+  } = useAllMappedEquipmentAssetMappings(cadAndPointCloudresources as AddModelOptions[], sdk, 1000);
```
christjt: This cast is not type safe

christjt: The default value of the limit is 1000, so you can omit the last param.

react-components/src/query/useFilterOnClassicAssetsInScene.ts
```diff
@@ -0,0 +1,107 @@
+/*!
+ * Copyright 2023 Cognite AS
+ */
+import { type AddModelOptions } from '@cognite/reveal';
```
christjt:
```suggestion
import type { AddModelOptions } from '@cognite/reveal';
```

react-components/src/query/useFilterOnClassicAssetsInScene.ts
```diff
@@ -0,0 +1,107 @@
...
+import { useContext, useEffect, useMemo } from 'react';
+import { type Asset, type CogniteClient, type DirectRelationReference } from '@cognite/sdk';
```
christjt:
```suggestion
import type { Asset, CogniteClient, DirectRelationReference } from '@cognite/sdk';
```

react-components/src/query/useFilterOnClassicAssetsInScene.test.tsx
```diff
@@ -0,0 +1,80 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { renderHook } from '@testing-library/react';
+import React from 'react';
+import { type Asset, type CogniteClient, type DirectRelationReference } from '@cognite/sdk';
```
christjt:
```suggestion
import type { Asset, CogniteClient, DirectRelationReference } from '@cognite/sdk';
```

react-components/src/query/useFilterOnClassicAssetsInScene.test.tsx
```diff
@@ -0,0 +1,80 @@
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { renderHook } from '@testing-library/react';
+import React from 'react';
```
christjt: Don't import all of react. Only import what you need

react-components/src/query/useFilterOnClassicAssetsInScene.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+        data: [createMockAsset(4)],
+        isLoading: false
+      })
+    };
+  });
+
+  it('returns a filter function that filters assets present in the asset map', () => {
+    const { result } = renderHook(() => useFilterOnClassicAssetsInScene(sdk, scene), { wrapper });
+
+    expect(result.current).toBeTypeOf('function');
```
christjt: ~This does not seem nessesary. This is handled by the type system. If the type was not a function, it would fail type checking further down when trying to invoke it.~

Turns out it can be undefined, though this was very much not obvious since you hid the `| undefined` inside the type.

react-components/src/query/useFilterOnClassicAssetsInScene.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const filterFn = result.current!;
```
christjt: ~result.current does not seem to be nullable, so the `!` operator seems unessesary.~

~This does not seem nessesary. This is handled by the type system. If the type was not a function, it would fail type checking further down when trying to invoke it.~

Turns out it can be undefined, though this was very much not obvious since you hid the `| undefined` inside the type.

react-components/src/query/useFilterOnClassicAssetsInScene.ts
```diff
@@ -0,0 +1,107 @@
...
+import {
+  type AddCadResourceOptions,
+  type AddImage360CollectionOptions,
+  type AddPointCloudResourceOptions
+} from '../components';
+import { is360ImageAddOptions } from '../components/Reveal3DResources/typeGuards';
+import { FilterOnClassicAssetsInSceneContext } from './FilterOnClassicAssetsInScene.context';
+
+export type FilterOnClassicAssetsInScene = ((assets: Asset[]) => Asset[]) | undefined;
```
christjt: This type makes it super hard to see that the useFilterOnClassicAssetsInScene hook can return undefined. I would much rather prefer:

```suggestion
export type FilterOnClassicAssetsInScene = (assets: Asset[]) => Asset[];
```

And then the hook returns `FilterOnClassicAssetsInScene | undefined`

react-components/src/query/useFilterOnClassicAssetsInScene.test.tsx
anders-hopland: Fixed in [40ad10b](https://github.com/cognitedata/reveal/pull/5150/commits/40ad10beb7bf836cd79ddbd44678152d7a2d622d)

react-components/src/query/useFilterOnClassicAssetsInScene.test.tsx
anders-hopland: Made the type more clear in [1c7dbff](https://github.com/cognitedata/reveal/pull/5150/commits/1c7dbffad72afb3f44436a89e6ffcb817da48255)

react-components/src/query/useFilterOnClassicAssetsInScene.test.tsx
anders-hopland: Made the type more clear in [1c7dbff](https://github.com/cognitedata/reveal/pull/5150/commits/1c7dbffad72afb3f44436a89e6ffcb817da48255)

react-components/src/query/useFilterOnClassicAssetsInScene.ts
anders-hopland: Fixed in [1c7dbff](https://github.com/cognitedata/reveal/pull/5150/commits/1c7dbffad72afb3f44436a89e6ffcb817da48255)

anders-hopland: Fixed in [1c7dbff](https://github.com/cognitedata/reveal/pull/5150/commits/1c7dbffad72afb3f44436a89e6ffcb817da48255)

christjt (approved):
risk review ok

PR #5155, opened 2025-06-03
refactor(react-component): Update most vitest describe by using name

nilscognite:
#### Type of change
Refactor 

## Description :pencil:

Nothing actually has change in the code. Only done this type of change:

`describe('LineView', ` => `describe(LineView.name, `


dmivankov:
:unicorn:

PR #5156, opened 2025-06-03
feat(react-components): hook for creating DM-filtering function

haakonflatval-cognite:
#### Type of change

Feat  

## Description :pencil:

Create and expose a hook that will create a callback that filters a given set of DM-nodes by connections to a given set of 3D-models.

It reuses parts of our DM-Search hook, but does not return values that are only indirectly connected to the 3D models.

This hook is intended for use in FRS in Unified 3D. 

This PR is not intended to change the behavior of existing code.

### DIFF
--- a/react-components/src/query/useSearchMappedEquipmentFDM.test.tsx
+++ b/react-components/src/query/useSearchMappedEquipmentFDM.test.tsx
@@ -9,7 +9,7 @@ import {
 } from '../query/useSearchMappedEquipmentFDM';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { cadNodesFixtures } from '#test-utils/fixtures/dm/nodeItems';
-import { Source, type FdmSDK } from '../data-providers/FdmSDK';
+import { type Source, type FdmSDK } from '../data-providers/FdmSDK';
 import { Mock } from 'moq.ts';
 import { type RevealRenderTarget } from '../architecture';
 import type { FC, PropsWithChildren } from 'react';
@@ -19,7 +19,7 @@ import { FdmSdkContext } from '../components/RevealCanvas/FdmDataProviderContext
 import { getMockViewByIdResponse } from '#test-utils/fixtures/dm/getMockViewByIdResponse';
 import { getMockViewItemFromSimpleSource } from '#test-utils/fixtures/dm/getMockViewItemFromSimpleSource';
 import { COGNITE_ASSET_SOURCE } from '../data-providers/core-dm-provider/dataModels';
-import { TableExpressionFilterDefinition } from '@cognite/sdk';
+import { type TableExpressionFilterDefinition } from '@cognite/sdk';
 
 const queryClient = new QueryClient();
 


react-components/src/query/useFilterOnClassicAssetsInScene.ts
```diff
@@ -116,9 +117,14 @@ export const useFilterOnClassicAssetsInScene = (
     isLoadingAllPointCloudAssets
   ]);
 
+  const filterAssetsCallback = useCallback(
+    (assets: Asset[]) => assets.filter((asset) => allAssetsMap?.has(asset.id) === true),
+    [allAssetsMap]
+  );
+
   if (allAssetsMap === undefined) {
```
haakonflatval-cognite, line 125: This was previously not memoized, causing it to update on every rerender

react-components/src/data-providers/core-dm-provider/check3dConnectedEquipmentQuery.ts
```diff
@@ -193,3 +197,48 @@ export function createCheck3dConnectedEquipmentQuery(
     }
   } as const satisfies Omit<QueryRequest, 'parameters' | 'cursor'>;
 }
+
+// We want to preserve most of the typing of the object returned here,
+// as it gives us better types in the result of the query call later on
+// Specifying a type here would either mean losing valuable type information
+// or an enourmous type
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+export function createCheck3dConnectedEquipmentQueryNoIndirect(
```
anders-hopland: non blocking
I would suggest to rename it to `createCheck3dConnectedEquipmentNoIndirectQuery` to get `query` at the end

react-components/src/data-providers/core-dm-provider/filterNodesByMappedTo3d.ts
```diff
@@ -31,7 +31,8 @@ export async function filterNodesByMappedTo3d(
   nodes: InstancesWithViewDefinition[],
   revisionRefs: DmsUniqueIdentifier[],
   _spacesToSearch: string[],
-  fdmSdk: FdmSDK
+  fdmSdk: FdmSDK,
+  includeIndirectRelations: boolean
```
anders-hopland, line 35: What do you think about adding a default value here?

react-components/src/data-providers/core-dm-provider/filterNodesByMappedTo3d.ts
```diff
@@ -128,16 +132,17 @@ function getRelevantObject3ds(
   connectionData: QueryResult<
     ReturnType<typeof createCheck3dConnectedEquipmentQuery>,
     SelectSourcesType
-  >
+  >,
+  includeIndirectRelations: boolean
```
anders-hopland, line 136: I think I would prefer 2 separate functions for `getRelevantObject3ds`. One for with indirect and one for no indirect nodes. The connectiondata type is currently wrong as non indirect query does not return non indirect clauses

react-components/src/query/useFilterOnClassicAssetsInScene.ts
anders-hopland: Nice catch

react-components/src/query/useSearchMappedEquipmentFDM.test.tsx
```diff
@@ -60,25 +62,25 @@ const wrapper: FC<PropsWithChildren> = ({ children }) => (
...
+];
 
-  const mockInstancesFilter = { equals: { property: ['key'], value: 'value' } };
+const mockInstancesFilter = { equals: { property: ['key'], value: 'value' } };
 
-  const mockViewsToSearch = [
-    { externalId: 'CogniteCADNode', space: 'cdf_cdm', version: 'v1', type: 'view' as const },
-    { externalId: 'CogniteCADNode', space: 'cdf_cdm', version: 'v1', type: 'view' as const }
-  ];
+const mockViewsToSearch = [
```
anders-hopland: I would prefer to add a type here

react-components/src/query/useSearchMappedEquipmentFDM.test.tsx
```diff
@@ -60,25 +62,25 @@ const wrapper: FC<PropsWithChildren> = ({ children }) => (
...
-    { modelId: 456, revisionId: 789 },
-    { modelId: 123, revisionId: 456 }
-  ];
+const mockModels: AddModelOptions[] = [
+  { modelId: 456, revisionId: 789 },
+  { modelId: 123, revisionId: 456 }
+];
 
-  const mockInstancesFilter = { equals: { property: ['key'], value: 'value' } };
+const mockInstancesFilter = { equals: { property: ['key'], value: 'value' } };
```
anders-hopland: I would prefer to add type here

anders-hopland (commented):
Looks good, only a few comments

react-components/src/data-providers/core-dm-provider/filterNodesByMappedTo3d.ts
haakonflatval-cognite: I don't think I want it - would rather have this explicit, with no mystical default values giving unexpected results at some point

react-components/src/data-providers/core-dm-provider/filterNodesByMappedTo3d.ts
haakonflatval-cognite: The `connectionData` type is correct, we just ignore the indirect fields. I agree this is messy, though, would you like to have it as a separat function regardless?

anders-hopland: I feel like this might be a footgun at some point, as the indirect properties does not exist. I will however let you take the final decision

haakonflatval-cognite: It definitely exists

anders-hopland:
🦄

PR #5158, opened 2025-06-04
doc(react-components): Remove Cognite copyright header in all files

nilscognite:
#### Type of change
Docs 

## Description :pencil:

Since I want the code consistent, and it is arbitrary where it is,  I have move the Copyright on top of all files in React-component.


christjt (approved):
risk review ok

PR #5160, opened 2025-06-04
test(react-components): Unit test for PlaneDragger and ExampleDragger

nilscognite:
#### Type of change
Test

## Description :pencil:

This is the first PR of 2. It contains some utility functions for testing the dragger and the unit test of the two simples draggers, `ExampleDragger `and `PlaneDragger`. The other PR is based on this and is found here: https://github.com/cognitedata/reveal/pull/5162

This test interactive editing of an object in space. The use can translate, rotate and resize interactively. This is what I test here. In the two draggers I test here, only translation is implemented.

I don't test the visualization. This is completely disconnected from the visualization since I use the the observer pattern in the architecture in general: https://en.wikipedia.org/wiki/Observer_pattern. I don't test the update either. This is in another abstraction.

I have made minor changes in the dragger code. This only relates to determine if the change is more than a small number. Before it was 0. If the change it tiny, it returns out as nothing have changed. Also I have removed an unnecessary statement.

## How to test:

**For example:**
`yarn storybook`
Open Architecture/Main
Click on the CircleIcon in the toolbar
Click at something to make the point
Grab it and move it around

**For Plane**
Click on the Cropbox in the toolbar
Select a plane (take the leftmost icon)
Click at something to make the plane
Grab it and move it around

PR #5162, opened 2025-06-05
test(react-components): Unit test from BoxDragger and CylinderDragger

nilscognite:
#### Type of change
Test

## Description :pencil:

**This is the hardest unit test I have written until now**. It test interactive editing of a Cylinder and Box in space.  The use can translate, rotate and change size of the objects interactively. This is what I test here.

The number seems are a bit magic,  but it is hard to do it in any other way, since I am simulate mouse movements, and had to calculate the expected result.  I have tried to give some description and explanation in the tests, but I can work 10 weeks with this of I should test all different ways to move or resize all corners, all faces, for geometries etc. Anyway, I found no bugs during this works.

I don't test the visualization. This is completely disconnected from the visualization since I use the the observer pattern in the architecture in general: https://en.wikipedia.org/wiki/Observer_pattern. I don't test the update either. This is in another abstraction.

I have made minor changes on the dragger code. This only relates to determine if the change is more than a small number. Before it was 0. If the change it tiny, it returns out as nothing have changed.

**Note: This PR has to be merged after https://github.com/cognitedata/reveal/pull/5160, since it is building on this.**

## How to test:

`yarn storybook`
Open Architecture/Main
Click on the Measurement in the toolbar
Select what to measure (cylinder or box)
Click at something to make the object
Grab it and move.

**For Cylinder:** You can change radius, height or translate
**For Box:** You can resize by move edges, translate, rotate or size by grabbing the corners

## Here is it in real life:

[screen-recorder-tue-jun-03-2025-10-48-51.webm](https://github.com/user-attachments/assets/e9c02957-db40-4024-b1f9-750dfedc4aef)


react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+  getTestCases,
+  getTestCasesWithSign,
+  type TestCase
+} from '#test-utils/architecture/baseDraggerUtil';
+import { Box } from '../../../base/utilities/primitives/Box';
+
+describe(BoxDragger.name, () => {
+  test('should not create correct dragger when focus type is None', () => {
+    const domainObject = createBoxDomainObject();
+    const startRay = new Ray(new Vector3(0, 0, 2), new Vector3(0, 0, -1));
```
danpriori, line 25: Can we have some const like "ARBITRARY...." for all these values with magic numbers to let us know that they're arbitrary? And for those that are not arbitrary, at least some explanation/comments about why these numbers?

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    );
+    expect(dragger).toBeUndefined();
+  });
+
+  test('should not change when illegal focus type', () => {
+    const focusType = FocusType.Focus;
+    const domainObject = createBoxDomainObject();
+    const direction = new Vector3(0, 0, -1);
+    const startRay = new Ray(new Vector3(0, 0, 2), direction);
+    const delta = new Vector3(1, 2, 0);
```
danpriori, line 40: same question here about these magic numbers + arbitrary consts

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+      const startRay = new Ray(new Vector3(0, 0, sign * 2), direction);
+      const delta = new Vector3();
+
+      const expectedBox = clone(domainObject.box);
+      if (testCase.expectedChange) {
+        delta.set(testCase.shiftKey ? 0 : 1, 2, 0); // Shift only translate in dominate direction
+        expectedBox.center.add(delta);
+      }
+      const dragger = domainObject.createDragger(
+        createCreateDraggerPropsMock(domainObject, startRay, focusType, sign === 1 ? 5 : 2)
```
danpriori: same question here about these magic numbers + arbitrary consts: why when sign is 1 we have 5 and if not then we have 2? Are they arbitrary?

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+      // Grab the box from the face and move it away from the center
+      const direction = new Vector3(-1, 0, -sign).normalize();
+      const startRay = new Ray(new Vector3(2, 0, 2 * sign), direction);
+      const delta = new Vector3();
+
+      const expectedBox = clone(domainObject.box);
+      if (testCase.expectedChange) {
+        delta.z = sign;
+        expectedBox.size.add(new Vector3(0, 0, 1));
+        expectedBox.center.addScaledVector(delta, 0.5);
```
danpriori, line 92: same question here about these magic numbers + arbitrary consts. Maybe just a comment here could be enough but can be also some arbitrary const if you prefer.

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    for (const testCase of getTestCasesWithSign()) {
+      const { sign } = testCase;
+      const domainObject = createBoxDomainObject();
+
+      // Grab the box from the face and move it away from the center
+      const direction = new Vector3(-1, 0, -sign).normalize();
+      const startRay = new Ray(new Vector3(2, 0, 2 * sign), direction);
+      const delta = new Vector3();
+
+      const expectedBox = clone(domainObject.box);
```
danpriori, line 88: I think we could use some shared clone function from some utilities, no? If we dont have it, maybe it is a good moment to create it, along with the other functions we have it here that could be in a shared place perhaps?

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+      if (testCase.expectedChange) {
+        delta.z = sign;
+        expectedBox.size.add(new Vector3(0, 0, 1));
+        expectedBox.center.addScaledVector(delta, 0.5);
+      }
+      const dragger = domainObject.createDragger(
+        createCreateDraggerPropsMock(domainObject, startRay, focusType, sign > 0 ? 2 : 5)
```
danpriori: same question here about these magic numbers + arbitrary consts

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    // Grab the box at the corner and move it away from the center
+    //            Y
+    //            ^ Face 1
+    //            +------+
+    //     Face 3 |      | Face 0
+    //            |      |
+    // (-1,-1,-1) +======+  --> X
+    //            ^ Face 4
+    //            | Dir = (0, 1, 0)
+    // (-1,-2,-1) *
```
danpriori, line 116: I like that you explained visually 😃 . Although it is a bit hard to understand what is going on here 🙈

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+      const delta = new Vector3(); // Movement in the XZ plane
+
+      const expectedBox = clone(domainObject.box);
+      if (testCase.expectedChange) {
+        delta.set(-1, 0, -2);
+        expectedBox.size.addScaledVector(delta, -1);
+        expectedBox.center.addScaledVector(delta, 0.5);
+      }
+      const dragger = domainObject.createDragger(
+        createCreateDraggerPropsMock(domainObject, startRay, focusType, 4, cornerSign)
```
danpriori: why 4?

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+      const domainObject = createBoxDomainObject();
+
+      const direction = new Vector3(0, 1, 0); // Along the y axis
+      const startRay = new Ray(new Vector3(-1, -2, -1), direction); // See figure
+      const cornerSign = new Vector3(-1, -1, -1); // This is the corner that is being dragged
+      const delta = new Vector3(); // Movement in the XZ plane
+
+      const expectedBox = clone(domainObject.box);
+      if (testCase.expectedChange) {
+        delta.set(-1, 0, -2);
```
danpriori: why setting the delta to this vector? Are they arbitrary also?

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    //            |      |
+    // (-1,-1,-1) +======+-----> X
+    //            ^ Face 4
+    //            | Dir = (0, 1, 0)
+    // (-1,-2,-1) *
+
+    const focusType = FocusType.Rotation;
+    for (const testCases of getTestCasesWithCanRotate()) {
+      const domainObject = createBoxDomainObject();
+      domainObject.canRotateComponent = (_component: number): boolean => testCases.canRotate;
```
danpriori, line 158: If you want to mock this method, I would suggest creating a mocked function (with some mocked name, likewise we do in any other mocked functions in other tests) and assigning it to canRotateComponent to explicitly indicate that it is a mocked version, intending to return the expected flag to perform the tests :)

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+  const intersection = createIntersectionMock(domainObject, focusType, face, cornerSign);
+  return { intersection, point, ray };
+}
+
+function createBoxDomainObject(): BoxDomainObject {
+  const domainObject = new MeasureBoxDomainObject(PrimitiveType.Box);
+  domainObject.box.size.set(2, 2, 2);
+  domainObject.box.center.set(0, 0, 0);
+  return domainObject;
+}
```
danpriori, line 201: Can we have this function in some shared utilities in case we want to use this in other tests?

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+
+function* getTestCasesWithCanRotate(): Generator<TestCase & { canRotate: boolean }> {
+  yield { expectedChange: true, shiftKey: false, canRotate: true };
+  yield { expectedChange: true, shiftKey: true, canRotate: true };
+  yield { expectedChange: false, shiftKey: false, canRotate: true };
+  yield { expectedChange: false, shiftKey: false, canRotate: false };
+}
```
danpriori, line 208: Can we have this function in some shared utilities in case we want to use this in other tests?
Maybe you can put this one in `baseDraggerUtils`, where it rests other similar functions.

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
```diff
@@ -0,0 +1,212 @@
...
+
+function clone(cylinder: Box): Box {
+  return new Box().copy(cylinder);
+}
```
danpriori, line 212: I guess we have a similar function out there, isnt? If not, then maybe we can put this one also in some shared utilities for testing.

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
```diff
@@ -0,0 +1,252 @@
...
+  createIntersectionMock,
+  drag,
+  getTestCases,
+  getTestCasesWithSign
+} from '#test-utils/architecture/baseDraggerUtil';
+
+describe(CylinderDragger.name, () => {
+  test('should not create correct dragger when focus type is None', () => {
+    const domainObject = createVerticalCylinderDomainObject();
+    const startRay = new Ray(new Vector3(0, 0, 2), new Vector3(0, 0, -1));
```
danpriori, line 23: same question here about these magic numbers + arbitrary consts + comments about these numbers

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    );
+    expect(dragger).toBeUndefined();
+  });
+
+  test('should not change when illegal focus type', () => {
+    const focusType = FocusType.Focus;
+    const domainObject = createVerticalCylinderDomainObject();
+    const direction = new Vector3(0, 0, -1);
+    const startRay = new Ray(new Vector3(0, 0, 2), direction);
+    const delta = new Vector3(1, 2, 0);
```
danpriori: same question here about these magic numbers + arbitrary consts + comments about these numbers

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
```diff
@@ -0,0 +1,252 @@
...
+  focusType: FocusType,
+  faceIndex: number = 0
+): CreateDraggerProps {
+  const point = domainObject.cylinder.intersectRay(ray, new Matrix4());
+  if (point === undefined) {
+    throw new Error('No intersection found for the ray and object');
+  }
+  const intersection = createIntersectionMock(domainObject, focusType, faceIndex);
+  return { intersection, point, ray };
+}
```
danpriori, line 226: is it possible to have a generic createCreateDraggerPropsMock for both box and cylinder (if ofc taking the special parameters in each case into account) ?

Also, the name `createCreateDraggerPropsMock` is also strange 🙈

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
```diff
@@ -0,0 +1,252 @@
...
+  test('should not change when illegal focus type', () => {
+    const focusType = FocusType.Focus;
+    const domainObject = createVerticalCylinderDomainObject();
+    const direction = new Vector3(0, 0, -1);
+    const startRay = new Ray(new Vector3(0, 0, 2), direction);
+    const delta = new Vector3(1, 2, 0);
+
+    const expectedBox = clone(domainObject.cylinder);
+    const dragger = domainObject.createDragger(
+      createCreateDraggerPropsMock(domainObject, startRay, focusType, 5)
```
danpriori: why 5 here?

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      const delta = new Vector3();
+
+      const expectedCylinder = clone(domainObject.cylinder);
+      if (testCase.expectedChange) {
+        delta.set(1, 2, 0);
+        expectedCylinder.centerA.add(delta);
+        expectedCylinder.centerB.add(delta);
+      }
+      const dragger = domainObject.createDragger(
+        createCreateDraggerPropsMock(domainObject, startRay, focusType, sign > 0 ? 5 : 2)
```
danpriori: same question here about these magic numbers + arbitrary consts + comments about these numbers

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      const domainObject = createVerticalCylinderDomainObject();
+
+      // Grab the cylinder at top cap from above and move it in the XY plane
+      const direction = new Vector3(0, 0, -sign);
+      const startRay = new Ray(new Vector3(0, 0, sign * 2), direction);
+      const delta = new Vector3();
+
+      const expectedCylinder = clone(domainObject.cylinder);
+      if (testCase.expectedChange) {
+        delta.set(1, 2, 0);
```
danpriori: same question here about these magic numbers + arbitrary consts + comments about these numbers

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
```diff
@@ -0,0 +1,252 @@
...
+      const startRay = new Ray(new Vector3(1, -2, 0), direction);
+      const delta = new Vector3();
+
+      const expectedCylinder = clone(domainObject.cylinder);
+      if (testCase.expectedChange) {
+        delta.x = 1;
+        expectedCylinder.radius += 1;
+      }
+      const dragger = domainObject.createDragger(
+        createCreateDraggerPropsMock(domainObject, startRay, focusType, 0)
```
danpriori: why 0 here? 
(Again: I know that it is the faceIndex but I'm thinking more why you're selecting the faceIndex 0 instead of B or C? If it is arbitrary then maybe it is good to explicitly say that :D )

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
```diff
@@ -0,0 +1,252 @@
...
+  const domainObject = new MeasureCylinderDomainObject(PrimitiveType.HorizontalCircle);
+  domainObject.cylinder.radius = 1;
+  domainObject.cylinder.centerA.set(0, 0, +Cylinder.HalfMinSize);
+  domainObject.cylinder.centerB.set(0, 0, -Cylinder.HalfMinSize);
+  return domainObject;
+}
+
+function clone(cylinder: Cylinder): Cylinder {
+  return new Cylinder().copy(cylinder);
+}
```
danpriori, line 246: I think this one could be under a shared utilities too, right?

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
```diff
@@ -0,0 +1,252 @@
...
+
+function createVerticalCylinderDomainObject(): CylinderDomainObject {
+  const domainObject = new MeasureCylinderDomainObject(PrimitiveType.VerticalCylinder);
+  domainObject.cylinder.radius = 1;
+  domainObject.cylinder.centerA.set(0, 0, +1);
+  domainObject.cylinder.centerB.set(0, 0, -1);
+  return domainObject;
+}
+
+function createHorizontalCircleDomainObject(): CylinderDomainObject {
```
danpriori, line 236: Shouldnt be: `createHorizontalCylinderDomainObject` ?
I suspect that these two functions could also be under a shared utilities, but maybe in this case it is too overkill if we dont have any expectation to be used elsewhere 🙈

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
nilscognite: This is needed here only. So I want add it to any utility file.

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
nilscognite: Yes. It is hard. Can't do it better without spending a lot of time on this, which is not very useful for Cognite

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
nilscognite: Not really. This is linked to the test and all 'magic' numbers

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
nilscognite: These are not been reused. So it is a local functions

react-components/src/architecture/concrete/primitives/box/BoxDragger.test.ts
nilscognite: No. This is unique. But I renamed cylinder -> box

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
nilscognite: Thanks, renamed to `createDraggerPropsMock`

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
nilscognite: This number is removed

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
nilscognite: Again. This is used only here

react-components/src/architecture/concrete/primitives/cylinder/CylinderDragger.test.ts
nilscognite: This is correct. I use the cylinder for measure horizontal circles (or discs). This ensure reuse of code. In fact. the discs has a thickness, so the height is > 0, in order to around z-fighting and make it possible to grab along the side.

nilscognite:
I have updated comments, Yes, this number seems are a bit magic,  but it is hard to do it in any other way, since I am simulate mouse movements, and have to calculate the expected result.  I have tried  to give some description and explanation in the tests, but I can work 10 weeks with this of I should test all different ways to move or resize all corners, all faces etc. Anywhay, I found no bugs during this works

danpriori (dismissed):
Ok. Now I think it is better with explanations about the numbers and the other changes made. Nice 👍

nilscognite:
> @nilscognite The PR is over the 500 line limit for risk review, needs exception from @thorkildcognite or split up 🙂

No. It was build on another PR so number of lines is less than 500. There is another PR that needed to be merge. This is now done and I have updated it with master.

danpriori (dismissed):
LGTM. 
Although, ofc, as the risk reviewer mentioned, please check with them in terms of needing split this up or having the exception.

react-components/src/architecture/concrete/primitives/box/BoxDragger.ts
```diff
@@ -135,8 +136,8 @@ export class BoxDragger extends BaseDragger {
       return false;
     }
     const deltaCenter = planeIntersect.sub(this.point);
-    if (deltaCenter.lengthSq() === 0) {
-      return false;
+    if (deltaCenter.length() < EPSILON) {
```
roligheten, line 139: Suggestion: For readability it would be nice if EPSILON was named something indication what it represents something like "TRANSLATE_SENSITIVITY_THRESHOLD" maybe? :slightly_smiling_face:

roligheten: Afaik the only functional change in this PR is that you are adding a sensitivity threshold that if the translation delta is less than the threshold it is not considered a change and the translation is essentially aborted.

There are test cases for "no translation" and "translation above the threshold" but I am wondering if it would be nice to test "translation below the threshold" also? (Ie. do translate, but translate only a tiny amount that should not be picked up) 

Asking since the logic for detecting if we are below the threshold does not seem trivial in every case and I could see a small mistake causing the logic to break that would not be picked up by either the "large" or "no" translation test cases :slightly_smiling_face:

roligheten: Also actually more concerning would be if there was a bug that caused the sensitivity check to be be too broad, ie if we are now translating very large values compared to the sensitivity in the test (Idk if we are) it might be good to either lower the test values or have separate tests for translations "just above the current threshold" to ensure that if there was some unintended changed that caused small translations that should be "significant" to be unintentionally considered "insignificant" and result in no translation that would be picked up by tests

roligheten: In summary: I think testing for translations below the thresholds is more of of a nice to have but I think testing that translations above but close to the threshold is important to not introduce potentially user affecting bugs in the future.

nilscognite: The only reason I have this threshold is to figure out that the mouse really moved. Moving the mouse one pixel is a lot more than the threshold. These is absolute not practical reason to add close to the threshold as a test, and it will be over engineered by doing this. The reason for using threshold instead of 0 is:

1. Numerical instabilities, as intersection and distance functions tends to give some very small value below or above 0, when the theoretical value is 0. (For instance plane-ray intersection or distances) This is the cause I had to change the production code, because it became a problem the testing when nothing really had changed..
2. Consistent in the code. So all test uses the same threshold.

By using large values: We are converting all coordinates to meters in the backend. The typical values of  measurements is 100 m to 0.1 m, but normally just around some meters.

nilscognite: `Epsilon `is a good name choice since it is widely used in the numerical community. But I can change it, if this is important, I really don't care that much.

roligheten: `These is absolute not practical reason to add close to the threshold as a test, and it will be over engineered by doing this`
I think what I am mainly concerned about is user actions being erroneously detected as numerical instabilities due to a future change to the code related to detecting and discarding insiginifcant translations, so testing movement close to the smallest change a user can reasonably make in a given scene would also suffice I think.

However if you think that this is not something that could practically happen or testing for this is not feasible I am ok skipping it.

nilscognite: I set it down to 0.000001. This is far beyond any possible movement you can no in a dragger and less then you can ever see with the eye. c59c9dc88297e923771a470e25dbc78cc31191f3

roligheten: Im not worried about the threshold itself, it's about having unit test coverage for these lines for example
 ```
    if (isAbsEqual(newRadius, cylinder.radius, EPSILON)) {
      return false; // Nothing has changed
    }
```
Ie. if someone were to change EPSILON for example or mess up the isAbsEqual function would we detect that this does not cause problems with with the dragger component?  :slightly_smiling_face: 

If you don't see the value or the possibility to test for values closer to the threshold you can disregard the suggestion.

nilscognite: It will just take long time for me to create those unit test.  Regard the number as 0. **Have you asked about that then?**

nilscognite: I can make a unit test of this: if (isAbsEqual(newRadius, cylinder.radius, EPSILON)) {

We have 60% coverage on the entire Reveal repo. It is more useful for the company that I use the time for making unit test for the these 40%, instead of getting this particular unit test up to 100%.

roligheten: Ok, sounds like its not worth the effort. :+1:

danpriori (approved):
👍

BugGambit:
:unicorn:

PR #5163, opened 2025-06-05
test(react-components): Unit test for BoxFace

nilscognite:
#### Type of change

Test 

## Description :pencil:

Unit test for BoxFace, which reparent a a face of a box, use as helper for some algorithms.
Finally found a bug (in the equals function)


react-components/src/architecture/concrete/primitives/common/BoxFace.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        expect(Math.abs(plane.constant)).toBeCloseTo(0.5);
+        sum.add(plane.normal);
+      }
+      expect(sum.length()).toBeCloseTo(0);
+      expect(isPointVisibleByPlanes(planes, new Vector3(0, 0, 0))).toBe(true);
+      expect(isPointVisibleByPlanes(planes, new Vector3(2, 2, 2))).toBe(false);
+    }
+  });
+
+  test('should compare two box faces whee some can be undefined', () => {
```
danpriori:
```suggestion
  test('should compare two box faces when some can be undefined', () => {
```

react-components/src/architecture/concrete/primitives/common/BoxFace.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      const sum = new Vector3();
+      for (const plane of planes) {
+        expect(plane.normal.length()).toBeCloseTo(1);
+        expect(Math.abs(plane.constant)).toBeCloseTo(0.5);
+        sum.add(plane.normal);
+      }
+      expect(sum.length()).toBeCloseTo(0);
+      expect(isPointVisibleByPlanes(planes, new Vector3(0, 0, 0))).toBe(true);
+      expect(isPointVisibleByPlanes(planes, new Vector3(2, 2, 2))).toBe(false);
+    }
```
danpriori, line 131: This test seems to check not only whether it creates 4 or 6 planes but other things as well (normals, etc.). Is it possible to separate it into other tests?

react-components/src/architecture/concrete/primitives/common/BoxFace.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  test('should get the tangents', () => {
+    for (let face = 0; face < 6; face++) {
+      const boxFace = new BoxFace(face);
+      const tangent1 = boxFace.getTangent1();
+      const tangent2 = boxFace.getTangent2();
+      expect(tangent1.getComponent(boxFace.tangentIndex1)).toBe(1);
+      expect(tangent2.getComponent(boxFace.tangentIndex2)).toBe(1);
+      expect(tangent1.length()).toBe(1);
+      expect(tangent2.length()).toBe(1);
+    }
```
danpriori, line 107: I see the creation of BoxFace via loop repeating itself in several tests. Is it possible to have a creation function and be reused? And maybe call it inside the beforeEach, if it does make sense?

danpriori (changes requested):
In general, the tests seem good and, in my view, easy to understand, due to the level of complexity that this type of test requires :)

However, I still feel that there are tests that are missing, maybe. Tests to check what happens with limiting or invalid values, for example:
What happens if we do BoxFace(1000) or BoxFace(-1)?
If does make sense to have it.

I added some more comments in the code

react-components/src/architecture/concrete/primitives/common/BoxFace.test.ts
nilscognite: This test that the received planes are correct. It has to be like this. But added another test for isPointVisibleByPlanes.

react-components/src/architecture/concrete/primitives/common/BoxFace.test.ts
nilscognite: Reuse the iterator

nilscognite:
> In general, the tests seem good and, in my view, easy to understand, due to the level of complexity that this type of test requires :)
> 
> However, I still feel that there are tests that are missing, maybe. Tests to check what happens with limiting or invalid values, for example: What happens if we do BoxFace(1000) or BoxFace(-1)? If does make sense to have it.
> 
> I added some more comments in the code

I address this in d65f6e8789902cd07b208565d7ecf1241e501156. Now illegal faces can not be given.

danpriori (dismissed):
LGTM

react-components/src/architecture/concrete/primitives/common/BoxFace.test.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  test('should copy and check if equal', () => {
+    const a = new BoxFace(2);
+    const b = new BoxFace(4);
+    expect(a.equals(b)).toBe(false);
+    b.copy(a);
+    expect(b.equals(a)).toBe(true);
+  });
+
+  test('should get correct position from fromPositionAtFace', () => {
```
pramod-cog, line 53: `fromPositionAtFace` is tested for happy path, I would also recommend to add a test case for handling ambiguous vector value

react-components/src/architecture/concrete/primitives/common/BoxFace.ts
```diff
@@ -5,13 +5,16 @@ import { type Matrix4, Plane, Vector2, Vector3 } from 'three';
...
+  private _face = 0;
 
   // ==================================================
   // CONSTRUCTOR
   // ==================================================
 
   public constructor(face: number = 0) {
+    if (face < 0 || face > 5) {
+      throw new Error('Invalid face ' + face);
+    }
```
pramod-cog: (suggestion)
This can be a function as same check is done in setter as well

nilscognite: Fixed here: 9654abb30474576df67063082432b45d666edec5

react-components/src/architecture/concrete/primitives/common/BoxFace.test.ts
nilscognite: Fixed here: 9654abb30474576df67063082432b45d666edec5

PR #5167, opened 2025-06-05
feat(react-components): imperative function to fetch classic CAD assets

haakonflatval-cognite:
#### Type of change

Feat  

## Description :pencil:

Implements an imperative version of the asset-fetching hooks. It will be a single function that, given a search query and a set of 3D resources, returns all assets relevant to the resources. It is modeled closely after the `useCombineAssetResultData` hook and related functions used in `ThreeDResults.tsx` in Fusion, which hopefully makes it easy to replace the `useSearchMappedEquipmentAssetMappings` hook (which fetches CAD asset mappings) and the other asset-fetching hooks in the future. 

For now however, the only intended usage of this hook is inside a `useInfiniteQuery`-call in our new search panel component. See related Fusion-PR.

It is called `searchClassicAssetsForModels`, although at the moment only the inner function `searchClassicAssetsForCadModels` is implemented (i.e. it will only work for CAD models). I have still chosen to expose it with this general name, and some commented-out function calls on the inside, as I intend to implement the rest shortly after this PR is finalized. This also makes it possible to test in Fusion at this stage.

Another interesting aspect of our classic asset-fetching functions is that we only paginate on CAD-asset mappings. This should eventually be addressed (so that we paginate on all other model types too), but I don't intend that to be part of the initial work on this function.

### Tests

I have only written one test in this PR, for the top-level function. It calls the other functions, so that almost all of the PR is covered by the tests. I have opted for this at the moment, since I'd rather not touch the file structure of the original implementations too much. 

Linked to related Fusion PR.


react-components/src/hooks/network/getAssetsList.ts
```diff
@@ -24,6 +24,8 @@ export const getAssetsList = async (
 ): Promise<{ items: Asset[]; nextCursor: string | undefined }> => {
   const advancedFilter = buildFilter(query);
 
+  // The `aggregatedProperties` parameter is not yet public, and hidden behind an alpha flag,
+  // which is why we cannot use the sdk.assets.list method directly
   return await sdk
     .post<CursorResponse<Asset[]>>(`/api/v1/projects/${sdk.project}/assets/list`, {
       headers: {
```
haakonflatval-cognite, line 28: This is not directly related to the rest of the PR, but I came across it when making tests

react-components/src/query/network/searchClassicAssetsForModels.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      .object()
+  )
+  .object();
+
+describe(searchClassicAssetsForModels.name, () => {
+  beforeEach(() => {
+    vi.resetAllMocks();
+    mockAssetMappings3dFilter.mockImplementation(
+      async () =>
+        await (Promise.resolve({ items: [] }) as unknown as CursorAndAsyncIterator<AssetMapping3D>)
```
haakonflatval-cognite: I have been unable to mock the `CursorAndAsyncIterator` return value from the SDK without resorting to `as unknown as CursorAndAsyncIterator`. I'd be glad to hear opinions on how to solve this better

react-components/src/query/network/searchClassicAssetsForModels.ts
```diff
@@ -0,0 +1,50 @@
...
+    searchQuery,
+    models,
+    limit,
+    cadAssetsCursor,
+    sdk,
+    assetMappingAndNode3DCache
+  );
+
+  /*
+  const isFirstPage = cadAssetsCursor === undefined;
```
haakonflatval-cognite: The comments in this file are stand-ins for upcoming implementation work

react-components/src/query/network/getAssetsFromAssetMappings.ts
haakonflatval-cognite, line 1: This function was moved out of /query/useSearchMappedEquipmentAssetMappings.ts

react-components/src/query/network/searchClassicAssetsForModels.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    );
+
+    mockAssetsRetrieve.mockImplementation(findIdEitherInAssetList);
+  });
+
+  describe('without query string', () => {
+    test('callls `sdk.assetMappings3D.filter` with input models and returns empty result when models have no mapped data', async () => {
```
anders-hopland:
```suggestion
    test('calls `sdk.assetMappings3D.filter` with input models and returns empty result when models have no mapped data', async () => {
```

react-components/src/query/network/fetchAllAssetsForCadModels.ts
```diff
@@ -0,0 +1,65 @@
...
+  limit: number,
+  cursor: string | undefined,
+  sdk: CogniteClient
+): Promise<SearchClassicCadAssetsResponse> {
+  const cursorsForModels =
+    cursor === undefined ? undefined : (JSON.parse(cursor) as CursorForModel[]);
+
+  const firstPage = cursorsForModels === undefined;
+
+  const currentPagesOfAssetMappingsPromises = models.map(async (model) => {
```
anders-hopland: We should chunk this in case we have a lot of models we try to fetch asset mappings for in parallell, as this can exceed the maximum number of concurrent requests

react-components/src/query/network/fetchAllAssetsForCadModels.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type CursorForModel = {
+  cursor: string | undefined;
+  model: AddModelOptions<ClassicDataSourceType>;
+};
+
+export async function fetchAllAssetsForCadModels(
+  models: Array<AddModelOptions<ClassicDataSourceType>>,
+  limit: number,
+  cursor: string | undefined,
```
anders-hopland: any reason wy this is passed as a string instead of as `CursorForModel`?

react-components/src/query/network/getAssetsFromAssetMappings.ts
```diff
@@ -0,0 +1,27 @@
+import { type AddModelOptions } from '@cognite/reveal';
+import { type AssetMapping3D, type CogniteClient, type ListResponse } from '@cognite/sdk';
+import { type ModelMappingsWithAssets } from '../useSearchMappedEquipmentAssetMappings';
+
+export async function getAssetsFromAssetMappings(
+  sdk: CogniteClient,
+  modelsMappings: Array<{ model: AddModelOptions; mappings: ListResponse<AssetMapping3D[]> }>
+): Promise<ModelMappingsWithAssets[]> {
+  const mappingsWithAssetsPromises = modelsMappings.map(async ({ mappings, model }) => {
```
anders-hopland: Should be chunked if modelMappings.length is greater than e.g. 10 or 20

react-components/src/query/network/fetchAllAssetsForCadModels.ts
```diff
@@ -0,0 +1,65 @@
...
+import { getAssetsFromAssetMappings } from './getAssetsFromAssetMappings';
+import { type ModelMappingsWithAssets } from '../useSearchMappedEquipmentAssetMappings';
+import { isSameModel } from '../../utilities/isSameModel';
+
+type CursorForModel = {
+  cursor: string | undefined;
+  model: AddModelOptions<ClassicDataSourceType>;
+};
+
+export async function fetchAllAssetsForCadModels(
```
anders-hopland, line 21: I would like to hae a test for this file

react-components/src/query/network/getAssetsFromAssetMappings.ts
```diff
@@ -0,0 +1,27 @@
+import { type AddModelOptions } from '@cognite/reveal';
+import { type AssetMapping3D, type CogniteClient, type ListResponse } from '@cognite/sdk';
+import { type ModelMappingsWithAssets } from '../useSearchMappedEquipmentAssetMappings';
+
+export async function getAssetsFromAssetMappings(
```
anders-hopland, line 8: I would like to have a test for this file

react-components/src/query/network/searchClassicAssetsForModels.ts
anders-hopland: Please add a comment in the code stating that this is for upcoming work

react-components/src/query/network/fetchAllAssetsForCadModels.ts
haakonflatval-cognite: Yes, the reason is that when using the `searchClassicAssetsForModels` function inside the `useInfiniteQuery` deep within FRS's search logic, we need the cursor to be a string, to avoid having to make large changes to the query logic. Whether we parse this string in this function on the function above is a bit the same to me. I can move it one step up if you prefer that.

react-components/src/query/network/fetchAllAssetsForCadModels.ts
haakonflatval-cognite: Added a comment in the PR to excuse myself

react-components/src/query/network/getAssetsFromAssetMappings.ts
haakonflatval-cognite: I'll have myself excused

anders-hopland (approved):
LGTM

PR #5175, opened 2025-06-10
fix(react-components): Poi share panel hidden behind Poi Info panel

pramod-cog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:
This PR fixes POI share panel hidden behind POI Info panel. Have tested all other UI for any regression

**Note**: Previous `cogs.js` package upgrade did seems to fix this issue, but after clearing cache, this bug re-appeared.

Fixed Ref:
Screenshot 2025-06-10 at 19 07 46
Issue Ref:
Screenshot 2025-06-10 at 19 08 22

In Search & unified-3d

- Link the local `react-component` build to `fusion`

### DIFF
--- a/react-components/src/components/Architecture/pointsOfInterest/PoiHeader.tsx
+++ b/react-components/src/components/Architecture/pointsOfInterest/PoiHeader.tsx
@@ -37,7 +37,9 @@ export const PoiHeader = (): ReactNode => {
         </Flex>
         <Divider direction="vertical" weight="2px" />
         <Flex direction="row" justifyContent="flex-start">
-          <Dropdown placement="bottom-end" content={<PoiSharePanel />}>
+          {/* zIndex is set to 10000 to ensure the dropdown appears above other elements
+              This is temporary fix until we have styleScope configured */}
+          <Dropdown placement="bottom-end" content={<PoiSharePanel />} zIndex={10000}>
             <Tooltip placement="top-end" content={t({ key: 'SHARE' })}>
               <Button icon={<ShareIcon />} type="ghost" />
             </Tooltip>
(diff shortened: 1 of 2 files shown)


react-components/src/components/Architecture/pointsOfInterest/PoiHeader.tsx
```diff
@@ -37,7 +37,9 @@ export const PoiHeader = (): ReactNode => {
         </Flex>
         <Divider direction="vertical" weight="2px" />
         <Flex direction="row" justifyContent="flex-start">
-          <Dropdown placement="bottom-end" content={<PoiSharePanel />}>
+          {/* zIndex is set to 10000 to ensure the dropdown appears above other elements
+              This is temporary fix until we have styleScope configured */}
+          <Dropdown placement="bottom-end" content={<PoiSharePanel />} zIndex={10000}>
```
danpriori, line 42: As you have explained in DM, it is not possible easily to have the zIndex in a styled component because it would be overwritten and the only option for now is to put it inline. 
If it has no other better option for now, then maybe the best would be to put the comment out of the template? Perhaps above the `return`. And maybe add a TODO to investigate that later.
But I dont have a strong opinion on where to put the comments. Just a suggestion.
```
  /**
   * The Dropdown component is using zIndex inline with 10000 to ensure the dropdown appears 
   * above other elements. This is temporary fix until we have styleScope configured
   */
```

danpriori (changes requested):
Just a comment about how to address this later and about the comment itself.

react-components/src/components/Architecture/pointsOfInterest/PoiHeader.tsx
pramod-cog: I prefer to have a comment at the fix itself, it gives point of correction for later

danpriori (approved):
👍

PR #5181, opened 2025-06-17
feat(react-components): Make it possible to change between meter, feet and inches.

nilscognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/CDFNEW-2685
https://cognitedata.atlassian.net/browse/BND3D-5808

See test instructions.
I have also updated all unit tests, so the converge covers the changes

## Description :pencil:

The length unit had before just feet and meter, and was controlled by a `Boolean`. Now it's a `enum`: Feet,  Inches and Meters.  It is a more scalable solution, since it is very easy to add new units. The class `UnitSystem `is therefore refactored a lot. 

Number of digits is still 2 for all units, since we can not guaranty more that 1 cm precision. For inches we round it to closest 0.25 inch which is is about 1 cm. (I couldn't find a good rounding number for feet that gives 1 cm precision)

I had to update the two commands to handle this:

* `SetLengthUnitCommand `(option for the unit)
* `CycleLengthUnitsCommand  `(cycling the available units)

These commands was renamed and moved to a unit folder. So these is quite a lot changes in the usage of those command.
No changes in the React/HTML layer as usual, regarding this. This is pure business code changes.

But there is a change in the `SettingsButton`. It relates to the layout to the popup. The designer wanted to get rid of the scroll bar, have less space between the element and place it to the right of the settings button, like this:

image

Before it was like this (when adding new options, a scrollbar appeared, not user friendly)

image

I updated the translation files. The translation is copied from Fusion, since the had units as a key.

All changes is approved visually with the designer @Umy-magic 

See test instructions. And it is under unit test.

1. yarn storybook
2. Active measurement tool in the toolbar to the left
3. Create a box by the tool at bottom.
4. Click on the measure icon in the card. It should cycle between the available units, see the updates at the measure you just did
5. Click at the setting button
6. Not a new unit settings.
7. Change it an see the updates at the measure you just did

## Todo:

The `lengthUnit `should be a signal. But it will require a lot of changes in 3D views. I don't want to do this in this PR. So I will do it in another.


react-components/src/common/i18n/de/reveal-react-components.json
```diff
@@ -149,6 +149,7 @@
   "TOUCH_ZOOM": "Vergrößern",
   "TRY_AGAIN": "Erneut versuchen",
   "UNDO": "Rückgängig machen",
+  "UNITS": "Einheiten",
```
haakonflatval-cognite, line 152: Just curious - how did we make these translations? :sweat_smile:

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -236,7 +244,7 @@ const StyledToggleContainer = styled(Flex).attrs({
 `;
 
 const StyledMenuPanel = styled.div<{ $flexDirection: FlexDirection }>`
-  max-height: 400px;
+  max-height: 600px;
```
haakonflatval-cognite, line 247: Just so I understand - we use this to increase the size of the dropdown panel?

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -40,6 +40,14 @@ export const SettingsButton = ({
   const flexDirection = getFlexDirection(placement);
   const isTooltipDisabled = isOpen || label === undefined;
 
+  // This ensures the rule for where the dropdown panel should open:
+  // Horizontal toolbar at top:    Below the button
+  // Horizontal toolbar at bottom: Above the button
+  // Vertical toolbar at left:     Right of the button
+  // Vertical toolbar at right:    Left of the button
+  const dropdownPlacement =
+    placement === 'top' || placement === 'bottom' ? 'bottom-end' : 'right-start';
```
haakonflatval-cognite, line 49: Not sure I understand if the code does what the comment says :thinking: Won't this place the tooltip at `bottom-end` both when placement is manually set to `top` or `bottom`?

haakonflatval-cognite (commented):
Looks good, but had a couple of comments

react-components/src/components/Architecture/SettingsButton.tsx
nilscognite: I didn't either. I experimented by placing the toolbar on different side and this worked. The `dropdownPlacement  `is also dependent on the offset settings where it actually is used. For me this is magic, and it was similar to this before , but it handled the toolbar on the left side only.

react-components/src/common/i18n/de/reveal-react-components.json
nilscognite: I took it from fusion. Unchanged

react-components/src/components/Architecture/SettingsButton.tsx
nilscognite: Yes. **Umy** wanted it bigger with no scrollbar visible for "normal use". But is might be too big. Not sure actually.

haakonflatval-cognite (dismissed):
LGTM :+1:

m-amaya:
Ping me once the build errors are resolved and I will review.

react-components/src/architecture/base/concreteCommands/KeyboardSpeedCommand.test.ts
```diff
@@ -38,9 +38,6 @@ describe(KeyboardSpeedCommand.name, () => {
   test('Should change the cameraKeyBoardSpeed in RevealSettingsController', () => {
     assert(command.children !== undefined);
     for (const option of command.children) {
-      expect(option).toBeDefined();
```
nilscognite, line 41: This is always define

react-components/src/architecture/base/concreteCommands/SettingsCommand.test.ts
```diff
@@ -1,9 +1,8 @@
 import { describe, expect, test } from 'vitest';
-import { CopyToClipboardCommand } from './CopyToClipboardCommand';
 import { isEmpty } from '../utilities/TranslateInput';
 import { SettingsCommand } from './SettingsCommand';
 
-describe(CopyToClipboardCommand.name, () => {
```
nilscognite, line 6: Used wrong name, so I renamed it

react-components/src/architecture/base/concreteCommands/SettingsCommand.ts
```diff
@@ -59,6 +60,9 @@ export class SettingsCommand extends BaseSettingsCommand {
     this.add(new SetPointColorTypeCommand());
     this.add(new SetPointShapeCommand());
     this.add(new PointCloudFilterCommand());
+
+    this.add(new DividerCommand());
+    this.add(new SetLengthUnitCommand());
```
nilscognite, line 65: Add SetLengthUnitCommand to the bottom

m-amaya:
LGTM

m-amaya:
🦄

anders-hopland (approved):

PR #5184, opened 2025-06-17
feat(react-components): filters in classic search function

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

## Description :pencil:

ntroduce custom filtering capabilities in the searchClassicAssetsForModels function. The filters correspond directly to those used in the asset-filter endpoint (https://api-docs.cognite.com/20230101/tag/Assets/operation/listAssets).

Note that I have simultaneously removed the searchQuery field in the options. Any search string should now be supplied through the filters. For internal use, this can be done by using the buildQueryFilter function.

In Fusion, Unified 3D

TBA


react-components/src/hooks/network/getAssetsList.ts
```diff
@@ -1,26 +1,24 @@
...
     limit = 1000,
     sort = sortOption,
+    filter,
     aggregatedProperties = ['path']
   }: {
-    query: string;
     cursor?: string;
     limit?: number;
     sort?: Array<{ property: string[] }>;
+    filter?: AllAssetFilterProps;
```
danpriori: Can we use the plural version of the word here? As we are dealing with both types of filters, I think it is better to keep the same pattern for consistency (like we have in other places for the same type)

react-components/src/query/network/getAssetMappedPointCloudAnnotations.ts
```diff
@@ -8,17 +8,19 @@ import {
...
-const MAX_PARALLEL_ASSET_REQUESTS = 5;
+import { getAssetsForIds } from './getAssetsForIds';
+import { toIdEither } from '../../utilities/instanceIds/toIdEither';
+import { type AllAssetFilterProps } from './types';
 
 export async function getAssetsMappedPointCloudAnnotations(
-  sdk: CogniteClient,
-  models: Array<AddPointCloudResourceOptions<ClassicDataSourceType>>
+  models: Array<AddPointCloudResourceOptions<ClassicDataSourceType>>,
+  filter: AllAssetFilterProps | undefined,
```
danpriori: Can we use the plural version of the word here? As we are dealing with both types of filters, I think it is better to keep the same pattern for consistency (like we have in other places for the same type)

react-components/src/query/network/searchClassicAssetsForModels.test.ts
```diff
@@ -288,37 +319,66 @@ describe(searchClassicAssetsForModels.name, () => {
...
+
+    test('calls post-endpoint with query filter and returns relevant assets when there is contextualization', async () => {
       const mockAnnotations = TEST_FILES.map((file) =>
         createClassic360AnnotationMock({ fileId: file.id, assetId: TEST_ASSETS[0].id })
       );
       mockAnnotationsList.mockReturnValue(
         createCursorAndAsyncIteratorMock({ items: mockAnnotations })
       );
+
+      const queryFilter = buildClassicAssetQueryFilter('asset0');
```
danpriori:
```suggestion
      const queryAdvancedFilter = buildClassicAssetQueryFilter('asset0');
```

react-components/src/query/network/searchClassicAssetsForModels.test.ts
```diff
@@ -234,42 +241,66 @@ describe(searchClassicAssetsForModels.name, () => {
...
+      );
 
       mockAnnotationsList.mockReturnValue(
         createCursorAndAsyncIteratorMock({
-          items: pointCloudAnnotations
+          items: TEST_POINT_CLOUD_ANNOTATIONS
         })
       );
 
+      const queryFilter = buildClassicAssetQueryFilter('asset1');
```
danpriori:
```suggestion
      const queryAdvancedFilter = buildClassicAssetQueryFilter('asset1');
```

react-components/src/query/network/searchClassicAssetsForModels.test.ts
```diff
@@ -234,42 +241,66 @@ describe(searchClassicAssetsForModels.name, () => {
     });
 
     test('returns empty result when no relevant results exist', async () => {
+      const queryFilter = buildClassicAssetQueryFilter('some-query');
```
danpriori:
```suggestion
      const queryAdvancedFilter = buildClassicAssetQueryFilter('some-query');
```

react-components/src/query/network/searchClassicAssetsForModels.test.ts
```diff
@@ -211,11 +218,11 @@ describe(searchClassicAssetsForModels.name, () => {
         );
         mockPostAssetList.mockResolvedValue(createHttpResponseObject({ items: TEST_ASSETS }));
 
+        const queryFilter = buildClassicAssetQueryFilter('some-query');
```
danpriori:
```suggestion
        const queryAdvancedFilter = buildClassicAssetQueryFilter('some-query');
```

react-components/src/query/network/searchClassicAssetsForModels.test.ts
```diff
@@ -193,11 +200,11 @@ describe(searchClassicAssetsForModels.name, () => {
     describe('with search query', () => {
       test('returns with no mappings when query has no result', async () => {
         mockGetAssetMappingsForModel.mockResolvedValue([]);
+        const queryFilter = buildClassicAssetQueryFilter('some-query');
```
danpriori:
```suggestion
        const queryAdvancedFilter = buildClassicAssetQueryFilter('some-query');
```

react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
```diff
@@ -71,10 +72,13 @@ export const useSearchMappedEquipmentAssetMappings = (
     accumulatedAssets: Asset[],
     mappedSearchedAssetIds: Set<number>
   ): Promise<{ assets: Asset[]; nextCursor: string | undefined }> => {
+    const advancedFilter = buildClassicAssetQueryFilter(query);
     const searchedAssetsResponse = await getAssetsList(sdk, {
-      query,
       limit: 1000,
-      cursor
+      cursor,
+      filter: {
```
danpriori: can we use the plural here as well?

react-components/src/hooks/network/getAssetsList.ts
```diff
@@ -1,26 +1,24 @@
...
     aggregatedProperties?: string[];
   }
```
danpriori: I would suggest to have a type definition for this

react-components/src/query/network/getAssetsForIds.ts
```diff
@@ -0,0 +1,82 @@
+import type { Asset, CogniteClient, IdEither } from '@cognite/sdk';
+import { chunk } from 'lodash';
+import { buildClassicAssetIdFilter, combineClassicAssetFilters } from './buildClassicAssetFilter';
+import { getAssetsList } from '../../hooks/network/getAssetsList';
+import type { AllAssetFilterProps } from './types';
+
+const MAX_PARALLEL_ASSET_REQUESTS = 5;
```
danpriori, line 7: This is a thing I would like to raise before. Why we have this limitation? Is it arbitrary limitation? Do we have any benchmark from our API/backend documentation/team on how many parallel requests it can handle based on our resources or what is the recommendation?
I mean, sometimes we can have more depending on the situation or our load balancer, or the recommendation is less 🤔

react-components/src/query/network/getAssetsForIds.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const MAX_LIMIT_ASSETS_BY_IDS = 1000;
+const MAX_LIMIT_ASSETS_BY_LIST_WITH_IDS = 100;
+
+export async function getAssetsForIds(
+  assetRefs: IdEither[],
+  filter: AllAssetFilterProps | undefined,
```
danpriori: As I mentioned before, can we have the plural here as it is including both filter types?

react-components/src/query/network/getAssetsForIds.ts
```diff
@@ -0,0 +1,82 @@
...
+      assetRefChunkBatch.map(async (assetRefChunk) => {
+        const idFilter = buildClassicAssetIdFilter(assetRefChunk);
+
+        const { items } = await getAssetsList(sdk, {
+          filter: {
+            ...filter,
+            advancedFilter: combineClassicAssetFilters([filter?.advancedFilter, idFilter])
+          },
+          limit: MAX_LIMIT_ASSETS_BY_LIST_WITH_IDS
+        });
```
danpriori:
```suggestion
        const advancedFilter = combineClassicAssetFilters([filter?.advancedFilter, idFilter]);

        const { items } = await getAssetsList(sdk, {
          filter: {
            ...filter?.filter,
            advancedFilter
          },
          limit: MAX_LIMIT_ASSETS_BY_LIST_WITH_IDS
        });
```

react-components/src/query/network/getAssetsForIds.ts
```diff
@@ -0,0 +1,82 @@
...
+        // can expect that there will be no need for pagination
+        return items;
+      })
+    );
+
+    allAssetResults.push(...assetsForChunkBatch.flat());
+  }
+
+  return allAssetResults;
+}
```
danpriori, line 66: I think it's more interesting to break this function into two parts to reduce the reading and structural complexity: one part focused on producing the chunks and within each chunk you call another function getAssetRefChunkBatch, something like this : 

```typescript
async function getAssetsForIdsWithFilter(
  assetRefs: IdEither[],
  filter: AllAssetFilterProps,
  sdk: CogniteClient
): Promise<Asset[]> {
  const assetRefChunks = chunk(assetRefs, MAX_LIMIT_ASSETS_BY_LIST_WITH_IDS);

  const chunkedAssetRefs =  chunk(assetRefChunks, MAX_PARALLEL_ASSET_REQUESTS);
  const allAssetResults: Asset[] = [];

  for (const assetRefChunkBatch of chunkedAssetRefs) {
    const assetsForChunkBatch = await getAssetRefChunkBatch(
      assetRefChunkBatch,
      filter,
      sdk
    );

    allAssetResults.push(...assetsForChunkBatch.flat());
  }

  return allAssetResults;
}

async function getAssetRefChunkBatch(
  assetRefChunk: IdEither[][],
  filter: AllAssetFilterProps,
  sdk: CogniteClient
): Promise<Asset[]> {
  const idFilter = buildClassicAssetIdFilter(assetRefChunk.flat());
  const advancedFilter = combineClassicAssetFilters([filter.advancedFilter, idFilter]);

  const { items } = await getAssetsList(sdk, {
    filter: {
      ...filter.filter,
      advancedFilter
    },
    limit: MAX_LIMIT_ASSETS_BY_LIST_WITH_IDS
  });

  // Since the filter narrows down the result to only the input IDs, we
  // can expect that there will be no need for pagination
  return items;
};
```

react-components/src/query/network/getAssetsForIds.ts
```diff
@@ -0,0 +1,82 @@
...
+
+async function getAssetsForIdsWithFilter(
+  assetRefs: IdEither[],
+  filter: AllAssetFilterProps | undefined,
```
danpriori: I think from here on the filter is not undefined anymore, right?

react-components/src/query/network/buildClassicAssetFilter.ts
```diff
@@ -0,0 +1,72 @@
...
+      property: ['externalId'],
+      values: externalIds
+    }
+  };
+}
+
+export function combineClassicAssetFilters(
+  filters: Array<AssetAdvancedFilterProps | undefined>
+): AssetAdvancedFilterProps | undefined {
+  const definedFilters = filters.filter((filter) => filter !== undefined);
```
danpriori: I guess it can be replaced using the isDefined:

`filters.filter(isDefined);`

react-components/src/query/network/buildClassicAssetFilter.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const externalIdsFilter =
+    externalIds.length > 0 ? buildAssetExternalIdFilter(externalIds) : undefined;
+  const internalIdsFilter =
+    internalIds.length > 0 ? buildAssetInternalIdFilter(internalIds) : undefined;
+
+  if (externalIdsFilter !== undefined && internalIdsFilter !== undefined) {
+    return { and: [externalIdsFilter, internalIdsFilter] };
+  } else {
+    return externalIdsFilter ?? internalIdsFilter;
```
danpriori: Just a question: is there any reason we do prefer the externalIds over internal ones in here? All the assets endpoints support both at the same level, right? Or usually the most inputs we get is external ids here ?

react-components/src/query/network/getClassicAssetMapped360Annotations.ts
```diff
@@ -18,28 +18,31 @@ import {
...
+import { type AllAssetFilterProps } from './types';
 
 const MAX_PARALLEL_ANNOTATION_QUERIES = 5;
 const MAX_PARALLEL_FILES_QUERIES = 5;
 
 export async function getClassicAssetMapped360Annotations(
   siteIds: string[],
-  sdk: CogniteClient,
-  image360AnnotationFilterOptions: Image360AnnotationFilterOptions
+  assetFilter: AllAssetFilterProps | undefined,
```
danpriori: If that plural suggestion does make sense, then I think we can have it here as well 🙈

react-components/src/query/network/getAssetsForIds.ts
```diff
@@ -0,0 +1,82 @@
...
+): Promise<Asset[]> {
+  const assetRefChunks = chunk(assetRefs, MAX_LIMIT_ASSETS_BY_IDS);
+
+  const allAssetResults: Asset[] = [];
+
+  for (const assetRefChunkBatch of chunk(assetRefChunks, MAX_PARALLEL_ASSET_REQUESTS)) {
+    const assetsForChunkBatch = await Promise.all(
+      assetRefChunkBatch.map(
+        async (assetRefChunk) =>
+          await sdk.assets.retrieve(assetRefChunk, { ignoreUnknownIds: true })
```
danpriori, line 40: Does make sense to isolate/move the sdk request into a separated function (or from the scope `const assetsForChunkBatch = await Promise.all`) to separate the concerns?

react-components/src/query/network/types.ts
```diff
@@ -1,6 +1,51 @@
...
+export function isOrFilter(filter: AssetAdvancedFilterProps): filter is AssetAdvancedFilterOr {
+  return 'or' in filter;
+}
+
+export type AllAssetFilterProps = {
+  filter?: AssetFilterProps;
+  advancedFilter?: AssetAdvancedFilterProps;
+};
+
+export function hasFilters(allAssetFilters: AllAssetFilterProps | undefined): boolean {
```
danpriori: I don't think it is a good idea to mix functions with type definitions within the same file when the file is primarily intended to accommodate only type definitions 🤔 
Is it possible to move out all these functions to a separate place, following some patterns we are already using somewhere else? Maybe a typeGuards.ts or the classic util.ts? Does make sense?

react-components/src/query/network/getAssetsForIds.ts
```diff
@@ -0,0 +1,82 @@
...
+export async function getAssetsForIds(
+  assetRefs: IdEither[],
+  filter: AllAssetFilterProps | undefined,
+  sdk: CogniteClient
+): Promise<Asset[]> {
+  if (assetRefs.length === 0) {
+    return [];
+  }
+
+  if (filter === undefined) {
```
danpriori: we have the util function hasFilters. Is it ok to use it here as well?

react-components/src/query/network/searchClassicAssetsForModels.test.ts
```diff
@@ -234,42 +241,66 @@ describe(searchClassicAssetsForModels.name, () => {
...
 
       expect(result).toEqual({ data: [], nextCursor: undefined });
     });
 
-    test('returns relevant assets when there is contextualization', async () => {
-      const pointCloudAnnotations = TEST_ASSETS.map((asset) =>
-        createPointCloudAnnotationMock({
-          assetId: asset.id,
-          modelId: taggedPointCloudModelOptions.addOptions.modelId
+    test('calls retrieve-endpoint and returns relevant assets when called without filter and there is contextulization', async () => {
```
danpriori:
```suggestion
    test('calls retrieve-endpoint and returns relevant assets when called without filter and there is contextualization', async () => {
```

react-components/src/query/network/searchClassicAssetsForModels.test.ts
```diff
@@ -288,37 +319,66 @@ describe(searchClassicAssetsForModels.name, () => {
...
+        { limit: ARBITRARY_SEARCH_LIMIT },
         mockSdk,
         mockRenderTarget
       );
 
       expect(result).toEqual({ data: [], nextCursor: undefined });
     });
 
-    test('returns relevant assets when there is contextualization', async () => {
+    test('calls retrieve-endpoint and returns relevant assets when called without filter and there is contextulization', async () => {
```
danpriori:
```suggestion
    test('calls retrieve-endpoint and returns relevant assets when called without filter and there is contextualization', async () => {
```

react-components/src/query/network/searchClassicAssetsForModels.ts
```diff
@@ -9,17 +8,28 @@ import { type Asset, type CogniteClient } from '@cognite/sdk';
...
+   * This contains both a `filter` and an `advancedFilter` field, which correspond
+   * to the fields of the same name that are used in the asset-filter endpoint
+   * https://api-docs.cognite.com/20230101/tag/Assets/operation/listAssets
+   */
+  filters?: AllAssetFilterProps;
+};
+
 export async function searchClassicAssetsForModels(
-  searchQuery: string,
   resources: TaggedAddResourceOptions[],
```
danpriori, line 31: So from now on we wont have the search by search string and only by filters (classic and advanced) and with the returned data we then filter the results out by the search string on client side?

react-components/src/query/useSearchAssetsMapped360Annotations.tsx
```diff
@@ -86,7 +86,8 @@ async function getAssetMapped360Annotations(
 
   return await getClassicAssetMapped360Annotations(
     classicSiteIds,
-    sdk,
-    image360AnnotationFilterOptions
+    undefined,
```
danpriori: I would suggest to transform the arguments as a single object to avoid forcing defining "undefined"'s but it is more a personal taste as this is not a long list of arguments yet :)

danpriori (changes requested):
A couple of comments, suggestions and questions. The rest looks good 👍

react-components/src/query/network/getAssetsForIds.ts
haakonflatval-cognite: Yes, this number is very much taken out of the air :sweat_smile: The primary function is just to avoid 429-errors. I'm not sure we currently should invest in optimizing this number - but it may be worthwhile to look into in the future :thinking:

react-components/src/query/network/getAssetsForIds.ts
haakonflatval-cognite: Makes sense :+1:

react-components/src/query/network/buildClassicAssetFilter.ts
haakonflatval-cognite: Ah, yeah, the ordering here is arbitrary. The if-statement ensures that only one of them can be defined (or none), so the ordering here wouldn't matter anyway

react-components/src/query/network/getAssetsForIds.ts
haakonflatval-cognite: Not sure if that would make things easier :thinking: I've already refactored this function now, so you can see if you want more changes :+1:

react-components/src/query/network/types.ts
haakonflatval-cognite: I moved everything filter-related (types and functions) to a `filters.ts` file in the same folder. What do you think? I personally think it's fine to have types and functions together, as long as they're closely related

react-components/src/query/network/getAssetsForIds.ts
haakonflatval-cognite: Ah, yes - good catch :ok_hand:

react-components/src/query/network/searchClassicAssetsForModels.ts
haakonflatval-cognite: No, the intended way to use this, is that the search string is being converted to a filter before passed into this function. Search passes it as a filter, so this makes it easy for us to use

react-components/src/query/network/getAssetsForIds.ts
danpriori: I agree

react-components/src/query/network/filters.ts
```diff
@@ -0,0 +1,54 @@
...
+  | AssetAdvancedFilterLeaf;
+
+export function isLeafFilter(filter: AssetAdvancedFilterProps): filter is AssetAdvancedFilterLeaf {
+  return !isBooleanFilter(filter);
+}
+
+function isBooleanFilter(
+  filter: AssetAdvancedFilterProps
+): filter is AssetAdvancedFilterAnd | AssetAdvancedFilterOr {
+  return 'and' in filter || 'or' in filter;
```
danpriori: I think we can have a composition here, isnt?
Like, `return isAndFilter(filter) || isOrFilter(filter);`

danpriori (changes requested):
Just a last comment. Sorry 🙈 
The rest looks good!

react-components/src/query/network/filters.ts
haakonflatval-cognite: Sure :+1:

react-components/src/query/network/buildClassicAssetFilter.ts
```diff
@@ -22,6 +22,10 @@ export function buildClassicAssetQueryFilter(query: string): AssetAdvancedFilter
 export function buildClassicAssetIdFilter(
   assetRefs: IdEither[]
 ): AssetAdvancedFilterProps | undefined {
+  if (assetRefs.length === 0) {
+    return undefined;
+  }
```
danpriori: What is the common pattern for early undefined returns? Are we gonna use it, or will we let the requester deal with empty/undefined vars/lists outside?
I'm fine with the early return here because it is under this function context (the function decides to return undefined if the list is empty, and the outside callers shouldn't care about it). But I saw some different opinions within the team on similar situations, so do we have a consensus on that?

haakonflatval-cognite: Hm, good question - I believe Pramod wanted me to put it here (this function isn't part of this PR, but was part of the previous one, and it pops up here due to the merge conflict. Pramod suggested having this, although I think it's a bit unnecessary)

danpriori (approved):
LGTM

PR #5185, opened 2025-06-18
fix(react-components): ensure computed model keys are unique across DM-models and bump version to 0.79.4

haakonflatval-cognite:
…M models

#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:

When finding the "classic" model ID and revision ID for a DM point cloud model, we create a query key for it (we don't want to use the model itself as a query key because `useQuery` serializes the key). In this computation we send the entire model, of type `CognitePointCloudModel`, into the function, which expects `AddModelOptions`. `AddModelOptions` either takes the form `{ modelId: number, revisionId: number }` or `{ revisionExternalId: string, revisionSpace: string }`. Incidentally, `CognitePointCloudModel` fits into the first of these two definitions, but not the second - but since it's a DM model, it has `modelId` and `revisionId` hard-coded to 0. 

The fix entails making sure we use the more correct `modelIdentifier` as input to the function.

In Fusion, 3d-dev-quickstart

In 3d-dev-quickstart:
- Go to Search (list view)
- Select location Office
- Go to 3D
- It should have selected the scene "Test point cloud scene"
- Click the breadcrumb, open 3D data modal
- Open the scene "Masa scene test"
- Show annotations
- Observe that clicking the annotation highlights it, unlike what happens in production


christjt (dismissed):
risk review ok

christjt (approved):
risk review ok

PR #5187, opened 2025-06-22
feat (react-components): Make possible to measure horizontal cylinder and change the way vertical cylinders is created.

nilscognite:
#### Type of change
Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/CDFNEW-2684
https://cognitedata.atlassian.net/browse/BND3D-5807

## Description :pencil:

There is added one extra tool for making the horizontal cylinder:

image

There is updates in the `MeasurmentTool`, `SetMeasurementTypeCommand`, `CylinderCreator `(for clicking around when creating a new cylinder), `CylinderDomainObject`, `CylinderDragger`, 

In the `CylinderView `there is a lot changes. I had to refactor the code to make it easier. Before the radius was only shown at top if it was a vertical cylinder. Now this is generalize, so it is shown at both ends for all types of cylinders.

`CylinderDragger`  is also refactored. Now, the first two point will calculate the **diameter**, rather than the **radius**. This reuqired a loot of refactoring. Also I have done simplification and added comment to better understanding.

The changes to the LineView is not related to this. I have done these changes just to improve the visualization (Setting down the radius of the lines). If this is a problem I can make another PR only for this. 

**Note:** This PR cannot be merge into master before the translation strings has been updated. It change the behavior, so the old tooltip, `MEASUREMENTS_ADD_CYLINDER `is renamed and change. In fact I need the following string to be translated:

`CYLINDER_VERTICAL, CYLINDER_HORIZONTAL, MEASUREMENTS_ADD_CIRCLE, MEASUREMENTS_ADD_CYLINDER_VERTICAL and MEASUREMENTS_ADD_CYLINDER_HORIZONTAL. 
`

There are several PR that unit test this code. These PR's are not merge yet due to slow reviewing process.

PR for testing the `CylinderView`, see https://github.com/cognitedata/reveal/pull/5188
PR for testing some measurement commands, see https://github.com/cognitedata/reveal/pull/5194

I also updated the unit test for the **CylinderCreator**

1. `yarn storybook`
2. Click measurement tool at left toolbar
3. Select horizontal cylinder creation method on the toolbar at bottom.
4. Create the horizontal  cylinder (see tooltip how to do it)
5. Try to edit it. By dragging the ring in the end you can rotate it along the Z-axis, so it is still horizontal .
6. You can also repeat this for the vertical cylinder.

image


react-components/src/architecture/concrete/primitives/cylinder/CylinderCreator.test.ts
```diff
@@ -6,81 +6,104 @@ import { FocusType } from '../../../base/domainObjectsHelpers/FocusType';
...
+      const y = 1;
+      const z = 2;
+      const radius = 1.5;
 
-    const center = new Vector3(dx, 0, 0);
-    const points = [center, new Vector3(dx, radius, 2)];
-    click(creator, points[0], direction, false, center);
-    click(creator, points[1], direction, true);
+      const points = [new Vector3(x, y - radius, z), new Vector3(x, y + radius, z)];
+      clickMe(creator, points[0], false, true);
```
anders-hopland, line 25: Why do we click here? I don't see any assertions checking if the click was successful or nor

anders-hopland (commented):
Looks mostly good. Please see my comment regarding the test
Would it be possible to make the initial creation of the cylinder a bit snappier. It's super responsive when resizing after the cylinder is created, but it's lagging a bit when initially creating the measurement cylinder

react-components/src/architecture/concrete/primitives/cylinder/CylinderCreator.test.ts
nilscognite: I don't understand?? There are several test after clicking is performed. These test that the clicking has created the expected cylinder. In this test, I pretend 2 click, the check the result.

```
      clickMe(creator, points[0], false, true);
      clickMe(creator, points[1], true, hasIntersectionPoint);

     -->  expect(creator.domainObject).toBe(domainObject);
     -->  expectEqualVector3(domainObject.cylinder.center, new Vector3(x, y, z));
     -->  expect(domainObject.cylinder.radius).toBeCloseTo(radius);
     -->  expect(domainObject.focusType).toBe(FocusType.Focus);
```

In addition I do a check inside `clickMe -> click -> IsFinished ` to check if the primitive is really created. It is created when the it is finished. Before it is finished is in a Pending state.

anders-hopland: I had a brain fart, the test here is of course correct. My bad

nilscognite:
> Looks mostly good. Please see my comment regarding the test Would it be possible to make the initial creation of the cylinder a bit snappier. It's super responsive when resizing after the cylinder is created, but it's lagging a bit when initially creating the measurement cylinder

This make it more snappier: 607dc838e95d183d6a1f5e5add95385b758c63cb

anders-hopland (approved):
🦄

PR #5188, opened 2025-06-23
test (react-component): Synchronize and refactor unit test for Views

nilscognite:
#### Type of change
Test Refactor

## Description :pencil:

This cleans up the unit tests for Views, since they were made in different PR's and at different time. 

### DIFF
--- a/react-components/src/architecture/base/utilities/sprites/createSprite.ts
+++ b/react-components/src/architecture/base/utilities/sprites/createSprite.ts
@@ -29,11 +29,8 @@ export function createSpriteWithText(
   worldHeight: number,
   color: Color,
   bgColor?: Color
-): Sprite | undefined {
+): Sprite {
   const canvas = createCanvasWithText(text, color, bgColor);
-  if (canvas === undefined) {
-    return undefined;
-  }
   return createSprite(canvas, worldHeight);
 }
 
@@ -92,11 +89,7 @@ function createTexture(canvas: HTMLCanvasElement): Texture {
   return texture;
 }
 
-function createCanvasWithText(
-  text: string,
-  color: Color,
-  bgColor?: Color
-): HTMLCanvasElement | undefined {
+function createCanvasWithText(text: string, color: Color, bgColor?: Color): HTMLCanvasElement {
   // https://www.javascripture.com/CanvasRenderingContext2D
   const borderSize = 2;
   const fontSize = 60; // This gives the resolution of the sprite, not real font side.
@@ -104,19 +97,22 @@ function createCanvasWithText(
 
   const canvas = document.createElement('canvas');
   const context = canvas.getContext('2d');
-  if (context === null) {
-    return undefined;
-  }
 
-  // Measure how long the name will be
-  context.font = font;
-  const textWidth = context.measureText(text).width;
+  if (context !== null) {
+    context.font = font;
+  }
+  const textWidth = context === null ? text.length * 10 : context.measureText(text).width;
   const doubleBorderSize = borderSize * 2;
   const width = textWidth + 2 * doubleBorderSize;
   const height = fontSize + doubleBorderSize;
-
-  canvas.width = width;
+  canvas.width = textWidth + 2 * doubleBorderSize;
   canvas.height = height;
+  if (context === null) {
+    // If context is null, we cannot draw anything
+    // This will happen when having unit test in the Views, since the views doesn't assume
+    // there is a browser. It will make the unit test more realistic.
+    return canvas;
+  }
   context.clearRect(0, 0, canvas.width, canvas.height);
 
   // Draw optional rounded rectangle
--- a/react-components/src/architecture/concrete/axis/AxisThreeView.test.ts
+++ b/react-components/src/architecture/concrete/axis/AxisThreeView.test.ts
@@ -2,10 +2,12 @@ import { assert, beforeEach, describe, expect, test } from 'vitest';
 import { AxisDomainObject } from './AxisDomainObject';
 import { AxisThreeView } from './AxisThreeView';
 import { createFullRenderTargetMock } from '../../../../tests/tests-utilities/fixtures/createFullRenderTargetMock';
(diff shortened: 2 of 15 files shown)


react-components/src/architecture/concrete/axis/AxisThreeView.test.ts
```diff
@@ -32,47 +34,75 @@ describe(AxisThreeView.name, () => {
 
   test('should have initial state', () => {
     expect(view.object).toBeDefined();
+    expectChildrenLength(view, 6 + 144 + 42);
```
pramod-cog: (must be fixed)
These numbers are not self explanatory, would like have constants for it or add comments for better under standing of these numbers

react-components/src/architecture/concrete/axis/AxisThreeView.test.ts
```diff
@@ -32,47 +34,75 @@ describe(AxisThreeView.name, () => {
...
+      x: number,
+      y: number,
+      z: number,
+      view: AxisThreeView
+    ): void {
+      camera.position.set(x, y, z);
+      camera.position.multiplyScalar(200);
       view.beforeRender(camera);
-      expectVisibleChildren(view, 26);
+      checkChildren(view, 21, 5, 48);
```
pramod-cog: Same here on magic numbers

react-components/src/architecture/concrete/axis/AxisThreeView.test.ts
```diff
@@ -32,47 +34,75 @@ describe(AxisThreeView.name, () => {
...
+      camera: PerspectiveCamera,
+      x: number,
+      y: number,
+      z: number,
+      view: AxisThreeView
+    ): void {
+      camera.position.set(x, y, z);
+      camera.position.multiplyScalar(200);
+      view.beforeRender(camera);
+      checkChildren(view, 18, 3, 72);
```
pramod-cog: Same here on magic numbers

react-components/src/architecture/concrete/axis/AxisThreeView.test.ts
```diff
@@ -32,47 +34,75 @@ describe(AxisThreeView.name, () => {
...
   test('should have initial state', () => {
     expect(view.object).toBeDefined();
+    expectChildrenLength(view, 6 + 144 + 42);
     expectChildrenOfTypeAndCount(view, Mesh, 6); // These are the sides (6 i a box)
+    expectChildrenOfTypeAndCount(view, Sprite, 144); // Number of labels
     expectChildrenOfTypeAndCount(view, LineSegments, 42); // There are 6 grids for the box, the rest is axis lines
   });
 
   test('should look towards all 6 sides', () => {
     const camera = new PerspectiveCamera();
```
pramod-cog, line 40: (clarification)
What should happen if we have a `OrthographicCamera` instead of `PerspectiveCamera`.
Can we add a test case for `OrthographicCamera`?

react-components/src/architecture/concrete/primitives/box/BoxView.test.ts
```diff
@@ -10,77 +10,115 @@ import { expectEqualVector3 } from '#test-utils/primitives/primitiveTestUtil';
...
 } from '#test-utils/architecture/viewUtil';
 import { isDomainObjectIntersection } from '../../../base/domainObjectsHelpers/DomainObjectIntersection';
 import { PrimitivePickInfo } from '../common/PrimitivePickInfo';
 import { MeasureBoxDomainObject } from '../../measurements/MeasureBoxDomainObject';
 import { PrimitiveType } from '../../../base/utilities/primitives/PrimitiveType';
 import { BoxFace } from '../common/BoxFace';
+import { Wireframe } from 'three/examples/jsm/lines/Wireframe.js';
+import { degToRad } from 'three/src/math/MathUtils.js';
 
 describe(BoxView.name, () => {
```
pramod-cog, line 24: (clarification)
This test file has too many magic numbers, hard to understand

react-components/src/architecture/concrete/axis/AxisThreeView.test.ts
nilscognite: Reveal doesn't support OrthographicCamera anyway

react-components/src/architecture/concrete/axis/AxisThreeView.test.ts
nilscognite: Made it slightly clearer

react-components/src/architecture/base/utilities/sprites/createSprite.ts
```diff
@@ -92,31 +89,30 @@ function createTexture(canvas: HTMLCanvasElement): Texture {
...
+  }
+  const textWidth = context === null ? text.length * 10 : context.measureText(text).width;
   const doubleBorderSize = borderSize * 2;
   const width = textWidth + 2 * doubleBorderSize;
   const height = fontSize + doubleBorderSize;
-
-  canvas.width = width;
+  canvas.width = textWidth + 2 * doubleBorderSize;
   canvas.height = height;
+  if (context === null) {
```
pramod-cog, line 110: Returning a blank canvas when `context` is `null` is safe for tests, but can hide real rendering problems in production. Any code using this canvas will not have expected data. This can cause confusing bugs if we expect a full rendering of sprite

pramod-cog (commented):
Have a comment, let me know what you think!

react-components/src/architecture/base/utilities/sprites/createSprite.ts
nilscognite: What do you mean my real rendering problems? Let us se what happen:

1) If it returns null/undefined: The sprite will not be there, so the text will not appear. The use think it is normal behavior, and don't complain. But the bug will be there for a while until we see it.
2) If it returns an empty sprite: We will notice the bug right away, and it will also be reported fast.

So both methods has benefits. I selected the method that gave less and easier coding, especially that the code will be the same for test and in production. Otherwise I have to hack it in some way.

pramod-cog (approved):
risk review ok

PR #5190, opened 2025-06-24
fix: remove usage of uuid package in favor of web api

christjt:
#### Type of change

Bug 

## Description :pencil:

Remove usage of UUID package. It exists as a webAPI, and we had not declared it as a dependency, so we unintentionally bundle the entire package.


danpriori (approved):
LGTM

christjt:
risk review ok

PR #5191, opened 2025-06-24
chore: preserve modules and outdir

christjt:
#### Type of change

Chore 

## Description :pencil:

Preserve modules in the lib output. This makes it easier to reason about the output of the lib and apps will bundle anyway, so likely no reason for us to do it. 

Also do not clear the out dir, as this will break the links (using the `file:` protocol).

Process for linking to fusion:
image

Search for all package.json files including the regex (remember to enable if in VSCode) `"@cognite/reveal-react-components":.*,` and replace with the absolute path to the react components lib. For me that is: `"@cognite/reveal-react-components": "file:///Users/ChristopherTannum/reveal/react-components",` this may or may not be different depending on OS etc. Then click replace all. 

When building react components, then output in node_modules in fusion should be automatically updated:

https://github.com/user-attachments/assets/98907ed5-284b-4a12-8907-f129de43152d

### DIFF
--- a/react-components/vite.config.ts
+++ b/react-components/vite.config.ts
@@ -1,85 +1,59 @@
 import { resolve } from 'path';
-import { type PluginOption, defineConfig } from 'vite';
+import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
 import dts from 'vite-plugin-dts';
 import { externalizeDeps } from 'vite-plugin-externalize-deps';
-import { exec } from 'node:child_process';
 import { coverageConfigDefaults } from 'vitest/config';
 
-export default defineConfig(({ command }) => {
-  return {
-    plugins: [
-      react(),
-      dts({ tsconfigPath: './tsconfig.build.json' }),
-      externalizeDeps({
-        devDeps: true
-      }),
-      yalcPush()
-    ],
-    build: {
-      lib: {
-        // Could also be a dictionary or array of multiple entry points
-        entry: resolve(__dirname, 'src/index.ts'),
-        name: 'Reveal react components',
-        // the proper extensions will be added
-        fileName: 'index',
-        formats: ['es']
-      },
-      sourcemap: command === 'build'
+export default defineConfig({
+  plugins: [
+    react(),
+    dts({ tsconfigPath: './tsconfig.build.json' }),
+    externalizeDeps({
+      devDeps: true
+    })
+  ],
+  build: {
+    emptyOutDir: false,
+    lib: {
+      entry: resolve(__dirname, 'src/index.ts'),
+      fileName: '[name]',
+      formats: ['es']
     },
-    test: {
-      include: ['src/**/*.{test,spec}.?(c|m)[jt]s?(x)'],
-      environment: 'happy-dom',
-      globals: true,
-      reporters: ['default'],
-      coverage: {
-        reportsDirectory: '../coverage/reveal-react-components',
-        exclude: [
-          ...coverageConfigDefaults.exclude,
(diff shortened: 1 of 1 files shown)


react-components/vite.config.ts
```diff
@@ -1,85 +1,59 @@
...
+export default defineConfig({
+  plugins: [
+    react(),
+    dts({ tsconfigPath: './tsconfig.build.json' }),
+    externalizeDeps({
+      devDeps: true
+    })
+  ],
+  build: {
+    emptyOutDir: false,
```
danpriori, line 17: AFAIK, this flag is to clean up the dist folder whenever we have a new build.
In general, it is fine for incremental changes. But if I'm not mistaken, with this flag false, we can have a situation where some files or chunks might turn out to be obsolete/outdated, and we can face some weird behavior, forcing us to reinstall everything from scratch or create some unnecessary garbage. Does make sense?

react-components/vite.config.ts
```diff
@@ -1,85 +1,59 @@
...
-    build: {
-      lib: {
-        // Could also be a dictionary or array of multiple entry points
-        entry: resolve(__dirname, 'src/index.ts'),
-        name: 'Reveal react components',
-        // the proper extensions will be added
-        fileName: 'index',
-        formats: ['es']
-      },
-      sourcemap: command === 'build'
```
danpriori, line 28: I was wondering if removing the sourcemap we can keep having the sourcemap for debugging. I dont see it anymore in the change. I tested here and I cant see the sourcemap 🤔

danpriori (changes requested):
Two questions. Probably the sourcemap is not working. 
The rest looks good.

react-components/vite.config.ts
christjt: Yeah, I totally understand. The reason for the change is that the file: protocol that we use with pnpm will create a hard link and seems to really dislike files getting deleted which seems to essentially break the updates. So for linking into fusion it must retain the build afaik. Any ideas?

react-components/vite.config.ts
christjt: I can try adding it back. I might have been too aggressive here as I typically don't leverage the source maps personally. So I did not want to have something no one uses. But I'll try to re-add it!

react-components/vite.config.ts
danpriori: Perhaps using the rollup-plugin-delete to remove the files right before a new build?

Something like this:

```typescript
import del from 'rollup-plugin-delete';
export default defineConfig({
  plugins: [
    del({ targets: 'dist' }),
    ...
  ],
  build: {
    emptyOutDir: false,
    ...
  }
});
```

danpriori: Hmm... I dont have any yet. However, as long as at least in the CI we can have a fresh build without any leftover, I'm good. And we can keep in mind to clean the `dist` folder manually if we see some inconsistencies anyway.

react-components/vite.config.ts
```diff
@@ -1,85 +1,59 @@
...
-            'lodash'
-          ]
-        }
       }
     }
-  };
-});
-
-function yalcPush(): PluginOption {
-  if (process.env.YALC !== 'true') {
```
haakonflatval-cognite, line 65: We have a script in the `package.json` that enables this option (`build:watch:yalc`). We should probably remove that one too

christjt: Done!

react-components/vite.config.ts
christjt: Done!

PR #5194, opened 2025-06-25
test(react-components): Add unit test for some measurement commands

nilscognite:
#### Type of change
Test 

## Description :pencil:

Add unit test for:

* `ShowMeasurementsOnTopCommand`
* `SetMeasurementTypeCommand`
* `MeasurementFolder`
* `installToolbars`

I needed this for another PR.

### DIFF
--- a/react-components/src/architecture/installToolbars.test.ts
+++ b/react-components/src/architecture/installToolbars.test.ts
@@ -16,7 +16,7 @@ describe(installToolbars.name, () => {
     expect(getCommandsInToolbarCount(new AnnotationsSelectTool())).toBe(6);
     expect(getCommandsInToolbarCount(new ClipTool())).toBe(11);
     expect(getCommandsInToolbarCount(new ExampleTool())).toBe(5);
-    expect(getCommandsInToolbarCount(new MeasurementTool())).toBe(10);
+    expect(getCommandsInToolbarCount(new MeasurementTool())).toBe(11);
   });
 
   test('Should not install toolbar for navigation ', () => {
--- a/react-components/src/architecture/concrete/measurements/commands/SetMeasurementTypeCommand.ts
+++ b/react-components/src/architecture/concrete/measurements/commands/SetMeasurementTypeCommand.ts
@@ -14,7 +14,8 @@ export const MEASURE_PRIMITIVE_TYPES = [
   PrimitiveType.HorizontalArea,
   PrimitiveType.Box,
   PrimitiveType.HorizontalCircle,
-  PrimitiveType.VerticalCylinder
+  PrimitiveType.VerticalCylinder,
+  PrimitiveType.HorizontalCylinder
 ];
 
 export class SetMeasurementTypeCommand extends RenderTargetCommand {


react-components/src/architecture/concrete/measurements/commands/SetMeasurementTypeCommand.test.ts
```diff
@@ -0,0 +1,76 @@
+import { beforeEach, describe, expect, test } from 'vitest';
+import { createFullRenderTargetMock } from '#test-utils/fixtures/createFullRenderTargetMock';
+import { getDefaultCommand } from '../../../../components/Architecture/utilities';
+import { isEmpty } from '../../../base/utilities/TranslateInput';
+import { PrimitiveType } from '../../../base/utilities/primitives/PrimitiveType';
+import { type RevealRenderTarget } from '../../../base/renderTarget/RevealRenderTarget';
+import { SetMeasurementTypeCommand } from './SetMeasurementTypeCommand';
+import { MeasurementTool } from '../MeasurementTool';
+
+describe(SetMeasurementTypeCommand.name, () => {
```
pramod-cog, line 10: (suggestion)
Can we add a test case where `command.invoke()` is called twice to check for false positive case

react-components/src/architecture/concrete/measurements/commands/SetMeasurementTypeCommand.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+function createActiveTool(renderTarget: RevealRenderTarget): MeasurementTool {
+  const tool = getDefaultCommand(new MeasurementTool(), renderTarget);
+  renderTarget.commandsController.setActiveTool(tool);
+  return tool;
+}
+
+const MeasurePrimitiveTypes = [
```
pramod-cog: (clarification)
Can an invalid primitive type be created?

react-components/src/architecture/concrete/measurements/commands/ShowMeasurementsOnTopCommand.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    expect(command.icon).toBe('EyeShow');
+    expect(command.buttonType).toBe('ghost');
+    expect(command.isEnabled).toBe(false);
+    expect(command.isToggle).toBe(true);
+    expect(command.getShortCutKeys()).toBeUndefined();
+  });
+
+  test('Should set all measurements on top', () => {
+    const command = new ShowMeasurementsOnTopCommand();
+    command.attach(renderTarget);
```
pramod-cog, line 45: (must be fixed)
Can we add a test where `renderTarget` is not attached to `command` and check if some measurement object can be added!

react-components/src/architecture/concrete/measurements/commands/ShowMeasurementsOnTopCommand.test.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+function* getAllRenderStylesForAllMeasurements(root: DomainObject): Generator<CommonRenderStyle> {
+  for (const descendant of root.getDescendants()) {
+    if (
+      isInstanceOf(descendant, MeasureBoxDomainObject) ||
+      isInstanceOf(descendant, MeasureLineDomainObject) ||
+      isInstanceOf(descendant, MeasureCylinderDomainObject)
+    ) {
+      const style = descendant.getRenderStyle() as CommonRenderStyle;
```
pramod-cog: (must be fixed)
using as `CommonRenderStyle` will not ensure `getRenderStyle()` return type of `CommonRenderStyle`, currently it only ensure it does not return `undefined`. It should be something like
```
const style = descendant.getRenderStyle();
assert(style && isCommonRenderStyle(style), 'Render style should be CommonRenderStyle');
```
where `isCommonRenderStyle` is a type guard function

react-components/src/architecture/installToolbars.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import { NavigationTool } from './base/concreteCommands/NavigationTool';
+
+describe(installToolbars.name, () => {
+  test('Should have toolbars with correct number of commands for the following tools', () => {
+    installToolbars();
+    expect(getCommandsInToolbarCount(new AnnotationsCreateTool())).toBe(5);
+    expect(getCommandsInToolbarCount(new AnnotationsSelectTool())).toBe(6);
+    expect(getCommandsInToolbarCount(new ClipTool())).toBe(11);
+    expect(getCommandsInToolbarCount(new ExampleTool())).toBe(5);
+    expect(getCommandsInToolbarCount(new MeasurementTool())).toBe(10);
```
pramod-cog: I am sure these numbers are correct count but most of the us are not aware of number of commands in each Tools, is there any way we can use existing variables/enums count rather than just integers?

pramod-cog (changes requested):
Overall looks good, have add some comments.
Let me your thoughts!

react-components/src/architecture/concrete/measurements/commands/SetMeasurementTypeCommand.test.ts
nilscognite: Fixed. Discovered a bug here

react-components/src/architecture/concrete/measurements/commands/SetMeasurementTypeCommand.test.ts
nilscognite: Done. Also done with the SetClipTypeCommand since it is similar

react-components/src/architecture/concrete/measurements/commands/ShowMeasurementsOnTopCommand.test.ts
nilscognite: I never make type guard function for classes. The language build-in `instanceof `is used for this.

nilscognite: Fixed

react-components/src/architecture/concrete/measurements/commands/ShowMeasurementsOnTopCommand.test.ts
nilscognite: `ShowMeasurementsOnTopCommand `changes the style of the visualization. It do not add anything new.

nilscognite: But I added that test

react-components/src/architecture/installToolbars.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
```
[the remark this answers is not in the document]
nilscognite: I think that is is meaningless and need just another implementation of the same thing in a sense. I just wanted that the toolbar wasn't changed. Thats all.

pramod-cog (approved):

PR #5197, opened 2025-06-25
fix: unmounting CadModelContainer does not properly remove cad model

christjt:
#### Type of change

Bug 

## Description :pencil:

Fixes a very ugly bug in the CadModelContainer due to closure not properly capturing the correct instance of the model in the remove model closure. I discovered this while creating a mount / unmount button in the storybook to provoke the race conditions that @pramod-cog 's workaround circumvents:
image

The fix is simply lines 63-65 in the diff, where we pass the actual instance. The rest is simply making this bug testable. We likely have the same issue in PointClouds and maybe 360s as well which needs to be investigated.

### DIFF
--- a/react-components/src/components/CadModelContainer/CadModelContainer.spec.tsx
+++ b/react-components/src/components/CadModelContainer/CadModelContainer.spec.tsx
@@ -1,0 +1,72 @@
+import { render, waitFor } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { CadModelContainer } from './CadModelContainer';
+import { CadModelContext, type CadModelContextDependencies } from './CadModelContainer.context';
+import type { PropsWithChildren, ReactElement } from 'react';
+import { Mock } from 'moq.ts';
+import type { RevealRenderTarget } from '../../architecture';
+import { type CogniteCadModel, type Cognite3DViewer, type CogniteModel } from '@cognite/reveal';
+
+describe('CadModelContainer', () => {
+  const deps: CadModelContextDependencies = {
+    useRevealKeepAlive: vi.fn(),
+    useRenderTarget: vi.fn(),
+    useReveal3DResourcesCount: vi.fn(),
+    useReveal3DResourceLoadFailCount: vi.fn(),
+    useApplyCadModelStyling: vi.fn(),
+    createCadDomainObject: vi.fn(),
+    removeCadDomainObject: vi.fn()
+  };
+
+  const wrapper = ({ children }: PropsWithChildren): ReactElement => (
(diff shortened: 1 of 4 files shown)


react-components/src/components/CadModelContainer/CadModelContainer.spec.tsx
```diff
@@ -0,0 +1,72 @@
+import { render, waitFor } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { CadModelContainer } from './CadModelContainer';
+import { CadModelContext, type CadModelContextDependencies } from './CadModelContainer.context';
+import type { PropsWithChildren, ReactElement } from 'react';
+import { Mock } from 'moq.ts';
+import type { RevealRenderTarget } from '../../architecture';
+import { type CogniteCadModel, type Cognite3DViewer, type CogniteModel } from '@cognite/reveal';
+
+describe('CadModelContainer', () => {
```
anders-hopland:
```suggestion
describe(CadModelContainer.name, () => {
```

react-components/src/components/CadModelContainer/CadModelContainer.spec.tsx
haakonflatval-cognite, line 1: I believe we prefer to use `.test` instead of `.spec` in Reveal

react-components/src/components/CadModelContainer/CadModelContainer.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const deps: CadModelContextDependencies = {
+    useRevealKeepAlive: vi.fn(),
+    useRenderTarget: vi.fn(),
```
haakonflatval-cognite: Nothing you need to do here, but when I declare dependencies like this, I usually declare the mocks as separate variables first (e.g.
```
const mockUseRevealKeepAlive = vi.fn<typeof useRevealKeepAlive>();
...
```
And then put them all in a dependencies-object afterwards. This has the benefit that you don't need to do `vi.mocked` (which is essentially a type cast) when mocking return values, but you must provide explicit types to the `vi.fn`s. We may want to agree on one of these approaches for consistency, if nothing else

haakonflatval-cognite (commented):
Looks good, a couple of comments

react-components/src/components/CadModelContainer/CadModelContainer.spec.tsx
christjt: (which is essentially a type cast), yes but it isn't type unsafe though. adding separate vars for each mock is sooo boilerplaty imo.

PR #5199, opened 2025-06-26
refactor(react-components): useModelIdRevisionIdFromModelOptions hook to return useQuery

pramod-cog:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5776

## Description :pencil:
Refactored `useModelIdRevisionIdFromModelOptions` hook to return `useQuery` instead of `useQueries`

In `Storybook`

- Run `SceneContainer` storybook example


react-components/src/components/Reveal3DResources/hooks/useClassicModelOptions.ts
haakonflatval-cognite, line 1: Nice!

react-components/src/hooks/useQueriedAddModelOptions.ts
```diff
@@ -16,49 +16,35 @@ import {
...
+        addModelOptionsArray.map(async (addModelOptions) => {
+          if (isClassicIdentifier(addModelOptions)) {
+            return addModelOptions;
+          }
+          if (isDM3DModelIdentifier(addModelOptions)) {
+            const { modelId, revisionId } = await getModelIdAndRevisionIdFromExternalId(
+              addModelOptions.revisionExternalId,
+              addModelOptions.revisionSpace,
+              fdmSdk
+            );
```
haakonflatval-cognite: I'll just make a mental note that this is probably a good candidate for reducing to a single query, instead of one query per model

react-components/src/hooks/useQueriedAddModelOptions.ts
```diff
@@ -16,49 +16,35 @@ import {
...
+  return useQuery({
+    queryKey: [queryKeys.modelRevisionId(keys)],
+    queryFn: async () => {
+      if (addModelOptionsArray === undefined || addModelOptionsArray.length === 0) {
+        return [];
+      }
+      return await Promise.all(
+        addModelOptionsArray.map(async (addModelOptions) => {
+          if (isClassicIdentifier(addModelOptions)) {
+            return addModelOptions;
```
haakonflatval-cognite: I think the lambda inside the `map` here would be nice to have as a separate function for clarity

react-components/src/hooks/useModelIdRevisionIdFromModelOptions.test.tsx
```diff
@@ -64,53 +58,73 @@ describe(useModelIdRevisionIdFromModelOptions.name, () => {
...
+      .setup((p) => p.data)
+      .returns([classicModelOption])
+      .setup((p) => p.isFetching)
+      .returns(false)
+      .setup((p) => p.isLoading)
+      .returns(false)
+      .setup((p) => p.isError)
+      .returns(false)
+      .setup((p) => p.isRefetching)
+      .returns(false)
```
haakonflatval-cognite: It seems to me that these fields are not necessary in the new implementation? (A few more occurences of this further down in this file as well)

I also think I'd prefer that the context of the hook could have a narrower type (something like `() => { data: Array<...> }`), so that we won't need `new Mock`, and can avoid all unnecessary fields

react-components/src/hooks/useModelIdRevisionIdFromModelOptions.test.tsx
```diff
@@ -64,53 +58,73 @@ describe(useModelIdRevisionIdFromModelOptions.name, () => {
...
 
   test('returns empty array if any query is still fetching', () => {
-    const mockDMModelResult = new Mock<UseQueryResult<AddModelOptions<ClassicDataSourceType>>>()
+    const resultMock = new Mock<UseQueryResult<Array<AddModelOptions<ClassicDataSourceType>>>>()
       .setup((p) => p.data)
-      .returns({
-        modelId: 987,
-        revisionId: 654
-      })
+      .returns([])
```
haakonflatval-cognite: I don't really understand why we return an empty array here - don't we want to still use the classic model ID? Same for a few more cases in this file.

haakonflatval-cognite (commented):
Looks very good! I had some questions about a test file

pramod-cog:
risk review ok

PR #5201, opened 2025-06-26
chore: add getMocksByDefaultDependencies test utility

christjt:
#### Type of change

Chore 

## Description :pencil:

Adds a test utility for auto applying vi.fn() typed correctly given the default dependencies of a context.


react-components/src/components/Reveal3DResources/Reveal3DResources.test.tsx
```diff
@@ -1,139 +1,103 @@
...
+      styledModels: [],
+      isModelMappingsLoading: false
+    }));
+    defaultDependencies.useReveal3DResourcesStylingLoadingSetter.mockImplementation(() => () => {});
+    defaultDependencies.useCalculatePointCloudStyling.mockImplementation(() => []);
+  });
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
```
haakonflatval-cognite: I think I'd personally want to use `vi.resetAllMocks()` in `beforeEach` and then put all the `mockImplementotions` etc. after that, to ensure no overridden mocks leak between tests.

react-components/src/components/RevealToolbar/LayersButton/LayersButton.test.tsx
```diff
@@ -14,52 +14,53 @@ import {
...
+        vi.fn()
+      ]
+    );
+    defaultDependencies.useReveal.mockImplementation(() => viewerMock);
+    defaultDependencies.use3dModels.mockImplementation(() => [cadMock, cadMock]);
+    defaultDependencies.ModelLayerSelection.mockImplementation(({ label }) => <div>{label}</div>);
+  });
+
+  beforeEach(() => {
+    vi.clearAllMocks();
```
haakonflatval-cognite: Same here

react-components/src/query/usePointCloudModelRevisionIdsFromReveal.test.tsx
```diff
@@ -29,35 +30,28 @@ describe(usePointCloudModelRevisionIdsFromReveal.name, async () => {
...
-        }}>
+      <UsePointCloudModelRevisionIdsFromRevealContext.Provider value={defaultDependencies}>
         {children}
       </UsePointCloudModelRevisionIdsFromRevealContext.Provider>
     </QueryClientProvider>
   );
 
   beforeEach(() => {
-    vi.resetAllMocks();
+    vi.clearAllMocks();
```
haakonflatval-cognite: I think we'll want `resetAllMocks` here too

react-components/src/query/usePointCloudModelRevisionIdsFromReveal.test.tsx
```diff
@@ -69,24 +63,23 @@ describe(usePointCloudModelRevisionIdsFromReveal.name, async () => {
...
       new Mock<FdmSDK>()
-        .setup((p) => p.queryNodesAndEdges<QueryRequest>)
-        .returns(mockFdmSdkQueryNodesAndEdgesMock)
+        .setup(async (p) => await p.queryNodesAndEdges<QueryRequest>(It.IsAny()))
+        .returnsAsync(
+          createDmsPointcloudRevisionResponse(
+            TEST_MODEL_EXTERNAL_ID0,
+            TEST_REVISION_EXTERNAL_ID0,
+            TEST_SPACE
+          )
```
haakonflatval-cognite, line 81: Just curious - is there a reason you prefer writing out the call to `queryNodesAndEdges` here, instead of leaving it as it was?

haakonflatval-cognite (changes requested):
Looks good, but had some more comments

react-components/src/query/usePointCloudModelRevisionIdsFromReveal.test.tsx
christjt: I just found it more confusing to add one mock lib into another in this "simple" case. I found it easier to reason about after the change. Instead of a mock containing a mock where its implementation is defined somewhere else.

react-components/src/components/Reveal3DResources/Reveal3DResources.test.tsx
christjt: Yup, I misunderstood the API here. I'll change it.

react-components/src/components/RevealToolbar/LayersButton/LayersButton.test.tsx
```diff
@@ -37,25 +36,20 @@ describe(LayersButton.name, () => {
...
-        },
-        vi.fn()
-      ]
-    );
-    defaultDependencies.useReveal.mockImplementation(() => viewerMock);
-    defaultDependencies.use3dModels.mockImplementation(() => [cadMock, cadMock]);
+    vi.resetAllMocks();
+    defaultDependencies.useModelHandlers.mockReturnValue([
+      {
+        cadHandlers: [createCadHandlerMock()],
```
haakonflatval-cognite: I think this should be a `beforeEach`

haakonflatval-cognite (changes requested):
Almost there :pray:

haakonflatval-cognite (approved):
LGTM :+1:

PR #5202, opened 2025-06-27
test(react-components): Make a new test for RootMainObject

nilscognite:
#### Type of change
Test 

## Description :pencil:

This tests that anything happens with some of the descendants of the root, the root will be notified, and the user can add a event listener to this.


m-amaya:
Nice!

m-amaya:
🦄

PR #5204, opened 2025-07-01
chore: automatically reset mocks between tests

christjt:
#### Type of change

Chore 

## Description :pencil:

I noticed that some tests did not properly reset mocks between each test, and now some tests are dependent on each other and the order in which they run. This PR resolves this by automatically resetting mocks between each run. Also fixes the test(s) which had these test ordering issues.


christjt:
risk review ok

PR #5206, opened 2025-07-01
fix(react-components): call domain object notify after delete

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

Part of the work for https://cognitedata.atlassian.net/browse/BND3D-5790

## Description :pencil:

Small change in the architecture: 

Changes up the notify-system slightly so that we send a notification _after_ a node has been deleted, rather than just before it is deleted.

Through unit tests


react-components/src/architecture/base/domainObjectsHelpers/Views.ts
```diff
@@ -74,6 +74,12 @@ export class Views {
     clear(this._listeners);
   }
 
+  public notifyListeners(domainObject: DomainObject, change: DomainObjectChange): void {
```
nilscognite, line 73: Reuse this x 2 in notify

react-components/src/architecture/base/domainObjectsHelpers/Changes.ts
```diff
@@ -27,6 +27,7 @@ export class Changes {
 
   // Parent-child relationship changed
   public static readonly added: symbol = Symbol('added'); // When added to the system
+  public static readonly deleting: symbol = Symbol('deleting'); // When deleted from the system
```
pramod-cog:
```suggestion
  public static readonly deleting: symbol = Symbol('deleting'); // When deleting from the system
```

pramod-cog (commented):
Looks good, one minor comment

PR #5207, opened 2025-07-01
feat(react-components): create useRevealDomainObjects function

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

Part of the work for https://cognitedata.atlassian.net/browse/BND3D-5790

## Description :pencil:

Introduces the hook `useRevealDomainObjects` which is implemented with a signal, and updates whenever (relevant) changes to the DomainObject hierarchy occurs. I intend this to be a generally useful function with a few use cases in existing hooks coming up in a subsequent PR.

This also introduces a pattern involving `DisposableSignal`, a signal that contains a callback and that comes with an attached dispose-function, which is relevant if e.g. the signal is based on an event listener to properly unregister the listener when the signal is no longer needed. The `useDisposableSignal` hook streamlines this part.

The reason for requiring `DisposableSignal` to return a callback in the signal, is that it allows us to avoid double-buffering values that should only exist one place, e.g. Reveal domain objects in the domain object hierarchy. Having this engraved in the type also lets us more easily implement `useDisposableSignal` without having the user put a `useMemo` at the end manually to actually trigger the callback themselves.

Through unit tests and, with upcoming changes, in Fusion

### DIFF
--- a/react-components/src/architecture/concrete/reveal/signal/getRevealDomainObjectsSignal.test.ts
+++ b/react-components/src/architecture/concrete/reveal/signal/getRevealDomainObjectsSignal.test.ts
@@ -1,7 +1,7 @@
 import { beforeEach, describe, expect, test, vi } from 'vitest';
 import { getRevealDomainUpdateSignal } from './getRevealDomainObjectsSignal';
 import { createRenderTargetMock } from '#test-utils/fixtures/renderTarget';
-import { RevealRenderTarget } from '../../../base/renderTarget/RevealRenderTarget';
+import { type RevealRenderTarget } from '../../../base/renderTarget/RevealRenderTarget';
 import { CadDomainObject } from '../cad/CadDomainObject';
 import { createCadMock } from '#test-utils/fixtures/cadModel';
 import { createImage360ClassicMock } from '#test-utils/fixtures/image360';
--- a/react-components/src/architecture/concrete/reveal/signal/getRevealDomainObjectsSignal.test.ts
+++ b/react-components/src/architecture/concrete/reveal/signal/getRevealDomainObjectsSignal.test.ts
@@ -1,0 +1,132 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { getRevealDomainUpdateSignal } from './getRevealDomainObjectsSignal';
+import { createRenderTargetMock } from '#test-utils/fixtures/renderTarget';
+import { RevealRenderTarget } from '../../../base/renderTarget/RevealRenderTarget';
+import { CadDomainObject } from '../cad/CadDomainObject';
+import { createCadMock } from '#test-utils/fixtures/cadModel';
+import { createImage360ClassicMock } from '#test-utils/fixtures/image360';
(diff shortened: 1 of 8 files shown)


react-components/src/architecture/concrete/reveal/signal/getRevealDomainObjectsSignal.ts
```diff
@@ -0,0 +1,54 @@
...
+  function updateDomainObjectListOnRelevantEvent(
+    domainObject: DomainObject,
+    change: DomainObjectChange
+  ): void {
+    const relevantChanges = [Changes.added, Changes.deleting, ...additionalChangeFlags];
+
+    const isRelevantChange = change.isChanged(...relevantChanges);
+
+    const isRelevantEvent = domainObject instanceof RevealDomainObject && isRelevantChange;
+    if (!isRelevantEvent) {
```
nilscognite: Use early return here!!!!! Test first `domainObject instanceof RevealDomainObject`.

react-components/src/architecture/concrete/reveal/signal/getRevealDomainObjectsSignal.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    revealRenderTarget.rootDomainObject.addChild(pointCloudDomainObject);
+    revealRenderTarget.rootDomainObject.addChild(image360DomainObject);
+    revealRenderTarget.rootDomainObject.addChild(dummyDomainObject);
+
+    const signalResult = getRevealDomainUpdateSignal(revealRenderTarget);
+
+    const value = signalResult.signal()();
+    expect(value[0]).toBe(cadDomainObject);
+    expect(value[1]).toBe(pointCloudDomainObject);
+    expect(value[2]).toBe(image360DomainObject);
```
nilscognite, line 39: Test length also:`  expect(initialResult).toHaveLength(3);`

react-components/src/architecture/concrete/reveal/signal/getRevealDomainObjectsSignal.test.ts
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  test('adds event listener, and calling dispose removes event listener', () => {
+    const addEventMock = vi.spyOn(revealRenderTarget.rootDomainObject.views, 'addEventListener');
+    const removeEventMock = vi.spyOn(
+      revealRenderTarget.rootDomainObject.views,
+      'removeEventListener'
+    );
+
+    const res = getRevealDomainUpdateSignal(revealRenderTarget);
```
nilscognite: res? Did you mean result?

nilscognite (changes requested):
Look good! See comments

react-components/src/architecture/concrete/reveal/signal/getRevealDomainObjectsSignal.ts
haakonflatval-cognite: I'm not sure I see how this improves the code :thinking: It will add a few more lines, but doesn't seem to make the code clearer. Am I misunderstanding something?

PR #5209, opened 2025-07-01
chore: add react hooks rules eslint plugin

christjt:
#### Type of change

Chore 

## Description :pencil:

Enable lint plugin for react hooks rules

### DIFF
--- a/.github/workflows/react-components-ci.yml
+++ b/.github/workflows/react-components-ci.yml
@@ -58,7 +58,7 @@ jobs:
       - name: Lint
         working-directory: react-components
         continue-on-error: false
-        run: yarn run lint --max-warnings 0
+        run: yarn run lint
         env:
           NPM_TOKEN: ${{ secrets.NPM_READONLY_TOKEN }}
 


react-components/src/components/Architecture/DropdownButton.tsx
```diff
@@ -87,7 +87,7 @@ const DropdownElement = ({
           </Button>
         </CogsTooltip>
       )}>
-      {command.children.map((child) => createMenuItem(child, setOpen))}
+      {command.children.map((child) => CreateMenuItem(child, setOpen))}
```
haakonflatval-cognite, line 90: This doesn't seem righ to me - `CreateMenuItem` contains a hook call, and when we call the component like a function like this, the hook will be counted as part of the parent components' hook. That means that if the number or order of commands change for some reason, the number of hook calls in this component will change, and React may explode

haakonflatval-cognite: I see that this may be the case in other components in this PR as well

PR #5212, opened 2025-07-02
fix: ignore dereferencing when group is already disposed

christjt:
#### Type of change

Bug 

#### Jira ticket :blue_book:

related to https://cognitedata.atlassian.net/browse/BND3D-5792

## Description :pencil:

Fixes an issue when CAD model is being disposed where a dereference of a geometry group is happening when it is already disposed.


anders-hopland:
🦄

PR #5213, opened 2025-07-02
chore: update rustc compiler version

christjt:
#### Type of change

Chore 

#### Jira ticket :blue_book:

related to https://cognitedata.atlassian.net/browse/BND3D-5792

## Description :pencil:

publishing failed due to `package `wasm-bindgen-wasm-conventions v0.2.95` cannot be built because it requires rustc 1.76 or newer, while the currently active rustc version is 1.73.0`

So updating the current rustc version.


anders-hopland (approved):
🦄

PR #5217, opened 2025-07-03
feat(react-components): create useRevealDomainObjects function, second take

haakonflatval-cognite:
#### Type of change

Chore 

#### Jira ticket :blue_book:

## Description :pencil:

Revitalization of https://github.com/cognitedata/reveal/pull/5207 after I merged it into the previous PR by mistake.

### Here is the original description:

Introduces the hook useRevealDomainObjects which is implemented with a signal, and updates whenever (relevant) changes to the DomainObject hierarchy occurs. I intend this to be a generally useful function with a few use cases in existing hooks coming up in a subsequent PR.

This also introduces a pattern involving DisposableSignal, a signal that contains a callback and that comes with an attached dispose-function, which is relevant if e.g. the signal is based on an event listener to properly unregister the listener when the signal is no longer needed. The useDisposableSignal hook streamlines this part.

The reason for requiring DisposableSignal to return a callback in the signal, is that it allows us to avoid double-buffering values that should only exist one place, e.g. Reveal domain objects in the domain object hierarchy. Having this engraved in the type also lets us more easily implement useDisposableSignal without having the user put a useMemo at the end manually to actually trigger the callback themselves. 

Through unit tests and, with upcoming changes, in Fusion


nilscognite (approved):
LGTM

react-components/src/utilities/signal/useDisposableSignal.ts
```diff
@@ -0,0 +1,12 @@
...
+import { type DisposableSignal } from './DisposableSignal';
+import { useEffect, useMemo } from 'react';
+
+export function useDisposableSignal<T>(disposableSignal: DisposableSignal<T>): T {
+  useEffect(() => {
+    return disposableSignal.dispose;
+  }, [disposableSignal]);
+
+  const valueCallback = useSignalValue(disposableSignal.signal);
+  return useMemo(() => valueCallback(), [valueCallback]);
```
m-amaya, line 11: `useSignalValue` is already memoized. We don't need another `useMemo` here. We can get rid of it.

m-amaya (commented):
Looks good for the most part. Requesting one change. Feel free to ping me directly once addressed.

react-components/src/utilities/signal/useDisposableSignal.ts
haakonflatval-cognite: I think it is still needed - `valueCallback` is not the value itself, but a callback to retrieve the value, which again is called inside the `useMemo`

m-amaya: Right, but in this hook, since we're only dealing with state derived from `disposableSignal`, `valueCallback()` only gets triggered whenever `disposableSignal` is changed from the `useSignalValue`. Since `useSignalValue` is already memoized, then `valueCallback()` is memoized, too. Put another way:

`disposableSignal` (input) -> `useSignalValue` (memoized) -> `valueCallback` (memoized) -> `valueCallback()` (only triggered when `valueCallback` changes, which is already memoized).

Therefore, you don't need the `useMemo`

m-amaya: I see now. `useMemo` is used in this case to optimize for re-rendering. 👍🏼

PR #5231, opened 2025-07-17
chore(react-components): implement hybrid functionality in classic asset cache

haakonflatval-cognite:
#### Type of change

Chore 

#### Jira ticket :blue_book:

Closes https://cognitedata.atlassian.net/browse/BND3D-4989

## Description :pencil:

Changes the classic CAD cache to also handle hybrid asset mappings; asset mappings that may contain DMS IDs in addition to / instead of classic asset IDs.

All changes:
- Introduced `rawAssetMappingTypes.ts` which defines the bridging between asset mappings from the CDF API and our internal types. It provides a more accurate CDF asset mapping type (where `assetId` is optional) and translation functions
- Refactored `assetMappingTypes` with reusable typeguards operating on wider types, used by the other type guards in the file. Also added a utility for getting the instance from an asset mapping.
- Renamed and added another model-key utility in `idAndKeyTranslation`
- Renamed `getNodesForAssetIds` on the `CadInstanceMappingsCache` to `getNodesForInstanceIds`
- Replaced `FdmKey | AssetId` with `InstanceKey` and `AssetId | DmsUniqueIdentifier` with `InstanceId` a few places (they are equivalent)
-  Added a type `HybridCadCacheIndexType = 'nodeIds' | 'assetIds'`, which is used some places for internal caches
- Changed types inside the `ClassicCadAssetMappingCacheImpl` class and related classes to accomodate hybrid instances. This involves replacing `AssetId` by `InstanceKey` a few places, with relevant transformations. 
- Actually ensure that we fetch hybrid asset mappings, by passing `getDmsInstances: true` to the asset mappings list endpoint
- Filtered away non-classic mappings a couple of places where we use the cache directly (and kept notes on these places)

Note that we don't have full feature parity with the classic asset mappings, since the backend does not (yet) support lookups of asset mappings with model/revision ID + DM instance ID, as it does with model/revision ID + classic asset ID. The upcoming front-end work to accomodate this is tracked here: https://cognitedata.atlassian.net/browse/BND3D-5837

I have also added a few more points to the issue https://cognitedata.atlassian.net/browse/BND3D-5836, noting places where we intentionally filter out hybrid mappings in order to appease the type checker. These will be fixed when the corresponding end-features are implemented.

There are no expected functional change from this PR. The functional changes will be introduced when each feature depending on this cache is addressed individually.

### DIFF
--- a/react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.test.ts
+++ b/react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.test.ts
@@ -8,9 +8,9 @@ import {
 } from '#test-utils/fixtures/sdk';
 import { createCadNodeMock } from '#test-utils/fixtures/cadNode';
 import { createCursorAndAsyncIteratorMock } from '#test-utils/fixtures/cursorAndIterator';
-import { InstanceId, isClassicInstanceId } from '../../../utilities/instanceIds';
-import { RawCdfHybridCadAssetMapping } from './rawAssetMappingTypes';
-import { AssetMapping3D, Node3D } from '@cognite/sdk';
+import { type InstanceId, isClassicInstanceId } from '../../../utilities/instanceIds';
+import { type RawCdfHybridCadAssetMapping } from './rawAssetMappingTypes';
+import { type AssetMapping3D, type Node3D } from '@cognite/sdk';
 
 describe(createClassicCadAssetMappingCache.name, () => {
   const MODEL_ID = 123;
--- a/react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.test.ts
+++ b/react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.test.ts
@@ -147,7 +147,7 @@ describe(createClassicCadAssetMappingCache.name, () => {
         createRawAssetMappingFromNodeAndInstance(CAD_NODES[0], ASSET_ID),
         createRawAssetMappingFromNodeAndInstance(CAD_NODES[2], 43)
       ];
-      assetMappings3DListMock.mockReturnValue(
+      assetMappings3DListMock.mockReturnValueOnce(
         createCursorAndAsyncIteratorMock({
           items: mappings as AssetMapping3D[]
         })
@@ -168,10 +168,14 @@ describe(createClassicCadAssetMappingCache.name, () => {
       const cache = createClassicCadAssetMappingCache(sdkMock);
 
       const result = await cache.getAssetMappingsForModel(MODEL_ID, REVISION_ID);
+
+      // The list endpoint is called twice, once for classic assets and once for DM
+      expect(assetMappings3DListMock).toHaveBeenCalledTimes(2);
+
       const result1 = await cache.getAssetMappingsForModel(MODEL_ID, REVISION_ID);
 
       expect(result).toEqual(result1);
-      expect(assetMappings3DListMock).toHaveBeenCalledTimes(1);
+      expect(assetMappings3DListMock).toHaveBeenCalledTimes(2);
     });
   });
 });
--- a/react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.test.ts
+++ b/react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.test.ts
@@ -1,0 +1,194 @@
+import { describe, expect, test } from 'vitest';
+import { createClassicCadAssetMappingCache } from './ClassicCadAssetMappingCacheImpl';
+import {
+  assetMappings3DFilterMock,
+  assetMappings3DListMock,
+  nodes3dRetrieveMock,
+  sdkMock
+} from '#test-utils/fixtures/sdk';
+import { createCadNodeMock } from '#test-utils/fixtures/cadNode';
+import { createCursorAndAsyncIteratorMock } from '#test-utils/fixtures/cursorAndIterator';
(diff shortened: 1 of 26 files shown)


react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.ts
```diff
@@ -297,9 +314,9 @@ class ClassicCadAssetMappingCacheImpl implements ClassicCadAssetMappingCache {
   private async fetchAndCacheMappingsForIds(
     modelId: ModelId,
     revisionId: RevisionId,
-    ids: number[],
-    filterType: string
-  ): Promise<ClassicCadAssetMapping[]> {
+    ids: InstanceKey[],
+    filterType: 'nodeIds' | 'assetIds'
```
anders-hopland:
```suggestion
    filterType: HybridCadCacheIndexType
```

react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.test.ts
```diff
@@ -0,0 +1,198 @@
...
+
+      expect(result).toEqual({ mappings: [] });
+    });
+
+    test('returns relevant mappings if exist', async () => {
+      const mapping = createRawAssetMappingFromNodeAndInstance(CAD_NODES[0], ASSET_ID);
+
+      assetMappings3DFilterMock.mockReturnValue(
+        createCursorAndAsyncIteratorMock({
+          // SDK method uses `AssetMapping3D', but its `assetId` may not always
```
anders-hopland: Is this because of the new hybrid asset mappings? If so, it would be nice to state that there are a disparency between the types

react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingPerModelCache.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  test('results for different models are cached separately', async () => {
+    const cache = new ClassicCadAssetMappingPerModelCache(sdkMock);
+
+    const modelId1 = 987;
+    const revisionId1 = 876;
+
+    const result = await cache.fetchAndCacheMappingsForModel(MODEL_ID, REVISION_ID);
+
+    assetMappings3DListMock.mockReturnValue(
+      createCursorAndAsyncIteratorMock({ items: [ASSET_MAPPINGS[1]] as AssetMapping3D[] })
```
anders-hopland: Please add comment for why we are doing this cast

react-components/src/components/CacheProvider/cad/rawAssetMappingTypes.ts
```diff
@@ -0,0 +1,82 @@
+import type { DmsUniqueIdentifier } from '../../../data-providers';
+import { isDefined } from '../../../utilities/isDefined';
+import type {
+  ClassicCadAssetMapping,
+  DmCadAssetMapping,
+  HybridCadAssetMapping
+} from './assetMappingTypes';
+
+export type RawCdfHybridCadAssetMappingBase = {
```
anders-hopland: Can an asset mapping only contain a nodeId, and have assetId and assetInstanceId set to undefined?

react-components/src/components/RuleBasedOutputs/hooks/useExtractUniqueAssetIdsFromMapped.tsx
```diff
@@ -3,13 +3,15 @@ import { type ModelWithAssetMappings } from '../../../hooks/cad/modelWithAssetMa
...
 import { isDefined } from '../../../utilities/isDefined';
+import { isClassicCadAssetMapping } from '../../CacheProvider/cad/assetMappingTypes';
 
 export const useExtractUniqueAssetIdsFromMapped = (
   assetMappings: ModelWithAssetMappings[] | undefined
 ): InternalId[] => {
   return useMemo(() => {
     const mappings = assetMappings?.map((item) => item.assetMappings).flat() ?? [];
     const assetIds: InternalId[] = mappings
+      .filter(isClassicCadAssetMapping)
```
anders-hopland, line 14: Should we consider to rename this function / file to `useExtractUniqueClassicAssetIdsFromMapped`

anders-hopland (commented):
Looks good, only minor comments

react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.ts
haakonflatval-cognite: Nice catch!

react-components/src/components/CacheProvider/cad/rawAssetMappingTypes.ts
haakonflatval-cognite: No, I don't think so :sweat_smile: Should I remove that possibility here?

react-components/src/components/CacheProvider/cad/ClassicCadAssetMappingCacheImpl.ts
```diff
@@ -253,7 +254,9 @@ class ClassicCadAssetMappingCacheImpl implements ClassicCadAssetMappingCache {
           filter
         })
         .autoPagingToArray({ limit: Infinity })
-    ).flatMap(extractHybridAssetMappings);
+    )
+      .map(convertToHybridAssetMapping)
```
anders-hopland, line 258: I would consider to do a flatmap where you return `[]` for each undefined element as that would be more performant

anders-hopland (approved):
🦄

PR #5233, opened 2025-07-18
fix(react-components): 360 image DM annotation on click selection

pramod-cog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5683

## Description :pencil:
If 360 image dataset contain very large amount of `annotation` and user select a associated `Asset` from `FRS` panel, `annotation` was not selected. The issue was that we used to fetch all annotations in the `Image360Collection` from `Reveal` which will trigger large number network request in parallel causing `429` error in `Image360AnnotationCache. getReveal360Annotations()` call.
This PR changes the logic by fetching annotations for selected Asset by using `Reveal's` `Image360Collection.findImageAnnotations()` which will make a single network request.

In `Unified-3D`

- link `react-component` to fusion branch `pramodcog/bump-reveal-react-component-v0.82.3`
- run `unified-3d` application with `jnc` project
- load `SP2` scene and select Asset `V3001` from `FRS` Panel

### DIFF
--- a/react-components/src/components/CacheProvider/Image360AnnotationCache.test.ts
+++ b/react-components/src/components/CacheProvider/Image360AnnotationCache.test.ts
@@ -1,60 +1,50 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { describe, it, expect, beforeEach } from 'vitest';
 import { Image360AnnotationCache } from './Image360AnnotationCache';
-import { type CogniteClient } from '@cognite/sdk';
-import { type InstanceReference } from '../../utilities/instanceIds';
-import { type Image360Collection } from '@cognite/reveal';
+import { viewerImage360CollectionsMock, viewerMock } from '#test-utils/fixtures/viewer';
+import { retrieveMock, sdkMock } from '#test-utils/fixtures/sdk';
+import { createImage360ClassicMock, findImageAnnotationsMock } from '#test-utils/fixtures/image360';
+import { type Asset } from '@cognite/sdk';
+import { type DataSourceType, type InstanceReference } from '@cognite/reveal';
+import { createAssetMock } from '#test-utils/fixtures/assets';
+import { mockImage360AnnotationAssetResult } from '#test-utils/fixtures/image360AssetWithAnnotation';
 
-function createMockViewer(collections: any[] = []) {
-  return {
-    get360ImageCollections: vi.fn(() => collections)
-  };
-}
-
-function createMockCollection(id: string, annotations: any[] = []) {
-  return {
-    id,
-    findImageAnnotations: vi.fn().mockResolvedValue(annotations)
-  };
-}
-
-const mockSdk = {} as CogniteClient;
-const mockAsset = { id: 'asset1' };
-const mockAssetInstance: InstanceReference = { id: 'asset1' };
-
-describe('Image360AnnotationCache', () => {
+describe(Image360AnnotationCache.name, () => {
   let cache: Image360AnnotationCache;
-  let mockViewer: any;
-  let mockCollection: any;
+  let assets: Asset[];
+  const mockImage360Collection = createImage360ClassicMock();
+
+  const mockAssetInstance: Array<InstanceReference<DataSourceType>> = [{ id: 1 }];
 
   beforeEach(() => {
-    mockCollection = createMockCollection('site1', [
-      { annotation: { annotation: { id: 'ann1' }, getCenter: () => ({ applyMatrix4: vi.fn(() => ({})) }) }, image: { transform: {} } }
-    ]);
-    mockViewer = createMockViewer([mockCollection]);
-    cache = new Image360AnnotationCache(mockSdk, mockViewer);
+    assets = [createAssetMock(1)];
+    viewerImage360CollectionsMock.mockReturnValue([mockImage360Collection]);
+    findImageAnnotationsMock.mockResolvedValue([mockImage360AnnotationAssetResult]);
   });
 
   it('returns empty array if viewer is undefined', async () => {
-    const cacheNoViewer = new Image360AnnotationCache(mockSdk, undefined);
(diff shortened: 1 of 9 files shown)


react-components/src/components/CacheProvider/Image360AnnotationCache.ts
```diff
@@ -2,141 +2,148 @@ import { type CogniteClient } from '@cognite/sdk';
...
-    const image360Collections = this._viewer.get360ImageCollections();
-
-    const annotationsInfoPromise = await Promise.all(
-      image360Collections.map(async (image360Collection: Image360Collection<DataSourceType>) => {
-        return await image360Collection.getAnnotationsInfo('all');
+  private async getReveal360AnnotationInfoForAssets(
+    assetInstances: InstanceReference[],
+    image360Collections: Array<Image360Collection<DataSourceType>>
+  ): Promise<Image360AnnotationAssetInfo[]> {
+    const image360AnnotationAssetNested = await Promise.all(
```
anders-hopland: Would consider early return if `assetInstances` or `image360Collections` are empty

react-components/src/components/CacheProvider/Image360AnnotationCache.ts
```diff
@@ -2,141 +2,148 @@ import { type CogniteClient } from '@cognite/sdk';
...
-
-    const assetIds = new Array<InstanceReference>();
-    annotationsInfo.forEach((annotation) => {
-      const assetRef = getInstanceReferenceFromImage360Annotation(annotation.annotationInfo);
-      if (assetRef !== undefined) {
-        assetIds.push(assetRef);
-      }
-    });
-
+    const image360AnnotationAssets = image360AnnotationAssetNested.flat();
```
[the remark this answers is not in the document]
pramod-cog: In this case, `getReveal360AnnotationInfoForAssets()` is only called for a small number of assets specifically for the currently selected `Asset` in the `FRS` panel which is always a single asset.  
Therefore, the number of concurrent `findImageAnnotations` requests will be just one and the risk of hitting the API rate limit will be rare or near to zero occurrence.

If the usage pattern changes in the future to support bulk or background annotation fetching for many assets, batching might be needed. For now, keeping the implementation simple and direct avoids unnecessary complexity and overhead for this specific, low-volume use case.

react-components/src/components/CacheProvider/Image360AnnotationCache.ts
```diff
@@ -2,141 +2,148 @@ import { type CogniteClient } from '@cognite/sdk';
...
-    const annotationsInfoPromise = await Promise.all(
-      image360Collections.map(async (image360Collection: Image360Collection<DataSourceType>) => {
-        return await image360Collection.getAnnotationsInfo('all');
+  private async getReveal360AnnotationInfoForAssets(
+    assetInstances: InstanceReference[],
+    image360Collections: Array<Image360Collection<DataSourceType>>
+  ): Promise<Image360AnnotationAssetInfo[]> {
+    const image360AnnotationAssetNested = await Promise.all(
+      assetInstances.map(async (assetInstance) => {
+        const results = await Promise.all(
```
anders-hopland: Would chunk this if the number of collections is large

react-components/src/components/CacheProvider/Image360AnnotationCache.ts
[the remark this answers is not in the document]
anders-hopland: I would prefer to handle this now even though we in throey should not have enough assets in our current use case to run into any issues for our current needs. Someone might use this function with a large number of assets down the road without knowing about this limitation and we will then get some issues

react-components/tests/tests-utilities/fixtures/assets.ts
```diff
@@ -2,14 +2,16 @@ import { type Asset } from '@cognite/sdk';
 import { type AssetProperties } from '../../../src/data-providers/core-dm-provider/utils/filters';
 import { type ExternalIdsResultList, type NodeItem } from '../../../src/data-providers/FdmSDK';
 
+const fixedDate = new Date('2025-01-01T00:00:00.000Z');
```
anders-hopland, line 5: :100

react-components/src/components/CacheProvider/Image360AnnotationCache.ts
```diff
@@ -2,141 +2,148 @@ import { type CogniteClient } from '@cognite/sdk';
...
   }
 
-  public async getReveal360Annotations(siteIds: string[]): Promise<Image360AnnotationAssetInfo[]> {
-    const key = siteIds.sort().join();
-    const cachedResult = this._annotationToAssetMappings.get(key);
+  public async getReveal360AnnotationsForAssets(
+    siteIds: string[],
+    assetInstances: InstanceReference[]
+  ): Promise<Image360AnnotationAssetInfo[]> {
+    if (this._viewer === undefined) {
```
anders-hopland: Would return early i siteIds or assetInstances are empty

react-components/src/components/CacheProvider/Image360AnnotationCache.test.ts
```diff
@@ -0,0 +1,50 @@
...
+import { Image360AnnotationCache } from './Image360AnnotationCache';
+import { viewerImage360CollectionsMock, viewerMock } from '#test-utils/fixtures/viewer';
+import { retrieveMock, sdkMock } from '#test-utils/fixtures/sdk';
+import { createImage360ClassicMock, findImageAnnotationsMock } from '#test-utils/fixtures/image360';
+import { type Asset } from '@cognite/sdk';
+import { type DataSourceType, type InstanceReference } from '@cognite/reveal';
+import { createAssetMock } from '#test-utils/fixtures/assets';
+import { mockImage360AnnotationAssetResult } from '#test-utils/fixtures/image360AssetWithAnnotation';
+
+describe(Image360AnnotationCache.name, () => {
```
anders-hopland, line 11: Could be nice with a test verifying that we throw if the API request fails

PR #5237, opened 2025-07-24
refactor(react-components): refactor in preparation for hybrid CAD search

haakonflatval-cognite:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

Part of the work for https://cognitedata.atlassian.net/browse/BND3D-5827

## Description :pencil:

Some preliminary refactoring before introducing the CAD hybrid search functionality. This includes
- Create the `cad` and `common` subdirectories under `queries/network`, since that folder started to grow out of control
- Removed and renamed a few types to make things simpler and more clear
- Factored out `fetchAllHybridAssetMappingsForModel` to a separate file (used by `fetchAllAssetsForCadModels`, and to be used by the upcoming hybrid functionality)
- Rewrote the asset-mappings-to-asset query in `fetchAllAssetsForAssetMappings` to use a loop instead of recursion
- Changed from `.filter` to `.list` when fetching asset mappings through the asset-mappings API

There are no intended functional changes.
The current tests already cover all files touched here, so coverage should be untouched.


react-components/src/query/useSearchMappedEquipmentAssetMappings.tsx
haakonflatval-cognite, line 1: Note that there are some minor changes in this file to maintain current functionality. However, it is not used in production and is scheduled for removal, so I don't think it's worth adding tests.

react-components/src/query/network/fetchAllAssetsForCadModels.ts
haakonflatval-cognite, line 1: This file was moved to `cad/fetchAllAssetsForCadModel`, but it is not recognized as a move because I factored out a major part of it (into the file `cad/fetchAllHybridAssetMappingsForModel`)

react-components/src/query/network/cad/fetchAllHybridAssetMappingsForModels.ts
```diff
@@ -0,0 +1,95 @@
...
+  const cursorForModel = cursorsForModels?.find((nextCursor) =>
+    isSameModel(nextCursor.model, model)
+  )?.cursor;
+
+  if (!isFirstPage && cursorForModel === undefined) {
+    return { mappings: { items: [] }, model };
+  }
+
+  // TODO(BND3D-5844): enable using the Asset mapping cache here by implementing cursoring in the cache
+  const mappings = (await sdk.assetMappings3D.list(model.modelId, model.revisionId, {
```
haakonflatval-cognite, line 82: Note that this call used to be a `.filter` instead of a `.list`. There is no reason we would need to use filter here, so I have reverted to `list`, which also supports the `getDmsInstances` parameter.

react-components/src/query/network/cad/searchClassicCadAssetsWithFilters.ts
```diff
@@ -37,57 +55,43 @@ export async function searchClassicCadAssetsWithFilters(
...
-    if (accumulatedAssets.length >= limit || searchedAssetsResponse.nextCursor === undefined) {
-      return { assets: accumulatedAssets, nextCursor: searchedAssetsResponse.nextCursor };
-    }
+    nextCursor = searchedAssetsResponse.nextCursor;
 
     filteredMappedSearchedAssets.forEach((asset) => {
-      mappedSearchedAssetIds.add(asset.id);
+      resultIds.add(asset.id);
     });
+  } while (accumulatedAssets.length < limit && nextCursor !== undefined);
```
anders-hopland, line 80: Do we have any usecases where we want to get more than one page (1000 elements) but not all elements?

haakonflatval-cognite: Yes, when the user is clicking "show more", we should be loading more elements, but not all elements

anders-hopland (approved):
🦄

PR #5238, opened 2025-07-24
feat(react-components): Hybrid CAD search

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

Solves https://cognitedata.atlassian.net/browse/BND3D-5827 (although, it needs to be used in Fusion too)

## Description :pencil:

Introduces and exposes the `searchHybridDmAssetsForModels` function for searching in hybrid CAD asset mappings with DM instances. It supports both search with and without applied filters.

I'll mention that I'd ideally make this more object-oriented, with at least _some_ common interface(s) between this and the implementation for classic asset mappings; as it would at least make the tests a lot simpler, and probably make the code more re-usable. However, the nature of the two search paths is different due to differing backend capabilities (especially for the query-search), so I have put it off.

Also introduces some minor utilities for tests, and changes/extends typing related to raw assets. This is only to enable finer typing on the function shared between hybrid-dm and classic search (`fetchAllHybridAssetMappingsForModels`)

In Fusion 

TBA

### DIFF
--- a/react-components/src/query/network/cad/searchHybridDmAssetsForCadModels.ts
+++ b/react-components/src/query/network/cad/searchHybridDmAssetsForCadModels.ts
@@ -1,13 +1,11 @@
-import { CogniteClient, FilterDefinition, ViewDefinition } from '@cognite/sdk';
-import { ClassicAdd3DModelOptions } from '../../../components/Reveal3DResources/types';
-import { RevealRenderTarget } from '../../../architecture';
+import { type CogniteClient, type FilterDefinition, type ViewDefinition } from '@cognite/sdk';
+import { type ClassicAdd3DModelOptions } from '../../../components/Reveal3DResources/types';
 import { searchHybridDmCadAssetMappingsWithFilters } from './searchHybridDmCadAssetMappingsWithFilters';
-import { NodeItem } from '../../../data-providers';
+import { type NodeItem } from '../../../data-providers';
 import { fetchAllHybridAssetMappingsForModels } from './fetchAllHybridAssetMappingsForModels';
 import { FdmSDK } from '../../../data-providers/FdmSDK';
-import { restrictToDmsId } from '../../../utilities/restrictToDmsId';
 import { restrictToViewReference } from '../../../utilities/restrictToViewReference';
-import { ClassicCadAssetMappingCache } from '../../../components/CacheProvider/cad/ClassicCadAssetMappingCache';
+import { type ClassicCadAssetMappingCache } from '../../../components/CacheProvider/cad/ClassicCadAssetMappingCache';
 
 export async function searchHybridDmAssetsForCadModels(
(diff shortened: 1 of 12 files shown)


react-components/src/query/network/cad/searchHybridDmAssetsForCadModels.ts
```diff
@@ -0,0 +1,59 @@
...
+  return await searchHybridDmCadAssetMappingsWithFilters(
+    models,
+    options,
+    view,
+    sdk,
+    classicCadCache
+  );
+}
+
+async function fetchAllHybridDmAssetsForCadModels(
```
anders-hopland: As we have a limit we are not neccessarily fetching all
```suggestion
async function fetchHybridDmAssetsForCadModels(
```

react-components/src/query/network/searchHybridDmAssetsForModels.test.ts
```diff
@@ -0,0 +1,245 @@
...
+    const TAGGED_MODEL = { type: 'cad', addOptions: MODEL } as const;
+
+    const HYBRID_ASSET_MAPPINGS = DMS_IDS.map((id) =>
+      createRawDmHybridAssetMappingMock({ assetInstanceId: id })
+    );
+
+    const mockSdkPost = vi.fn<CogniteClient['post']>(
+      defaultPostImplementation as CogniteClient['post']
+    );
+    const mockAssetMappingsList = vi.fn<CogniteClient['assetMappings3D']['list']>(
```
anders-hopland, line 54: type error

react-components/src/query/network/searchHybridDmAssetsForModels.ts
```diff
@@ -0,0 +1,44 @@
...
+export type SearchSort = {
+  property: string[];
+  direction: 'ascending' | 'descending';
+};
+
+export type DMSView = {
+  rawView: ViewDefinition;
+};
+
+export async function searchHybridDmAssetsForModels(
```
anders-hopland, line 19: This will probably only ever work for CAD. Should we reflect that in the name?

react-components/src/query/network/cad/searchHybridDmAssetsForCadModels.ts
haakonflatval-cognite: Good point :sweat_smile:

react-components/src/query/network/searchHybridDmAssetsForModels.test.ts
haakonflatval-cognite: Nooooo

react-components/src/query/network/searchHybridDmAssetsForModels.ts
haakonflatval-cognite: It's intended to be extended for the other models in the future

PR #5248, opened 2025-07-31
feat(react-component): Set names to the Reveal domain objects

nilscognite:
#### Type of change
Feat  

## Description :pencil:

When loading a Reveal domain object, it also set the name of it. This will be used the the tree control or other places.

Also changed `getByRevealModel ` to use `getDescendantsByType` instead of `getChildrenByType`, since this is more general.

By unit test. In production the names are not used yet.

### DIFF
--- a/react-components/src/architecture/concrete/reveal/RevealModelsUtils.ts
+++ b/react-components/src/architecture/concrete/reveal/RevealModelsUtils.ts
@@ -101,7 +101,7 @@ export class RevealModelsUtils {
     try {
       domainObject.name = await RevealModelsUtils.getName(renderTarget.rootDomainObject.sdk, model);
     } catch (error) {
-      console.error('Can not get name if point cloud', error);
+      console.error('Can not get name of point cloud', error);
     }
     return model;
   }
--- a/react-components/src/architecture/concrete/reveal/RevealModelsUtils.ts
+++ b/react-components/src/architecture/concrete/reveal/RevealModelsUtils.ts
@@ -78,7 +78,7 @@ export class RevealModelsUtils {
     try {
       domainObject.name = await RevealModelsUtils.getName(renderTarget.rootDomainObject.sdk, model);
     } catch (error) {
-      console.error('Can not get Cad model name', error);
+      console.error('Can not get name of Cad model', error);
     }
     return model;
   }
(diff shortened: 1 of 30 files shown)


react-components/src/architecture/concrete/reveal/RevealModelsUtils.ts
```diff
@@ -49,14 +49,24 @@ export class RevealModelsUtils {
...
     });
+    if (domainObject !== undefined) {
+      const sdk = renderTarget.rootDomainObject.sdk;
+      await sdk.models3D.retrieve(model.modelId).then((model) => {
+        if (domainObject !== undefined) {
+          domainObject.name = model.name;
+        }
+      });
+    }
+    return model;
```
haakonflatval-cognite: I think this an be simplified quite a bit by dropping the use of `.then` and instead use `await`s everywhere. Also, can we separate out the name-fetching functiononality in a separate function? It's being used by both point clouds and CAD

haakonflatval-cognite (changes requested):
Looks good - one suggestion

react-components/src/architecture/concrete/reveal/RevealModelsUtils.ts
pramod-cog: @haakonflatval-cognite The visual test is failing!! 
We need to fix it first before moving forward

nilscognite: Fixed here 4d68e4073c4a6034b5d373c90ba790f5aa605001

react-components/src/architecture/concrete/reveal/RevealModelsUtils.ts
```diff
@@ -41,40 +44,66 @@ export class RevealModelsUtils {
...
+    root.addChildInteractive(domainObject);
+    if (model.visible) {
+      domainObject.setVisibleInteractive(true);
+    }
+
+    try {
+      domainObject.name = await RevealModelsUtils.getName(renderTarget.rootDomainObject.sdk, model);
+    } catch (error) {
+      console.error('Can not get name if point cloud', error);
+    }
```
haakonflatval-cognite: Would be nice to have a fallback model name if this throws (e.g. `Point cloud modelId/revisionId`), but not critical

haakonflatval-cognite (dismissed):

PR #5252, opened 2025-08-03
refactor(react-components): Use UnitSyste.lengthUnit as signal instaed of just the value

nilscognite:
#### Type of change

Refactor 

## Description :pencil:

See header. No changes in the user experience.

By unit tests.

`yarn storybook`
Click on **Measurments** on the left toolbar
Click on box on the bottom toolbar
Click in the 3D window 3 times to maker a box
Open setting on the left toolbar.
Change the unit
See it changes in label on the box


react-components/src/components/Architecture/DomainObjectPanel.tsx
```diff
@@ -22,18 +26,19 @@ export const DomainObjectPanel = (): ReactElement => {
   useSignalValue(DomainObjectPanelUpdater.update);
   const domainObject = useSignalValue(DomainObjectPanelUpdater.selectedDomainObject);
   const commands = useMemo(() => domainObject?.getPanelToolbar(), [domainObject]);
+  const root = domainObject === undefined ? undefined : getRoot(domainObject);
+  const unitSystem = root === undefined ? UNDEFINED_UNIT_SYSTEM : root.unitSystem;
```
nilscognite, line 32: When no root, just use the predefined UNDEFINED_UNIT_SYSTEM. This will normally not being used, but  it makes compiler/react happy

react-components/src/architecture/base/domainObjects/RootDomainObject.ts
```diff
@@ -43,6 +44,14 @@ export class RootDomainObject extends DomainObject {
...
     this._fdmSdk = new FdmSDK(sdk);
     this.addChild(new AxisDomainObject());
+
+    this.addEffect(() => {
+      this.unitSystem.lengthUnit();
+
+      // Make sure all views are notified when unit change
+      this.notify(Changes.unit);
+      this.notifyDescendants(Changes.unit);
+    });
```
haakonflatval-cognite, line 54: Just wondering - do we want to get rid of this code at some point and let all domain objects and commands rely directly on the signal? Or will we keep this in the foreseeable future?

haakonflatval-cognite (approved):
LGTM 👍  Just had a question to a small part of the code

react-components/src/architecture/base/domainObjects/RootDomainObject.ts
nilscognite: I will have a fix for this later, in another PR. But I want to keep the existing system as it is,  because the benefits are huge and it scales a lot better than signal. A can explain to you why.

pramod-cog (dismissed):
risk review ok

PR #5254, opened 2025-08-06
feat(react-components): initial support for hybrid cad in rule based coloring

danpriori:
#### Type of change

Feat  

#### Jira ticket :blue_book:

## Description :pencil:
Initial support for hybrid cad in rule base by fetching the hybrid data from cache and transforming and adding data to be a fdmMapping similar with the current fdm mappings.
It is coming along with the PRs: 
https://github.com/cognitedata/reveal/pull/5269
https://github.com/cognitedata/reveal/pull/5273

Using the hybrid projects that have hybrid cad mappings and creating rule based coloring for these mappings (cog-demo and cog-test). 
Performed regression tests as well (DM project and no hybrid mappings)

1- In cog-demo, open the Fornebu scene
2- Open the Rule Based Tool and select the last rule "Test Hybrid"
3- Observe the hybrid mappings with the description property that contains the string "SEPAR GAS" are highlighted

image

- [X] I am happy with this implementation.

### DIFF
--- a/react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
+++ b/react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
@@ -1,5 +1,4 @@
 import { type UseQueryResult, useQuery } from '@tanstack/react-query';
-import { useFdmSdk } from '../components/RevealCanvas/SDKProvider';
 import { type FdmKey, type FdmConnectionWithNode } from '../components/CacheProvider/types';
 import { type InstanceType } from '@cognite/sdk';
 import { chunk, uniqBy } from 'lodash';
@@ -7,17 +6,19 @@ import {
   type FdmInstanceNodeWithConnectionAndProperties,
   type FdmInstanceWithPropertiesAndTyping
 } from '../components/RuleBasedOutputs/types';
-import { useMemo } from 'react';
+import { useContext, useMemo } from 'react';
 import { createFdmKey } from '../components/CacheProvider/idAndKeyTranslation';
 import { executeParallel } from '../utilities/executeParallel';
 import { isDefined } from '../utilities/isDefined';
 import { concatenateMapValues } from '../utilities/map/concatenateMapValues';
+import { UseAll3dDirectConnectionsWithPropertiesContext } from './useAll3dDirectConnectionsWithProperties.context';
 
 const MAX_PARALLEL_QUERIES = 4;
 
 export function useAll3dDirectConnectionsWithProperties(
   connectionWithNodeAndView: FdmConnectionWithNode[]
 ): UseQueryResult<FdmInstanceNodeWithConnectionAndProperties[]> {
+  const { useFdmSdk } = useContext(UseAll3dDirectConnectionsWithPropertiesContext);
   const fdmSdk = useFdmSdk();
 
   const connectionKeys = useMemo(() => {
--- a/react-components/package.json
+++ b/react-components/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cognite/reveal-react-components",
-  "version": "0.84.1",
+  "version": "0.84.2",
   "exports": {
     ".": {
       "import": "./dist/index.js",
@@ -38,7 +38,7 @@
     "sort-keys": "cdf-i18n-utils-cli sort-local-keys --namespace reveal-react-components --path ./src/common/i18n"
   },
   "peerDependencies": {
-    "@cognite/reveal": "4.25.3",
+    "@cognite/reveal": "4.25.4",
     "@cognite/sdk": "^10.0.0",
     "react": ">=18",
     "react-dom": ">=18",
@@ -56,7 +56,7 @@
   "devDependencies": {
     "@cognite/cdf-i18n-utils": "^0.7.5",
     "@cognite/cdf-utilities": "^3.6.0",
-    "@cognite/reveal": "4.25.3",
+    "@cognite/reveal": "4.25.4",
     "@cognite/sdk": "^10.0.0",
     "@playwright/test": "1.49.0",
     "@storybook/addon-essentials": "8.4.5",
--- a/react-components/src/architecture/base/concreteCommands/pointCloud/PointCloudFilterCommand.ts
+++ b/react-components/src/architecture/base/concreteCommands/pointCloud/PointCloudFilterCommand.ts
@@ -102,7 +102,7 @@ export class PointCloudFilterCommand extends BaseFilterCommand {
 
 // Note: This is not exported, as it is only used internally
 
-class FilterItemCommand extends BaseFilterItemCommand {
+export class FilterItemCommand extends BaseFilterItemCommand {
   private readonly _pointClass: PointClass;
   private readonly _currentUniqueId: UniqueId;
 
@@ -153,7 +153,7 @@ class FilterItemCommand extends BaseFilterItemCommand {
   // ==================================================
 }
 
-class PointClass {
+export class PointClass {
   name: string;
   code: number | WellKnownAsprsPointClassCodes;
   color: Color;
--- a/react-components/src/query/useAll3dDirectConnectionsWithProperties.context.ts
+++ b/react-components/src/query/useAll3dDirectConnectionsWithProperties.context.ts
@@ -1,0 +1,16 @@
+import { createContext } from 'react';
+import { useFdmSdk } from '../components/RevealCanvas/SDKProvider';
+
+export type UseAll3dDirectConnectionsWithPropertiesDependencies = {
+  useFdmSdk: typeof useFdmSdk;
+};
+
+export const defaultUseAll3dDirectConnectionsWithPropertiesDependencies: UseAll3dDirectConnectionsWithPropertiesDependencies =
+  {
+    useFdmSdk
+  };
+
+export const UseAll3dDirectConnectionsWithPropertiesContext =
+  createContext<UseAll3dDirectConnectionsWithPropertiesDependencies>(
+    defaultUseAll3dDirectConnectionsWithPropertiesDependencies
+  );
--- a/react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
+++ b/react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
@@ -1,0 +1,282 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { useAll3dDirectConnectionsWithProperties } from './useAll3dDirectConnectionsWithProperties';
+import {
+  defaultUseAll3dDirectConnectionsWithPropertiesDependencies,
+  UseAll3dDirectConnectionsWithPropertiesContext
+} from './useAll3dDirectConnectionsWithProperties.context';
+import { type FdmConnectionWithNode } from '../components/CacheProvider/types';
+import {
+  type ExternalIdsResultList,
+  type InspectResultList,
+  type NodeItem,
+  type FdmSDK
+} from '../data-providers/FdmSDK';
+import { Mock } from 'moq.ts';
+import type { FC, PropsWithChildren } from 'react';
+import { getMocksByDefaultDependencies } from '#test-utils/vitest-extensions/getMocksByDefaultDependencies';
+import { FdmSdkContext } from '../components/RevealCanvas/FdmDataProviderContext';
+import { createCadNodeMock } from '#test-utils/fixtures/cadNode';
+import { type FdmInstanceNodeWithConnectionAndProperties } from '../components/RuleBasedOutputs/types';
+
+describe(useAll3dDirectConnectionsWithProperties.name, () => {
+  const queryClient = new QueryClient();
+
+  const ARBITRARY_DATA_SET_SIZE = 1500;
+  const ARBITRARY_TREE_INDEX = 789;
+  const ARBITRARY_MODEL_ID = 123;
+  const ARBITRARY_REVISION_ID = 456;
+
+  const baseDmsUniqueIdentifier = {
+    space: 'test-space',
+    externalId: 'test-id'
+  };
+
+  const node3d1 = createCadNodeMock({
+    id: 100,
+    treeIndex: ARBITRARY_TREE_INDEX,
+    subtreeSize: 1
+  });
+
+  const node3d2 = createCadNodeMock({
+    id: 101,
+    treeIndex: ARBITRARY_TREE_INDEX + 1,
+    subtreeSize: 1
+  });
+
+  const mockConnectionWithNode: FdmConnectionWithNode = {
+    connection: {
+      instance: baseDmsUniqueIdentifier,
+      modelId: ARBITRARY_MODEL_ID,
+      revisionId: ARBITRARY_REVISION_ID,
+      treeIndex: ARBITRARY_TREE_INDEX
+    },
+    cadNode: node3d1,
+    views: [{ space: 'view-space', externalId: 'view-id', version: '1', type: 'view' as const }]
+  };
+
+  const mockConnectionWithNodeSameInstance: FdmConnectionWithNode = {
+    connection: {
+      instance: baseDmsUniqueIdentifier,
+      modelId: ARBITRARY_MODEL_ID,
+      revisionId: ARBITRARY_REVISION_ID,
+      treeIndex: ARBITRARY_TREE_INDEX + 1
+    },
+    cadNode: node3d2,
+    views: [{ space: 'view-space', externalId: 'view-id', version: '1', type: 'view' as const }]
+  };
+
+  const mockConnectionWithNode2: FdmConnectionWithNode = {
+    connection: {
+      instance: { space: 'test-space', externalId: 'test-id-2' },
+      modelId: ARBITRARY_MODEL_ID,
+      revisionId: ARBITRARY_REVISION_ID,
+      treeIndex: ARBITRARY_TREE_INDEX + 1
+    },
+    cadNode: node3d2,
+    views: [{ space: 'view-space-2', externalId: 'view-id-2', version: '1', type: 'view' as const }]
+  };
+
+  const mockFdmInstanceData = {
+    ...baseDmsUniqueIdentifier,
+    instanceType: 'node' as const,
+    version: 1,
+    createdTime: 11111,
+    lastUpdatedTime: 11111,
+    properties: { testProperty: { value: 'testValue' } }
+  };
+
+  const mockInspectResultItem = {
+    ...baseDmsUniqueIdentifier,
+    instanceType: 'node' as const,
+    inspectionResults: {
+      involvedViews: [
+        { space: 'view-space', externalId: 'view-id', version: '1', type: 'view' as const }
+      ],
+      involvedContainers: []
+    }
+  };
+
+  const mockInspectResultList = [mockInspectResultItem];
+
+  const mockFdmSdk = new Mock<FdmSDK>()
+    .setup((p) => p.getByExternalIds)
+    .returns(vi.fn<() => Promise<ExternalIdsResultList<any>>>())
+    .setup((p) => p.inspectInstances)
+    .returns(vi.fn<() => Promise<InspectResultList>>())
+    .object();
+
+  const mockDependencies = getMocksByDefaultDependencies(
+    defaultUseAll3dDirectConnectionsWithPropertiesDependencies
+  );
+
+  const wrapper: FC<PropsWithChildren<unknown>> = ({ children }) => (
+    <QueryClientProvider client={queryClient}>
+      <FdmSdkContext.Provider value={{ fdmSdk: mockFdmSdk }}>
+        <UseAll3dDirectConnectionsWithPropertiesContext.Provider value={mockDependencies}>
+          {children}
+        </UseAll3dDirectConnectionsWithPropertiesContext.Provider>
+      </FdmSdkContext.Provider>
+    </QueryClientProvider>
+  );
+
+  beforeEach(() => {
+    queryClient.clear();
+
+    mockDependencies.useFdmSdk.mockReturnValue(mockFdmSdk);
+
+    vi.mocked(mockFdmSdk.getByExternalIds).mockResolvedValue({
+      items: [mockFdmInstanceData],
+      typing: {}
+    });
+
+    vi.mocked(mockFdmSdk.inspectInstances).mockResolvedValue({
+      items: mockInspectResultList
+    });
+  });
+
+  it('should return empty result when no connections provided', async () => {
+    const { result } = renderHook(() => useAll3dDirectConnectionsWithProperties([]), { wrapper });
+
+    expect(result.current.data).toBeUndefined();
+    expect(result.current.isLoading).toBe(false);
+  });
+
+  it('should process connections and return multiple instance data', async () => {
+    const { result } = renderHook(
+      () =>
+        useAll3dDirectConnectionsWithProperties([
+          mockConnectionWithNode,
+          mockConnectionWithNodeSameInstance
+        ]),
+      { wrapper }
+    );
+
+    const expectedData = [
+      createFdmInstanceNodeWithConnectionDataMock(mockConnectionWithNode, mockFdmInstanceData),
+      createFdmInstanceNodeWithConnectionDataMock(
+        mockConnectionWithNodeSameInstance,
+        mockFdmInstanceData
+      )
+    ];
+
+    await waitFor(() => {
+      expect(result.current.isSuccess).toBe(true);
+    });
+
+    expect(mockFdmSdk.getByExternalIds).toHaveBeenCalled();
+    expect(mockFdmSdk.inspectInstances).toHaveBeenCalled();
+    expect(result.current.data).toEqual(expectedData);
+  });
+
+  it('should handle empty instance data response', async () => {
+    vi.mocked(mockFdmSdk.getByExternalIds).mockResolvedValue({
+      items: [],
+      typing: {}
+    });
+
+    vi.mocked(mockFdmSdk.inspectInstances).mockResolvedValue({
+      items: []
+    });
+
+    const { result } = renderHook(
+      () => useAll3dDirectConnectionsWithProperties([mockConnectionWithNode]),
+      { wrapper }
+    );
+
+    await waitFor(() => {
+      expect(result.current.isSuccess).toBe(true);
+    });
+
+    expect(result.current.data).toEqual([]);
+  });
+
+  it('should try to fetch fdm nodes without views', async () => {
+    const connectionWithoutViews: FdmConnectionWithNode = {
+      ...mockConnectionWithNode,
+      views: undefined
+    };
+
+    const { result } = renderHook(
+      () => useAll3dDirectConnectionsWithProperties([connectionWithoutViews]),
+      { wrapper }
+    );
+
+    await waitFor(() => {
+      expect(result.current.isSuccess).toBe(true);
+    });
+
+    expect(mockFdmSdk.getByExternalIds).toHaveBeenCalledWith(expect.any(Array), undefined);
+  });
+
+  it('should chunk large datasets correctly', async () => {
+    const connections = Array.from({ length: ARBITRARY_DATA_SET_SIZE }, (_, i) => ({
+      ...mockConnectionWithNode,
+      connection: {
+        ...mockConnectionWithNode.connection,
+        instance: { space: 'space', externalId: `id${i}` },
+        treeIndex: ARBITRARY_TREE_INDEX + i
+      }
+    }));
+
+    vi.mocked(mockFdmSdk.getByExternalIds).mockResolvedValue({
+      items: [],
+      typing: {}
+    });
+
+    vi.mocked(mockFdmSdk.inspectInstances).mockResolvedValue({
+      items: []
+    });
+
+    const { result } = renderHook(() => useAll3dDirectConnectionsWithProperties(connections), {
+      wrapper
+    });
+
+    await waitFor(() => {
+      expect(result.current.isSuccess).toBe(true);
+    });
+
+    // Should call getByExternalIds multiple times due to chunking
+    expect(mockFdmSdk.getByExternalIds).toHaveBeenCalledTimes(2); // 2 chunks of 1000 each
+  });
+
+  it('should refetch when connections change', async () => {
+    const initialCallCount = 1; // Initial call to getByExternalIds
+    const extraCallCount = 3; // After rerender with new connections
+
+    const { result, rerender } = renderHook(
+      ({ connections }) => useAll3dDirectConnectionsWithProperties(connections),
+      {
+        wrapper,
+        initialProps: { connections: [mockConnectionWithNode] }
+      }
+    );
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(vi.mocked(mockFdmSdk.getByExternalIds)).toHaveBeenCalledTimes(initialCallCount);
+
+    rerender({ connections: [mockConnectionWithNode, mockConnectionWithNode2] });
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(vi.mocked(mockFdmSdk.getByExternalIds)).toHaveBeenCalledTimes(extraCallCount);
+  });
+});
+
+function createFdmInstanceNodeWithConnectionDataMock(
+  connection: FdmConnectionWithNode,
+  mockFdmInstanceData: NodeItem<unknown>
+): FdmInstanceNodeWithConnectionAndProperties {
+  return {
+    ...connection,
+    items: [mockFdmInstanceData],
+    instanceType: 'node' as const,
+    typing: {}
+  };
+}
--- a/react-components/yarn.lock
+++ b/react-components/yarn.lock
@@ -610,7 +610,7 @@ __metadata:
     "@cognite/cdf-utilities": "npm:^3.6.0"
     "@cognite/cogs-lab": "npm:9.0.0-alpha.160"
     "@cognite/cogs.js": "npm:^10.36.0"
-    "@cognite/reveal": "npm:4.25.3"
+    "@cognite/reveal": "npm:4.25.4"
     "@cognite/sdk": "npm:^10.0.0"
     "@cognite/signals": "npm:^0.0.4"
     "@floating-ui/dom": "npm:^1.0.0"
@@ -667,7 +667,7 @@ __metadata:
     vite-plugin-externalize-deps: "npm:^0.8.0"
     vitest: "npm:^3.0.6"
   peerDependencies:
-    "@cognite/reveal": 4.25.3
+    "@cognite/reveal": 4.25.4
     "@cognite/sdk": ^10.0.0
     react: ">=18"
     react-dom: ">=18"
@@ -675,9 +675,9 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@cognite/reveal@npm:4.25.3":
-  version: 4.25.3
-  resolution: "@cognite/reveal@npm:4.25.3"
+"@cognite/reveal@npm:4.25.4":
+  version: 4.25.4
+  resolution: "@cognite/reveal@npm:4.25.4"
   dependencies:
     "@rajesh896/broprint.js": "npm:^2.1.1"
     "@tweenjs/tween.js": "npm:^25.0.0"
@@ -697,7 +697,7 @@ __metadata:
   peerDependencies:
     "@cognite/sdk": ^10.0.0
     three: 0.176.0
-  checksum: 10/41c8ebe1256397f771fcd1b4aa311b159e8c4c9276edba0ad902c1eba2d600d30597f254ba09a41beb7e66d086bff7cec2971b0a923a6480de3613a47d750646
+  checksum: 10/1f2a7d6b874bc5a3566746000c07dead780a2aa090013e07bee7118d45fd1216ce4add89d0867d01b761eff721a413c7d934326587fbda004a0a08c03d227898
   languageName: node
   linkType: hard
 
--- a/viewer/package.json
+++ b/viewer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cognite/reveal",
-  "version": "4.25.3",
+  "version": "4.25.4",
   "description": "WebGL based 3D viewer for CAD and point clouds processed in Cognite Data Fusion.",
   "homepage": "https://github.com/cognitedata/reveal/tree/master/viewer",
   "repository": {


react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
```diff
@@ -27,12 +27,17 @@ export function useAll3dDirectConnectionsWithProperties(
...
+    connectionWithNodeAndView.forEach((item) => {
+      const fdmKey = createFdmKey(item.connection.instance);
+      const currentMap = dataMap.get(fdmKey);
+      if (currentMap !== undefined) {
+        currentMap.push(item);
+      } else {
+        dataMap.set(fdmKey, [item]);
+      }
+    });
+    return dataMap;
```
haakonflatval-cognite: We have a function called `concatenateMapValues` which you may be able to use here to simplify this

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
```diff
@@ -145,24 +150,26 @@ export function useAll3dDirectConnectionsWithProperties(
...
-
-          return {
-            ...connectionFound,
-            ...itemsData
-          };
-        }) ?? [];
+          }).flat().filter(isDefined);
+          return dataFound;
+        }
+      ) ?? [];
```
haakonflatval-cognite: I think the `??` is unnecessary now
```suggestion
      );
```

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
```diff
@@ -145,24 +150,26 @@ export function useAll3dDirectConnectionsWithProperties(
...
           .filter((item) => item.items.length > 0);
 
-      const instanceWithData =
-        instanceItemsAndTyping?.flatMap((itemsData) => {
-          let connectionFound: FdmConnectionWithNode | undefined;
-
-          itemsData.items.every((itemData) => {
+      const instanceWithData = instanceItemsAndTyping.flatMap(
+        (itemsData) => {
+          let connectionsFound: FdmConnectionWithNode[] | undefined;
```
haakonflatval-cognite: Can we remove this variable declaration? I don't see why it needs to be declared outside the `map` where it's used 🤔

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
```diff
@@ -145,24 +150,26 @@ export function useAll3dDirectConnectionsWithProperties(
...
+          const dataFound = itemsData.items.map((itemData) => {
             const fdmKey = createFdmKey(itemData);
             if (connectionWithNodeAndViewMap.has(fdmKey)) {
-              connectionFound = connectionWithNodeAndViewMap.get(fdmKey);
-              return false;
+              connectionsFound = connectionWithNodeAndViewMap.get(fdmKey);
```
haakonflatval-cognite: I have a feeling inverting the condition and do an early return could make the code a little easier to read

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,86 @@
...
+                  name: cadNodeItem.name,
+                  modelId,
+                  revisionId,
+                  nodeId: cadNodeItem.id,
+                  treeIndex: cadNodeItem.treeIndex,
+                  subtreeSize: cadNodeItem.subtreeSize,
+                  instanceType: 'node',
+                  version: 1,
+                  properties: cadNodeItem.properties
+                },
```
haakonflatval-cognite:
```suggestion
                cadNode: cadNodeItem,
```
can we just do this?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -137,12 +151,9 @@ export function RuleBasedOutputsSelector({
...
-    if (!ruleSetStylingCache.has(ruleSet.id)) {
-      void ruleBasedInitilization();
-    } else {
-      onAllMappingsFetched(true);
-      if (onRuleSetChanged !== undefined) onRuleSetChanged(ruleSetStylingCache.get(ruleSet.id));
-    }
+    void ruleBasedInitilization();
+    onAllMappingsFetched(true);
+    if (onRuleSetChanged !== undefined) onRuleSetChanged(ruleSetStylingCache.get(ruleSet.id));
   }, [
```
haakonflatval-cognite: Just so I understand - why are we removing this condition?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -64,14 +70,22 @@ export function RuleBasedOutputsSelector({
...
   const fdmConnectionWithNodeAndViewList = useMemo(() => {
     return fdmMappedEquipmentEdges !== undefined
       ? Array.from(fdmMappedEquipmentEdges.values()).flat()
       : [];
   }, [fdmMappedEquipmentEdges]);
 
   const { data: fdmMappings, isLoading: isFdmMappingsLoading } =
-    useAll3dDirectConnectionsWithProperties(fdmConnectionWithNodeAndViewList);
+    useAll3dDirectConnectionsWithProperties(
+      fdmConnectionWithNodeAndViewList.concat(dmConnectionWithNodeFromHybridDataList ?? [])
```
haakonflatval-cognite: It's a little annoying - but I think we should put this concatination in a `useMemo`

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -64,14 +70,22 @@ export function RuleBasedOutputsSelector({
   const { data: fdmMappedEquipmentEdges, isLoading: isFdmMappingsEdgesLoading } =
     useMappedEdgesForRevisions(cadModels, true);
 
+  const { data: dmConnectionWithNodeFromHybridDataList } =
+    useGetDMConnectionWithNodeFromHybridMappingsQuery(
+      nodeWithDmIdsFromHybridMappings ?? [],
```
haakonflatval-cognite: Can we use `?? EMPTY_ARRAY`? This will create a new reference every rerender if `nodeWithDmIdsFromHybridMappings` is undefined

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,86 @@
...
+  const fdmSDK = useFdmSdk();
+  const dmIds = nodeWithDmIdsFromHybridMappings.map((item) => item.instanceId).filter(isDefined);
+
+  return useQuery({
+    queryKey: ['fdm-mappings', dmIds, models],
+    queryFn: async () => {
+      const dataResult = await Promise.all(
+        models.flatMap(async ({ modelId, revisionId }) => {
+          const assetMappingDataPerModelAndInstance =
+            await assetMappingCache.getNodesForInstanceIds(modelId, revisionId, dmIds);
```
haakonflatval-cognite: Is it possible to put the function inside the `flatMap` in a separate function? I think that would make this section much easier to read 😅

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,86 @@
...
+
+          if (assetMappingDataPerModelAndInstance.size === 0) {
+            return [];
+          }
+          const dmNodesInspect = await inspectNodes(fdmSDK, dmIds);
```
haakonflatval-cognite: Here we perform the same inspect call once for every model, which seems wasteful. I suppose we can just perform this once outside the loop?

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,86 @@
...
+export function useGetDMConnectionWithNodeFromHybridMappingsQuery(
+  nodeWithDmIdsFromHybridMappings: DmCadAssetMapping[],
+  models: CadModelOptions[]
+): UseQueryResult<FdmConnectionWithNode[]> {
+  const { useClassicCadAssetMappingCache, useFdmSdk } = useContext(
+    UseGetDMConnectionWithNodeFromHybridMappingsQueryContext
+  );
+  const assetMappingCache = useClassicCadAssetMappingCache();
+  const fdmSDK = useFdmSdk();
+  const dmIds = nodeWithDmIdsFromHybridMappings.map((item) => item.instanceId).filter(isDefined);
```
haakonflatval-cognite: Maybe this should be in a `useMemo`? 🤔

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
haakonflatval-cognite: Note to self - we should have a cache for DM inspect-calls - but that's for another PR, I guess 😅

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+          name: 'Test Node',
+          modelId: modelsMock[0].modelId,
+          revisionId: modelsMock[0].revisionId,
+          nodeId: firstNode3DMock.id,
+          treeIndex: firstNode3DMock.treeIndex,
+          subtreeSize: firstNode3DMock.subtreeSize,
+          instanceType: 'node',
+          version: 1,
+          properties: {}
+        },
```
haakonflatval-cognite: Isn't this a normal `Node3D`? Why do we need e.g. `version` and `properties`? There is also a `createCadNodeMock` function you may be able to use to create this

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,86 @@
...
+
+            const views = dmNode.inspectionResults.involvedViews ?? [];
+            return cadNodesData.map((cadNodeItem) => {
+              return {
+                connection: {
+                  instance: {
+                    externalId: dmNode.externalId,
+                    space: dmNode.space
+                  },
+                  instanceType: 'node',
```
haakonflatval-cognite:
```suggestion
                  instance: restrictToDmsId(dmNode)
```
I think we can remove `instanceType` and simplify the `instance` definition here

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+  const modelsMock: CadModelOptions[] = [
+    {
+      type: 'cad',
+      modelId: 123,
+      revisionId: 456
+    }
+  ];
+
+  const firstNode3DMock: Node3D = {
```
haakonflatval-cognite: Use `createCadNodeMock` here and below

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+    }
+  ];
+
+  const mockAssetMappingData = new Map<InstanceKey, Node3D[]>([
+    [
+      `${mockDmCadAssetMappings[0].instanceId.space}/${mockDmCadAssetMappings[0].instanceId.externalId}` as InstanceKey,
+      [firstNode3DMock]
+    ],
+    [
+      `${mockDmCadAssetMappings[1].instanceId.space}/${mockDmCadAssetMappings[1].instanceId.externalId}` as InstanceKey,
```
haakonflatval-cognite:
```suggestion
      `${mockDmCadAssetMappings[0].instanceId.space}/${mockDmCadAssetMappings[0].instanceId.externalId}` as const,
      [firstNode3DMock]
    ],
    [
      `${mockDmCadAssetMappings[1].instanceId.space}/${mockDmCadAssetMappings[1].instanceId.externalId}` as const,
```

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+  const mockDmCadAssetMappings: DmCadAssetMapping[] = [
+    {
+      instanceId: {
+        space: 'test-space',
+        externalId: 'instance-1'
+      },
+      treeIndex: firstNode3DMock.treeIndex,
+      nodeId: firstNode3DMock.id,
+      subtreeSize: firstNode3DMock.subtreeSize
+    },
```
haakonflatval-cognite: Can we have a function that takes in a DMS ID and a Node3D to create this object? We do this construction four times in this file

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+  );
+
+  beforeEach(() => {
+    queryClient.clear();
+
+    mockDependencies.useFdmSdk.mockReturnValue(mockFdmSdk);
+  });
+
+  it('should return an empty array if no matched asset mappings are found from the mappings cache', async () => {
+    const mockClassicCadAssetMappingCache = new Mock<ClassicCadAssetMappingCache>()
```
haakonflatval-cognite: `ClassicCadAssetMappingCache` is an interface, so perhaps we can just construct it directly with e.g. `const cache: ClassicCadAssetMappingCache = { getNodesForInstanceIds: vi.fn(), ... };`? At least I think that's a little simpler 😅

haakonflatval-cognite (changes requested):
I think the overall approach makes sense 👍 I had a few code-stylistic suggestions

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: Hmmmm... TBH, I don't think it is a good idea to use this global constant as shared across different components that have a mutation potential it as this const can be turned into a regular variable down the road, but its reference keeps the same.
I saw this happening in some other components and it was hard to detect that. 
The only way is to create a new empty reference on any renderer, which will be garbage collected eventually anyway. So I don't think creating a new empty array on every render (which is not that large amount of rendering here) to keep the mutation away is not a bad idea, not because the usage of it in this component in isolation but rather than where this constant is used and transformed at the end of the road that could affect other components or being affected by other components that are using this constant at the same time. 
Does it make sense? Are you sure that it is still safe to use ?🤔

haakonflatval-cognite: Yes, I see the point.. However, I think mutating an array that comes from elsewhere (either as argument or as return value from a function call) is very risky and should be avoided anyway. We can use [`Object.freeze`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) to throw a runtime error if we try to do this anywhere in the code.

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: The rule needs to be recalculated when a late fetch of classic or fdm mappings occurs. That condition was predicting that a recalculation would occur for hybrid projects.

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
danpriori: Then I have to mock all the other functions defined in the interface. Using new Mock I dont need to declare it.

```typescript
    const mockClassicCadAssetMappingCache: ClassicCadAssetMappingCache = {
      getNodesForInstanceIds: vi.fn(async () => await Promise.resolve(mockNoMatchedAssetMappingData)),
      getAssetMappingsForLowestAncestor: vi.fn(),
      generateNode3DCachePerItem: vi.fn(),
      generateAssetMappingsCachePerItemFromModelCache: vi.fn(),
      getAssetMappingsForModel: vi.fn()
    };

```

danpriori: As we discussed, we decided to keep as it is to avoid mocking all the other not-related methods from the interface

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
danpriori: yes sure
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
danpriori: changed here:
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
danpriori: nice!
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: cool
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
danpriori: nice but I'mn reusing the nodes that are mocked above which are using the `createCadNodeMock` already:

https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: nice catch
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: nice catch
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: done
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: changed here:
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: yep
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
danpriori: https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
danpriori: yep
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: yep
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
danpriori: yeah
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
danpriori: yes. Sorry, forgot to update this one as well

here:
https://github.com/cognitedata/reveal/pull/5254/commits/d149b00a940c611419b44fc2b81a4983ca004eb3#diff-358f296dc194c0822e64fcf160cce88cd5d9e43f6e985a6b423d481137b76d1bR33

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,97 @@
...
+    staleTime: Infinity
+  });
+}
+
+function uniqueFdmConnectionWithNodeItem(item: FdmConnectionWithNode): string {
+  return `${item.connection.modelId}/${item.connection.revisionId}/${item.connection.instance.space}/${item.connection.instance.externalId}/${item.cadNode.id}/${item.cadNode.treeIndex}`;
+}
+
+function generateFdmConnectionItemFromDmNodeAndMappingCache(
+  dmNodesInspect: InspectResultList,
```
haakonflatval-cognite:
```suggestion
function generateFdmConnectionsFromInspectResultAndMappings(
  dmNodesInspect: InspectResultList,
```

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,97 @@
...
+  assetMappingDataPerModelAndInstance: Map<InstanceKey, Node3D[]>,
+  modelId: number,
+  revisionId: number
+): FdmConnectionWithNode[] {
+  return dmNodesInspect.items.flatMap((dmNode) => {
+    if (!assetMappingDataPerModelAndInstance.has(`${dmNode.space}/${dmNode.externalId}`)) {
+      return [];
+    }
+    const cadNodesData = assetMappingDataPerModelAndInstance.get(
+      `${dmNode.space}/${dmNode.externalId}`
```
haakonflatval-cognite:
```suggestion
    if (!assetMappingDataPerModelAndInstance.has(createFdmKey(dmNode))) {
      return [];
    }
    const cadNodesData = assetMappingDataPerModelAndInstance.get(
      createFdmKey(dmNode)
```
Or alternatively put the key in a constant.. That's probably cleaner 🤔

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
danpriori: https://github.com/cognitedata/reveal/pull/5254/commits/bdf9f63c38395952b402b71ef8561ab90098e93e

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+      }
+    );
+
+    const expectedData = [
+      {
+        connection: {
+          instance: {
+            externalId: mockDmCadAssetMappingsList[0].instanceId.externalId,
+            space: mockDmCadAssetMappingsList[0].instanceId.space
+          },
```
haakonflatval-cognite:
```suggestion
          instance: mockDmCadAssetMappingsList[0].instanceId,
```

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+        },
+        cadNode: firstNode3DMock,
+        views: mockDmNodesInspectList.items[0].inspectionResults.involvedViews
+      },
+      {
+        connection: {
+          instance: {
+            externalId: mockDmCadAssetMappingsList[1].instanceId.externalId,
+            space: mockDmCadAssetMappingsList[1].instanceId.space
+          },
```
haakonflatval-cognite:
```suggestion
          instance: mockDmCadAssetMappingsList[1].instanceId,
```

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+      { space: 'test-space', externalId: 'other-instance-2' }
+    ],
+    [firstNode3DMock, secondNode3DMock]
+  );
+
+  const mockDmNodesInspectList: InspectResultList = {
+    items: [
+      {
+        space: mockDmCadAssetMappingsList[0].instanceId.space,
+        externalId: mockDmCadAssetMappingsList[0].instanceId.externalId,
```
haakonflatval-cognite:
```suggestion
        ...mockDmCadAssetMappingsList[0].instanceId,
```
I think this would be nicer - but let me know if you disagree!

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+          involvedViews: [
+            { type: 'view', version: '1', space: 'test-space', externalId: 'view-external-id-1' }
+          ],
+          involvedContainers: []
+        },
+        instanceType: 'node'
+      },
+      {
+        space: mockDmCadAssetMappingsList[1].instanceId.space,
+        externalId: mockDmCadAssetMappingsList[1].instanceId.externalId,
```
haakonflatval-cognite:
```suggestion
        ...mockDmCadAssetMappingsList[1].instanceId,
```

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,97 @@
...
+      if (dmIds.length === 0 || models.length === 0) {
+        return EMPTY_ARRAY;
+      }
+      const dmNodesInspect = await inspectNodes(fdmSDK, dmIds);
+      if (dmNodesInspect.items.length === 0) {
+        return EMPTY_ARRAY;
+      }
+      const dataResult = await Promise.all(
+        models.flatMap(async ({ modelId, revisionId }) => {
+          const assetMappingDataPerModelAndInstance =
```
haakonflatval-cognite: I imagine there's a chance we can get 429-problems here if we have many models.. But not sure if it's worth dealing with now

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,97 @@
...
+      const uniqueDataResult = uniqBy(dataResult.flat(), uniqueFdmConnectionWithNodeItem);
+      return uniqueDataResult;
+    },
+    enabled: dmIds.length > 0 && models.length > 0,
+    staleTime: Infinity
+  });
+}
+
+function uniqueFdmConnectionWithNodeItem(item: FdmConnectionWithNode): string {
+  return `${item.connection.modelId}/${item.connection.revisionId}/${item.connection.instance.space}/${item.connection.instance.externalId}/${item.cadNode.id}/${item.cadNode.treeIndex}`;
```
haakonflatval-cognite: Strictly speaking, I don't think we need `treeIndex` here, since `cadNode.id` is always unique within a model 😅

haakonflatval-cognite (changes requested):
Nice, it's getting a lot better! I've got a few more small comments - but then I think we're golden 👌

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: Yeah, might be overkill here.
changed:
https://github.com/cognitedata/reveal/pull/5254/commits/f0692ce4845f1e1b3a62aad2593a50796b3b2c4f

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: added execute parallel. Wdyt?
https://github.com/cognitedata/reveal/pull/5254/commits/f0692ce4845f1e1b3a62aad2593a50796b3b2c4f

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: done
https://github.com/cognitedata/reveal/pull/5254/commits/f0692ce4845f1e1b3a62aad2593a50796b3b2c4f

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -37,23 +39,28 @@ export function useGetDMConnectionWithNodeFromHybridMappingsQuery(
       if (dmNodesInspect.items.length === 0) {
         return EMPTY_ARRAY;
       }
-      const dataResult = await Promise.all(
-        models.flatMap(async ({ modelId, revisionId }) => {
+
+      const dataResult = await executeParallel(
+        models.flatMap(({ modelId, revisionId }) => async () => {
```
haakonflatval-cognite: I think this `flatMap` can be a `map`

haakonflatval-cognite (approved):
LGTM :+1: A small comment, but nothing critical

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: done here:
https://github.com/cognitedata/reveal/pull/5254/commits/8e61d2a874e67679de77259e86e291daeb26c918

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -63,9 +65,9 @@ export function RuleBasedOutputsSelector({
   }, [assetMappings]);
 
   const {
-    data: mappedAssets,
-    isLoading: isAssetMappedLoading,
-    isFetched: isAssetMappingsFetched
+    data: allClassicAssetConnections,
+    isLoading: isAllClassicAssetConnectionsLoading,
+    isFetched: isAllClassicAssetConnectionsFetched
   } = useAssetsByIdsQuery(assetIdsFromMapped);
```
haakonflatval-cognite, line 76: Not critical for this PR - but is there a reason we need to keep track of both `isLoading` and `isFetched` here?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -118,11 +110,50 @@ export function RuleBasedOutputsSelector({
 
   const flatAssetsMappingsListPerModel = useCreateAssetMappingsMapPerModel(models, assetMappings);
 
+  const uniqueRuleSetKey = useMemo(() => {
+    return (generateUniqueRuleSetKey({
+      ruleSet,
```
haakonflatval-cognite: I'm wondering if it simplifies things to just get rid of this `useMemo` and compute the `uniqueRuleSetKey` inside the `useEffect` where it's used below?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -195,3 +237,57 @@ async function initializeRuleBasedOutputs({
...
+    assetMappings === undefined
+  )
+    return false;
+  if (allFdmConnections.length > 0 && fdmMappings === undefined) return false;
+  if (models === undefined || models.length === 0) return false;
+  if (timeseriesExternalIds.length > 0 && isLoadingAssetIdsAndTimeseriesData) return false;
+  if (ruleSet === undefined) return false;
+  if (!allDataLoaded) return false;
+
+  return true;
```
haakonflatval-cognite: Would this be easier to read if we could put these conditions into variables with explaining names? :thinking:

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -149,20 +180,31 @@ export function RuleBasedOutputsSelector({
...
+      onAllMappingsFetched(true);
     };
-    void ruleBasedInitilization();
-    onAllMappingsFetched(true);
-    if (onRuleSetChanged !== undefined) onRuleSetChanged(ruleSetStylingCache.get(ruleSet.id));
+
+    if (ruleSetStylingCache.has(uniqueRuleSetKey)) {
+      const cachedStylings = ruleSetStylingCache.get(uniqueRuleSetKey);
+      if (cachedStylings && onRuleSetChanged) {
+        onRuleSetChanged(cachedStylings);
```
haakonflatval-cognite: Nitpick - I think this could be done a little simpler by doing `ruleSetStylingCache.get(uniqueRuleSetKey)` first (instead of `ruleSetStylingCache.has`), and then checking if the result is `undefined`. If you use `onRuleSetChanged?.(cachedStylings)` you get rid of the second `if`-statement too

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -195,3 +237,57 @@ async function initializeRuleBasedOutputs({
...
+  ruleSet,
+  assetMappings,
+  fdmMappings,
+  contextualizedAssetNodes,
+  assetIdsWithTimeseriesData,
+  timeseriesExternalIds
+}: {
+  ruleSet: RuleOutputSet;
+  assetMappings: ModelWithAssetMappings[] | undefined;
+  fdmMappings: FdmInstanceNodeWithConnectionAndProperties[] | undefined;
```
haakonflatval-cognite: I think we should just have these fields as function arguments directly, it will save a few lines of code.

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -195,3 +237,57 @@ async function initializeRuleBasedOutputs({
...
+  timeseriesExternalIds: string[];
+}): string {
+  return `${ruleSet.id}-${assetMappings?.length}-${fdmMappings?.length}-${contextualizedAssetNodes.length}-${assetIdsWithTimeseriesData?.assetIdsWithTimeseries.length}-${timeseriesExternalIds.length}`;
+}
+
+function isReadyToInitializeRules({
+  ruleSet,
+  assetMappings,
+  fdmMappings,
+  timeseriesExternalIds,
```
haakonflatval-cognite: Same here

haakonflatval-cognite (changes requested):
Nice! A few nitpicks

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: the fetched is being used to update to call the `setAllContextualizedAssets` in order to generate all the metadata in lower case here `useConvertAssetMetadatasToLowerCase`

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: I usually do prefer using arguments object when we have too many arguments, but it is fine to change it here.
https://github.com/cognitedata/reveal/pull/5254/commits/303bcdc94ec7b26364c6ec0ffc2e0fa46ff1fe1b

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: https://github.com/cognitedata/reveal/pull/5254/commits/303bcdc94ec7b26364c6ec0ffc2e0fa46ff1fe1b

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: yes.. sure. Good option
https://github.com/cognitedata/reveal/pull/5254/commits/303bcdc94ec7b26364c6ec0ffc2e0fa46ff1fe1b

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: sure!
https://github.com/cognitedata/reveal/pull/5254/commits/303bcdc94ec7b26364c6ec0ffc2e0fa46ff1fe1b

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: Hmm yes. It is also a good option indeed.

https://github.com/cognitedata/reveal/pull/5254/commits/303bcdc94ec7b26364c6ec0ffc2e0fa46ff1fe1b

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -153,10 +135,19 @@ export function RuleBasedOutputsSelector({
...
     )
       return;
 
+    const uniqueRuleSetKey = generateUniqueRuleSetKey(
+      ruleSet,
+      assetMappings,
+      fdmMappings,
+      contextualizedAssetNodes,
+      assetIdsWithTimeseriesData,
+      timeseriesExternalIds
```
haakonflatval-cognite, line 150: I don't see this variable (`timeseriesExternalIds`) being used inside this function (and it's also not in the dependency array), should we leave it out, or do you want to keep it?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -240,56 +221,46 @@ async function initializeRuleBasedOutputs({
...
+  fdmMappings: FdmInstanceNodeWithConnectionAndProperties[] | undefined,
+  timeseriesExternalIds: string[],
+  isLoadingAssetIdsAndTimeseriesData: boolean,
+  allDataLoaded: boolean,
+  allClassicAssetConnections: Asset[] | undefined,
+  allFdmConnections: FdmConnectionWithNode[],
+  models: Array<CogniteModel<DataSourceType>> | undefined
+): boolean {
+  const modelsReady = models === undefined || models.length === 0;
+  if (!modelsReady) return false;
```
haakonflatval-cognite: Nice! I was thinking that we could just put all of the extracted variables into one big `if`-statement to make the code less noisy, but not critical

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: I would like to keep it because is used to reinforce whenever if it has timeseries but the assets with timeseries were not loaded yet.

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: hmm yeah. 
Done here:
https://github.com/cognitedata/reveal/pull/5254/commits/4c09145645a00614d55da8429a65233a0344f234

haakonflatval-cognite (dismissed):
LGTM 👍

haakonflatval-cognite (dismissed):
LGTM again  👍

danpriori:
> @danpriori This is not targeting master/main, so it won't necessarily be picked up by risk review. I also can't see another PR for the branch you are targeting? What's the plan to get this mergeable?

We decided to split this PR and move it partially to an initial PR and update this target

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -181,3 +234,49 @@ async function initializeRuleBasedOutputs({
...
+    ruleSet === undefined ||
+    !allDataLoaded ||
+    classicAssetMappingsNotReady ||
+    fdmAssetMappingsNotReady ||
+    timeseriesNotReady
+  )
+    return false;
+
+  return true;
+}
```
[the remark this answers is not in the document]
danpriori: The comments are not exactly true.
`classicAssetMappingsNotReady` and `fdmAssetMappingsNotReady` are not directly covered by the `allDataLoaded` because it checks whether the mapping data has been loaded, but not necessarily the correlation between the loaded mappings and the possible existing connections, which can be asynchronous.

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
```diff
@@ -145,27 +144,27 @@ export function useAll3dDirectConnectionsWithProperties(
...
+                instanceType: 'node' as const,
+                ...connectionFound,
+                ...itemsData
+              };
+            });
+          })
+          .flat()
+          .filter(isDefined);
+        return dataFound;
+      });
```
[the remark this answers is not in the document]
danpriori: https://github.com/cognitedata/reveal/pull/5254/commits/200ae46f66dbeb2fc74ea36fad8704f6d2595298

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
```diff
@@ -0,0 +1,367 @@
...
+    name: ''
+  };
+
+  const node3d2: Node3D = {
+    id: 200,
+    treeIndex: 200,
+    subtreeSize: 1,
+    parentId: 0,
+    depth: 0,
+    name: ''
```
haakonflatval-cognite: Use `createCadNodeMock` here and above

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
```diff
@@ -0,0 +1,367 @@
...
+  const mockClassicAssetConnections: Asset[] = [
+    {
+      id: 1,
+      externalId: 'asset-1',
+      name: 'Asset 1',
+      description: '',
+      lastUpdatedTime: new Date(),
+      createdTime: new Date(),
+      rootId: 0
+    },
```
haakonflatval-cognite: Use `createAssetMock` here

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
```diff
@@ -0,0 +1,367 @@
...
+      cadNode: node3d1
+    },
+    {
+      connection: mockFdmCadConnections[1],
+      cadNode: node3d2
+    }
+  ];
+  const mockFdmMappedEquipmentEdges: ModelRevisionToConnectionMap = new Map([
+    ['1/10', [{ connection: fdmConnections[0].connection, cadNode: fdmConnections[0].cadNode }]],
+    ['2/20', [{ connection: fdmConnections[1].connection, cadNode: fdmConnections[1].cadNode }]]
```
haakonflatval-cognite: Where does `1/10` come from? Can we use the relevant `modelId`/`revisionId` directly here, so it's obvious where they are from?

Also - is there a reason the lists cannot be just `[fdmConnections[0]]` and `[fdmConnections[1]]` respectively?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
```diff
@@ -0,0 +1,367 @@
...
+  ]);
+  const mockDmConnectionWithNodeFromHybridDataList: FdmConnectionWithNode[] = [
+    { connection: fdmConnections[1].connection, cadNode: fdmConnections[1].cadNode }
```
haakonflatval-cognite: Here too, can we just write `[fdmConnections[1]]`?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
```diff
@@ -0,0 +1,367 @@
...
+  ];
+  const mockFdmMappings: FdmInstanceNodeWithConnectionAndProperties[] = [
+    {
+      instanceType: 'node',
+      items: [],
+      typing: {},
```
haakonflatval-cognite: I would consider using a util function to create these mocks, but not critical

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
```diff
@@ -0,0 +1,367 @@
...
+    vi.clearAllMocks();
+
+    defaultDependencies.use3dModels.mockReturnValue([cadMock]);
+
+    const mockuseAssetMappedNodesForRevisionsResult = {
+      data: [mockAssetMappings, mockHybridAssetMappings],
+      isLoading: false
+    };
+    defaultDependencies.useAssetMappedNodesForRevisions.mockReturnValue(
+      mockuseAssetMappedNodesForRevisionsResult as UseQueryResult<ModelWithAssetMappings[]>
```
haakonflatval-cognite: Rather than casting, it would be nicer if we modified the type of `useAssetMappedNodesForRevisions` in the component context to only have the necessary fields, e.g. `useAssetMappedNodesForRevisions: (...) => { data: ModelWithAssetMappings[] | undefined, isLoading: boolean }`. Same for the casts below

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
```diff
@@ -145,27 +144,28 @@ export function useAll3dDirectConnectionsWithProperties(
...
-
-          itemsData.items.every((itemData) => {
-            const fdmKey = createFdmKey(itemData);
-            if (connectionWithNodeAndViewMap.has(fdmKey)) {
-              connectionFound = connectionWithNodeAndViewMap.get(fdmKey);
-              return false;
-            }
-            return true;
-          });
+      const instanceWithData = ((map) =>
```
haakonflatval-cognite: I don't understand this declaration - are we making an anonymous function and calling it directly on a clone of `connectionWithNodeAndViewMap`? Can't we just make the clone first and then do what's inside this function as normal?

Also - although the cloning makes it less problematic, I think it's a bit strange having to delete the entries as we traverse the map here. Do we really need to do it?

danpriori: This IIFE is a Gemini suggestion, which is I think does make sense (in rare cases, but anyway):

https://github.com/cognitedata/reveal/pull/5254#discussion_r2278564860

About the deletion, I think you're right. It is not necessary here.

danpriori: removed the deletion and the IIFE as there is no way for mutation here anymore.
https://github.com/cognitedata/reveal/pull/5254/commits/f595c34af86102e1af791ccd674594a6ac7ffcfd

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
danpriori: done
https://github.com/cognitedata/reveal/pull/5254/commits/f595c34af86102e1af791ccd674594a6ac7ffcfd

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
danpriori: https://github.com/cognitedata/reveal/pull/5254/commits/f595c34af86102e1af791ccd674594a6ac7ffcfd

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
danpriori: yep
https://github.com/cognitedata/reveal/pull/5254/commits/f595c34af86102e1af791ccd674594a6ac7ffcfd

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
danpriori: yes. It is just a random id. Also, it can use the same model and revision Ids for both. 
Changed here:
https://github.com/cognitedata/reveal/pull/5254/commits/f595c34af86102e1af791ccd674594a6ac7ffcfd

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
danpriori: https://github.com/cognitedata/reveal/pull/5254/commits/f595c34af86102e1af791ccd674594a6ac7ffcfd

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
danpriori: https://github.com/cognitedata/reveal/pull/5254/commits/f595c34af86102e1af791ccd674594a6ac7ffcfd

haakonflatval-cognite (dismissed):
LGTM 👍

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.context.ts
```diff
@@ -0,0 +1,58 @@
...
+import { type CadModelOptions } from '../Reveal3DResources';
+import { type IdEither } from '@cognite/sdk';
+import { type DmCadAssetMapping } from '../CacheProvider/cad/assetMappingTypes';
+import { type FdmConnectionWithNode } from '../CacheProvider/types';
+
+export type RuleBasedOutputsSelectorDependencies = {
+  use3dModels: typeof use3dModels;
+  useAssetsByIdsQuery: (
+    ids: IdEither[]
+  ) => Pick<ReturnType<typeof useAssetsByIdsQuery>, 'data' | 'isLoading' | 'isFetched'>;
```
gregfagan, line 19: Suggestion: I wrote a type utility that you might consider copying to Reveal if you're going to be doing this a lot:

https://github.com/cognitedata/fusion/blob/master/libs/industry-canvas/src/lib/utils/types.ts#L1-L37

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.context.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const RuleBasedOutputsSelectorContext = createContext<RuleBasedOutputsSelectorDependencies>({
+  use3dModels,
+  useAssetsByIdsQuery,
+  useAssetMappedNodesForRevisions,
+  useMappedEdgesForRevisions,
+  useAll3dDirectConnectionsWithProperties,
+  useGetDMConnectionWithNodeFromHybridMappingsQuery,
+  generateRuleBasedOutputs
+});
```
gregfagan:
```suggestion
export const RuleBasedOutputsSelectorContext = createContext<RuleBasedOutputsSelectorDependencies>(
  defaultRuleBasedOutputsSelectorDependencies
);
```

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
```diff
@@ -145,27 +144,26 @@ export function useAll3dDirectConnectionsWithProperties(
...
             const fdmKey = createFdmKey(itemData);
-            if (connectionWithNodeAndViewMap.has(fdmKey)) {
-              connectionFound = connectionWithNodeAndViewMap.get(fdmKey);
-              return false;
-            }
-            return true;
-          });
 
-          if (connectionFound === undefined) return [];
+            if (!connectionWithNodeAndViewMap.has(fdmKey)) return undefined;
```
gregfagan: Suggestion:

Since we're within a `flatMap` iteration, returning `[]` will flatten to nothing, meaning the calls to `flat` and `filter(isDefined)` afterwards would be redundant (isn't `flat` already redundant?).

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
gregfagan, line 1: This looks like a fairly large and complicated hook that doesn't appear to have tests.

The change here is not cosmetic, since it now supports multiple connections per item.

Do you think we should add some tests for this change?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -118,58 +164,65 @@ export function RuleBasedOutputsSelector({
...
-  timeseriesDatapoints
-}: {
-  assetMappings: ClassicCadAssetMapping[];
-  fdmMappings: FdmInstanceNodeWithConnectionAndProperties[];
-  contextualizedAssetNodes: Asset[];
-  ruleSet: RuleOutputSet;
-  assetIdsAndTimeseries: AssetIdsAndTimeseries[];
-  timeseriesDatapoints: Datapoints[] | undefined;
-}): Promise<AllMappingStylingGroupAndStyleIndex[]> {
+async function initializeRuleBasedOutputs(
```
gregfagan: Can't this entire function just be removed? It appears to be entirely equivalent to just calling `generateRuleBasedOutputs` inline on 167.

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
```diff
@@ -181,3 +234,49 @@ async function initializeRuleBasedOutputs({
...
+
+function generateUniqueRuleSetKey(
+  ruleSet: RuleOutputSet,
+  assetMappings: ModelWithAssetMappings[] | undefined,
+  fdmMappings: FdmInstanceNodeWithConnectionAndProperties[] | undefined,
+  contextualizedAssetNodes: Asset[],
+  assetIdsWithTimeseriesData: AssetIdsAndTimeseriesData | undefined,
+  timeseriesExternalIds: string[]
+): string {
+  return `${ruleSet.id}-${assetMappings?.length}-${fdmMappings?.length}-${contextualizedAssetNodes.length}-${assetIdsWithTimeseriesData?.assetIdsWithTimeseries.length}-${timeseriesExternalIds.length}`;
```
gregfagan: are there scenarios where the array length wouldn't change but the contents would, but this key (using `.length`) wouldn't bust the cache?

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
gregfagan, line 1: I don't see major issues with the changes, but the component itself has some pretty fundamental issues:

1. It doesn't render anything, so it should probably be a hook instead of a component
2. [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect)

@gemini-code-assist can you make suggestions based on the linked page?

@danpriori no need to refactor for this PR, but I think the useEffect situation is one y'all should be watching out for in code reviews.

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
```diff
@@ -0,0 +1,340 @@
...
+            <RuleBasedOutputsSelectorContext.Provider value={defaultDependencies}>
+              {children}
+            </RuleBasedOutputsSelectorContext.Provider>
+          </ViewerContextProvider>
+        </QueryClientProvider>
+      </SDKProvider>
+    );
+  };
+
+  beforeEach(() => {
```
gregfagan, line 180: just a minor suggestion to think about, but for all the same reasons they're good in production code, I like to write tests as pure functions that don't need to manipulate shared state and reset it.

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
danpriori: Yes, sure. I can add tests. The only concern for me would be that since this file doesn't have a test yet, it will be necessary to create all the scaffolding in addition to the new tests and it will increase the code lines. If we can't keep it in this same PR, we will have to create a new PR just with this hook and point this PR to the hook's PR and restart all the review cycle, and we are short on time 🙈 😄 . I was thinking to add tests only targeting the changes and open a ticket to cover also the other not-changes cases to keep it on track and keep the tests (for this case) in here in this PR to avoid exploding the lines and speed up the process, if it is fine by you. Does it make sense?
What do you suggest?

For now, I added the tests in the PR (along with the changes made for this hook): https://github.com/cognitedata/reveal/pull/5273

I can target this current PR to this PR specific for the all3d...' hook (+ tests), if make sense.

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: Yes. It is in our plans for refactoring this component here:
https://cognitedata.atlassian.net/browse/BND3D-5919

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.context.ts
danpriori: Thanks for the suggestion. I will bring this suggestion up to the team :)

react-components/src/query/useAll3dDirectConnectionsWithProperties.ts
danpriori: Nice catch!
It's actually a three-dimensional array that needs to be flattened twice. Perhaps to simplify things, I'll replace it with flatMap in the second dimension and return [] to discard it when flattening is done.

https://github.com/cognitedata/reveal/pull/5254/commits/c43ac6afc7d5bd93faceeb684741f0db554d0ff8

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.context.ts
danpriori: https://github.com/cognitedata/reveal/pull/5254/commits/b64bbefbf6c324da171f8ccf9f954c9c56a9cc27

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.test.tsx
danpriori: Yeah make sense. We in the team tried to follow a pattern to avoid replicating dependency mocks if they dont need to change the returning value cross tests, so we put those there to reduce the amount of lines. 
If we go to the same direction as you suggesting, then we have to discuss internally and decide together. But thanks for your suggestion and bring this up!

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: Yes. I was thinking to change that. It is really rare but in theory can happen. I added a simple hash creator as an utility function and replace this old generator in order to be more precise based on the elements for all the arrays. It is safer in cases the length is the same but the objects are different. Wdyt?
https://github.com/cognitedata/reveal/pull/5254/commits/f7b3c001421d0c7ae1e0830077c87fcd47c15990

react-components/src/components/RuleBasedOutputs/RuleBasedOutputsSelector.tsx
danpriori: Yep. 
https://github.com/cognitedata/reveal/commit/f7b3c001421d0c7ae1e0830077c87fcd47c15990

PR #5257, opened 2025-08-06
chore(react-components): bump @tanstack/react-query to 5.84.1

haakonflatval-cognite:
#### Type of change

Chore 

#### Jira ticket :blue_book:

## Description :pencil:

Bumps react-query to 5.84.1. This entails an improvement in type checking, which revealed a few issues in our code base. I've fixed most of them 

### DIFF
--- a/react-components/src/components/Architecture/pointsOfInterest/useCommentsForPoiQuery.ts
+++ b/react-components/src/components/Architecture/pointsOfInterest/useCommentsForPoiQuery.ts
@@ -6,14 +6,14 @@ import { usePoiDomainObject } from './usePoiDomainObject';
 
 export function useCommentsForPoiQuery(
   poi: PointOfInterest<unknown> | undefined
-): UseQueryResult<CommentProperties[]> {
+): UseQueryResult<CommentProperties[] | null> {
   const domainObject = usePoiDomainObject();
 
   return useQuery({
     queryKey: queryKeys.poiCommentsById(poi?.id),
     queryFn: async () => {
       if (poi === undefined || domainObject === undefined) {
-        return undefined;
+        return null;
       }
       return [...(await domainObject.getCommentsForPoi(poi))];
     },
--- a/react-components/src/query/useHybridMappingsForAssetInstances.ts
+++ b/react-components/src/query/useHybridMappingsForAssetInstances.ts
@@ -5,6 +5,8 @@ import { type ThreeDModelFdmMappings } from '../hooks';
 import { queryKeys } from '../utilities/queryKeys';
 import { UseHybridMappingsForAssetInstancesContext } from './useHybridMappingsForAssetInstances.context';
 import { useContext } from 'react';
+import { FdmKey } from '../components/CacheProvider/types';
+import { Node3D } from '@cognite/sdk';
 
 export const useHybridMappingsForAssetInstances = (
   models: AddModelOptions[],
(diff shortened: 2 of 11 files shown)


react-components/src/components/CacheProvider/cad/CadInstanceMappingsCache.ts
```diff
@@ -12,5 +12,5 @@ export type CadInstanceMappingsCache = {
     instances: InstanceId[],
     models: ModelRevisionId[]
   ) => Promise<CadModelMappingsWithNodes>;
-  getAllModelMappings: (models: ModelRevisionId[]) => Promise<CadModelTreeIndexMappings>;
```
nilscognite, line 15: Remove CadModelTreeIndexMappings , not in use anymore

haakonflatval-cognite: Ah, nice catch :raised_hands: Thanks

react-components/src/query/useHybridMappingsForAssetInstances.ts
```diff
@@ -26,7 +28,11 @@ export const useHybridMappingsForAssetInstances = (
           assetInstanceIds
         );
 
-        return { modelId: model.modelId, revisionId: model.revisionId, mappings };
+        return {
+          modelId: model.modelId,
+          revisionId: model.revisionId,
+          mappings: mappings as Map<FdmKey, Node3D[]>
```
BugGambit: Can we avoid `as` here?

react-components/src/components/Architecture/pointsOfInterest/useCommentsForPoiQuery.ts
```diff
@@ -6,14 +6,14 @@ import { usePoiDomainObject } from './usePoiDomainObject';
 
 export function useCommentsForPoiQuery(
   poi: PointOfInterest<unknown> | undefined
-): UseQueryResult<CommentProperties[]> {
+): UseQueryResult<CommentProperties[] | null> {
```
BugGambit, line 9: FYI, you would likely not return `UseQueryResult` here as `UseQueryResult` contains a wide API surface, and you would likely control which part of the response you want to expose. E.g.:
```ts
const { data: models, isFetching, error } = useQuery(...);
// ...
return { models, isFetching, error); // Here you control the API surface returned.
```

haakonflatval-cognite: Yes, I see :thinking: We return `UseQueryResult` from very many functions, so this sounds like it should be done in a separate PR. I'll create a ticket for it

react-components/src/query/useHybridMappingsForAssetInstances.ts
haakonflatval-cognite: I added some code to filter out non-`FdmKey` keys from the Map (although they cannot appear in the current implementation, but nice for resilience to change), and added a utility: https://github.com/cognitedata/reveal/pull/5257/commits/21b28b213ccdbb2fc34a5f723aaa454a3aee346a Perhaps a little overengineered, but I think the utility is worthwhile to have for future use too

react-components/src/components/Architecture/pointsOfInterest/useCommentsForPoiQuery.ts
haakonflatval-cognite: Created the ticket here: https://cognitedata.atlassian.net/browse/BND3D-5935

BugGambit:
:unicorn:

nilscognite (approved):

PR #5259, opened 2025-08-07
fix(react-components): generalize `getInstancesFromClick`

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

Fixes point 2 in https://cognitedata.atlassian.net/browse/BND3D-5836

## Description :pencil:

Ensures that we return all relevant IDs, including hybrid ones, from `getInstancesFromClick`. This is used in the PoI-implementation, primarily.

In Fusion

- Open scene with hybrid-mapped instances
- Click "create PoI" in toolbar
- Click on a Hybrid-mapped CAD node
- Fill out dialog and create PoI
- Observe that the POI is connected to the instance


danpriori (dismissed):
👍

react-components/src/components/CacheProvider/idAndKeyTranslation.test.ts
```diff
@@ -85,6 +86,16 @@ describe('idAndKeyTranslation', () => {
     });
   });
 
+  describe(instanceIdToInstanceReference, () => {
```
notrixbe: question: 

should this not be?
```
instanceIdToInstanceReference.name
```

react-components/src/components/CacheProvider/idAndKeyTranslation.ts
```diff
@@ -42,6 +42,14 @@ export function createInstanceKey(id: InstanceId): InstanceKey {
   }
 }
 
+export function instanceIdToInstanceReference(id: InstanceId): InstanceReference {
```
notrixbe, line 45: I was a bit confused about the naming of the function. 
Because it sometimes returned an ID and other times an object. 

Don't know if we can make it clear that it is sometimes id:number and sometimes object

react-components/src/utilities/getInstancesFromClick.ts
```diff
@@ -159,7 +160,7 @@ async function getAssetMappingPromise(
       ancestors
     );
 
-  return nodeAssetResult.mappings
-    .filter(isClassicCadAssetMapping)
-    .map((mapping) => ({ id: mapping.assetId }));
+  return nodeAssetResult.mappings.map((mapping) =>
+    instanceIdToInstanceReference(getMappingInstanceId(mapping))
```
notrixbe, line 164:
```
  if (isClassicCadAssetMappingInstance(mappingInstance)) {
    return mappingInstance.assetId;
  }

  if (isDmCadAssetMappingInstance(mappingInstance)) {
    return mappingInstance.instanceId;
  }
```

Do you expect you will ever come in MappingInstance? 

Because from what I see you map of those types
```
[{id: 123 } or { externalId: 'externalId', space: 'space' }]
```

Could it be that you also need ```assertNever```
Or did I miss something?

notrixbe (commented):
Reviewed: 

have added some questions.
let me know if i can help !

react-components/src/components/CacheProvider/idAndKeyTranslation.test.ts
haakonflatval-cognite: Ah, yes, thanks :sweat_smile:

react-components/src/components/CacheProvider/idAndKeyTranslation.ts
haakonflatval-cognite: I agree that the naming here isn't great - but `InstanceId` is a `CogniteInternalId | DmsID` and `InstanceReference` is `IdEither | DmsID`. We prefer `InstanceId` in most cases, but some APIs operate with `InstanceReference`, which is why we have the separation. But given this information, I'm not sure I know what to call this function other than `instanceIdToInstanceReference`

react-components/src/utilities/getInstancesFromClick.ts
haakonflatval-cognite: Not sure I understand this comment - there is an `assertNever` inside `getMappingInstanceId` - but do you mean I should have one somewhere else too?

notrixbe: @haakonflatval-cognite 
could it be that it never gets in the isClassicCadAssetMappingInstance or isDmCadAssetMappingInstance?

haakonflatval-cognite: Right now, no - the type is either a classic Classic or DM asset mapping

react-components/src/components/CacheProvider/idAndKeyTranslation.ts
notrixbe: Thanks for answer

PR #5262, opened 2025-08-09
refactor(react-components): Remove static lass OnIdleHandler and use signal instead

nilscognite:
#### Type of change
Refactor 

## Description :pencil:

Cool stuff. The signal system can use to a defferedUpdate. In this case it updates all commands.

By unit test.


react-components/src/architecture/base/renderTarget/CommandsController.ts
```diff
@@ -197,16 +200,31 @@ export class CommandsController extends PointerEvents {
     return true;
   }
 
-  public update(): void {
+  /**
+   * Updates all registered commands by invoking their `update` method.
+   * Do not call this directly, (except in test code), instead use the deferredUpdate(), which will be called
+   * this function only once, even if called multiple times.
+   */
+  public update = (): void => {
```
haakonflatval-cognite: Should we consider leaving this method private if it's not supposed to be called directly from the outside?

haakonflatval-cognite: (Also - is there a reason it's declared using `update = (): void => { ...` instead of `update(): void { ...` ? 😅

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
```diff
@@ -246,6 +246,28 @@ describe(CommandsController.name, () => {
...
+      controller.deferredUpdate();
+      controller.deferredUpdate();
+
+      expect(updateTester.calledTimes).toBe(0);
+
+      // This simulate OnIdle
+      vi.advanceTimersByTime(1);
+
+      expect(updateTester.calledTimes).toBe(1);
+      vi.useRealTimers();
```
haakonflatval-cognite: Would it be tidier to put `vi.useFakeTimers` and `vi.useRealTimers` into `beforeEach` and `afterEach` as in the other test?

react-components/src/architecture/base/renderTarget/CommandsController.ts
nilscognite: 1. Done. Not even need to be private. Added this in the constructor.
2. I pass this function as an argument. Then it has to be a lambda func

react-components/src/architecture/base/renderTarget/CommandsController.test.ts
nilscognite: Done

react-components/src/architecture/base/renderTarget/CommandsController.ts
nilscognite: Fixes here: 918652a08cc943461021709772d1e72ec45bc523

PR #5264, opened 2025-08-11
refactor(react-components): Fix unit test if language is norwegian or similar

nilscognite:
#### Type of change
Refactor 

## Description :pencil:

Fore to use` en-US` as language for test, to match the test exceptions.

Try to run on your home computer of it is Norwegian.


christjt (approved):
risk review ok

PR #5265, opened 2025-08-11
refactor(react-components): Make signal in 360 image settings + Use domain object instead of Reveal models in the architecture code

nilscognite:
#### Type of change
Refactor 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5908

## Description :pencil:

* Use the same pattern on  360 images as for point Cloud.
* Updated some Mock mode to have the visual setting methods and property.
* Simplify `RevealSettingsController `so more code can be reused.
* Added unit test for all 360 settings commands
* Did the same thing for Cad, but here wasn't it any settings command. So just a few lines change in the `CadDividerCommand`, and made unit test for it. 
* Removed unnecessary functions in `RevealRenderTarget`

No real UX changes.

By unit tests


react-components/tests/tests-utilities/fixtures/image360.ts
```diff
@@ -52,34 +58,70 @@ export function createImage360ClassicMock(parameters?: { visible?: boolean }): I
...
+      // Get and set ImagesOpacity
+      .setup((p) => p.getImagesOpacity)
+      .callback(() => {
+        return () => imagesOpacity;
+      })
+      .setup((p) => p.setImagesOpacity)
+      .returns((value: number) => {
+        imagesOpacity = value;
+      })
+      .object()
```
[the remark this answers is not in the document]
nilscognite: Yes it will for sure, but the existing pattern in over-engineered. I want it to be simple!

react-components/src/architecture/base/concreteCommands/image360Collection/Image360ImagesDividerCommand.ts
```diff
@@ -0,0 +1,8 @@
+import { Image360CollectionDomainObject } from '../../../concrete/reveal/Image360Collection/Image360CollectionDomainObject';
+import { DividerCommand } from '../../commands/DividerCommand';
+
+export class Image360ImagesDividerCommand extends DividerCommand {
```
pramod-cog: why the renamed to `Image360ImagesDividerCommand` from `Image360CollectionDividerCommand`, isn't this collection?

react-components/src/architecture/concrete/reveal/RevealSettingsController.ts
```diff
@@ -23,20 +24,25 @@ export class RevealSettingsController {
...
-  private _pointShapeInitialized = false;
-  private _pointColorTypeInitialized = false;
+  // Settings for the 360 images
+  private readonly _isIconsVisible = signal(false);
+  private readonly _isOccludedIconsVisible = signal(false);
+  private readonly _iconsOpacity = signal(0);
+  private readonly _imagesOpacity = signal(0);
+
+  // If the signal is the set, the signal is initialized with a value from the first domain object of that type.
+  private readonly _isInitialized = new Set<Signal<any>>();
```
pramod-cog, line 45: Cannot have `any`, need to have specific type

react-components/tests/tests-utilities/fixtures/image360.ts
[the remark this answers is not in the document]
pramod-cog: We should be using either the existing pattern or new pattern you added, having both is confusing and not maintainable.

react-components/src/architecture/concrete/reveal/RevealSettingsController.ts
nilscognite: Here I need it.

nilscognite: If you don't like., please suggest. **You have more experience in TypeScript than I**. I just want to get the pointer of a signal, but it is a temple class.

react-components/tests/tests-utilities/fixtures/image360.ts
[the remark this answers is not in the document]
nilscognite: I see. But I will not use the existing one. The existing seems overengineered and confuses me. I stick to a simple solution that works in test.

react-components/src/architecture/base/concreteCommands/image360Collection/Image360ImagesDividerCommand.ts
nilscognite: Sorry. Renamed back again

react-components/tests/tests-utilities/fixtures/image360.ts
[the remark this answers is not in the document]
nilscognite: The thing is that in the  previous code, there are a lot `MockFunction`, for set and get that is used to check whether the the function is called or not. This make more complexity that just using the doing it in my way.

For instance:

`model.isVisible = true;`

In the code, normally people are checking that the `isVisible `code is called. I check the `isVisible `is in fact changed from false to true. In way way, the test are more realistic, since I actually charge the visible state. I can then use it for more advanced tests.

I am in fact very surprise that way some of you do testing like this. I would never approved a PR like that.

PR #5269, opened 2025-08-14
feat(react-components): add the hook to get the dm connection with node from hybrid mappings by query hook and other improvements

danpriori:
#### Type of change

Feat  

#### Jira ticket :blue_book:

This add the auxiliary hook to get the dm connection with node data from hybrid mappings used by the hybrid support in rule base coloring in this PR: https://github.com/cognitedata/reveal/pull/5254

It is also including one improvement on querykey related to fetch data for the rule base coloring.

### DIFF
--- a/react-components/src/components/CacheProvider/idAndKeyTranslation.test.ts
+++ b/react-components/src/components/CacheProvider/idAndKeyTranslation.test.ts
@@ -7,6 +7,7 @@ import {
   createModelInstanceIdKey,
   revisionKeyToIds,
   createModelNodeIdKey,
+  instanceIdToInstanceReference,
   fdmKeyToId,
   isFdmKey
 } from './idAndKeyTranslation';
@@ -82,6 +83,16 @@ describe('idAndKeyTranslation', () => {
       const result = createInstanceKey(INSTANCE);
       expect(result).toBeTypeOf('string');
       expect(result).toBe(createFdmKey(INSTANCE));
+    });
+  });
+
+  describe(instanceIdToInstanceReference.name, () => {
+    test('correctly translates classic ID', () => {
+      expect(instanceIdToInstanceReference(ASSET_ID)).toEqual({ id: ASSET_ID });
+    });
+
+    test('correctly preserves DM ID', () => {
+      expect(instanceIdToInstanceReference(INSTANCE)).toEqual(INSTANCE);
     });
   });
 
--- a/react-components/src/components/CacheProvider/idAndKeyTranslation.ts
+++ b/react-components/src/components/CacheProvider/idAndKeyTranslation.ts
@@ -1,5 +1,5 @@
 import { type DmsUniqueIdentifier } from '../../data-providers/FdmSDK';
-import type { InstanceId, InstanceKey } from '../../utilities/instanceIds/types';
+import type { InstanceId, InstanceKey, InstanceReference } from '../../utilities/instanceIds/types';
 import { isClassicInstanceId } from '../../utilities/instanceIds/typeGuards';
 import {
   type FdmKey,
@@ -39,6 +39,14 @@ export function createInstanceKey(id: InstanceId): InstanceKey {
     return id;
   } else {
     return createFdmKey(id);
+  }
+}
+
+export function instanceIdToInstanceReference(id: InstanceId): InstanceReference {
+  if (isClassicInstanceId(id)) {
+    return { id };
+  } else {
+    return id;
   }
 }
 
--- a/react-components/src/utilities/getInstancesFromClick.ts
+++ b/react-components/src/utilities/getInstancesFromClick.ts
@@ -14,7 +14,8 @@ import { isDM3DModelIdentifier } from '../components/Reveal3DResources/typeGuard
 import { type RevealRenderTarget } from '../architecture';
 import { getInstanceReferenceFromImage360Annotation } from '../components/CacheProvider/utils';
 import { type InstanceReference, isIdEither } from './instanceIds';
-import { isClassicCadAssetMapping } from '../components/CacheProvider/cad/assetMappingTypes';
+import { getMappingInstanceId } from '../components/CacheProvider/cad/assetMappingTypes';
+import { instanceIdToInstanceReference } from '../components/CacheProvider/idAndKeyTranslation';
 
 export async function getInstancesFromClick(
   renderTarget: RevealRenderTarget,
@@ -159,7 +160,7 @@ async function getAssetMappingPromise(
       ancestors
     );
 
-  return nodeAssetResult.mappings
-    .filter(isClassicCadAssetMapping)
-    .map((mapping) => ({ id: mapping.assetId }));
+  return nodeAssetResult.mappings.map((mapping) =>
+    instanceIdToInstanceReference(getMappingInstanceId(mapping))
+  );
 }
--- a/react-components/yarn.lock
+++ b/react-components/yarn.lock
@@ -11784,9 +11784,9 @@ __metadata:
   linkType: hard
 
 "tmp@npm:^0.2.0":
-  version: 0.2.3
-  resolution: "tmp@npm:0.2.3"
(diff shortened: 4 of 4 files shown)


react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+  });
+});
+
+function createAssetMappingDataMapMock(
+  instanceIds: DmsUniqueIdentifier[],
+  node3D: Node3D[]
+): Map<InstanceKey, Node3D[]> {
+  const map = new Map<InstanceKey, Node3D[]>();
+  instanceIds.forEach((instanceId, index) => {
+    map.set(`${instanceId.space}/${instanceId.externalId}` as const, [node3D[index]]);
```
haakonflatval-cognite: Can we use `createFdmKey` here?

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+    [firstNode3DMock, secondNode3DMock]
+  );
+
+  const mockNoMatchedAssetMappingDataMap = createAssetMappingDataMapMock(
+    [
+      { space: 'test-space', externalId: 'other-instance-1' },
+      { space: 'test-space', externalId: 'other-instance-2' }
+    ],
+    [firstNode3DMock, secondNode3DMock]
+  );
```
haakonflatval-cognite: Seems like these are only used inside one test each - can they be moved into respective tests?

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,103 @@
...
+  );
+  const assetMappingCache = useClassicCadAssetMappingCache();
+  const fdmSDK = useFdmSdk();
+  const dmIds = useMemo(
+    () => nodeWithDmIdsFromHybridMappings.map((item) => item.instanceId).filter(isDefined),
+    [nodeWithDmIdsFromHybridMappings]
+  );
+
+  return useQuery({
+    queryKey: ['fdm-mappings', dmIds, models],
```
haakonflatval-cognite: Should we be using the centralized `queryKeys` here?

haakonflatval-cognite (commented):
Some comments, otherwise looks good 👍

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
danpriori: sure
https://github.com/cognitedata/reveal/pull/5269/commits/38c3a6e69331e62f355146cb7305ecd82ad3fe0b

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
danpriori: yep
https://github.com/cognitedata/reveal/pull/5269/commits/38c3a6e69331e62f355146cb7305ecd82ad3fe0b

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.test.tsx
danpriori: yeah
https://github.com/cognitedata/reveal/pull/5269/commits/38c3a6e69331e62f355146cb7305ecd82ad3fe0b

haakonflatval-cognite (approved):
LGTM 👍

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
```diff
@@ -0,0 +1,104 @@
...
+  );
+  const assetMappingCache = useClassicCadAssetMappingCache();
+  const fdmSDK = useFdmSdk();
+  const dmIds = useMemo(
+    () => nodeWithDmIdsFromHybridMappings.map((item) => item.instanceId).filter(isDefined),
+    [nodeWithDmIdsFromHybridMappings]
+  );
+
+  return useQuery({
+    queryKey: queryKeys.fdmConnectionWithNode(dmIds, models),
```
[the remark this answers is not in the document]
gregfagan: @danpriori is there any particular reason why we should expect these arrays to reorder between renders? Gemini's suggestion seems like unnecessary work if the scenario doesn't arise.

react-components/src/query/useAssetsAndTimeseriesLinkageDataQuery.ts
```diff
@@ -43,9 +43,11 @@ export function useAssetsAndTimeseriesLinkageDataQuery({
...
   return useQuery({
     queryKey: [
-      queryKeys.timeseriesLinkedToAssets(),
-      timeseriesExternalIds,
-      relationshipResourceTypes
+      queryKeys.assetsAndTimeseriesLinkageData(
+        timeseriesExternalIds,
+        relationshipResourceTypes,
+        assetNodes.map((asset) => asset.externalId).filter(isDefined)
+      )
```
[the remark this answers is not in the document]
gregfagan: Same as above, by default I would assume these arrays would be stable and thus not need to be sorted, will that be the case?

gregfagan:
This PR LGTM. I think Gemini's suggestions are overkill, my default assumption when reading React code is that array order is stable between renders (unless explicitly changed by an event).

Is that not reasonable? @m-amaya @christjt

gregfagan:
🦄

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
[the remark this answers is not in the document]
danpriori: Hmmm in general, it shouldn't happen AFAIK. 

But I honestly don't rule out the possibility of this happening, as the component could end up receiving a new list in a different order from the backend in a new reference (if at some point a cache expired due to a staleTime setting) in case the component is called after a long pause.
I would add a sort with useMemo as Gemini suggested (not exactly the same but kinda similar) in the rare case this happens and end up not propagating the refetch for nothing. I prefer to be safe.

But I'm open to opinions from @m-amaya and @christjt if they also think this is overkill in this case.

react-components/src/query/useAssetsAndTimeseriesLinkageDataQuery.ts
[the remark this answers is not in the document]
danpriori: Same comment like this:
https://github.com/cognitedata/reveal/pull/5269#discussion_r2277347453

react-components/src/components/RuleBasedOutputs/hooks/useGetDMConnectionWithNodeFromHybridMappingsQuery.ts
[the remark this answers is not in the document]
christjt: Seems to me that this is very much theoretical, not sure if this is a problem in practice. And even if it is, it will just result in cache miss and refetch. So I am fine with either solution here.

PR #5270, opened 2025-08-14
feat(react-components): make URLs in PoI comments clickable

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

Closes https://cognitedata.atlassian.net/browse/BND3D-5958

## Description :pencil:

Makes URLs in comments in the PoI side panel clickable:

bg0

In Fusion

- Open Unified 3D
- Go to Håkon's scene
- Go to Points of Interest in FRS
- Click point "Too many points"
- Click the URL in one of the comments, and it will take you to a new site

You can also try to create new comments in other PoIs with URL(s) if you're not convinced


react-components/src/utilities/TextWithClickableLink.tsx
```diff
@@ -0,0 +1,41 @@
+import { type ReactElement, useMemo } from 'react';
+import styled from 'styled-components';
+import { isDefined } from './isDefined';
+
+export function TextWithClickableLink({ content }: { content: string }): ReactElement {
+  // Adapted from https://gist.github.com/kiennt2/c9a489369562c424c793b8883b98802e
+  const URL_REGEX = /(https?:\/\/[-A-Z0-9+&@#/%?=~_|!:,.;]+[-A-Z0-9+&@#/%=~_|])/gi;
```
danpriori: Maybe one "cleaner" option would be to use some simple lib like `validator` (https://www.npmjs.com/package/validator) or `is-url` (https://www.npmjs.com/package/is-url) or even use the URL API (`new URL()` ) instead of using regex. And it might simplify the verification down below.

danpriori: As we've discussed in DM, one another option would be to use Linkifyjs in case for multiple URLs inside the same string. 
But it is, ofc, an option and not crucial.

danpriori (dismissed):
LGTM 👍

react-components/src/utilities/TextWithClickableLink.test.tsx
```diff
@@ -0,0 +1,68 @@
+import { describe, expect, test } from 'vitest';
+import { TextWithClickableLink } from './TextWithClickableLink';
+import { render, screen } from '@testing-library/react';
+
+describe(TextWithClickableLink.name, () => {
```
pramod-cog, line 5: If I run below test case
```
test('malformed URLs', () => {
    const malformedUrls = [
      'http:// shouldnotlink.com',
      'http://a.b-.co'
    ];

    malformedUrls.forEach(url => {
      const { container } = render(<TextWithClickableLink content={url} />);
      const span = container.children[0];
      
      expect(span.children).toHaveLength(1);
      expect(span.children[0].tagName).toBe('A');
      expect(span.children[0].textContent).toBe(url);
    });
});
```

The test fails, basic invalid URLs test

react-components/src/utilities/TextWithClickableLink.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  const urlComponents = useMemo(
```
pramod-cog, line 6: I feel below is better approach where it removes empty string up front, handles single regex test

```
const urlComponents = useMemo(() => {
  const parts = content.split(URL_REGEX);
  return parts
    .filter(part => part !== '')
    .map((part, index) => {
      if (URL_REGEX.test(part)) {
        URL_REGEX.lastIndex = 0;
        return (
          <LinkText href={part} key={index}>
            {part}
          </LinkText>
        );
      }
      return <span key={index}>{part}</span>;
    });
}, [content]);
```

You also add URL validation in the `if` check

pramod-cog (changes requested):
Added some comments and gemini pointed some interesting comment regaring parentheses in regex, let me know your thoughts

react-components/src/utilities/TextWithClickableLink.test.tsx
haakonflatval-cognite: The second of the URLs seems to actually be valid :thinking: But the first one should be filtered out. Fixing :+1:

pramod-cog (dismissed):
risk review ok

pramod-cog:
@haakonflatval-cognite need to fix the CI error

pramod-cog (changes requested):
Need to fix the CI

PR #5273, opened 2025-08-17
feat(react-components): update fdm hook fetch all threed connections with properties rule based coloring

danpriori:
#### Type of change

Feat  

#### Jira ticket :blue_book:

It is part of the Hybrid for rule base coloring support: https://github.com/cognitedata/reveal/pull/5254

## Description :pencil:
This is an update of the hook `useAll3dDirectConnectionsWithProperties` to support multiple connections for a single `instanceWithData` and also add a unit test file for the main cases

### DIFF
--- a/react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
+++ b/react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
@@ -61,7 +61,6 @@ describe(useAll3dDirectConnectionsWithProperties.name, () => {
     views: [{ space: 'view-space', externalId: 'view-id', version: '1', type: 'view' as const }]
   };
 
-
   const mockConnectionWithNode2: FdmConnectionWithNode = {
     connection: {
       instance: { space: 'test-space', externalId: 'test-id-2' },
@@ -141,7 +140,11 @@ describe(useAll3dDirectConnectionsWithProperties.name, () => {
 
   it('should process connections and return multiple instance data', async () => {
     const { result } = renderHook(
-      () => useAll3dDirectConnectionsWithProperties([mockConnectionWithNode, mockConnectionWithNodeSameInstance]),
+      () =>
+        useAll3dDirectConnectionsWithProperties([
+          mockConnectionWithNode,
+          mockConnectionWithNodeSameInstance
+        ]),
       { wrapper }
     );
 
--- a/react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
+++ b/react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
@@ -22,6 +22,11 @@ describe(useAll3dDirectConnectionsWithProperties.name, () => {
   const ARBITRARY_MODEL_ID = 123;
   const ARBITRARY_REVISION_ID = 456;
 
+  const baseDmsUniqueIdentifier = {
+    space: 'test-space',
+    externalId: 'test-id'
+  };
+
   const node3d1 = createCadNodeMock({
     id: 100,
     treeIndex: ARBITRARY_TREE_INDEX,
@@ -36,7 +41,7 @@ describe(useAll3dDirectConnectionsWithProperties.name, () => {
 
   const mockConnectionWithNode: FdmConnectionWithNode = {
     connection: {
-      instance: { space: 'test-space', externalId: 'test-id' },
+      instance: baseDmsUniqueIdentifier,
       modelId: ARBITRARY_MODEL_ID,
       revisionId: ARBITRARY_REVISION_ID,
       treeIndex: ARBITRARY_TREE_INDEX
@@ -44,6 +49,18 @@ describe(useAll3dDirectConnectionsWithProperties.name, () => {
     cadNode: node3d1,
     views: [{ space: 'view-space', externalId: 'view-id', version: '1', type: 'view' as const }]
   };
+
+  const mockConnectionWithNodeSameInstance: FdmConnectionWithNode = {
+    connection: {
+      instance: baseDmsUniqueIdentifier,
+      modelId: ARBITRARY_MODEL_ID,
+      revisionId: ARBITRARY_REVISION_ID,
+      treeIndex: ARBITRARY_TREE_INDEX + 1
+    },
+    cadNode: node3d2,
+    views: [{ space: 'view-space', externalId: 'view-id', version: '1', type: 'view' as const }]
+  };
+
 
   const mockConnectionWithNode2: FdmConnectionWithNode = {
     connection: {
@@ -57,9 +74,8 @@ describe(useAll3dDirectConnectionsWithProperties.name, () => {
   };
 
   const mockFdmInstanceData = {
+    ...baseDmsUniqueIdentifier,
     instanceType: 'node' as const,
-    space: 'test-space',
-    externalId: 'test-id',
     version: 1,
     createdTime: 11111,
     lastUpdatedTime: 11111,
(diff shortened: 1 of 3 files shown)


react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+    .returns(vi.fn())
+    .setup((p) => p.inspectInstances)
+    .returns(vi.fn())
+    .object();
+
+  const mockDependencies = getMocksByDefaultDependencies(
+    defaultUseAll3dDirectConnectionsWithPropertiesDependencies
+  );
+
+  mockDependencies.useFdmSdk.mockReturnValue(mockFdmSdk);
```
haakonflatval-cognite: We have configured `react-components` to automatically clear all `vi.fn` mocks implicitly in `beforeEach` - so this one will be overridden before the first test. I'd suggest moving this line into `beforeEach`

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+  const wrapper: FC<PropsWithChildren<unknown>> = ({ children }) => (
+    <QueryClientProvider client={queryClient}>
+      <FdmSdkContext.Provider value={{ fdmSdk: mockFdmSdk }}>
+        {children}
+        <UseAll3dDirectConnectionsWithPropertiesContext.Provider value={mockDependencies}>
+          {children}
```
haakonflatval-cognite: Are we mounting `children` twice? 🧐

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+        typing: {}
+      },
+      {
+        connection: mockConnectionWithNodeSameInstance.connection,
+        cadNode: mockConnectionWithNodeSameInstance.cadNode,
+        views: mockConnectionWithNodeSameInstance.views,
+        items: [mockFdmInstanceData],
+        instanceType: 'node' as const,
+        typing: {}
+      }
```
haakonflatval-cognite: I would consider a utility function for creating these, but not critical. Also - should this declaration be moved above the `waitFor` further up? Makes the separation between declaration and assertions a bit clearer 😅

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+    }
+  };
+
+  const mockInspectResultList = [mockInspectResultItem];
+
+  const mockFdmSdk = new Mock<FdmSDK>()
+    .setup((p) => p.getByExternalIds)
+    .returns(vi.fn())
+    .setup((p) => p.inspectInstances)
+    .returns(vi.fn())
```
haakonflatval-cognite: Personally, I would have liked to declare the `vi.fn`s outside of this `new Mock`. Then you don't need to write `vi.mocked` when mocking them

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+        views: mockConnectionWithNodeSameInstance.views,
+        items: [mockFdmInstanceData],
+        instanceType: 'node' as const,
+        typing: {}
+      }
+    ];
+
+    await waitFor(() => {
+      expect(result.current.data).toEqual(expectedData);
+    });
```
haakonflatval-cognite: Are the `waitFor`s necessary? I mean - probably the first one is, but can the others be removed?

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
```diff
@@ -0,0 +1,281 @@
...
+        wrapper,
+        initialProps: { connections: [mockConnectionWithNode] }
+      }
+    );
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    const initialCallCount = vi.mocked(mockFdmSdk.getByExternalIds).mock.calls.length;
```
haakonflatval-cognite: Is it better to assert that `mockFdmSdk.getByExternalIds` has a specific number of calls? And same at the end of this test?

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
danpriori: Ok, but in `useHybridMappingsForAssetInstances.test`, for example, (which is a recent code), we are using this approach with vi.mocked. 🤔 
So if we keep a unique pattern, then I think it is better to define together what pattern we should use for mocking strategies in general going forward first, and then we start to use it and replace it wherever we don't have it. Otherwise, we will have a mix of patterns again. 
I don't have a strong opinion on which one to use. As long as it is simple, slim, and we can have type safe, I'm good. The only concern is to decide on a unique pattern. Does make sense?

haakonflatval-cognite: Yes, I see - I guess we can argue that this is a bit up to taste, and having multiple different conventions will be annoying :sweat_smile: We can leave this be :+1:

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
danpriori: yes. Agree
https://github.com/cognitedata/reveal/pull/5273/commits/5cee78700f58e4abfcf735354e2990e686a023e2

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
danpriori: https://github.com/cognitedata/reveal/pull/5273/commits/5cee78700f58e4abfcf735354e2990e686a023e2

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
danpriori: yep!
https://github.com/cognitedata/reveal/pull/5273/commits/5cee78700f58e4abfcf735354e2990e686a023e2

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
danpriori: ops! 🙈 
https://github.com/cognitedata/reveal/pull/5273/commits/5cee78700f58e4abfcf735354e2990e686a023e2

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
danpriori: Oh yeah! Forgot that.
https://github.com/cognitedata/reveal/pull/5273/commits/5cee78700f58e4abfcf735354e2990e686a023e2

react-components/src/query/useAll3dDirectConnectionsWithProperties.test.tsx
```diff
@@ -0,0 +1,282 @@
...
+        wrapper,
+        initialProps: { connections: [mockConnectionWithNode] }
+      }
+    );
+
+    await waitFor(() => {
+      expect(result.current.isLoading).toBe(false);
+    });
+
+    expect(vi.mocked(mockFdmSdk.getByExternalIds).mock.calls.length).toEqual(initialCallCount);
```
haakonflatval-cognite: Can we use `expect(...).toHaveBeenCalledTimes(...)` here?

danpriori: https://github.com/cognitedata/reveal/pull/5273/commits/ad60d705258d6a68551d05fd91ee65f8a232eeba

PR #5308, opened 2025-09-08
feat(react-components): Adding calculation of angular Coverage in the cylinder fit

nilscognite:
#### Type of change

Feat  

## Description :pencil:

In order to filter out the false detection, I also added angular coverage as the result. It will give a value between 0 an 1, where 1 is the case where the points are semi equal distributed round the axis. Closer to 0 point are clustered to a small angle interval.

In case of a straight line, it will actually be a perfect match for a cylinder, but can be detected a<s false, because of low angular distribution.

By unit tests

### DIFF
--- a/react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
+++ b/react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
@@ -37,7 +37,7 @@ export function bestFitCylinder(
     return undefined;
   }
   let bestCylinder: LeastSquareCylinderResult | undefined;
-  const angularHistogram = new Array<number>(points.length / 2); // Just to avoid to many allocation
+  const angularHistogram = new Array<number>(Math.ceil(points.length / 2)); // Just to avoid to many allocation
 
   // Try with all available axis as a starting point, and keep the best one
   for (const axis of MAIN_AXES) {
@@ -288,8 +288,6 @@ function addToAngularHistogram(point: Vector3, histogram: number[]): void {
   // Fold into range 0..length-1
   if (index < 0) {
     index += histogram.length;
-  } else if (index >= histogram.length) {
-    index -= histogram.length;
   }
   histogram[index]++;
 }


react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
```diff
@@ -252,16 +257,36 @@ function computeGaussNewton(points: Vector3[], cylinder: LeastSquareCylinderResu
...
 
-function getRms(points: Vector3[], cylinder: LeastSquareCylinderResult): number {
+function calculateFitness(
+  cylinder: LeastSquareCylinderResult,
+  points: Vector3[],
+  angleHistogram: number[]
+): void {
   const matrix = cylinder.getTranslationRotationMatrix();
   let sumError = 0;
+  angleHistogram.fill(0);
```
haakonflatval-cognite: I would personally prefer that the caller of this function takes care of cleaning the array, but not critical

react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
```diff
@@ -53,6 +55,9 @@ export function bestFitCylinder(
     }
     bestCylinder = cylinder;
   }
+  console.log(
+    `Best cylinder RMS: ${bestCylinder?.rms}, angle coverage: ${bestCylinder?.angularCoverage}`
+  );
```
haakonflatval-cognite: 🧐

react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
```diff
@@ -252,16 +257,36 @@ function computeGaussNewton(points: Vector3[], cylinder: LeastSquareCylinderResu
...
     // Find relative radius error
     const radius = horizontalLength(point);
     const error = cylinder.radius > EPSILON ? 1 - radius / cylinder.radius : radius;
+    addToAngleHistogram(point, angleHistogram);
     sumError += error * error;
   }
-  return Math.sqrt(sumError / points.length);
+  const nonZeroCount = count(angleHistogram, (value: number) => value > 0);
+  cylinder.rms = Math.sqrt(sumError / points.length);
+  cylinder.angularCoverage = nonZeroCount / angleHistogram.length;
```
haakonflatval-cognite: I see that we set the `angularCoverage` here, but should we be checking that it fits in a particular threshold too, or is that for another PR?

haakonflatval-cognite (commented):
Some comments, but otherwise looks great 👍

react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
nilscognite: Dismissed. I disagree. The reason for it is outside is only to avoid allocations

react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
nilscognite: Fix

react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
nilscognite: The purpose of this is to check  it in the accept function in this:

```
export function bestFitCylinder(
  points: Vector3[],
  accept?: AcceptCylinder
): LeastSquareCylinderResult | undefined {
```

So it is outside this PR of you mend that. I also have this check in the unit test:

```
  test('should not get a cylinder if the angularCoverage is too small', () => {
    const expectedCylinder = getRandomCylinder(random);
    const points = createPoints(expectedCylinder, 100, random, 0, 0.4);
    
    const accept = (cylinder: LeastSquareCylinderResult): boolean =>
      cylinder.rms < 0.05 && cylinder.angularCoverage > 0.5;      <==========================
      
    const cylinder = bestFitCylinder(points, accept);
    expect(cylinder).toBeUndefined();
  });

```

react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
```diff
@@ -36,6 +37,7 @@ export function bestFitCylinder(
     return undefined;
   }
   let bestCylinder: LeastSquareCylinderResult | undefined;
+  const angularHistogram = new Array<number>(points.length / 2); // Just to avoid to many allocation
```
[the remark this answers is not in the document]
nilscognite: Fixed. Use ceil instead

react-components/src/architecture/base/utilities/cylinderFit/bestFitCylinder.ts
```diff
@@ -252,16 +254,44 @@ function computeGaussNewton(points: Vector3[], cylinder: LeastSquareCylinderResu
...
+
+function addToAngularHistogram(point: Vector3, histogram: number[]): void {
+  const angle = Math.atan2(point.y, point.x);
+  let index = Math.floor((histogram.length * angle) / (2 * Math.PI));
+  // Fold into range 0..length-1
+  if (index < 0) {
+    index += histogram.length;
+  } else if (index >= histogram.length) {
+    index -= histogram.length;
+  }
```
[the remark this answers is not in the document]
nilscognite: Fixed

haakonflatval-cognite (approved):

pramod-cog (commented):
risk review ok

PR #5310, opened 2025-09-08
fix: allow cached sectors to be used across different models with same modelId/revisionId part2

pramod-cog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5696

## Description :pencil:
This is second part where textured mesh were not loaded as Mesh objects are not shared. Here changes are made to use geometry buffer, texture and bounding box for textured data in CadNode to generate the mesh data so the each data requested contain its own textured mesh data, even though it uses same cached data.

In examples

- run `yarn && yarn build` in `viewer` folder
- run `yarn && yarn start` in `examples` folder
- In browser, load an cad model [example model](https://localhost:3549/?project=3d-test&env=greenfield-3d&modelId=4727509756716142&revisionId=5448061448970587)
- After the 3d model is loaded, in `Models UI` load same model with an `offset`
- Now remove first loaded model by clicking `Remove last model`
- Open `Network` tab in developer tools, now move the camera in the viewer with only second model load, there should no call to `network` call requesting for the sector data. 


viewer/packages/cad-model/src/wrappers/CadNode.ts
```diff
@@ -209,14 +228,146 @@ export class CadNode extends Object3D<Object3DEventMap & { update: undefined }>
...
+  }
+
   public setCacheSize(sectorCount: number): void {
     this._sectorRepository.setCacheSize(sectorCount);
   }
 
   public dispose(): void {
     this.nodeAppearanceProvider.dispose();
     this.materialManager.off('materialsChanged', this._setModelRenderLayers);
+    this._sectorRepository.clearCache();
```
haakonflatval-cognite: 🧐

viewer/packages/cad-model/src/wrappers/CadNode.ts
```diff
@@ -209,6 +228,115 @@ export class CadNode extends Object3D<Object3DEventMap & { update: undefined }>
     this._geometryBatchingManager?.removeSectorBatches(sectorId);
   }
 
+  public removeSectorMeshGroup(sectorId: number): void {
+    const meshGroup = this._sectorMeshGroups.get(sectorId);
+    if (meshGroup) {
```
haakonflatval-cognite, line 233: This might be a bit more readable with early return

viewer/packages/sector-loader/src/GltfSectorRepository.ts
```diff
@@ -5,17 +5,16 @@
...
 export class GltfSectorRepository implements SectorRepository {
   private readonly _gltfSectorLoader: GltfSectorLoader;
   private readonly _gltfCache: MemoryRequestCache<string, ConsumedSector>;
 
-  constructor(sectorFileProvider: BinaryFileProvider, materialManager: CadMaterialManager) {
-    this._gltfSectorLoader = new GltfSectorLoader(sectorFileProvider, materialManager);
-    this._gltfCache = new MemoryRequestCache(200, consumedSector => consumedSector.group?.dereference(), 50);
+  constructor(sectorFileProvider: BinaryFileProvider) {
+    this._gltfSectorLoader = new GltfSectorLoader(sectorFileProvider);
+    this._gltfCache = new MemoryRequestCache(200, () => {}, 50);
```
haakonflatval-cognite, line 17: Can we change the type signature of  `MemoryRequestCache` to avoid having to send it an empty function?

viewer/packages/cad-model/src/wrappers/CadNode.ts
```diff
@@ -217,6 +345,28 @@ export class CadNode extends Object3D<Object3DEventMap & { update: undefined }>
...
+        if (child instanceof THREE.Mesh && child.geometry !== undefined) {
+          child.geometry.dispose();
+        }
+      }
+
+      // Dispose textures
+      meshGroup.textures.forEach(texture => texture.dispose());
+
+      // Clear the group
+      meshGroup.clear();
```
haakonflatval-cognite, line 366: Can we replace this with a call to `this.removeSectorMeshGroup`?

viewer/packages/cad-geometry-loaders/src/CadManager.ts
```diff
@@ -84,20 +85,30 @@ export class CadManager {
...
         cadModel.removeBatchedSectorGeometries(sector.metadata.id);
+        // Also clean up any mesh groups created from parsed geometries
+        cadModel.removeSectorMeshGroup(sector.metadata.id);
+      }
+
+      // Create meshes from parsedMeshGeometries data
+      let meshGroup: AutoDisposeGroup | undefined = undefined;
+      if (sector.parsedMeshGeometries && sector.parsedMeshGeometries.length > 0) {
+        meshGroup = cadModel.createMeshesFromParsedGeometries(sector.parsedMeshGeometries, sector.metadata.id);
       }
```
haakonflatval-cognite, line 96: It seems a bit unnecessary to declare the variable here - can we instead just move these lines down to where `meshGroup` is being used, and have `meshGroup` be a `const`?

viewer/packages/cad-geometry-loaders/src/CadManager.ts
```diff
@@ -254,18 +265,8 @@ export class CadManager {
...
-    if (!treeIndices) {
+    if (sector.parsedMeshGeometries?.length !== 1) {
       return;
     }
-
-    for (const treeIndex of treeIndices.keys()) {
-      cadModel.treeIndexToSectorsMap.set(treeIndex, sector.metadata.id);
-    }
-    cadModel.treeIndexToSectorsMap.markCompleted(sector.metadata.id, RevealGeometryCollectionType.TriangleMesh);
   }
```
haakonflatval-cognite, line 270: This function seems to do nothing now. Have we removed the functionality associated with `updateTreeIndexToSectorsMap`completely?

viewer/packages/cad-model/src/wrappers/CadNode.ts
```diff
@@ -41,6 +57,9 @@ export class CadNode extends Object3D<Object3DEventMap & { update: undefined }>
 
   private _needsRedraw: boolean = false;
 
+  // Track mesh groups by sector ID for proper cleanup when sectors are unloaded
+  private readonly _sectorMeshGroups: Map<number, AutoDisposeGroup> = new Map();
```
haakonflatval-cognite, line 61: We don't use need the "auto-dispose" functonality of `AutoDisposeGroup` anywhere here, so we could probably use another, perhaps local, type. That would also mean we could remove the texture-related functionality from `AutoDisposeGroup`

viewer/packages/cad-model/src/wrappers/CadNode.ts
```diff
@@ -209,6 +228,115 @@ export class CadNode extends Object3D<Object3DEventMap & { update: undefined }>
     this._geometryBatchingManager?.removeSectorBatches(sectorId);
   }
 
+  public removeSectorMeshGroup(sectorId: number): void {
```
haakonflatval-cognite, line 231: I wonder if I'd prefer putting `removeSectorMeshGroup` and `createMeshesFromParsedGeometries` in a separate class, to keep `CadNode` somewhat slim. Then that class could handle the mesh creation and deletion internally.

viewer/packages/cad-geometry-loaders/src/CadManager.ts
haakonflatval-cognite: Upon further investigation, I think this functionality should be added back - we don't update tree indices for triangle mesh geometry anywhere else

viewer/packages/cad-model/src/wrappers/CadNode.test.ts
```diff
@@ -1,29 +1,253 @@
...
+    const mesh = getMeshFromGroup(result, 0);
+    expect(mesh.userData.treeIndices).toEqual(new Map());
+  });
+
+  test('should skip textured geometry without texture', () => {
+    const cadNode = createCadNode(3, 3);
+    const geometry = createBasicGeometry();
+
+    // Create textured geometry without texture property in a type-safe way
+    const parsedGeometries: ParsedMeshGeometry[] = [
```
haakonflatval-cognite, line 157: In my head, if our behavior is to ignore geometry of type `TexturedTriangleMesh`without a texture, it would be best not to allow them to be constructed.

I'm also feeling that we should output a warning or something in that case. What do you think?

viewer/packages/cad-model/src/wrappers/CadNode.test.ts
```diff
@@ -1,29 +1,253 @@
...
+
+  test('should remove sector mesh group when sector is unloaded', () => {
+    const cadNode = createCadNode(3, 3);
+    const sectorId = 123;
+    const meshGroup = createTestMeshGroup(cadNode, sectorId);
+
+    expectMeshGroup(meshGroup, 1);
+
+    cadNode.removeSectorMeshGroup(sectorId);
+    expect(meshGroup.children.length).toBe(0); // AutoDisposeGroup clears children on dispose
```
haakonflatval-cognite, line 203: As mentioned, we don't really use `AutoDisposeGroup`'s functionality, so I suggest removing the comment

viewer/packages/cad-model/src/wrappers/CadNode.test.ts
```diff
@@ -1,29 +1,253 @@
...
+
+    cadNode.removeSectorMeshGroup(100);
+    expect(meshGroup1.children.length).toBe(0);
+    expect(meshGroup2.children.length).toBe(1);
+
+    cadNode.removeSectorMeshGroup(200);
+    expect(meshGroup2.children.length).toBe(0);
+  });
+
+  test('should replace existing mesh group when sector is updated', () => {
```
haakonflatval-cognite, line 222: Just wondering - can this happen? That a sector is updated before the old sector data/mesh group is removed? 🤔

viewer/packages/cad-model/src/wrappers/CadNode.test.ts
```diff
@@ -1,29 +1,253 @@
...
+};
+
+const createTestMeshGroup = (cadNode: CadNode, sectorId: number, vertices?: number[]) => {
+  const geometry = createBasicGeometry(vertices);
+  const parsedGeometries = [createParsedGeometry(RevealGeometryCollectionType.TriangleMesh, geometry)];
+  return cadNode.createMeshesFromParsedGeometries(parsedGeometries, sectorId);
+};
+
+const isMesh = (object: unknown): object is Mesh => {
+  return object instanceof Mesh;
```
haakonflatval-cognite, line 60: Can we just do this check inline instead of defining this type guard separately?

viewer/packages/cad-model/src/wrappers/CadNode.test.ts
```diff
@@ -1,29 +1,253 @@
...
 
 import { jest } from '@jest/globals';
 import { createCadNode } from '../../../../test-utilities/src/createCadNode';
+import { SectorRepository } from '@reveal/sector-loader';
+import { Mock } from 'moq.ts';
 import { createMockedConsumedSector, createWantedSectorMock } from '@reveal/sector-loader/tests/mockSectorUtils';
 
+const createBasicGeometry = (vertices: number[] = [0, 0, 0, 1, 0, 0, 0, 1, 0], treeIndices?: number[]) => {
+  const geometry = new BufferGeometry();
+  geometry.setAttribute('position', new BufferAttribute(new Float32Array(vertices), 3));
```
haakonflatval-cognite, line 18: Can we have all the utility functions here at the bottom of the test file? It's a little distracting having to scroll past all of them to get to the actual tests

viewer/packages/cad-geometry-loaders/src/CadManager.test.ts
```diff
@@ -0,0 +1,110 @@
...
+      .setup(p => p.addModel(It.IsAny()))
+      .returns()
+      .setup(p => p.removeModel(It.IsAny()))
+      .returns()
+      .setup(p => p.updateCamera(It.IsAny(), It.IsAny()))
+      .returns()
+      .setup(p => (p.clippingPlanes = It.IsAny()))
+      .callback(() => {});
+
+    cadManager = new CadManager(materialManagerMock.object(), cadModelFactoryMock.object(), updateHandlerMock.object());
```
haakonflatval-cognite, line 52: Personally, I would like `cadManager` to be created separately in each test explicitly. But not critical

viewer/packages/cad-geometry-loaders/src/CadManager.test.ts
```diff
@@ -0,0 +1,110 @@
...
+    expect(cadManager.materialManager).toBe(materialManagerMock.object());
+  });
+
+  test('should create CadManager with all required components', () => {
+    // Test that CadManager can be created with valid components
+    const validCadManager = new CadManager(
+      materialManagerMock.object(),
+      cadModelFactoryMock.object(),
+      updateHandlerMock.object()
+    );
```
haakonflatval-cognite, line 103: How is this test different from the first test in this file?

viewer/packages/sector-loader/tests/GltfSectorRepository.test.ts
```diff
@@ -52,13 +47,123 @@ describe(GltfSectorRepository.name, () => {
...
   });
+
+  // Helper functions to reduce code duplication
+  const createWantedSectorWithMetadata = (
+    metadata: Partial<SectorMetadata>,
+    levelOfDetail = LevelOfDetail.Detailed,
+    modelIdentifier = new LocalModelIdentifier('test_model')
+  ) => {
+    const mockMetadata = new Mock<SectorMetadata>();
+    Object.entries(metadata).forEach(([key, value]) => {
```
haakonflatval-cognite, line 66: I would prefer to have these helper functions at the bottom of the file

viewer/packages/utilities/src/cache/MemoryRequestCache.ts
```diff
@@ -35,19 +35,13 @@ export class MemoryRequestCache<Key, Data> implements RequestCache<Key, Data> {
...
-    maxElementsInCache: number = 50,
-    removeCallback?: (value: Data) => void,
-    defaultCleanupCount: number = 10
-  ) {
+  constructor(maxElementsInCache: number = 50, defaultCleanupCount: number = 10) {
     this._data = new Map();
     this._maxElementsInCache = maxElementsInCache;
     this._defaultCleanupCount = Math.max(defaultCleanupCount, 1);
-    this._removeCallback = removeCallback;
   }
```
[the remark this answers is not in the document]
pramod-cog: addressed in [part3 PR](https://github.com/cognitedata/reveal/pull/5353)

viewer/packages/cad-geometry-loaders/src/CadManager.ts
```diff
@@ -84,20 +85,29 @@ export class CadManager {
...
         throw new Error(`Could not find 3D node for sector ${sector.metadata.id} - invalid id?`);
       }
-      if (sector.group) {
-        sectorNode.add(sector.group);
+
+      // Create meshes from parsedMeshGeometries data
+      const meshGroup: AutoDisposeGroup | undefined =
+        sector.parsedMeshGeometries && sector.parsedMeshGeometries.length > 0
+          ? cadModel.createMeshesFromParsedGeometries(sector.parsedMeshGeometries, sector.metadata.id)
+          : undefined;
```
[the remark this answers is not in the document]
pramod-cog: This would be remove buffer geometry, which we do not want

PR #5315, opened 2025-09-09
feat(react-component): standardize date and time in Image 360 historical panel and POI details panel

pramod-cog:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5134

## Description :pencil:
The Date and Time format in Image 360 historical details panel and POI details panel are updated to `@cognite/cdf-utilities` date format. Have refactored `Image360HistoricalDetails` to MVVM pattern and added test.

Previous Date and Time format
Screenshot 2025-09-09 at 16 32 04

Update Date and Tie format
Screenshot 2025-09-09 at 16 24 55

In storybook and unified-3d through local build

- yarn && yarn build && yarn storybook
- Load Image360Details storybook example
- Click on Details button at the right bottom corner


react-components/vite.config.ts
```diff
@@ -16,6 +16,15 @@ export default defineConfig(({ command }) => {
...
     ],
+    resolve: {
+      alias: {
+        // Fix dayjs plugin imports for @cognite/cdf-utilities
+        'dayjs/plugin/isToday': 'dayjs/plugin/isToday.js',
+        'dayjs/plugin/isTomorrow': 'dayjs/plugin/isTomorrow.js',
+        'dayjs/plugin/isYesterday': 'dayjs/plugin/isYesterday.js',
+        'dayjs/plugin/relativeTime': 'dayjs/plugin/relativeTime.js'
+      }
+    },
```
haakonflatval-cognite: Why do we need this?

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+  formatDateTime: vi.fn(({ date }) => date.toISOString().split('T')[0])
+}));
+
+vi.mock('./utils/getStationIdentifier', () => ({
+  getStationIdentifier: vi.fn(() => 'station-123')
+}));
+
+vi.mock('lodash', () => ({
+  uniqueId: vi.fn(() => 'test-id')
+}));
```
haakonflatval-cognite: Since we have refactored to the MVVM-pattern, shouldn't we be able to just inject these dependencies if we need to, instead of using `vi.mock`?

react-components/tests/tests-utilities/fixtures/image360Station.ts
```diff
@@ -0,0 +1,25 @@
...
+    .setup((p) => p.label)
+    .returns('Test Station Label')
+    .setup((p) => p.getRevisions)
+    .returns(() => revisions ?? [])
+    .object();
+
+export const createMockImage360Revision = (
+  date: Date,
+  thumbnailUrl: string
+): Image360Revision<ClassicDataSourceType> =>
```
haakonflatval-cognite: Nice! Could we make these parameters optional and make them part of one input object, instead of separate arguments? It makes the mock more flexible, if we want to use it and mock other properties in the future

react-components/tests/tests-utilities/fixtures/image360Station.ts
```diff
@@ -0,0 +1,25 @@
+import { type Image360Revision, type Image360, type ClassicDataSourceType } from '@cognite/reveal';
+import { Mock } from 'moq.ts';
+
+export const createMockImage360Entity = (
+  revisions?: Image360Revision[]
+): Image360<ClassicDataSourceType> =>
```
haakonflatval-cognite, line 11: Can we put this in an input object as well?

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.viewmodel.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const newScrollPosition = useRef(0);
+
+  const stationId = image360Entity !== undefined ? getStationIdentifier(image360Entity) : undefined;
+  const stationName = image360Entity?.label;
+
+  useEffect(() => {
+    const fetchRevisionCollection = async (): Promise<void> => {
+      if (image360Entity !== undefined) {
+        const revisions = image360Entity.getRevisions();
+        const revisionDates = revisions.map((revision) => revision.date);
```
haakonflatval-cognite, line 29: I know it's not your code, but it looks really brittle.. If we trigger this twice in a small amount of time, we may end up with a race condition

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.viewmodel.tsx
```diff
@@ -0,0 +1,82 @@
...
+    };
+  }, [image360Entity]);
+
+  useEffect(() => {
+    const newMinWidth = revisionDetailsExpanded ? '100%' : '100px';
+    setMinWidth(newMinWidth);
+    if (onExpand !== undefined) {
+      onExpand(revisionDetailsExpanded);
+    }
+  }, [revisionDetailsExpanded, onExpand]);
```
haakonflatval-cognite: Again, I know it's not your code, but is it possible to rewrite this? E.g. have a `useMemo` to evaluate the `revisionDetailsExpanded` - perhaps the `onExpand` should be part of some callback controlling expansion of the panel?

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  const defaultProps: Image360HistoricalDetailsProps = {
+    viewer: viewerMock,
+    image360Entity: undefined,
+    onExpand: mockOnExpand,
+    fallbackLanguage: 'en'
+  };
+
+  beforeEach(() => {
+    vi.clearAllMocks();
```
haakonflatval-cognite: `vi.resetAllMocks` is called implicitly in all our tests in react-components, so this is unnecessary

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+    // Mock URL methods to avoid errors in test environment
+    globalThis.URL.revokeObjectURL = vi.fn();
+  });
```
haakonflatval-cognite: Is this necessary? What kind of error do we get?

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+    const mockEntity = createMockImage360Entity(testRevisions);
+
+    render(<Image360HistoricalDetails {...defaultProps} image360Entity={mockEntity} />);
+
+    expect(screen.getByRole('button')).toBeTruthy();
+  });
+
+  test('expands to show historical summary when panel is clicked', async () => {
+    const user = userEvent.setup();
```
haakonflatval-cognite: Is this needed? I see that we don't use this the other places we use `userEvent` (e.g. `InputField.test.tsx`)

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+        fallbackLanguage="en"
+      />
+    );
+
+    const expandButton = screen.getByRole('button');
+
+    // Should not throw when onExpand is not provided
+    expect(async () => {
+      await user.click(expandButton);
+    }).not.toThrow();
```
haakonflatval-cognite: Don't we need to `await` this?

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
pramod-cog: It is need, without it we get an error `TypeError: URL.revokeObjectURL is not a function` because the ViewModel calls `URL.revokeObjectURL()` during image URL cleanup (line 31 in viewmodel), but Node.js test environment doesn't have this browser API available by default. The mock prevents the runtime error while allowing tests to focus on component behavior.

react-components/vite.config.ts
pramod-cog: This is needed to resolve  ESM module resolution issues with `@cognite/cdf-utilities` which uses `dayjs` modules, if you notice the CI for the [commit](https://github.com/cognitedata/reveal/pull/5315/commits/8316707cd12978829a075c740de1da2e5dff4c4a), it fails because of this issue

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
pramod-cog: Lint warns `'await' has no effect on the type of this expression.`

haakonflatval-cognite: I think you need to write `await expect(async () => {
      await userEvent.click(expandButton);
    }).resolves.not.toThrow();`

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
haakonflatval-cognite: I see... I'm a little skeptical to overriding a global function like this directly in the test, since it can affect other tests if run in parallel.. Can we provide the `revokeObjectURL` function through a context?

react-components/package.json
```diff
@@ -26,7 +26,7 @@
     "typecheck:source": "tsc --noEmit --project tsconfig.build.json",
     "typecheck:test": "tsc --noEmit --project tsconfig.test.json",
     "lint": "eslint --cache --max-warnings 0 --ext=.tsx,.ts .",
-    "test": "vitest run --dom --silent=true",
+    "test": "vitest run --dom",
```
haakonflatval-cognite: Is this intended to be part of this PR?

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.viewmodel.tsx
```diff
@@ -15,56 +16,62 @@ export function useImage360HistoricalDetailsViewModel({
...
+      setActiveRevision(0);
+    }
+  }, [image360Entity, imageUrls]);
+
+  useMemo(() => {
+    if (previousImage360Entity.current !== image360Entity) {
+      previousImage360Entity.current = image360Entity;
+      void fetchRevisionCollection();
+    }
+  }, [image360Entity, fetchRevisionCollection]);
```
haakonflatval-cognite: I don't feel that a `useMemo` here is better than using `useEffect` 😅 If it's hard to get rid of the `useEffect`, we can just keep it the way it was

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.viewmodel.tsx
```diff
@@ -15,56 +16,62 @@ export function useImage360HistoricalDetailsViewModel({
...
+
+  const minWidth = useMemo(() => {
     const newMinWidth = revisionDetailsExpanded ? '100%' : '100px';
-    setMinWidth(newMinWidth);
+
     if (onExpand !== undefined) {
       onExpand(revisionDetailsExpanded);
```
haakonflatval-cognite: Here too - ideally `useMemo` function should be pure, but this one calls the `onExpand` on the inside, sometimes. Is it possible to instead send the `onExpand` function to the view and trigger it from there?

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.viewmodel.context.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type Image360HistoricalDetailsViewModelDependencies = {
+  formatDateTime: typeof formatDateTime;
+  revokeObjectUrl: typeof revokeObjectUrl;
+};
+
+export const defaultImage360HistoricalDetailsViewModelDependencies: Image360HistoricalDetailsViewModelDependencies =
+  {
+    formatDateTime,
+    revokeObjectUrl
```
haakonflatval-cognite: In my head, we could just have written `revokeObjectUrl: globalThis.URL.revokeObjectURL`, so that we wouldn't have to create a new file for `revokeObjectUrl`. Or wouldn't that work?

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -8,72 +8,77 @@ import {
...
+} from './Image360HistoricalDetails.viewmodel.context';
+import { type ReactElement, type ReactNode } from 'react';
 
 describe(Image360HistoricalDetails.name, () => {
   const mockOnExpand = vi.fn();
 
+  const mockContextValue: Image360HistoricalDetailsViewModelDependencies = {
+    formatDateTime: vi.fn(({ date }) => date.toISOString().split('T')[0]),
+    revokeObjectUrl: vi.fn()
+  };
```
haakonflatval-cognite: You can use `getMockByDefaultDependencies` to create this - although that means you will have to mock `formatDateTime` in `beforeEach`. Not critical, if you prefer this

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -101,29 +107,107 @@ describe(Image360HistoricalDetails.name, () => {
...
+      <Image360HistoricalDetails {...defaultProps} image360Entity={undefined} />,
+      { wrapper }
+    );
+
+    rerender(<Image360HistoricalDetails {...defaultProps} image360Entity={initialEntity} />);
+
+    await waitFor(() => {
+      expect(mockContextValue.formatDateTime).toHaveBeenCalled();
+    });
+    vi.clearAllMocks();
```
haakonflatval-cognite: If you want to clear just a single function, you can use `mockContextValue.formatDateTime.mockClear()`. But not critical

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -101,29 +107,107 @@ describe(Image360HistoricalDetails.name, () => {
...
+    const { rerender } = render(
+      <Image360HistoricalDetails {...defaultProps} image360Entity={undefined} />,
+      { wrapper }
+    );
+
+    // Then render with entity to trigger the fetch
+    rerender(<Image360HistoricalDetails {...defaultProps} image360Entity={mockEntity} />);
+
+    // Wait for async operations to complete
+    await waitFor(() => {
```
haakonflatval-cognite: This feels like a `// This is a comment`-comment 😅

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -0,0 +1,213 @@
...
+      expect(mockContextValue.formatDateTime).toHaveBeenCalled();
+    });
+    vi.clearAllMocks();
+
+    // Change entity to trigger new fetch
+    rerender(<Image360HistoricalDetails {...defaultProps} image360Entity={newEntity} />);
+
+    await waitFor(() => {
+      expect(mockContextValue.formatDateTime).toHaveBeenCalledWith({
+        date: new Date('2024-02-20T14:45:00Z')
```
haakonflatval-cognite: I would perhaps prefer that date objects are stored in variables, so that it's easy to see which date object is asserted where without trying to parse the date manually, but not critical

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
haakonflatval-cognite, line 1: I see that most of this file is testing behavior in the `viewmodel`, and not the `view`.. I'm wondering if it would be cleaner if we separated those? I.e. move most of the tests from this file into a dedicated `viewmodel` test file? It could also make it easier to test it, as we can observe the return values of the view model directly.

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const mockEntity = createMockImage360Entity([]);
+
+  // Create a proper mutable ref object for React
+  const mockRef = { current: 100 };
+  Object.defineProperty(mockRef, 'current', {
+    writable: true,
+    enumerable: true,
+    configurable: true,
+    value: 100
+  });
```
haakonflatval-cognite: Is this `Object.defineProperty` necessary? What is the purpose?

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  const mockDependencies: Image360HistoricalDetailsDependencies = {
+    Image360HistoricalPanel: mockImage360HistoricalPanel,
+    Image360HistoricalSummary: mockImage360HistoricalSummary,
+    useImage360HistoricalDetailsViewModel: mockUseImage360HistoricalDetailsViewModel
+  };
+
+  beforeEach(() => {
+    mockUseImage360HistoricalDetailsViewModel.mockReturnValue(mockViewModelResult);
+  });
```
haakonflatval-cognite: I think this is unneeded, since we define `mockUseImage360HistoricalDetailsViewModel` as `vi.fn(() => mockViewModelResult);`

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+            index: 1,
+            image360Entity: mockEntity
+          }),
+          expect.objectContaining({
+            date: '2024-03-10',
+            imageUrl: 'thumb3.jpg',
+            index: 2,
+            image360Entity: mockEntity
+          })
+        ]),
```
haakonflatval-cognite: It seems like this can be shortened to `revisionCollection: mockViewModelResult.revisionCollection`, which is also clearer in my opinion

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.viewmodel.test.tsx
```diff
@@ -0,0 +1,359 @@
...
+    createMockImage360Revision({
+      date: new Date(mockDefaultDateTime),
+      thumbnailUrl: mockDefaultThumbnailUrl
+    })
+  ];
+
+  const mockEntity = createMockImage360Entity(mockRevisions);
+
+  beforeEach(() => {
+    defaultDependencies.revokeObjectUrl.mockReturnValue(undefined);
```
haakonflatval-cognite: I think all functions in `defaultDependencies` are just `vi.fn()`, so they will have `undefined` as return value by default. But it's fine by me if you want to keep this for clarity

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.viewmodel.test.tsx
```diff
@@ -0,0 +1,359 @@
...
+          viewer: viewerMock,
+          image360Entity: mockEntity,
+          onExpand: mockOnExpand,
+          fallbackLanguage: 'en'
+        }),
+      { wrapper }
+    );
+
+    expect(defaultDependencies.formatDateTime({ date: undefined })).toBe('Date not available');
+    expect(defaultDependencies.formatDateTime).toHaveBeenCalledWith({ date: undefined });
```
haakonflatval-cognite: Is the first statement here needed? It seems to make the second line a potentially false positive

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.tsx
```diff
@@ -1,80 +1,37 @@
...
+    stationName
+  } = useImage360HistoricalDetailsViewModel({
+    viewer,
+    image360Entity,
+    onExpand,
+    fallbackLanguage
+  });
 
   return (
     <DetailsContainer style={{ minWidth }}>
```
haakonflatval-cognite: Personally, I think `minWidth` should be calculated in the view from the state, rather than the view-model. It is a view-specific property in my opinion

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.test.tsx
pramod-cog: Not needed, oversight!

PR #5327, opened 2025-09-15
feat(react-component): Add Help and Shared button in the architecture with unit tests

nilscognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

image

https://cognitedata.atlassian.net/browse/BND3D-4835
https://cognitedata.atlassian.net/browse/BND3D-4833

## Description :pencil:

Added the new buttons using the Architecture.

**Deprecate** the old ones. According to Haakon I can't remove them and replace them with the new.

Se below and Unit tests

1. yarn storybook
2. Open Architecture/Main
3. Click at the help button at the left toolbar
4. Play with it
5. Click on the share icon
6. Paste the copy paste buffer to somewhere

### DIFF
--- a/react-components/src/architecture/base/concreteCommands/general/ShareCommand.ts
+++ b/react-components/src/architecture/base/concreteCommands/general/ShareCommand.ts
@@ -25,6 +25,6 @@ export class ShareCommand extends BaseCommand {
       .catch((error) => {
         console.error(error);
       });
-    return true; // Return true regardless, use is isDone
+    return false; // Do not need another update
   }
 }


react-components/src/architecture/base/concreteCommands/general/ShareCommand.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    expect(isEmpty(command.tooltip)).toBe(false);
+    expect(command.icon).toBe('Share');
+    expect(command.isEnabled).toBe(true);
+  });
+
+  test('should copy the URL to clipboard', async () => {
+    const command = new ShareCommand();
+    expect(command.invoke()).toBe(true);
+    const text = await navigator.clipboard.readText();
+    expect(text).toBe('http://localhost:3000/');
```
nilscognite: Is this correct in all test env?

react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -73,8 +69,8 @@ export const SettingsButton = ({
       onHide={(open) => {
         setOpen(open);
       }}
-      placement={dropdownPlacement}
-      offset={{ mainAxis: TOOLBAR_HORIZONTAL_PANEL_OFFSET, crossAxis: 0 }}>
+      placement={getDropdownPlacement(placement)}
```
nilscognite, line 72: Reused code. Common with HelpButton

react-components/src/components/Architecture/utilities.ts
```diff
@@ -27,6 +29,18 @@ export function getTooltipPlacement(toolbarPlacement: PlacementType): PlacementT
   }
 }
 
+export const DROP_DOWN_OFFSET = { mainAxis: TOOLBAR_HORIZONTAL_PANEL_OFFSET, crossAxis: 0 };
```
nilscognite, line 32: Move the common code here to place the popup equally

react-components/src/architecture/base/concreteCommands/general/ShareCommand.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  });
```
christjt: I recommend doing it like this. Then it will work everywhere. 
```suggestion
  test('should copy the URL to clipboard', async () => {
    vi.spyOn(window, 'location', 'get').mockReturnValue({
      ...window.location,
      href: 'http://cognite.test'
    });

    const command = new ShareCommand();
    expect(command.invoke()).toBe(true);
    const text = await navigator.clipboard.readText();
    expect(text).toBe('http://cognite.test');
  });
```

react-components/src/architecture/base/concreteCommands/general/ShareCommand.test.ts
```diff
@@ -0,0 +1,19 @@
...
+
+describe(ShareCommand.name, () => {
+  test('should have following default behavior', () => {
+    const command = new ShareCommand();
+    expect(isEmpty(command.tooltip)).toBe(false);
+    expect(command.icon).toBe('Share');
+    expect(command.isEnabled).toBe(true);
+  });
+
+  test('should copy the URL to clipboard', async () => {
```
christjt, line 14: I am a little worried about this test being flaky. The share command is asynchronous but doesn't return the promise. So `const text = await navigator.clipboard.readText();` may run before the command has pasted the url to the clip board. Would it not make sense do have an AsyncCommand where the consumer could await the `invoke()`?

react-components/src/architecture/base/concreteCommands/general/ShareCommand.ts
```diff
@@ -0,0 +1,21 @@
...
+    return 'Share';
+  }
+
+  protected override invokeCore(): boolean {
+    const url = window.location.href;
+    void navigator.clipboard.writeText(url).catch((error) => {
+      console.error(error);
+    });
+    return true;
+  }
```
christjt: As mentioned in the test, this command will finish and return true before the actual clip board has been written to, this could cause some strange issues where f.ex. they are scripting reading from the clipboard after the command has run, right? Would it not make sense to have an async command abstraction or something like that?

react-components/src/components/Architecture/HelpButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { type PlacementType } from './types';
+
+export function createHelpButton(
+  command: BaseCommand,
+  placement: PlacementType
+): ReactElement | undefined {
+  if (command instanceof HelpCommand) {
+    return <HelpButton key={command.uniqueId} inputCommand={command} placement={placement} />;
+  }
+  return undefined;
```
christjt, line 21: Why is it better to return undefined than throwing an error here?

christjt (changes requested):

react-components/src/components/Architecture/HelpButton.tsx
nilscognite: When returning undefined, it goes to the next product in the factory.

```
export class ComponentFactory {
  private readonly _items = new Array<{ createFunc: CreateElementFunc; order: number }>();

 public createElement(command: BaseCommand, placement: PlacementType): ReactElement {
    if (this._items.length === 0) {
      throw new Error('No element creators installed. Please call installReactElement() first.');
    }
    for (const item of this._items) {
      const element = item.createFunc(command, placement);
      if (element !== undefined) {
        return element;
      }
    }
    throw new Error(`No element creator found for command: ${command.constructor.name}`);
  }
```

react-components/src/architecture/base/concreteCommands/general/ShareCommand.ts
nilscognite: Fix cf8e83b80f9f053532d77a4b13fdfef8d72975ea  and dac7658a47d44cc21844947cba7950a684f9725d

react-components/src/architecture/base/concreteCommands/general/ShareCommand.test.ts
nilscognite: Fix cf8e83b80f9f053532d77a4b13fdfef8d72975ea  and dac7658a47d44cc21844947cba7950a684f9725d

react-components/src/architecture/base/concreteCommands/general/ShareCommand.test.ts
nilscognite: Done

PR #5328, opened 2025-09-16
Taras.khomyn/bnd3 d 3698 internalize broprint js to avoid tiny dependency

taraskhomyn0333:
#### Type of change

Feat  
Bug 
Test 
Docs 
Style 
Refactor 
Chore 

#### Jira ticket :blue_book:


viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
haakonflatval-cognite, line 1: We usually give files the name of what it exports - so this file should be called e.g. `generateAudioFingerprint` - and the same for the rest of the files here 👍 I realize that the file you replaced (`Fingerprint.ts`) didn't follow that rule, but it is the convention almost everywhere else in the code base, I think 😅

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
```diff
@@ -0,0 +1,62 @@
+function setCompressorValueIfDefined( 
+  compressor: DynamicsCompressorNode,
```
haakonflatval-cognite: A small nitpick: We often place "utility" functions that are not exported at the bottom of the file, so that the important, exported function is at the top

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
```diff
@@ -0,0 +1,62 @@
...
+) {
+  const param = compressor[item] as AudioParam | undefined;
+    if (param?.setValueAtTime) {
+        param.setValueAtTime(value, context.currentTime);
+    }
+}
+
+export async function generateAudioFingerprint(): Promise<string> {
+    return new Promise((resolve, reject) => {
+        try {
```
haakonflatval-cognite: I'm curious if the `Promise` + `try-catch` is necessary here? Is it any different than just throwing errors in this function directly?

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
```diff
@@ -0,0 +1,62 @@
...
+            // OfflineAudioContext renders audio in memory
+            const OfflineAudioContext = (window.OfflineAudioContext || (window as any).webkitOfflineAudioContext);
+            
+            if (!OfflineAudioContext) {
+                return reject(new Error("Offline Audio Context is not supported."));
+            }
+
+            // 44100 is the 44.1kHz
+            const context = new OfflineAudioContext(1, 44100, 44100);
```
haakonflatval-cognite: Not sure if this comment is explaining much 😅 Maybe it can just be removed?

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (param?.setValueAtTime) {
+        param.setValueAtTime(value, context.currentTime);
+    }
+}
+
+export async function generateAudioFingerprint(): Promise<string> {
+    return new Promise((resolve, reject) => {
+        try {
+            // OfflineAudioContext renders audio in memory
+            const OfflineAudioContext = (window.OfflineAudioContext || (window as any).webkitOfflineAudioContext);
```
haakonflatval-cognite: I see that `OfflineAudioContext` is already a defined global class in Javascript. Are we sure we need to look for `webkitOfflineAudioContext` too?

viewer/packages/metrics/src/Fingerprint/index.ts
```diff
@@ -0,0 +1,27 @@
...
+import { generateCanvasFingerprint } from "./CanvasFingerprint";
+import { generateAudioFingerprint } from "./AudioFingerprint";
+import { hashFingerprint } from "./FingerprintHash";
+
+/**
+ * Returns a fingerprint based on the current browser. This is can be used to
+ * distinguish between different users. However, it is not 100% reliable,
+ * and can produce duplicates for distinct users.
+ */
+
```
haakonflatval-cognite:
```suggestion
```

viewer/packages/metrics/src/Fingerprint/index.ts
haakonflatval-cognite, line 1: Our convention is that `index.ts` files don't define any functions/classes itself, but only exports them from other modules. We can keep the `index.ts`, but the `getUserFingerprint` function can be defined in its own file, and `index.ts` can export it (i.e. `export { getUserFingerprint } from './getUserFingerprint';`)

viewer/packages/metrics/src/Fingerprint/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export async function getUserFingerprint(): Promise<string> {
```
haakonflatval-cognite: We should try to test this. It can be a little hard, but to begin with, can we just make a test that ensures that we get a valid output (i.e. not `'unknown'`) when calling it in CI?

haakonflatval-cognite (changes requested):
I had a few comments, but otherwise looks nice 👍

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
taraskhomyn0333: done!

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
taraskhomyn0333: done!

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
taraskhomyn0333: yes, we need Promise here, because this function depends on Web API (OfflineAudioContext)

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
taraskhomyn0333: done!

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
taraskhomyn0333: done!

viewer/packages/metrics/src/Fingerprint/index.ts
taraskhomyn0333: done!

viewer/packages/metrics/src/Fingerprint/index.ts
taraskhomyn0333: done!

viewer/packages/metrics/src/Fingerprint/AudioFingerprint.ts
haakonflatval-cognite: I'm still not convinced that there is a difference :thinking: Why would it not work to just remove the `new Promise` and `try`-clause? Then we would still return a `Promise<string>` which would reject if something went wrong?

viewer/packages/metrics/src/Fingerprint/index.ts
taraskhomyn0333: done!

viewer/packages/metrics/src/Fingerprint/types.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+export type PartialOfflineAudioContext = {
+  new (numberOfChannels: number, length: number, sampleRate: number): PartialOfflineAudioContextInstance;
+};
+
+export type GlobalDependencies = {
+  document: PartialDocument;
+  OfflineAudioContext: PartialOfflineAudioContext;
+};
```
haakonflatval-cognite, line 43: I will just mention that it might have made it easier to mock if the `GlobalDependencies` contained a function for creating the `PartialOfflineAudioContext`, instead of an actual constructor. But no need to change it, it's fine as it is 👍

haakonflatval-cognite (approved):
LGTM 👍

viewer/packages/metrics/src/Fingerprint/types.ts
taraskhomyn0333: got it!

PR #5330, opened 2025-09-16
feat(react-components): support for applying scene quality settings

pramod-cog:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5998

## Description :pencil:
This PR adds support for applying quality settings from the scene data to the respective scenes when `Scene` is loaded. We have used wild card properties for scene so as to support existing scene without new Quality properties, this is safe guarded by validating the response with type guard.

Note:
The test coverage is comparatively low due to missing test for changes in `use3dScenes` and `useSceneConfig` hooks which does not have test currently.
I will have to refactor the both the hooks to make it testable which I have to do it separately in new PR, as this task work has already made me to create 3 different PR's of refactoring, I am inclined towards not adding test to these file in this PR.
- Refactoring of use3dScenes hook [Draft PR](https://github.com/cognitedata/reveal/pull/5339)
- Refactoring of useSceneConfig hook [Draft PR](https://github.com/cognitedata/reveal/pull/5340)

In fusion

- Link local `react-components` build with `fusion` code base and run `unified-3d`
- Load `3d-test-aws` project and load `Anders test 55` scene
- Click on `Setting` button from `Toolbar`, we should be having `Rendering Quality` set to `4` and `point size` set to `0.5`
- Switch to any other existing test which does not have new Quality properties (e.g `new test scene`), the applicato should load the scene with out breaking


react-components/src/components/SceneContainer/hooks/useQualitySettingsFromScene.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+  useEffect(() => {
+    const qualitySettings = scene?.sceneConfiguration.qualitySettings;
+
+    if (qualitySettings === undefined) {
+      return;
+    }
+
+    applyQualitySettingsToRenderTarget(renderTarget, qualitySettings);
+  }, [scene?.sceneConfiguration.qualitySettings, renderTarget]);
```
danpriori, line 28: Is it possible to not use `useEffect` here and use probably `useMemo`? I think there is no need to use the effect hook here as this is just updating the renderTarget settings. Or involving peeking the value from the signals it needs to be under an effect hook?

react-components/src/components/SceneContainer/hooks/useQualitySettingsFromScene.tsx
```diff
@@ -0,0 +1,38 @@
...
+  applyQualitySettingsToRenderTarget
+} from '../../../utilities/pointCloudSettings';
+
+type UseQualitySettingsFromSceneResult = {
+  onPointCloudSettingsCallback: () => void;
+};
+
+export const useQualitySettingsFromScene = (
+  sceneExternalId: string,
+  sceneSpaceId: string
```
danpriori, line 14: To keep the consistency, maybe is it better to use the `DMSUniqueIdentifier` instead of separating into two arguments here? Not a big deal but as we use the identifier in most of the places, maybe it will be good to use it here 🤔

react-components/src/components/SceneContainer/SceneContainer.test.tsx
```diff
@@ -95,11 +99,58 @@ describe(SceneContainer.name, () => {
...
+      onPointCloudSettingsCallback: mockOnPointCloudSettingsCallback
+    });
+
+    render(<SceneContainer {...mockProps} />, { wrapper });
+
+    const reveal3DResourcesCall = sceneContainerDependencies.Reveal3DResources.mock.calls[0];
+    const passedProps = reveal3DResourcesCall[0];
+    const combinedCallback = passedProps.onResourcesAdded;
+
+    assert(combinedCallback !== undefined, 'onResourcesAdded should be defined');
```
danpriori, line 151: This test is a bit confusing 🙈 .
The title said "even when onRersourcesAdded is not provided" but then it checks the combinedCallback from within the onRsourcesAdded and then it asserts with the message "onResourcesAdded should be defined"

🤔 
Am I missing something?

react-components/src/components/SceneContainer/SceneContainer.viewmodel.tsx
```diff
@@ -22,10 +23,16 @@ export function useSceneContainerViewModel({
   useSkyboxFromScene(sceneExternalId, sceneSpaceId);
   useLoadPoisForScene(sceneExternalId, sceneSpaceId);
 
+  const { onPointCloudSettingsCallback } = useQualitySettingsFromScene(
+    sceneExternalId,
+    sceneSpaceId
```
danpriori, line 28: If you agree with using the dms unique identifier here

react-components/src/hooks/scenes/sceneResponseTypeGuard.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      expect(
+        isSceneConfigurationProperties({
+          ...createValidSceneConfig(),
+          ...allUndefinedOverrides
+        })
+      ).toBe(true);
+    });
+
+    it('should validate quality settings properties with valid values', () => {
```
danpriori: I guess this test is covering the same as the first ones on the top for valid values 'should return true for valid scene configuration', isnt?

react-components/src/hooks/scenes/types.ts
```diff
@@ -156,27 +156,7 @@ export const transformationSourceWithProperties = [
...
-      'updatedAt',
-      'cadBudget',
-      'pointCloudBudget',
-      'maxRenderResolution',
-      'movingCameraResolutionFactor',
-      'pointCloudPointSize',
-      'pointCloudPointShape',
-      'pointCloudColor'
-    ]
+    properties: ['*']
```
danpriori, line 159: Just to confirm: this wildcard it's for the dms query, right?

react-components/src/hooks/scenes/use3dScenes.tsx
```diff
@@ -112,62 +122,56 @@ export const use3dScenes = (
...
-            const skyboxProperties = Object.values(Object.values(connectedSkybox.properties)[0])[0];
-            skyboxObject = skyboxProperties;
-          }
+  return scenes.reduce((acc: Record<Space, Record<ExternalId, SceneData>>, scene: SceneNode) => {
+    const { space, externalId } = scene;
+    const properties = scene.properties.scene['SceneConfiguration/v1'];
+    if (acc[space] === undefined) {
+      acc[space] = {};
+    }
+    if (acc[space][externalId] === undefined) {
```
danpriori, line 133:
```suggestion
    const { space, externalId } = scene;
    if (acc[space] === undefined) {
      acc[space] = {};
    }
    if (acc[space][externalId] === undefined) {
      const properties = scene.properties.scene['SceneConfiguration/v1'];
```

react-components/src/hooks/scenes/useSceneConfig.ts
```diff
@@ -147,18 +154,27 @@ async function sceneViewsExist(fdmSdk: FdmSDK): Promise<boolean> {
   return await fdmViewsExist(fdmSdk, neededViews);
 }
 
+function extractSceneProperties(
+  scenesProperties: Record<string, Record<string, unknown>>
+): SceneConfigurationProperties {
+  const currentSceneProperties = scenesProperties.scene['SceneConfiguration/v1'];
+  const sceneConfigurationProperties = isSceneConfigurationProperties(currentSceneProperties);
+  if (!sceneConfigurationProperties) {
```
danpriori:
```suggestion
  const currentSceneProperties = scenesProperties.scene['SceneConfiguration/v1'];
  if (isSceneConfigurationProperties(currentSceneProperties) === false) {
```

react-components/src/components/SceneContainer/hooks/useQualitySettingsFromScene.tsx
pramod-cog: `useEffect` is needed here as the Scene quality settings are to be assigned to `renderTarget`

react-components/src/components/SceneContainer/hooks/useQualitySettingsFromScene.tsx
pramod-cog: As other hooks (`useGroundPlaneFromScene`, `useSkyboxFromScene`, `useLoadPoisForScene`) use same pattern, I have followed the same here

react-components/src/components/SceneContainer/SceneContainer.viewmodel.tsx
pramod-cog: as above comment

react-components/src/hooks/scenes/types.ts
pramod-cog: Yes

PR #5338, opened 2025-09-19
fix(react-components): port datetime format utils from cdf utilities and replace related places to fix cdf-utilities package issues

danpriori:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:

We're having compatibility issues with the cdf-utilities package between react-components and some subapps in Fusion, such as fdx-pages, which also uses a datetime function. So, the solution for now is to port the function used in react-components to avoid importing cdf-utilities within react-components and avoid compatibility issues in tests.

The code is practically the same as that already in production for cdf-utilities.
This PR also redirects the locations that use the function that was previously imported from cdf-utilities.

It was tested in Fusion (in the locations where the Poi Header and image 360 ​​historical details are located), where datetimes are displayed. 
Unit tests were also tested in both react-components and Fusion.

**Note**: Most of the lines changed are in relation to yarn.lock, the rest is basically porting the code from cdf-utilities into react-components as explained above.

### DIFF
--- a/react-components/package.json
+++ b/react-components/package.json
@@ -45,13 +45,13 @@
     "styled-components": "^6.1.12"
   },
   "dependencies": {
-    "@cognite/cdf-utilities": "^5.0.0",
     "@cognite/cogs-lab": "9.0.0-alpha.160",
     "@cognite/cogs.js": "^10.36.0",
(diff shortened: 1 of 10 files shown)


christjt (approved):
risk review ok

PR #5339, opened 2025-09-19
refactor(react-components): refactor use3dscenes hook and added test

pramod-cog:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

[BND3D-5322](https://cognitedata.atlassian.net/browse/BND3D-5322)

## Description :pencil:
Refactored `use3dscenes` hook and added test. This is first part of [BND3D-5322](https://cognitedata.atlassian.net/browse/BND3D-5322) tasks

In fusion

- Link local build of `react-components` to `fusion`
- Load `unified-3d` application
- Test listing of all scenes and select individual scenes from `Resource` selector

[BND3D-5322]: https://cognitedata.atlassian.net/browse/BND3D-5322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react-components/src/components/Image360CollectionContainer/Image360CollectionContainer.context.ts
+++ b/react-components/src/components/Image360CollectionContainer/Image360CollectionContainer.context.ts
@@ -1,0 +1,33 @@
+import { createContext } from 'react';
+
+import {
+  useReveal3DResourceLoadFailCount,
+  useReveal3DResourcesCount
+} from '../Reveal3DResources/Reveal3DResourcesInfoContext';
+import { useRenderTarget } from '../RevealCanvas';
+import { useRevealKeepAlive } from '../RevealKeepAlive/RevealKeepAliveContext';
+import { RevealModelsUtils } from '../../architecture/concrete/reveal/RevealModelsUtils';
+
+export type Image360CollectionContainerContextDependencies = {
+  useRenderTarget: typeof useRenderTarget;
+  useRevealKeepAlive: typeof useRevealKeepAlive;
+  useReveal3DResourcesCount: typeof useReveal3DResourcesCount;
+  useReveal3DResourceLoadFailCount: typeof useReveal3DResourceLoadFailCount;
+  createImage360CollectionDomainObject: typeof RevealModelsUtils.addImage360Collection;
+  removeImage360CollectionDomainObject: typeof RevealModelsUtils.remove;
+};
+
+export const defaultImage360CollectionContainerContextDependencies: Image360CollectionContainerContextDependencies =
+  {
+    useRenderTarget,
+    useRevealKeepAlive,
+    useReveal3DResourcesCount,
+    useReveal3DResourceLoadFailCount,
+    createImage360CollectionDomainObject: RevealModelsUtils.addImage360Collection.bind(this),
+    removeImage360CollectionDomainObject: RevealModelsUtils.remove.bind(this)
+  };
+
+export const Image360CollectionContainerContext =
+  createContext<Image360CollectionContainerContextDependencies>(
+    defaultImage360CollectionContainerContextDependencies
+  );
--- a/react-components/src/components/Image360CollectionContainer/Image360CollectionContainer.test.tsx
+++ b/react-components/src/components/Image360CollectionContainer/Image360CollectionContainer.test.tsx
@@ -1,0 +1,130 @@
+import { render, waitFor } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { Image360CollectionContainer } from './Image360CollectionContainer';
+import {
+  Image360CollectionContainerContext,
+  defaultImage360CollectionContainerContextDependencies
+} from './Image360CollectionContainer.context';
+import type { PropsWithChildren, ReactElement } from 'react';
+import { Mock } from 'moq.ts';
+import type { RevealRenderTarget } from '../../architecture';
+import {
+  type Image360Collection,
+  type Cognite3DViewer,
+  type ClassicDataSourceType
+} from '@cognite/reveal';
+import { getMocksByDefaultDependencies } from '#test-utils/vitest-extensions/getMocksByDefaultDependencies';
+import type { AddImage360CollectionOptions } from '../Reveal3DResources/types';
+import * as useApply360AnnotationStylingModule from './useApply360AnnotationStyling';
+
+describe(Image360CollectionContainer.name, () => {
+  const deps = getMocksByDefaultDependencies(defaultImage360CollectionContainerContextDependencies);
+
+  const wrapper = ({ children }: PropsWithChildren): ReactElement => (
+    <Image360CollectionContainerContext.Provider value={deps}>
+      {children}
+    </Image360CollectionContainerContext.Provider>
+  );
+
+  const collectionsGetter = vi.fn<() => Array<Image360Collection<ClassicDataSourceType>>>();
+  const image360Collection = new Mock<Image360Collection<ClassicDataSourceType>>()
+    .setup((x) => x.id)
+    .returns('test-site-id')
+    .setup((x) => x.setModelTransformation)
+    .returns(vi.fn())
+    .setup((x) => x.set360IconCullingRestrictions)
+    .returns(vi.fn())
+    .object();
+
+  const renderTargetMock = new Mock<RevealRenderTarget>()
+    .setup((x) => x.viewer)
+    .returns(
+      new Mock<Cognite3DViewer>()
+        .setup((x) => x.models)
+        .returns([])
+        .setup((x) => x.get360ImageCollections)
+        .returns(() => collectionsGetter())
+        .object()
+    );
+
+  beforeEach(() => {
+    deps.useRenderTarget.mockReturnValue(renderTargetMock.object());
+
+    deps.useRevealKeepAlive.mockReturnValue({
+      renderTargetRef: { current: renderTargetMock.object() },
+      isRevealContainerMountedRef: { current: true },
+      sceneLoadedRef: { current: undefined }
+    });
+
+    deps.useReveal3DResourcesCount.mockReturnValue({
+      reveal3DResourcesCount: 0,
+      setRevealResourcesCount: vi.fn()
+    });
+
+    deps.useReveal3DResourceLoadFailCount.mockReturnValue({
+      reveal3DResourceLoadFailCount: 0,
+      setReveal3DResourceLoadFailCount: vi.fn()
+    });
+
+    deps.createImage360CollectionDomainObject.mockResolvedValue(image360Collection);
+    collectionsGetter.mockReturnValue([]);
+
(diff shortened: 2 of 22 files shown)


react-components/src/hooks/scenes/allScenesQuery.ts
```diff
@@ -0,0 +1,128 @@
...
+  groundPlaneSourceWithProperties,
+  image360CollectionSourceWithProperties,
+  revisionSourceWithProperties,
+  SCENE_QUERY_LIMIT,
+  sceneSourceWithProperties,
+  transformationSourceWithProperties
+} from './types';
+
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
+export function createGetScenesQuery(limit: number = SCENE_QUERY_LIMIT, sceneCursor?: string) {
```
haakonflatval-cognite: There is a trick we have used in 3d-management where we declare a non-exported function in the module without type annotation, and then call it from the exported function, and annotate the exported function with `ReturnType<typeof nonExportedFunction>`. It could be done here as well, I guess it's nice for enabling some TS compiler flags, but not sure how relevant they are here (but I know it's relevant in the Fusion repo)

react-components/src/hooks/scenes/use3dScenes.context.ts
```diff
@@ -0,0 +1,24 @@
+import { createContext } from 'react';
+import { type CogniteClient } from '@cognite/sdk';
+import { type QueryFunction, type UseQueryResult, useQuery } from '@tanstack/react-query';
+import { useSDK } from '../../components/RevealCanvas/SDKProvider';
+import { FdmSDK, type FdmSDK as FdmSDKType } from '../../data-providers/FdmSDK';
+
+export type Use3dScenesDependencies = {
+  useSDK: (userSdk?: CogniteClient) => CogniteClient;
+  useQuery: <T>(options: {
```
haakonflatval-cognite: I would personally rather just add a `QueryProvider` in the test instead of mocking out the entire `useQuery` - then we are able to test the logic inside the `useQuery` as well

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+  const mockCreateFdmSdk = vi.fn();
+  const mockQueryNodesAndEdges = vi.fn();
+
+  const mockFdmSdkInstance = new Mock<FdmSDK>()
+    .setup((instance) => instance.queryNodesAndEdges)
+    .returns(mockQueryNodesAndEdges)
+    .object();
+
+  const mockDependencies: Use3dScenesDependencies = {
+    useSDK: mockUseSDK,
```
haakonflatval-cognite: You could also create this mock dependency object with `getMocksByDefaultDependencies`

react-components/src/hooks/scenes/types.ts
```diff
@@ -182,3 +192,31 @@ export const environmentMapSourceWithProperties = [
     properties: ['label', 'file', 'isSpherical']
   }
 ] as const satisfies SourceSelectorV3;
+
+export type Space = string;
+export type ExternalId = string;
+
```
haakonflatval-cognite: I'm fairly sure we have these types other places in the code as well, can we reuse those?

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+
+  test('should return empty data when no scenes', () => {
+    mockUseQuery.mockReturnValue(createMockQueryResult({}));
+    const { result } = renderHook(() => use3dScenes(), { wrapper });
+
+    expect(result.current.data).toEqual({});
+  });
+
+  test('should handle refetch functionality', () => {
+    const mockRefetch = vi.fn();
```
haakonflatval-cognite: Is there a particular reason this is tested? `refetch` is never mentioned in the code, and I would personally argue that it's just an attribute of `useQuery`, which `use3dScenes` shouldn't need to think too much about

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+
+    expect(result).toBeDefined();
+    expect(result['test-space']).toBeDefined();
+    expect(result['test-space']['test-scene-1']).toBeDefined();
+    expect(result['test-space']['test-scene-1'].name).toBe('Scene test-scene-1');
+  });
+
+  test('should handle pagination', async () => {
+    const createScenes = (count: number, start: number = 0): SceneNode[] =>
+      Array.from({ length: count }, (_, i) => createSceneNode(`scene-${start + i}`, 'test-space'));
```
haakonflatval-cognite, line 98: I didn't know one could do this 👀 Nice!

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+          scene360Collections: [],
+          sceneGroundPlanes: [],
+          sceneGroundPlaneEdges: [],
+          sceneSkybox: []
+        },
+        nextCursor: { scenes: 'cursor-token' }
+      })
+      .mockResolvedValueOnce({
+        items: {
+          scenes: createScenes(50, 100),
```
haakonflatval-cognite: Could we use `setupMockResponse` to simplify this code too, or wouldn't that work?

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+                revisionId: 1,
+                translationX: 10,
+                translationY: 20,
+                translationZ: 30,
+                eulerRotationX: 0,
+                eulerRotationY: 0,
+                eulerRotationZ: 0,
+                scaleX: 1,
+                scaleY: 1,
+                scaleZ: 1
```
haakonflatval-cognite: Can we put these transformation fields in a separate variable and reuse it across tests?

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+    });
+
+    const result = await executeQueryFn();
+    const scene = result['test-space']['scene-with-model'];
+
+    expect(scene.modelOptions).toHaveLength(1);
+    expect(scene.modelOptions[0]).toEqual({
+      modelId: 123,
+      revisionId: 1,
+      transform: expect.any(Object)
```
haakonflatval-cognite: ... Then we could also check that the transformation is right on the other end

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+      });
+
+    const result = await executeQueryFn();
+    const allScenes = Object.values(result).flatMap((spaceScenes) => Object.values(spaceScenes));
+
+    expect(allScenes).toHaveLength(150);
+    expect(mockQueryNodesAndEdges).toHaveBeenCalledTimes(2);
+  });
+
+  test('should populate scene with models', async () => {
```
haakonflatval-cognite: Nitpick: The phrasing here is a bit weird - what do you mean by `should populate scene with models`? (Same for the next few tests too)

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+  test('should populate scene with ground planes', async () => {
+    setupMockResponse({
+      scenes: [createSceneNode('scene-with-groundplane', 'test-space')],
+      sceneGroundPlanes: [
+        {
+          externalId: 'ground-plane-1',
+          space: 'test-space',
+          properties: {
+            'GroundPlane/v1': {
+              'GroundPlane/v1': {
```
haakonflatval-cognite: Nitpick: This response structure seems weird (i.e. `GroundPlane/v1` is used twice as a key) - I assume it's not how the real data looks?

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+  });
+
+  test('should handle complex scene with all components', async () => {
+    setupMockResponse({
+      scenes: [
+        createSceneNode('complex-scene', 'test-space', {
+          properties: {
+            scene: {
+              'SceneConfiguration/v1': {
+                name: 'Complex Scene',
```
haakonflatval-cognite: Would be nice if this test could re-use fixtures from the previous tests, so that it's becomes a bit more concise

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+    let capturedQueryFunction!: QueryFunction<ScenesMap>;
+    mockUseQuery.mockImplementation((options) => {
+      capturedQueryFunction = options.queryFn;
+      return createMockQueryResult({});
+    });
+
+    renderHook(() => use3dScenes(), { wrapper });
+    assert(capturedQueryFunction !== undefined);
+    return await capturedQueryFunction(mockContext);
+  };
```
haakonflatval-cognite: I hope we can just remove this part if we use an actual `QueryClient` object instead of mocking `useQuery` directly

react-components/src/hooks/scenes/use3dScenes.test.tsx
```diff
@@ -0,0 +1,503 @@
...
+
+  const setupMockResponse = (responseData: any): void => {
```
haakonflatval-cognite: `any`?? 👀

react-components/tests/tests-utilities/fixtures/queryResult.ts
```diff
@@ -1,21 +1,29 @@
-import { type UseQueryResult } from '@tanstack/react-query';
+import { QueryClient, type QueryFunctionContext, type UseQueryResult } from '@tanstack/react-query';
 import { Mock } from 'moq.ts';
 import { vi } from 'vitest';
 
-export function createMockQueryResult<T>(data: T, isFetching = false): UseQueryResult<T> {
+export function createMockQueryResult<T>(
+  data: T,
+  isFetching = false,
+  isLoading = false,
```
haakonflatval-cognite: We may also be able to get away from changes in this file by using a real `QueryClient` in the tests.

haakonflatval-cognite (changes requested):
Had a few comments, but otherwise looks nice 👍

react-components/src/hooks/scenes/use3dScenes.test.tsx
haakonflatval-cognite: I don't see that this was addressed :eyes: Not cruical, though

haakonflatval-cognite (approved):
LGTM again :+1:

PR #5343, opened 2025-09-23
feat(react-components): Make custom top banner in settings button

anders-hopland:
#### Type of change

Feat  

## Description :pencil:
Adds custom banner command and uses it for enabling setting a top banner in the settings button. See image below for how this looks
Also setting a `max-width` for the Settings panel and increasing the `max-height` by `100px` to accomodate for more settings. This has been discussed with the designer

Has been tested in Fusion

## Test instructions
Try it in either storybook or in fusion. Code for using it in settings button
```
<RevealButtons.Settings
  toolbarPlacement="right-end"
  topBanner={{
    text: 'Testing top banner content',
  }}
/>
```

## Image
image


react-components/src/components/Architecture/SettingsButton.tsx
```diff
@@ -252,6 +252,7 @@ const StyledToggleContainer = styled(Flex).attrs({
 const StyledMenuPanel = styled.div<{ $flexDirection: FlexDirection }>`
   max-height: 600px;
   min-width: 340px;
+  max-width: 400px;
```
haakonflatval-cognite, line 255: We should align with the designer on exactly how the panel should behave when we add new, broader elements

react-components/src/architecture/base/concreteCommands/general/SettingsCommand.test.ts
```diff
@@ -11,17 +13,41 @@ describe(SettingsCommand.name, () => {
...
+  test('should only show default settings when no extra settings are provided', async () => {
+    const command = new SettingsCommand(false, false, undefined);
     expect(command.hasChildren).toBe(true);
     expect(command.children.length).toBe(11);
   });
 
+  test('should show 360 image settings when include360Images is true', async () => {
+    const command = new SettingsCommand(true, false, undefined);
+    expect(command.hasChildren).toBe(true);
+    expect(command.children.length).toBe(18);
```
haakonflatval-cognite, line 36: Not a big fan of expecting an exact amount of elements in the test - then we have to change all the test if we add a new common command 🙈 But I see that that's the norm in this file..

react-components/src/components/Architecture/SettingsButton.tsx
anders-hopland: I have talked with design. `max-width` of `400px` is good. I have also increase the `max-height`by `100px`

haakonflatval-cognite (approved):
LGTM 👍

PR #5353, opened 2025-09-25
fix: allow cached sectors to be used across different models with same modelId/revisionId part3

pramod-cog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5696

## Description :pencil:
This is third part where textured mesh creation is handle in `CadMeshManager` 


viewer/packages/cad-model/src/wrappers/CadMeshManager.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    return group;
+  }
+
+  /**
+   * Removes mesh group for the given sector ID.
+   * Note: This does not dispose the underlying geometries as they may be shared with other models.
+   * @param sectorId The sector ID whose mesh group should be removed
+   */
+  public removeSectorMeshGroup(sectorId: number): void {
```
haakonflatval-cognite: I personally think this should be private - we shouldn't be able to remove a mesh group without also dereferencing it, right?

viewer/packages/sector-loader/src/GltfSectorRepository.ts
```diff
@@ -78,6 +93,19 @@ export class GltfSectorRepository implements SectorRepository {
...
   }
 
+  /**
+   * Dereferences a sector when a model stops using it.
+   * The cache handles reference counting and automatic disposal when count reaches zero.
+   * @param modelIdentifier The model identifier that was using the sector
+   * @param sectorId The sector ID to dereference
+   */
+  dereferenceSector(modelIdentifier: ModelIdentifier, sectorId: number): void {
+    const cacheKey = modelIdentifier.sourceModelIdentifier() + '.' + sectorId;
```
haakonflatval-cognite: Use `wantedSectorCacheKey` here

viewer/packages/utilities/src/cache/MemoryRequestCache.ts
```diff
@@ -108,6 +110,42 @@ export class MemoryRequestCache<Key, Data> implements RequestCache<Key, Data> {
     }
   }
 
+  /**
+   * Adds a reference to the cached item. Items with active references will not be disposed
+   * until all references are removed.
+   * @param id The cache key to add a reference to
+   */
+  addReference(id: Key): void {
+    const currentCount = this._referenceCounts.get(id) || 0;
```
haakonflatval-cognite: I think this should be 
```suggestion
    const currentCount = this._referenceCounts.get(id) ?? 0;
```
, same with the other places this is used in this file

viewer/packages/cad-geometry-loaders/src/CadManager.ts
```diff
@@ -86,13 +85,8 @@ export class CadManager {
...
         // Also clean up any mesh groups created from parsed geometries
-        cadModel.removeSectorMeshGroup(sector.metadata.id);
-      }
-
-      // Create meshes from parsedMeshGeometries data
-      let meshGroup: AutoDisposeGroup | undefined = undefined;
-      if (sector.parsedMeshGeometries && sector.parsedMeshGeometries.length > 0) {
-        meshGroup = cadModel.createMeshesFromParsedGeometries(sector.parsedMeshGeometries, sector.metadata.id);
+        cadModel.removeSectorMeshGroupWithDereferencing(sector.metadata.id);
+        return;
```
haakonflatval-cognite: Not sure if adding `return` here is safe.. In particular because of the `sectorNode.updateGeometry(...)` call we do further down in this function

viewer/packages/cad-geometry-loaders/src/CadManager.ts
```diff
@@ -101,14 +95,21 @@ export class CadManager {
...
+      // Update the sector node with the mesh group
       if (meshGroup) {
         sectorNode.add(meshGroup);
       }
       sectorNode.updateGeometry(meshGroup, sector.levelOfDetail);
 
+      // Apply render layers to the sector node
       if (meshGroup) {
-        cadModel.setModelRenderLayers(meshGroup);
+        cadModel.setModelRenderLayers(sectorNode);
```
haakonflatval-cognite: Is this change intended? It seems strange

viewer/packages/cad-model/src/wrappers/CadNode.ts
```diff
@@ -229,112 +226,27 @@ export class CadNode extends Object3D<Object3DEventMap & { update: undefined }>
...
-    const treeIndexSet = new Map<number, number>();
+  public createMeshesFromParsedGeometries(parsedMeshGeometries: ParsedMeshGeometry[], sectorId: number): Group {
+    const managedSectorIds = this._meshManager.getManagedSectorIds();
 
-    for (let i = 0; i < treeIndexAttribute.count; i++) {
-      incrementOrInsertIndex(treeIndexSet, (treeIndexAttribute as BufferAttribute).getX(i));
+    // Check if we already have meshes for this sector - if so, we need to dereference the old sector first
+    if (managedSectorIds.includes(sectorId)) {
+      // This is a sector update/replacement - dereference the old one first
+      this._meshManager.removeSectorMeshGroupAndDereference(sectorId, this._sectorRepository, this._modelIdentifier);
```
haakonflatval-cognite: In what cases do we need to replace a sector mesh group? Can the new mesh group be different from the old?

viewer/packages/cad-model/src/wrappers/CadMeshManager.ts
```diff
@@ -0,0 +1,210 @@
...
+   */
+  public createMeshesFromParsedGeometries(parsedMeshGeometries: ParsedMeshGeometry[], sectorId: number): Group {
+    const materials = this._materialManager.getModelMaterials(this._modelIdentifier);
+    const group = new Group();
+    const allTreeIndices = new Set<number>();
+
+    parsedMeshGeometries.forEach(geometryData => {
+      if (geometryData.type === RevealGeometryCollectionType.TriangleMesh) {
+        this.createMeshFromGeometry(
+          group,
```
haakonflatval-cognite: The branches of this if-else seems to be mostly the same - can we merge them? E.g. define the material as 
```
const material = geometryData.texture ? this._materialManager.addTexturedMeshMaterial(
            this._modelIdentifier,
            sectorId,
            geometryData.texture
          ) : this._materialManager.getModelMaterials(this._modelIdentifier).triangleMesh;
```
and reuse the rest?

viewer/packages/cad-model/src/wrappers/CadNode.ts
```diff
@@ -209,6 +225,30 @@ export class CadNode extends Object3D<Object3DEventMap & { update: undefined }>
     this._geometryBatchingManager?.removeSectorBatches(sectorId);
   }
 
+  public removeSectorMeshGroup(sectorId: number): void {
```
haakonflatval-cognite: I think this can be removed

viewer/packages/sector-loader/src/GltfSectorRepository.ts
```diff
@@ -5,17 +5,27 @@
...
+
+    // Create cache with disposal callback that properly disposes GPU resources
+    this._gltfCache = new MemoryRequestCache(200, 50, (sector: ConsumedSector) => {
+      sector.parsedMeshGeometries?.forEach(mesh => {
+        mesh.geometryBuffer.dispose();
+        mesh.texture?.dispose();
+      });
+      sector.geometryBatchingQueue?.forEach(batch => {
+        batch.geometryBuffer.dispose();
+        batch.texture?.dispose();
```
haakonflatval-cognite, line 26: Not sure if `texture` will ever be defined here - but we can keep it if you prefer that

viewer/packages/cad-model/src/wrappers/CadNode.ts
```diff
@@ -209,6 +225,30 @@ export class CadNode extends Object3D<Object3DEventMap & { update: undefined }>
...
+   */
+  public removeSectorMeshGroupWithDereferencing(sectorId: number): void {
+    this._meshManager.removeSectorMeshGroupAndDereference(sectorId, this._sectorRepository, this._modelIdentifier);
+  }
+
+  public createMeshesFromParsedGeometries(parsedMeshGeometries: ParsedMeshGeometry[], sectorId: number): Group {
+    const managedSectorIds = this._meshManager.getManagedSectorIds();
+
+    // Check if we already have meshes for this sector - if so, we need to dereference the old sector first
+    if (managedSectorIds.includes(sectorId)) {
```
haakonflatval-cognite: I think I would prefer e.g. a `hasManagedSector(...)` method on the mesh manager to simplify and speed up this, but not critical

viewer/packages/utilities/src/cache/MemoryRequestCache.test.ts
```diff
@@ -44,20 +44,318 @@ describe('MemoryRequestCache', () => {
...
+  test('dispose callback is not called when not provided', () => {
+    const cache = new MemoryRequestCache<number, string>(10, 5);
+
+    cache.insert(1, 'test1');
+    cache.remove(1);
+    cache.insert(2, 'test2');
+    cache.clear();
+
+    // Should not throw or cause any issues
+    expect(true).toBe(true);
```
haakonflatval-cognite: Can we instead have e.g. a `expect(() => { ... }).not.toThrow()` to test this?

viewer/packages/utilities/src/cache/MemoryRequestCache.test.ts
```diff
@@ -44,20 +44,318 @@ describe('MemoryRequestCache', () => {
...
+
+    cache.insert(1, 'test1');
+    const result = cache.get(1);
+
+    expect(result).toBe('test1');
+    expect(disposeCallback).not.toHaveBeenCalled();
+  });
+
+  // Reference counting tests
+  describe('Reference Counting', () => {
```
haakonflatval-cognite: I think this comment is unnecessary 😅

viewer/packages/utilities/src/cache/MemoryRequestCache.test.ts
```diff
@@ -44,20 +44,318 @@ describe('MemoryRequestCache', () => {
...
+      // Force cache cleanup by inserting when full
+      cache.forceInsert(3, 'test3');
+
+      // test1 should be disposed (unreferenced), test2 should remain (referenced)
+      expect(cache.has(2)).toBe(true); // Still referenced
+      expect(cache.has(3)).toBe(true); // Newly inserted
+      expect(disposeCallback).toHaveBeenCalledWith('test1');
+    });
+
+    test('removeReference should not dispose item with active references', () => {
```
haakonflatval-cognite: This test and the one on line 245 seem very similar - can they be merged? (I think it would also be good to test the case where everything in the cache is referenced, and an element is force-inserted, which means the cache will contain more elements than its capacity)

viewer/packages/sector-loader/tests/GltfSectorRepository.test.ts
```diff
@@ -52,13 +47,306 @@ describe(GltfSectorRepository.name, () => {
...
+
+  test('cache should work with different model identifiers', async () => {
+    const [modelId1, modelId2] = ['model1', 'model2'].map(id => new LocalModelIdentifier(id));
+    const testMetadata = { id: 1, sectorFileName: 'test.glb', downloadSize: 100 };
+
+    const [sector1, sector2] = [modelId1, modelId2].map(modelId =>
+      createWantedSectorWithMetadata(testMetadata, LevelOfDetail.Detailed, modelId)
+    );
+
+    const loaderSpy = jest.spyOn(binaryFileProvider.object(), 'getBinaryFile');
```
haakonflatval-cognite: Instead of spying like this, can we just write `vi.mocked(binaryFileProvider.object().getBinaryFile)`?

viewer/packages/cad-geometry-loaders/src/CadManager.ts
pramod-cog: `sectorNode` is extends `threejs` `Group`, I do not see an issue passing it

viewer/packages/sector-loader/tests/GltfSectorRepository.test.ts
```diff
@@ -52,13 +47,306 @@ describe(GltfSectorRepository.name, () => {
...
+      geometryBatchingQueue: [
+        {
+          geometryBuffer: mockBatchGeometryBuffer,
+          texture: mockBatchTexture
+        }
+      ]
+    };
+
+    // Access private cache to directly test disposal callback
+    const cache = (sectorRepository as any)._gltfCache;
```
haakonflatval-cognite: Is there any way we can avoid accessing this private member? Do we really need it to perform the test? I.e. could we instead call `loadSector` on the `GltfSectorRepository` instance with different sector IDs to force the cache items to be evicted?

viewer/packages/cad-model/src/wrappers/CadNode.ts
pramod-cog: Sector replacement scenarios includes LOD changes - sector loaded at different detail levels and cache eviction and reload

viewer/packages/sector-loader/src/GltfSectorRepository.ts
pramod-cog: `geometryBatchingQueue` contains `texture` property, even though texture object is not filled, I would prefer have the clean, just in case

viewer/packages/cad-model/src/wrappers/CadMeshManager.test.ts
```diff
@@ -0,0 +1,407 @@
...
+    const geometry = createBasicGeometry();
+    const parsedGeometries: ParsedMeshGeometry[] = [
+      {
+        type: RevealGeometryCollectionType.TexturedTriangleMesh,
+        geometryBuffer: geometry,
+        wholeSectorBoundingBox: new Box3(new Vector3(-1, -1, -1), new Vector3(1, 1, 1))
+      }
+    ];
+
+    const consoleSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
```
haakonflatval-cognite: I don't think you need to mock the implementation in this case? 🤔

viewer/packages/cad-model/src/wrappers/CadMeshManager.test.ts
```diff
@@ -0,0 +1,407 @@
...
+    expect(firstGroup.children.length).toBe(0); // First group should be disposed
+
+    const secondGroup = meshManager.createMeshesFromParsedGeometries(parsedGeometries, 150);
+    expect(secondGroup.children.length).toBe(1); // Second group should be active
+    expect(secondGroup).not.toBe(firstGroup);
+  });
+
+  test('should remove mesh group from scene without disposing shared geometries', () => {
+    const geometry = createBasicGeometry();
+    const disposeSpy = jest.spyOn(geometry, 'dispose');
```
haakonflatval-cognite: I would prefer to just use e.g. `vi.mocked(geometry.dispose)` (provided that the `dispose` method is mocked), but it's alright if you prefer this

viewer/packages/cad-model/src/wrappers/CadMeshManager.ts
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    // Update tree index mapping if we have indices
+    if (allTreeIndices.size > 0) {
+      for (const treeIndex of allTreeIndices) {
+        this._treeIndexToSectorsMap.set(treeIndex, sectorId);
+      }
+
+      // Mark the sector as completed for triangle mesh geometry
+      this._treeIndexToSectorsMap.markCompleted(sectorId, RevealGeometryCollectionType.TriangleMesh);
```
haakonflatval-cognite: Just a small general observation - I feel like a lot of the comments in this PR are stating the obvious and don't add much value

viewer/packages/cad-model/src/wrappers/CadMeshManager.ts
```diff
@@ -0,0 +1,210 @@
...
+        }
+      }
+    });
+
+    // Track this mesh group by sector ID for cleanup when sector is unloaded
+    this._sectorMeshGroups.set(sectorId, group);
+
+    // Update tree index to sector mapping only if there's exactly one parsed geometry
+    // (matching original behavior from CadManager)
+    if (parsedMeshGeometries.length === 1) {
```
haakonflatval-cognite: After reading through the original files, I think this condition should be ignored - we should process all mesh geometries no matter how many. Although I don't think there will ever be more than one in the current setup.

viewer/packages/cad-model/src/wrappers/CadMeshManager.test.ts
```diff
@@ -28,13 +30,28 @@ describe(CadMeshManager.name, () => {
...
+      dereferenceSector: jest.fn(),
+      loadSector: jest.fn(),
+      clearCache: jest.fn(),
+      setCacheSize: jest.fn()
+    } as unknown as SectorRepository;
+
+    mockModelIdentifier = {
+      sourceModelIdentifier: jest.fn().mockReturnValue('test-model'),
+      revealInternalId: Symbol('test-model-internal')
+    } as unknown as ModelIdentifier;
```
haakonflatval-cognite: I would prefer `new Mock<...>` instead of this unsafe casting

viewer/packages/cad-model/src/wrappers/CadMeshManager.ts
```diff
@@ -45,44 +45,34 @@ export class CadMeshManager {
...
-            geometryData.wholeSectorBoundingBox
-          );
-
-          // Collect tree indices for mapping
-          const treeIndices = this.createTreeIndexSet(geometryData.geometryBuffer);
-          for (const treeIndex of treeIndices.keys()) {
-            allTreeIndices.add(treeIndex);
-          }
-        } else {
+        } else if (isTexturedTriangleMesh && !geometryData.texture) {
```
haakonflatval-cognite, line 60: In my opinion, I think it's fine to just not check for this case, (and then we could just have a ternary operator rather than an if-else statement), but we can keep it if you want.

viewer/packages/sector-loader/src/GltfSectorRepository.ts
```diff
@@ -104,13 +104,17 @@ export class GltfSectorRepository implements SectorRepository {
...
   }
 
   private wantedSectorCacheKey(wantedSector: WantedSector) {
-    return wantedSector.modelIdentifier.sourceModelIdentifier() + '.' + wantedSector.metadata.id;
+    return this.createSectorCacheKey(wantedSector.modelIdentifier, wantedSector.metadata.id);
+  }
+
+  private createSectorCacheKey(modelIdentifier: ModelIdentifier, sectorId: number): string {
+    return modelIdentifier.sourceModelIdentifier() + '.' + sectorId;
   }
```
haakonflatval-cognite: These two functions now do exactly the same - can we remove one of them?

viewer/packages/utilities/src/cache/MemoryRequestCache.ts
```diff
@@ -116,7 +116,7 @@ export class MemoryRequestCache<Key, Data> implements RequestCache<Key, Data> {
    * @param id The cache key to add a reference to
    */
   addReference(id: Key): void {
-    const currentCount = this._referenceCounts.get(id) || 0;
+    const currentCount = this._referenceCounts.get(id) ?? 0;
```
haakonflatval-cognite, line 119: There are couple more of these `||`s in this file that should be changed

viewer/packages/cad-model/src/wrappers/CadNode.ts
haakonflatval-cognite: Ah, okay :+1:

viewer/packages/cad-geometry-loaders/src/CadManager.ts
haakonflatval-cognite: Right - it just seemed strange that this was changed for no apparent reason

viewer/packages/sector-loader/tests/GltfSectorRepository.test.ts
pramod-cog: We using `jest` in `reveal` remember, not sure if we have similar support as `vi.mocked` in `jest`

haakonflatval-cognite: Ah right - I thought there was a `jest.mocked` too, though :thinking:

pramod-cog: your right, `jest.mocked` is supported, fixed it now

viewer/packages/cad-parsers/src/sector/SectorNode.test.ts
```diff
@@ -78,10 +78,12 @@ describe(SectorNode.name, () => {
 
     // Force a small delay to ensure timestamp difference
     const now = Date.now();
-    jest.spyOn(Date, 'now').mockReturnValue(now + 100);
+    const mockDateNow = jest.fn<() => number>().mockReturnValue(now + 100);
+    (Date.now as jest.MockedFunction<() => number>) = mockDateNow;
```
haakonflatval-cognite: We should not override this - if we run tests without isolation, this will affect other tests using the `Date.now` function

haakonflatval-cognite: I think the `jest.spyOn` approach you had earlier is fine :+1: At least that one will be reset when `jest.restoreAllMocks` is called at the end of the test

PR #5364, opened 2025-10-02
fix(react-components): RevealToolbar buttons not active when toggled

michalzagojski-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:
https://cognitedata.atlassian.net/browse/](https://cognitedata.atlassian.net/browse/BND3D-4269)

## Description :pencil:

- Fixes a bug where the reveal toolbar buttons would not become active (indicated by blue background color) after being toggled. Instead, their background color would remain default grey. Now, the buttons become active as expected until clicked again or until their dropdown / selection / menu is closed (depends on the type of the button).
- Updates LayersButton test file

- Pick any model using industrial tools. 
- Notice the toolbar on the left side of 3D view.
- Click on any button which opens up a modal / selection list / dropdown
- Observe that it is now active as expected, which was not the case before the PR

Note: buttons which do not toggle any kind of modal / dropdown will not become active - this is expected behaviour.


react-components/src/components/RevealToolbar/LayersButton/LayersButton.tsx
```diff
@@ -18,21 +18,39 @@ export const LayersButton = ({
 }: LayersButtonProps): ReactElement => {
   const { t } = useTranslation();
 
+  const [layersActive, setLayersActive] = useState<boolean>(false);
+  const [isTriggerHovered, setIsTriggerHovered] = useState<boolean>(false);
+
```
michalzagojski-cognite, line 22: The hover flag is to prevent a bug where the button, after clicked when active, would become inactive after **onClickOutside** on `<SelectPanel>` and then immediately active again. This would lead to discrepancy between the selection list visibility and button toggled state.

haakonflatval-cognite: I see - I tried to fiddle with this a bit myself to see if we could simplify it, but I'm very confused about how this component actually works 😅 But this works fine, so it's good

haakonflatval-cognite (approved):
LGTM - great job 🙌

react-components/src/components/RevealToolbar/RuleBasedOutputsButton/RuleBasedOutputsButton.tsx
```diff
@@ -56,6 +56,8 @@ export const RuleBasedOutputsButton = ({
   const { data: ruleInstancesResult } = useFetchRuleInstances();
 
   const disabled = isAssetMappingsLoading && !isAssetMappingsFetched;
+  const noRuleSetSelected =
+    currentRuleSetEnabled === undefined || Boolean(emptyRuleSelected?.isEnabled);
```
michalzagojski-cognite, line 64: Decided to extract it here as this is used in different parts of the component.

haakonflatval-cognite (dismissed):

react-components/src/components/RevealToolbar/RuleBasedOutputsButton/RuleBasedOutputsButton.tsx
```diff
@@ -43,6 +43,8 @@ export const RuleBasedOutputsButton = ({
 
   const [isRuleLoading, setIsRuleLoading] = useState(false);
 
+  const [isExpanded, setIsExpanded] = useState<boolean>(false);
```
anders-hopland, line 50: Would it be very difficult to add a test for this file?

anders-hopland (changes requested):
Looks good, great job :100:
It would be great if you can address the Gemini comments by which advice us to test toggling on and off
It would also be good if you can externalize teh dependencies of SlicerButton and SettngsButton into a context so we don't have to vi.mock on a module level. This makes it easier to make the mocking type safe.
Please see `react-components/src/components/SceneContainer/SceneContainer.context.tsx` for an example of how to create a context, `react-components/src/components/SceneContainer/SceneContainer.test.tsx` for how to use it in a test and `react-components/src/components/SceneContainer/SceneContainer.tsx` for how to use the externalized dependencies in the component itself

react-components/src/components/RevealToolbar/RuleBasedOutputsButton/RuleBasedOutputsButton.tsx
michalzagojski-cognite: @anders-hopland done :)

react-components/src/components/RevealToolbar/RuleBasedOutputsButton/index.ts
```diff
@@ -0,0 +1,6 @@
+export { RuleBasedOutputsButton } from './RuleBasedOutputsButton';
+export {
+  RuleBasedOutputsButtonContext,
+  defaultRuleBasedOutputsButtonDependencies
+} from './RuleBasedOutputsButton.context';
+export type { RuleBasedOutputsButtonDependencies } from './RuleBasedOutputsButton.context';
```
anders-hopland: non critical, but could probably be exported in the export clause above

react-components/src/components/RevealToolbar/RuleBasedOutputsButton/RuleBasedOutputsButton.test.tsx
```diff
@@ -0,0 +1,162 @@
...
+    {
+      rule: {
+        properties: {
+          id: 'rule2',
+          name: 'Test Rule 2'
+        }
+      },
+      isEnabled: false
+    }
+  ] as any;
```
anders-hopland: Please don't cast to any as it's not type safe. If you want a partial mock I would recomment to use `moq.ts`. If you see that you only need part of the return type of a hook or util in a component I can also recommend to narrow down the type in the context

react-components/src/components/RevealToolbar/HelpButton/HelpButton.tsx
```diff
@@ -27,7 +27,7 @@ export const HelpButton = ({ fallbackLanguage, placement }: HelpButtonProps): Re
 
   return (
     <Dropdown
-      onClickOutside={() => {
+      onHide={() => {
```
anders-hopland, line 30: Just to double check, this will also trigger on click outside right?

react-components/src/components/RevealToolbar/HelpButton.tsx
```diff
@@ -45,7 +45,9 @@ export const HelpButton = ({ fallbackLanguage, placement }: HelpButtonProps): Re
           aria-label="help-button"
           toggled={helpActive}
           onClick={() => {
-            setHelpActive((prevState) => !prevState);
+            if (!helpActive) {
```
anders-hopland: The conditional if (!helpActive) is not needed because calling setHelpActive(true) when helpActive is already true has no effect in React—state updates to the same value are ignored

react-components/src/components/RevealToolbar/HelpButton/HelpButton.tsx
michalzagojski-cognite: Yes, the onHide callback will trigger when clicked outside of the button and the modal

anders-hopland (approved):
🦄

PR #5367, opened 2025-10-03
feat(react-components): point cloud hybrid styling

haakonflatval-cognite:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5953

## Description :pencil:

Styling for hybrid point clouds.

Note that this is awkward to test currently because clicking functionality is not ready yet.

The PR contains a bunch of smaller refactoring, which should perhaps be ported to the base branch instead.

Tests are also lacking.


react-components/src/components/Reveal3DResources/hooks/useCalculateCadStyling.tsx
haakonflatval-cognite, line 1: Changes in this file is purely a simplification of types, no functional change

react-components/src/components/Reveal3DResources/hooks/useCalculatePointCloudStyling.tsx
haakonflatval-cognite, line 1: This file has been simplified massively. The most important factor is that now both classic and DM-based annotations and stylings are treated in the same code path

react-components/src/components/Reveal3DResources/types.ts
haakonflatval-cognite, line 1: I have removed the `Image360*AssetStylingGroup` types here to simplify the typing. It is a breaking change, but I have the corresponding changes for Fusion ready when needed

react-components/src/utilities/instanceIds/toKey.ts
```diff
@@ -3,15 +3,15 @@ import { assertNever } from '../assertNever';
 import { isDmsInstance, isExternalId, isInternalId } from './typeGuards';
 import { type InstanceReference } from './types';
 
-export type InstanceReferenceKey = string;
+export type InstanceReferenceKey = string | number;
```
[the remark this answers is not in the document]
haakonflatval-cognite: The type checker has caught the cases where this is was relevant. After manual inspection I also found no other places needed to update - and I expect no other consequences of this change

haakonflatval-cognite, line 6: The reason for this change is that it simplifies the interaction between `InstanceIdKey` and `InstanceReferenceKey` (`InstanceIdKey` can already be a `number`, and allowing `InstanceReferenceKey` to be that as well makes them easier to interchange.)

react-components/src/components/CacheProvider/Image360AnnotationCache.ts
```diff
@@ -12,10 +12,15 @@ import {
 } from '@cognite/reveal';
 import { fetchAssetsForAssetReferences } from './annotationModelUtils';
 import { isDefined } from '../../utilities/isDefined';
-import { assetInstanceToKey } from '../../utilities/assetInstanceToKey';
 import { isInternalId, type InstanceReference } from '../../utilities/instanceIds';
```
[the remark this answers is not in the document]
haakonflatval-cognite: Usage is also removed in this PR

react-components/src/components/CacheProvider/utils.ts
```diff
@@ -33,6 +41,45 @@ export function getInstanceReferencesFromPointCloudAnnotation(
...
+    }
+  }
+  if (isDMPointCloudVolume(volume) && volume.assetRef !== undefined) {
+    instances.push(volume.assetRef);
+  }
+
+  return instances;
+}
+
+export function getVolumeAnnotationId(
```
danpriori: If we are going to name the annotation id as either `AnnotationId` (number) or a `DMSUniqueIdentifier` as a `PointCloudVolumeId`, then maybe it is better to assume everything as a `PointCloudVolumeId`. So I would suggest to change the function name: `getPointCloudVolumeId`

react-components/src/components/CacheProvider/PointCloudAnnotationCache.ts
```diff
@@ -186,9 +186,10 @@ function buildAssetKeyToAnnotationIdsMap(
     for (const asset of assets) {
       const keys: InstanceReferenceKey[] = [];
       if (isClassicAsset(asset)) {
-        keys.push(String(asset.id));
+        // Make sure we propagate both `externalId` and `id` independently
+        keys.push(createInstanceReferenceKey({ id: asset.id }));
         if (asset.externalId !== undefined) {
-          keys.push(asset.externalId);
+          keys.push(createInstanceReferenceKey({ externalId: asset.externalId }));
```
danpriori, line 192: nice catch, btw :D

react-components/src/components/CacheProvider/utils.ts
```diff
@@ -33,6 +41,45 @@ export function getInstanceReferencesFromPointCloudAnnotation(
...
+  volume: PointCloudObjectMetadata<DataSourceType>
+): PointCloudVolumeId {
+  if (isClassicPointCloudVolume(volume)) {
+    return volume.annotationId;
+  } else {
+    return volume.volumeInstanceRef;
+  }
+}
+
+export function createPointCloudVolumeIdKey(id: PointCloudVolumeId): string {
```
danpriori: Three questions related to this function 😅 :
1) Is it not better to move this key creation into the `idAndKeyTranslation` ?
2) I dont see any usage of this function yet. It will come later in some another PR or in Fusion?
3) Returnig a string is it mandatory here (as I dont see where would call it, I dont know the usage context yet :D)? I'm asking this because I saw a change in the `createInstanceReferenceKey` that change the returning from as String to as number, so that is why I'm raising this question 😅

react-components/src/components/Reveal3DResources/hooks/useCalculateImage360Styling.ts
danpriori, line 1: I know this is out of this PR's scope, but this file is one of these good/easy candidates to be downgraded as a regular function once we start this movement 😅

react-components/src/components/Reveal3DResources/hooks/useCalculatePointCloudStyling.tsx
```diff
@@ -12,32 +9,36 @@ import {
...
+import { UseCalculatePointCloudStylingContext } from './useCalculatePointCloudStyling.context';
 import { useMatchedPointCloudModels } from './useMatchedPointCloudModels';
-import { createFdmKey } from '../../CacheProvider/idAndKeyTranslation';
-import { usePointCloudAnnotationMappingsForModels } from '../../../hooks/pointClouds';
 
 type PointCloudVolumeWithModel = {
   model: PointCloudModelOptions;
-  pointCloudVolumes: Array<number | DMInstanceRef>;
-  asset?: Array<number | string | DMInstanceRef | undefined>;
+  volumesWithAsset: Array<{ pointCloudVolume: PointCloudVolumeId; instance: InstanceReference }>;
```
danpriori:
```suggestion
  volumesWithAsset: Array<{ pointCloudVolumeId: PointCloudVolumeId; instance: InstanceReference }>;
```

react-components/src/components/Reveal3DResources/hooks/useCalculatePointCloudStyling.tsx
```diff
@@ -68,145 +69,55 @@ function useCalculateInstanceStyling(
...
-      return acc;
-    },
-    []
+  const assetInstancesSet = new Set<InstanceReferenceKey>(
+    getInstanceKeysFromStylingGroup(instanceGroup)
   );
 
+  const matchedVolumeRefModels = modelWithVolumes.volumesWithAsset
+    .filter(({ instance }) => assetInstancesSet.has(createInstanceReferenceKey(instance)))
+    .map(({ pointCloudVolume }) => pointCloudVolume);
```
danpriori:
```suggestion
    .map(({ pointCloudVolumeId }) => pointCloudVolumeId);
```

Note: I would prefer to avoid unpacking here as it wont creates a temporary binding (for performance reasons, but it is only critical for millions of items)... but this is not crucial here IMO.

react-components/src/components/Reveal3DResources/hooks/useCalculatePointCloudStyling.tsx
```diff
@@ -308,7 +210,7 @@ function getMappedStyleGroupFromPointCloudVolume(
   nodeAppearance: NodeAppearance
 ): PointCloudVolumeStylingGroup {
   const pointCloudVolumes = pointCloudVolumeData.flatMap((data) => {
-    return data.pointCloudVolumes;
+    return data.volumesWithAsset.map((volumeWithAsset) => volumeWithAsset.pointCloudVolume);
```
danpriori:
```suggestion
    return data.volumesWithAsset.map((volumeWithAsset) => volumeWithAsset.pointCloudVolumeId);
```

react-components/src/components/Reveal3DResources/hooks/useCalculatePointCloudStyling.tsx
```diff
@@ -235,47 +146,38 @@ function useCalculateMappedPointCloudStyling(
...
-        return acc;
-      }, []);
-      const pointCloudVolumes = pointCloudVolumesWithAsset.map((data) => data.pointCloudVolume);
-      const asset = pointCloudVolumesWithAsset.map((data) => data.asset);
-      return { model: model as PointCloudModelOptions, pointCloudVolumes, asset };
+      const pointCloudVolumesWithAsset = viewerModel.stylableObjects.flatMap(
+        (pointCloudObjectData) =>
+          getInstanceReferencesFromPointCloudVolume(pointCloudObjectData).map(
+            (instanceReference) => ({
+              pointCloudVolume: getVolumeAnnotationId(pointCloudObjectData),
```
danpriori:
```suggestion
              pointCloudVolumeId: getVolumeAnnotationId(pointCloudObjectData),
```

danpriori (changes requested):
Some comments and suggestions.
The rest looks good!
Of course we still lack of some testings, but the tests for the styling created are very good: concise and straight to the point.

react-components/src/components/CacheProvider/utils.ts
haakonflatval-cognite: Agreed :+1:

react-components/src/components/CacheProvider/utils.ts
haakonflatval-cognite: Ah, woops - I should just remove this function :see_no_evil: There is no follow-up-PR, so I don't have any plans for it :sweat_smile:

react-components/src/components/Reveal3DResources/hooks/useCalculateImage360Styling.ts
haakonflatval-cognite: Ah, yes - totally agree :+1: It doesn't even use hooks internally :monocle_face: Strange that it's a hook at all

react-components/src/components/Reveal3DResources/hooks/useCalculatePointCloudStyling.tsx
haakonflatval-cognite: Agree :+1:

danpriori (approved):
LGTM. Nice work !

react-components/src/components/Reveal3DResources/hooks/useCalculateImage360Styling.ts
```diff
@@ -14,22 +14,24 @@ export type Image360StyledGroup = Image360PolygonStylingGroup & {
...
+      if (isClassicAssetMappingStylingGroup(group)) {
         return { assetRefs: group.assetIds, style: group.style.image360 };
       }
-      if (isImage360DMAssetStylingGroup(group)) {
-        return { assetRefs: group.assetRefs, style: group.style.image360 };
+      if (isFdmAssetStylingGroup(group)) {
+        return { assetRefs: group.fdmAssetExternalIds, style: group.style.image360 };
       }
       return undefined;
     })
```
haakonflatval-cognite, line 30: The changes in this file are only necessary because I have removed the `Image360StylingGroup` type.

react-components/src/components/Reveal3DResources/utils.ts
```diff
@@ -1,14 +1,24 @@
...
-): Array<AssetId | FdmKey> {
+export function getInstanceKeysFromStylingGroup(stylingGroup: InstanceStylingGroup): InstanceKey[] {
+  return getInstanceReferencesFromStylingGroup(stylingGroup).map(createInstanceKey);
+}
+
+export function getInstanceReferencesFromStylingGroup(
+  stylingGroup: InstanceStylingGroup
+): InstanceId[] {
   if (isClassicAssetMappingStylingGroup(stylingGroup)) {
     return stylingGroup.assetIds;
```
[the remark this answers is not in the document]
haakonflatval-cognite: This is false - `InstanceId` and `InstanceReference` are not completely interchangable. `InstanceId` can be a number, `InstanceReference` cannot.

react-components/src/components/Image360CollectionContainer/types.ts
```diff
@@ -1,6 +0,0 @@
-import { type DMInstanceRef, type PointCloudAppearance } from '@cognite/reveal';
-
-export type Image360PolygonStylingGroup = {
```
haakonflatval-cognite, line 3: This type is no longer needed, and I have confirmed that it's also unused in Fusion

react-components/src/components/Reveal3DResources/utils.ts
```diff
@@ -1,14 +1,24 @@
...
+import { createInstanceKey } from '../CacheProvider/idAndKeyTranslation';
 
-export function getInstanceKeysFromStylingGroup(
-  stylingGroup: FdmInstanceStylingGroup | ClassicAssetStylingGroup
-): Array<AssetId | FdmKey> {
+export function getInstanceKeysFromStylingGroup(stylingGroup: InstanceStylingGroup): InstanceKey[] {
+  return getInstanceReferencesFromStylingGroup(stylingGroup).map(createInstanceKey);
+}
+
+export function getInstanceReferencesFromStylingGroup(
```
anders-hopland: Do we need to export this function? Could we instead only test the function above?

react-components/src/components/Reveal3DResources/hooks/useCalculatePointCloudStyling.tsx
```diff
@@ -235,47 +146,38 @@ function useCalculateMappedPointCloudStyling(
...
-  const classicAddModelOptions = useModelIdRevisionIdFromModelOptions(models);
 
   const pointCloudViewerModels = viewerModels.filter(
     (model): model is CognitePointCloudModel<DataSourceType> => model.type === 'pointcloud'
   );
-  const matchedPointCloudModels = useMatchedPointCloudModels(
-    pointCloudViewerModels,
-    classicAddModelOptions
-  );
+  const matchedPointCloudModels = useMatchedPointCloudModels(pointCloudViewerModels, models);
```
anders-hopland, line 156: Would it be nice to pull this into the context?

anders-hopland (commented):
Two minor questons, other than that it looks good

react-components/src/components/Reveal3DResources/utils.ts
haakonflatval-cognite: Ah, good point :eyes: I'll un-export it

react-components/src/components/Reveal3DResources/hooks/useCalculatePointCloudStyling.tsx
haakonflatval-cognite: Yes, I did it early on, but it turns out that `useMatchedPointCloudModels` is actually pure (it doesn't use any other hooks internally, except `useMemo`). It shouldn't be a hook in the first place, really. In the end it would be more work to mock it (and getting the output right with respect to input), than just using the original implementation.

anders-hopland (dismissed):
🦄

PR #5369, opened 2025-10-06
feat(react-components): pc hybrid annotation on click support and foundation for hybrid annotations part1

danpriori:
#### Type of change

Feat  

#### Jira ticket :blue_book:
https://cognitedata.atlassian.net/browse/BND3D-5949

## Description :pencil:
This is the implementation for the click functionality on a Point Cloud annotation for hybrid annotations.

The functionality also involves the fundamental part of loading and caching hybrid annotations and some hooks involved that are also used for other hybrid annotation search and highlighting features.

It is also related to the PR in Fusion to connect the use of click in Unified 3D.

This is a stacked PR:
- #5369
- #5370
- #5371
- #5373

It was tested in Unified 3D in 3d-test and in the Point Clouds, where I created some hybrid annotation samples: gonza:pointcolor-test and mariudu/test-new-classificatio.

It also needs to be tested with the PR in Fusion to connect this functionality.

[screen-capture (26).webm](https://github.com/user-attachments/assets/345ba944-00e4-4e5d-9f19-9213aad4a7f4)

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -1,0 +1,37 @@
+# Code Best Practices
+
+## Scope Note
+Unless otherwise specified, these rules apply to all code changes in the repository. The rules under "Mocking Practices" are specifically limited to files within the `react-components/` directory.
+
+---
+
+## 1. Mocking Practices (React-Components Scope)
+
+### Rule: Avoid 'vi.mock' for Module Mocks
+- **Scope:** This rule ONLY applies to files located within the `react-components/` directory and its subdirectories (e.g., test files).
+- **Guidance:** In files under `react-components/`, DO NOT use the global `vi.mock` function. This often leads to unnecessary global state and confusion, especially in a modular component library.
+- **Action:** Flag any usage of `vi.mock` that occurs in files under the `react-components/` path.
+
+### Rule: Avoid 'vi.mocked'
+- **Scope:** This rule ONLY applies to files located within the `react-components/` directory and its subdirectories.
+- **Guidance:** In files under `react-components/`, DO NOT use the `vi.mocked` helper. Prefer using explicitly defining the type of the mock object for clearer intent.
+- **Action:** Flag any usage of `vi.mocked` that occurs in files under the `react-components/` path.
+
+---
+
+## 2. TypeScript Type Safety (Global Scope)
+
+### Rule: Avoid Explicit Type Casting using "as any"
+- **Guidance:** Explicit type casting to `as any` bypasses TypeScript's type checking system and introduces potential runtime errors. This practice should be strictly avoided as it sacrifices type safety.
+- **Preferred Alternative:** Find the correct type, use utility types like `Partial<T>` or `Omit<T, K>`, or use a type assertion to a more specific type if you are certain of the type (e.g., `as T`).
+- **Action:** Flag any occurrence of the `as any` pattern in the codebase.
+
+### Rule: Avoid Double Type Casting: "as unknown as SomeType"
+- **Guidance:** Using the pattern **`as unknown as SomeType`** is a red flag that forces a type assignment. It is often used to silence the compiler when a genuine type mismatch issue exists. This practice must be avoided.
+- **Preferred Alternative:** Use proper runtime validation (e.g., with a validation library) to assert the type correctly, or fix the source of the type mismatch.
+- **Action:** Flag any occurrence of the sequential cast pattern **`as unknown as`** in the codebase.
+
+### Rule: Avoid Variables Defined as "any" Type
+- **Guidance:** Variables, function arguments, or return types explicitly defined as `any` defeat the purpose of using TypeScript. This should be avoided to maintain strong type safety across the application.
+- **Preferred Alternative:** Always define a specific type, even if it requires creating a new interface or type. Use `unknown` or `T` where the type cannot be determined.
+- **Action:** Flag any declaration where a variable, parameter, or return type is explicitly set to the `any` keyword.
--- a/react-components/src/components/Image360CollectionContainer/Image360CollectionContainer.context.ts
+++ b/react-components/src/components/Image360CollectionContainer/Image360CollectionContainer.context.ts
@@ -1,0 +1,33 @@
+import { createContext } from 'react';
+
+import {
+  useReveal3DResourceLoadFailCount,
+  useReveal3DResourcesCount
+} from '../Reveal3DResources/Reveal3DResourcesInfoContext';
+import { useRenderTarget } from '../RevealCanvas';
+import { useRevealKeepAlive } from '../RevealKeepAlive/RevealKeepAliveContext';
+import { RevealModelsUtils } from '../../architecture/concrete/reveal/RevealModelsUtils';
+
+export type Image360CollectionContainerContextDependencies = {
+  useRenderTarget: typeof useRenderTarget;
+  useRevealKeepAlive: typeof useRevealKeepAlive;
+  useReveal3DResourcesCount: typeof useReveal3DResourcesCount;
+  useReveal3DResourceLoadFailCount: typeof useReveal3DResourceLoadFailCount;
+  createImage360CollectionDomainObject: typeof RevealModelsUtils.addImage360Collection;
+  removeImage360CollectionDomainObject: typeof RevealModelsUtils.remove;
+};
+
+export const defaultImage360CollectionContainerContextDependencies: Image360CollectionContainerContextDependencies =
+  {
+    useRenderTarget,
+    useRevealKeepAlive,
+    useReveal3DResourcesCount,
+    useReveal3DResourceLoadFailCount,
+    createImage360CollectionDomainObject: RevealModelsUtils.addImage360Collection.bind(this),
+    removeImage360CollectionDomainObject: RevealModelsUtils.remove.bind(this)
+  };
+
+export const Image360CollectionContainerContext =
+  createContext<Image360CollectionContainerContextDependencies>(
+    defaultImage360CollectionContainerContextDependencies
+  );
--- a/react-components/src/components/Image360CollectionContainer/Image360CollectionContainer.test.tsx
+++ b/react-components/src/components/Image360CollectionContainer/Image360CollectionContainer.test.tsx
@@ -1,0 +1,130 @@
+import { render, waitFor } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { Image360CollectionContainer } from './Image360CollectionContainer';
+import {
+  Image360CollectionContainerContext,
+  defaultImage360CollectionContainerContextDependencies
+} from './Image360CollectionContainer.context';
+import type { PropsWithChildren, ReactElement } from 'react';
+import { Mock } from 'moq.ts';
+import type { RevealRenderTarget } from '../../architecture';
+import {
+  type Image360Collection,
+  type Cognite3DViewer,
+  type ClassicDataSourceType
+} from '@cognite/reveal';
+import { getMocksByDefaultDependencies } from '#test-utils/vitest-extensions/getMocksByDefaultDependencies';
+import type { AddImage360CollectionOptions } from '../Reveal3DResources/types';
+import * as useApply360AnnotationStylingModule from './useApply360AnnotationStyling';
+
+describe(Image360CollectionContainer.name, () => {
+  const deps = getMocksByDefaultDependencies(defaultImage360CollectionContainerContextDependencies);
+
+  const wrapper = ({ children }: PropsWithChildren): ReactElement => (
+    <Image360CollectionContainerContext.Provider value={deps}>
+      {children}
+    </Image360CollectionContainerContext.Provider>
+  );
+
+  const collectionsGetter = vi.fn<() => Array<Image360Collection<ClassicDataSourceType>>>();
+  const image360Collection = new Mock<Image360Collection<ClassicDataSourceType>>()
+    .setup((x) => x.id)
+    .returns('test-site-id')
+    .setup((x) => x.setModelTransformation)
+    .returns(vi.fn())
+    .setup((x) => x.set360IconCullingRestrictions)
+    .returns(vi.fn())
+    .object();
+
+  const renderTargetMock = new Mock<RevealRenderTarget>()
+    .setup((x) => x.viewer)
+    .returns(
+      new Mock<Cognite3DViewer>()
+        .setup((x) => x.models)
+        .returns([])
+        .setup((x) => x.get360ImageCollections)
+        .returns(() => collectionsGetter())
+        .object()
+    );
+
+  beforeEach(() => {
+    deps.useRenderTarget.mockReturnValue(renderTargetMock.object());
+
+    deps.useRevealKeepAlive.mockReturnValue({
+      renderTargetRef: { current: renderTargetMock.object() },
+      isRevealContainerMountedRef: { current: true },
+      sceneLoadedRef: { current: undefined }
+    });
+
+    deps.useReveal3DResourcesCount.mockReturnValue({
+      reveal3DResourcesCount: 0,
+      setRevealResourcesCount: vi.fn()
+    });
+
+    deps.useReveal3DResourceLoadFailCount.mockReturnValue({
+      reveal3DResourceLoadFailCount: 0,
+      setReveal3DResourceLoadFailCount: vi.fn()
+    });
+
+    deps.createImage360CollectionDomainObject.mockResolvedValue(image360Collection);
+    collectionsGetter.mockReturnValue([]);
+
+    vi.spyOn(useApply360AnnotationStylingModule, 'useApply360AnnotationStyling').mockImplementation(
+      () => {}
+    );
+  });
+
+  it('should properly add and remove Image360 Collection when mounting and unmounting Image360CollectionContainer', async () => {
+    const addImage360CollectionOptions: AddImage360CollectionOptions = {
+      source: 'events',
+      siteId: 'test-site-id'
+    };
+
+    const { unmount } = render(
+      <Image360CollectionContainer
+        addImage360CollectionOptions={addImage360CollectionOptions}
+        onLoad={vi.fn()}
+      />,
+      { wrapper }
+    );
+
+    await waitFor(() => {
+      expect(deps.createImage360CollectionDomainObject).toHaveBeenCalled();
+    });
+
+    collectionsGetter.mockReturnValue([image360Collection]);
+
+    unmount();
+
+    await waitFor(() => {
+      expect(deps.removeImage360CollectionDomainObject).toHaveBeenCalled();
+    });
+  });
+
+  it.each([undefined, false, true])(
+    'should propagate defaultVisible flag %s to `createImage360CollectionDomainObject`',
+    async (defaultVisibleFlag) => {
+      const addImage360CollectionOptions: AddImage360CollectionOptions = {
+        source: 'events',
+        siteId: 'test-site-id'
+      };
+
+      render(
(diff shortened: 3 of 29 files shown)


react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
```diff
@@ -0,0 +1,76 @@
...
+} from '@cognite/reveal';
+import {
+  isClassicModelIdentifier,
+  isDMModelIdentifier,
+  isDmsInstance,
+  isIdEither
+} from '../../utilities/instanceIds';
+
+type InstanceData = {
+  classicModelIdentifier: { modelId: number; revisionId: number } | undefined;
```
anders-hopland: non blocking
Could b nice to pull `{ modelId: number; revisionId: number }` into a type called `ClassicModelIdentifier` or something similar

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
```diff
@@ -0,0 +1,76 @@
...
+): InstanceData | undefined {
+  const emptyResult = {
+    classicModelIdentifier: undefined,
+    dmsModelUniqueIdentifier: undefined,
+    reference: undefined
+  };
+  const isPointCloudIntersection = intersection?.type === 'pointcloud';
+  if (!isPointCloudIntersection) return emptyResult;
+
+  if (
```
anders-hopland: Would be great to do early return on each of the conditional clauses instead of having lots of `if` - `else if`

react-components/src/utilities/getInstancesFromClick.ts
```diff
@@ -81,24 +80,19 @@ async function getPointCloudAnnotationMappingsFromIntersection(
...
-
-  const annotations = await fetchAnnotationsForModel(
-    intersection.model.modelIdentifier.modelId,
-    intersection.model.modelIdentifier.revisionId,
-    [assetExternalId],
-    caches.pointCloudAnnotationCache
-  );
-
-  return annotations?.map((annotation) => ({ id: annotation.asset.id })) ?? EMPTY_ARRAY;
+  return [];
```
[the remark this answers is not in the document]
anders-hopland: As an annotation can be connected to both a classic and a DMS instance, but the DMS instance being the newest and most interesting one, I think that the DMS one should take precedence

react-components/src/utilities/getInstancesFromClick.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+vi.mock(import('../components/CacheProvider/requests'), () => ({
+  fetchAncestorNodesForTreeIndex: vi.fn<typeof fetchAncestorNodesForTreeIndex>()
+}));
+
+vi.mock(import('../hooks/pointClouds/fetchAnnotationsForModel'), () => ({
+  fetchAnnotationsForModel: vi.fn<typeof fetchAnnotationsForModel>()
+}));
+
+vi.mock(import('../components/CacheProvider/utils'), () => ({
```
anders-hopland: non blocking
Would be great to externalize these dependencies so we don't have to use vi.mock

react-components/src/utilities/getInstancesFromClick.test.ts
```diff
@@ -0,0 +1,221 @@
...
+  getInstanceReferenceFromImage360Annotation:
+    vi.fn<typeof getInstanceReferenceFromImage360Annotation>()
+}));
+
+describe(getInstancesFromClick.name, () => {
+  const mockEvent = {
+    offsetX: 100,
+    offsetY: 200
+  } as const as PointerEvent;
```
anders-hopland: Please don't cast here. I can recommend to use moq.ts if you only want a partial mock

react-components/src/utilities/getInstancesFromClick.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        annotatedResourceId: 1,
+        annotationType: 'image360',
+        status: 'approved',
+        annotatedResourceType: 'threedmodel',
+        creatingApp: 'test-app',
+        creatingAppVersion: '1.0.0',
+        creatingUser: 'test-user',
+        data: {}
+      }
+    } as const as Image360Annotation<DataSourceType>,
```
anders-hopland: Please don't cast here

react-components/src/utilities/getInstancesFromClick.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    expect(result).toEqual([dmInstanceRef]);
+  });
+
+  test('returns undefined for unknown intersection type', async () => {
+    const unknownIntersection = {
+      type: 'unknown' as any,
+      point: new Vector3(1, 2, 3),
+      distanceToCamera: 10
+    } as const as AnyIntersection;
```
anders-hopland: Let's not cast here

react-components/src/utilities/getInstancesFromClick.test.ts
```diff
@@ -0,0 +1,221 @@
...
+    .returns(sdkMock)
+    .setup((p) => p.classicCadAssetMappingCache)
+    .returns(mockClassicCadAssetMappingCache)
+    .object();
+
+  viewerMock.getPixelCoordinatesFromEvent = vi
+    .fn<typeof viewerMock.getPixelCoordinatesFromEvent>()
+    .mockReturnValue(new Vector2(100, 200));
+
+  const mockRenderTarget = new Mock<RevealRenderTarget>()
```
nilscognite, line 101: Reuse: createRenderTargetMock. Also check similar places.

anders-hopland (changes requested):
Code looks good. Minot comments. I am not able to click hybrid instances when linking this react components version to fusion

anders-hopland (changes requested):
Code looks good. We only need minor changes. I'm not able to click hybrid instances in Fusion. We need to look into that

react-components/src/utilities/getInstancesFromClick.ts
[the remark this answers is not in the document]
haakonflatval-cognite: Should we return both in that case? The function (`getInstancesFromClick`) supports returning multiple results

react-components/src/utilities/getInstancesFromClick.test.ts
danpriori: https://github.com/cognitedata/reveal/pull/5369/commits/a8cc9aa6866b60de26d3b894f0f00545a2e901cc

react-components/src/utilities/getInstancesFromClick.test.ts
danpriori: https://github.com/cognitedata/reveal/pull/5369/commits/48f85d63adcfe5883a5c91124278c338325e5350

react-components/src/utilities/getInstancesFromClick.test.ts
danpriori: https://github.com/cognitedata/reveal/pull/5369/commits/48f85d63adcfe5883a5c91124278c338325e5350

react-components/src/utilities/getInstancesFromClick.test.ts
danpriori: https://github.com/cognitedata/reveal/pull/5369/commits/48f85d63adcfe5883a5c91124278c338325e5350

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
danpriori: it is already doing early return, the only thing is the else if which is kinda useless in this case.
Removed here:
https://github.com/cognitedata/reveal/pull/5369/commits/48f85d63adcfe5883a5c91124278c338325e5350

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
danpriori: https://github.com/cognitedata/reveal/pull/5369/commits/48f85d63adcfe5883a5c91124278c338325e5350

react-components/src/utilities/getInstancesFromClick.ts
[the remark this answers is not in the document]
danpriori: changing the precedence and support for multiple results

https://github.com/cognitedata/reveal/pull/5369/commits/48f85d63adcfe5883a5c91124278c338325e5350

react-components/src/utilities/getInstancesFromClick.test.ts
danpriori: In here I think it will be hard to use the global shared createRenderTargetMock as it is creating a renderTarget mock and returning its object, which blocks me to customize it based on the tests that I need. So I think the best here is to use a separated renderTarget mock locally. Unless we change the createRenderTargetMock to support more customization or not return the object to allow us to override for each tests later. But this is out of this scope, IMO.
Unless I'm forgetting something here or if you have any suggestion to use it based on this local tests needs.

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
```diff
@@ -0,0 +1,79 @@
...
+
+type InstanceData = {
+  classicModelIdentifier: ClassicModelIdentifierType | undefined;
+  dmsModelUniqueIdentifier: DMModelIdentifierType | undefined;
+  reference: IdEither | DmsUniqueIdentifier | undefined;
+};
+export function getInstanceDataFromIntersection(
+  intersection: AnyIntersection | undefined
+): InstanceData | undefined {
+  const emptyResult = {
```
pramod-cog: Why do we need to return object with `undefined` values in it instead of just `undefined`?

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
```diff
@@ -0,0 +1,79 @@
...
+  if (!isPointCloudIntersection) return emptyResult;
+
+  if (
+    isIdEither(intersection.volumeMetadata?.assetRef) &&
+    isClassicModelIdentifier(intersection.model.modelIdentifier)
+  ) {
+    return {
+      classicModelIdentifier: intersection.model.modelIdentifier,
+      dmsModelUniqueIdentifier: undefined,
+      reference: intersection.volumeMetadata?.assetRef
```
pramod-cog: I would like to reduce duplication of creating data by having a function

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
```diff
@@ -0,0 +1,79 @@
...
+  const emptyResult = {
+    classicModelIdentifier: undefined,
+    dmsModelUniqueIdentifier: undefined,
+    reference: undefined
+  };
+  const isPointCloudIntersection = intersection?.type === 'pointcloud';
+  if (!isPointCloudIntersection) return emptyResult;
+
+  if (
+    isIdEither(intersection.volumeMetadata?.assetRef) &&
```
pramod-cog: we can destructure intersection data like below
```
const { model, volumeMetadata } = intersection;
const { modelIdentifier } = model;
```
This would be much cleaner

react-components/tests/tests-utilities/fixtures/dm/viewDefinitions.ts
```diff
@@ -226,7 +226,9 @@ export const viewDefinitionMock: ViewDefinition = {
   createdTime: 0,
   lastUpdatedTime: 0,
   writable: false,
+  queryable: false,
```
pramod-cog, line 229: type error

react-components/src/utilities/getInstancesFromClick.ts
```diff
@@ -8,18 +8,22 @@ import {
...
 
+type InstancesFromClickDependencies = {
+  fetchAncestorNodesTreeIndex: typeof fetchAncestorNodesForTreeIndex;
+};
+
 export async function getInstancesFromClick(
   renderTarget: RevealRenderTarget,
-  event: PointerEvent
+  event: PointerEvent,
+  dependencies: Partial<InstancesFromClickDependencies> = {}
```
pramod-cog, line 26: I would prefer to have dependencies as optional rather than partial Type

react-components/tests/tests-utilities/fixtures/assets.ts
```diff
@@ -48,3 +52,25 @@ export function createDMAssetMock(externalId: string): ExternalIdsResultList<Ass
     typing: {}
   };
 }
+
+export function createFdmNodeItem(id: {
```
pramod-cog, line 56: I am guessing this would be used in other PR's?

react-components/src/utilities/getInstancesFromClick.ts
danpriori: this pattern is initially used in getUserFingerPrint and discussed with @haakonflatval-cognite. 
I think it is better to follow the same pattern. Or change there as well.
Wdyt?

react-components/tests/tests-utilities/fixtures/assets.ts
danpriori: Yes. It is used here for instance: https://github.com/cognitedata/reveal/pull/5371/files#diff-da7346d4a5924efa6da165bd0ea0562045644e980268a9ab4ae525d006ac4f6c

react-components/tests/tests-utilities/fixtures/dm/viewDefinitions.ts
danpriori: Sorry, I'm not getting any type error with these changes 🤔 can yo elaborate?
It will get type error if I remove these added parameters 🤔

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
danpriori: unpacking here:

https://github.com/cognitedata/reveal/pull/5369/commits/01e9b7d03c7df92e093b0931cf1125a82560b7c4 

Note: 
but in the last statement I kept the full intersection because the type checker is not getting into account the instanceRef even returning the correct volumeData type (classic one) which includes the instanceRef from the unpacked references.

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
danpriori: https://github.com/cognitedata/reveal/pull/5369/commits/01e9b7d03c7df92e093b0931cf1125a82560b7c4

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
danpriori: changed here
https://github.com/cognitedata/reveal/pull/5369/commits/01e9b7d03c7df92e093b0931cf1125a82560b7c4

react-components/src/utilities/getInstancesFromClick.ts
haakonflatval-cognite: I might agree with Pramod that not having it partial is a bit cleaner, but I don't have a strong opinion 😅

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
```diff
@@ -0,0 +1,62 @@
...
+  isDmsInstance,
+  isIdEither
+} from '../../utilities/instanceIds';
+
+type InstanceData = {
+  classicModelIdentifier: ClassicModelIdentifierType | undefined;
+  dmsModelUniqueIdentifier: DMModelIdentifierType | undefined;
+  reference: IdEither | DmsUniqueIdentifier | undefined;
+};
+export function getInstanceDataFromIntersection(
```
pramod-cog: I think the function to be named as `getPointCloudInstanceFromIntersection`

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+    revisionId: 789
+  });
+
+  const pointCloudDMMock = createPointCloudDMMock({
+    revisionExternalId: 'revision-ext-id',
+    revisionSpace: 'test-space'
+  });
+
+  // Create test data that the real type guards will recognize
+  const classicIntersection: PointCloudIntersection<ClassicDataSourceType> = {
```
pramod-cog: we can reduce code if we can pull put these values into a function because there are some common values used in several mock data

react-components/src/utilities/getInstancesFromClick.ts
pramod-cog: Having `Partial` types can be an issue, take below example
```
type UserType = {
  requiredType1: string;        // This should be required!
  requiredType2: number;       // This should be required!
  requiredType3: boolean;       // This might be optional
};

function createType(type: Partial<UserType>) {
  // Typescript allows this call which is an issue not cuaght
  // createType({}) - no requiredType provided!
}
```
Having `Partial` type is not always safe, so I would prefer us to not use it unless necessary which is not in this case

react-components/src/utilities/instanceIds/typeGuards.ts
```diff
@@ -34,3 +35,18 @@ export function isDmsInstance(instance: unknown): instance is DmsUniqueIdentifie
...
+    (modelIdentifier as DMModelIdentifierType).revisionExternalId !== undefined &&
+    (modelIdentifier as DMModelIdentifierType).revisionSpace !== undefined
+  );
+}
+
+export function isClassicModelIdentifier(
+  modelIdentifier: DMModelIdentifierType | ClassicModelIdentifierType
+): modelIdentifier is ClassicModelIdentifierType {
+  return (modelIdentifier as ClassicModelIdentifierType).modelId !== undefined;
+}
```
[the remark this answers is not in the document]
pramod-cog: We can disregard this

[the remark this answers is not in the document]
danpriori: https://github.com/cognitedata/reveal/pull/5369/commits/6e12e4d975dd7fb61d87ef3d3394a5a6cf68ff84

react-components/src/utilities/getInstancesFromClick.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      fetchAncestorNodesTreeIndex: mockfetchAncestorNodesTreeIndex
+    });
+
+    expect(result).toEqual([dmInstanceRef]);
+  });
+
+  test('returns undefined for unknown intersection type', async () => {
+    const unknownIntersectionMock = new Mock<AnyIntersection>()
+      .setup((p) => p.type)
+      .returns('unknown' as any) // Force as any to simulate unknown type
```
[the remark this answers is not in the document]
danpriori: Actually this is an unnecessary test for unknown type that can be handled by the type checker anyways. Removed.
 
https://github.com/cognitedata/reveal/pull/5369/commits/6e12e4d975dd7fb61d87ef3d3394a5a6cf68ff84

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.ts
danpriori: https://github.com/cognitedata/reveal/pull/5369/commits/fb42c8291f685964d4cd8f85386d85afcc6f1640

react-components/src/hooks/pointClouds/getInstanceDataFromIntersection.test.tsx
danpriori: sure
https://github.com/cognitedata/reveal/pull/5369/commits/fb42c8291f685964d4cd8f85386d85afcc6f1640

PR #5377, opened 2025-10-08
[BND3D-5801]: removed all errors and warning from the linting

rostyslavnahornyi-cognite:
#### Type of change
Refactor 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/jira/software/c/projects/BND3D/boards/857?assignee=712020%3Adb01e922-3ce5-4213-b5cf-b7bc4bd7350b&selectedIssue=BND3D-5801

## Description :pencil:
Removed all errors and warnings from the linting. It is about up to 10 errors and 20 warnings.
Took the part of commits from "pramodcog/fix-all-react-lint-warnings".
This change has made due to reducing the tech debt.

- cd react-components;
-  yarn lint (check 0 errors and warnings);
-  check storybook stories not to broke smth.

### DIFF
--- a/react-components/src/components/Architecture/DomainObjectPanelInput.tsx
+++ b/react-components/src/components/Architecture/DomainObjectPanelInput.tsx
@@ -1,5 +1,5 @@
 import { Input } from '@cognite/cogs.js';
-import { useState, type ReactElement, useEffect } from 'react';
+import { useState, type ReactElement, useEffect, useCallback } from 'react';
 import { type UnitSystem, type LengthUnit } from '../../architecture/base/renderTarget/UnitSystem';
 import { useSignalValue } from '@cognite/signals/react';
 import { type NumberPanelItem } from '../../architecture/base/domainObjectsHelpers/PanelInfo';
@@ -10,14 +10,15 @@ type Props = {
 };
 
 export function DomainObjectPanelInput({ item, unitSystem }: Props): ReactElement {
-  function getOriginalValue(): string {
+  const getOriginalValue = useCallback(() => {
     return unitSystem.toString(item.value, item.quantity, false);
-  }
+  }, [unitSystem, item.value, item.quantity]);
+
   const [value, setValue] = useState(getOriginalValue());
   const lengthUnit = useSignalValue<LengthUnit>(unitSystem.lengthUnit);
   useEffect(() => {
     setValue(getOriginalValue());
-  }, [lengthUnit, item.value]);
+  }, [lengthUnit, item.value, getOriginalValue]);
 
   function onChange(newStringValue: string): void {
     if (newStringValue === '-') {


react-components/src/components/ViewerAnchor/ViewerAnchor.tsx
```diff
@@ -132,7 +132,10 @@ function useStickyCompensation(
     sticky,
     stickyMargin,
     viewerDomElement.clientWidth,
-    viewerDomElement.clientHeight
+    viewerDomElement.clientHeight,
+    htmlRef,
+    stickyCompensation,
```
michalzagojski-cognite: It seems like this `stickyCompensation` variable in dependencies in combination with `setStickyCompensation` calls within the same useEffect is causing an infinite loop. Does it have to be there?

react-components/src/components/RevealToolbar/LayersButton/hooks/useModelHandlers.tsx
```diff
@@ -46,7 +43,7 @@ export const useModelHandlers = (
   useEffect(() => {
     const newHandlers = createHandlers(models, modelNames.data, image360Collections, viewer);
     setModelHandlers(newHandlers);
-  }, [models, modelNames.data, image360Collections, viewer]);
+  }, [models, modelNames.data, image360Collections, viewer, modelHandlers]);
```
michalzagojski-cognite: Again, this looks like an infinite loop because of the `modelHandlers` depencency in a useEffect with a `setModelHandlers` call.

react-components/src/components/Image360HistoricalDetails/Image360HistoricalDetails.viewmodel.tsx
```diff
@@ -58,7 +58,7 @@ export function useImage360HistoricalDetailsViewModel({
       });
       setImageUrls([]);
     };
-  }, [image360Entity, revokeObjectUrl, formatDateTime]);
+  }, [image360Entity, imageUrls, revokeObjectUrl, formatDateTime]);
```
michalzagojski-cognite: This could also potentially be an infinite loop because we both set `imageUrls` and have them as a dependency at the same time.

react-components/src/components/i18n/I18n.tsx
```diff
@@ -38,7 +38,7 @@ const useTranslationContent = (
     return () => {
       window.removeEventListener('languagechange', handleLanguageChange);
     };
-  }, []);
+  }, [currentLanguage, enabled]);
```
michalzagojski-cognite, line 41: Im not sure about the context here, but does this useEffect need those dependencies here? I can see that there is a guarding check before we `setCurrentLanguage` but a small change can again lead to an infinite loop here. Perhaps it's not necessary to have those dependencies here if this code has been working as expected?

PR #5389, opened 2025-10-13
chore: cleanup JSX usage

christjt:
#### Type of change

Chore 

## Description :pencil:

Cleanup JSX usages. Typically when setting JSX in props you should encompass it in brackets. Apperently it works without, but at least my LSP does not like it and I get a bunch of red syntax highlighting. This PR simply wraps JSX props with brackets.


christjt:
risk review ok

PR #5381, opened 2025-10-10
fix(react-components): use DomainObjects in LayersButton

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-6147

## Description :pencil:

We have experienced some bugs related to FRS in Fusion not updating its item list properly when models are turned visible/invisible. This is because those state changes are not automatically propagated from the model layers button to the 3D data fetcher. 

The solution chosen is to use the `DomainObjects` which serve as the ground truth for model visibility, and already has some utilities for automatic querying/updating on visibility changes. 

The Fusion issue will need some additional changes in that code base as well.

## How has this been tested? :mag:

In Fusion and Storybook

## Test instructions :information_source:

Go to Storybook, Layers-example. Make sure you have a valid Bearer token in the `token` URL parameter. See that updating the model layers still works.

## Checklist :ballot_box_with_check:

- [x] I am happy with this implementation.
- [ ] I have performed a self-review of my own code.
- [ ] I have added unit and visuals tests to prove that my feature works to the best of my ability.
- [ ] I have added documentation to new and changed elements; both public and internally shared ones
- [ ] I have refactored the code for testability, readability and extendibility to the best of my ability.
- [ ] I have listed the JIRA tasks covering remaining work or tech debt related to this PR in the description.

PR #5392, opened 2025-10-15
refactor(react-components): BND3D-6147 part 1 - foundational hooks and functions

haakonflatval-cognite:
#### Type of change

Refactor 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-6147

## Description :pencil:

This is a partial recreation of https://github.com/cognitedata/reveal/pull/5381

Due to the size of the PR, I have decided to reimplement the new functions from scratch through a new PR stack. The combined end result will be identical to the above PR. Note that files having "2" in their name will be renamed to replace their non-numbered counterparts in the final PR.

PRs in this stack:
- https://github.com/cognitedata/reveal/pull/5392 (This PR)
- https://github.com/cognitedata/reveal/pull/5393
- https://github.com/cognitedata/reveal/pull/5394
- https://github.com/cognitedata/reveal/pull/5395
- https://github.com/cognitedata/reveal/pull/5396

### Original PR description:

We have experienced some bugs related to FRS in Fusion not updating its item list properly when models are turned visible/invisible. This is because those state changes are not automatically propagated from the model layers button to the 3D data fetcher.

The solution chosen is to use the DomainObjects which serve as the ground truth for model visibility, and already has some utilities for automatic querying/updating on visibility changes.

The Fusion issue will need some additional changes in that code base as well.

### Content of this PR

- Add functions `updateExternalStateFromLayerContent` and `updateViewerFromExternalStateState` for synchronizing visibility state changes that either happen in the URL or in the viewer's models.
- Add hook `useModelsVisibilityState` which is used to construct a model layers object for use in synchronization between the internal model visibility state (in the 3D viewer), and the one in the external state (which may e.g. be the URL or localstorage, depending on the application.

Unit tests


nilscognite (dismissed):
See comment in the original PR  https://github.com/cognitedata/reveal/pull/5381

react-components/src/components/RevealToolbar/LayersButton/hooks/useModelsVisibilityState.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  test('calling update updates external layer state', () => {
+    const setExternalState = vi.fn<Dispatch<SetStateAction<LayersUrlStateParam | undefined>>>();
+
+    renderHook(() => useModelsVisibilityState(setExternalState, renderTarget), {
+      wrapper
+    });
+
+    expect(setExternalState).toHaveBeenCalledOnce();
+    const returnedObject = setExternalState.mock.calls[0][0] as LayersUrlStateParam;
```
christjt:
```suggestion
    const returnedObject = setExternalState.mock.calls[0][0];
    expect(typeof returnedObject !== 'function');
    assert(typeof returnedObject !== 'function');
```

react-components/src/components/RevealToolbar/LayersButton/hooks/useModelsVisibilityState.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    expect(returnedObject?.cadLayers?.[0].applied).toBe(cadObject.isVisible(renderTarget));
+
+    expect(returnedObject?.pointCloudLayers?.[0].applied).toBe(
+      pointCloudObject.isVisible(renderTarget)
+    );
+
+    expect(returnedObject?.image360Layers?.[0].applied).toBe(
+      image360CollectionObject.isVisible(renderTarget)
+    );
```
christjt:
```suggestion
    expect(returnedObject).toBeDefined();
    assert(returnedObject !== undefined);

    expect(returnedObject.cadLayers).toHaveLength(1);
    expect(returnedObject.pointCloudLayers).toHaveLength(1);
    expect(returnedObject.image360Layers).toHaveLength(1);

    expect(returnedObject.cadLayers?.[0].applied).toBe(cadObject.isVisible(renderTarget));

    expect(returnedObject.pointCloudLayers?.[0].applied).toBe(
      pointCloudObject.isVisible(renderTarget)
    );

    expect(returnedObject.image360Layers?.[0].applied).toBe(
      image360CollectionObject.isVisible(renderTarget)
    );
```

react-components/src/components/RevealToolbar/LayersButton/hooks/useModelsVisibilityState.tsx
```diff
@@ -0,0 +1,64 @@
...
+  );
+
+  const domainObjects = useRevealDomainObjects();
+  const domainObjectsByVisibility = useVisibleRevealDomainObjects();
+
+  useEffect(() => {
+    setExternalLayersState?.(
+      createExternalStateFromLayers(createModelLayersObject(domainObjects), renderTarget)
+    );
+  }, [domainObjectsByVisibility]);
```
christjt: Can we discuss the need for useEffect here? Initially, this feels wrong.

react-components/src/components/RevealToolbar/LayersButton/updateExternalStateFromLayerContent.test.ts
```diff
@@ -0,0 +1,91 @@
...
+    expect(result.image360Layers?.[0].applied).toBe(
+      image360CollectionDomainObject0.isVisible(renderTargetMock)
+    );
+    expect(result.image360Layers?.[0].siteId).toBe(image360CollectionDomainObject0.model.id);
+
+    expect(result.image360Layers?.[1].applied).toBe(
+      image360CollectionDomainObject1.isVisible(renderTargetMock)
+    );
+    expect(result.image360Layers?.[1].siteId).toBe(image360CollectionDomainObject1.model.id);
+  });
```
christjt: Same changes as above, so assert the result as not function, and expect+assert layers to not be undefined and remove null propagation.

react-components/src/components/RevealToolbar/LayersButton/updateExternalStateFromLayerContent.tsx
```diff
@@ -0,0 +1,30 @@
...
+      index
+    })),
+    image360Layers: modelLayerContent.image360Collections.map((collection) => ({
+      applied: collection.isVisible(renderTarget),
+      siteId: collection.model.id
+    }))
+  };
+
+  setExternalLayersState?.(newState);
+}
```
christjt: Should we pass the set state here? To me it makes more sense just to return the new state and let the consumer set it. 

If we do actually want the setState passed here, we surely should not accept it as optional, right? It does not do anything when the set state is undefined.

react-components/src/components/RevealToolbar/LayersButton/updateViewerFromExternalState2.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { type RevealRenderTarget } from '../../../architecture';
+
+export function updateViewerFromExternalState(
+  layersState: LayersUrlStateParam | undefined,
+  modelLayerContent: ModelLayerContent,
+  renderTarget: RevealRenderTarget
+): void {
+  if (layersState === undefined) {
+    return;
+  }
```
christjt: I personally don't love this pattern. Why can't the consumer just not call this if layerState is undefined? It clearly doesn't work / do anything if layerState is undefined, so I would just not allow that state.

christjt (changes requested):

react-components/src/components/RevealToolbar/LayersButton/updateViewerFromExternalState2.tsx
haakonflatval-cognite: Fair :+1:

react-components/src/components/RevealToolbar/LayersButton/updateExternalStateFromLayerContent.tsx
haakonflatval-cognite: Yes, you're right. I have decided to remove this file and instead extract the `createExternalStateFromLayerContent` function from `useModelsVisibilityState` which essentially does this transformation. I'll update the future PRs to use that function instead.

react-components/src/components/RevealToolbar/LayersButton/hooks/useModelsVisibilityState.tsx
haakonflatval-cognite: After some investigation, I found that this `useEffect` is actually completely redundant. this synchronization will happen in `useSyncExternalLayersState` in one of the future PRs.

christjt (approved):
risk review ok

PR #5403, opened 2025-10-20
Fix some icons and upgrade new version of Cogs (10.51.0) and make tests work again!

nilscognite:
#### Type of change
Feat  

## Description :pencil:

* Use the new icons for Measurement diameter
* Change icons for the root and the tree (previously the wrong icons was used and this is not in production yet)
* Sort icons in the **DefaultIcon.txs**
* Added better unit test on icon checks

And

I added some new utility functions for testing React. The function `queryByRole `don't work for disabled segmented buttons. I tried different solution and ending up with :

```
export function getButtonsInContainer(container: HTMLElement): HTMLButtonElement[] {
  return Array.from(container.querySelectorAll('button'));
}
```
I added some utility functions in the same file, for better standardizing the way we test the state of an element. I tried the same functions in 'testing-library/jest-dom', but they didn't always work either. **The testing environment we have in React is very unstable and unpredictable.**

I have tested in storybook


react-components/src/components/Architecture/Factories/DefaultIcons.tsx
```diff
@@ -15,6 +15,7 @@ import {
   CubeFrontLeftIcon,
   CubeFrontRightIcon,
   CubeIcon,
+  CubesIcon,
```
nilscognite, line 18: Changes here is because of sorting

react-components/src/components/Architecture/Factories/DefaultIcons.tsx
```diff
@@ -67,24 +68,24 @@ export const DefaultIcons: Array<[IconName, IconType]> = [
   ['ArrowLeft', ArrowLeftIcon],
   ['ArrowRight', ArrowRightIcon],
   ['Axis3D', Axis3DIcon],
-  ['Bug', BugIcon],
   ['BorderHorizontal', BorderHorizontalIcon],
   ['BorderVertical', BorderVerticalIcon],
+  ['Bug', BugIcon],
```
nilscognite, line 73: Changes here is because of sorting

react-components/src/architecture/concrete/measurements/commands/SetMeasurementTypeCommand.test.ts
```diff
@@ -25,6 +26,7 @@ describe(SetMeasurementTypeCommand.name, () => {
       const command = getDefaultCommand(newCommand, renderTarget);
       expect(isEmpty(command.tooltip)).toBe(false);
       expect(command.icon).not.toBe('');
+      expect(command.icon).toBe(getIconByPrimitiveType(command.primitiveType));
```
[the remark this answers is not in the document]
nilscognite: I will not do this...

react-components/tests/tests-utilities/cogs/htmlTestUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+  return element.getAttribute('aria-label') ?? '';
+}
+
+export function isToggled(element: HTMLButtonElement): boolean {
+  const attribute = element.getAttribute('class');
+  if (attribute === null) {
+    return false;
+  }
+  return attribute.includes('toggled');
+}
```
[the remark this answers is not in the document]
nilscognite: Ok, Fixed

react-components/tests/tests-utilities/cogs/cogsTestUtils.ts
```diff
@@ -0,0 +1,31 @@
+export function getButtonsInContainer(container: HTMLElement): HTMLButtonElement[] {
+  return Array.from(container.querySelectorAll('button'));
+}
+
+export function isEnable(element: HTMLElement): boolean {
+  return element.getAttribute('aria-disabled') === 'false';
+}
```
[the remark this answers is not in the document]
nilscognite: Good catch! Fixed

react-components/tests/tests-utilities/cogs/cogsTestUtils.ts
```diff
@@ -0,0 +1,27 @@
+export function getButtonsInContainer(container: HTMLElement): HTMLButtonElement[] {
+  return Array.from(container.querySelectorAll('button'));
+}
+
+export function isEnable(element: HTMLElement): boolean {
```
anders-hopland:
```suggestion
export function isEnabled(element: HTMLElement): boolean {
```

react-components/tests/tests-utilities/cogs/cogsTestUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+  return !isDisable(element);
+}
+
+export function isDisable(element: HTMLElement): boolean {
```
anders-hopland:
```suggestion
export function isDisabled(element: HTMLElement): boolean {
```

react-components/tests/tests-utilities/cogs/htmlTestUtils.ts
```diff
@@ -0,0 +1,27 @@
+export function getButtonsInContainer(container: HTMLElement): HTMLButtonElement[] {
```
anders-hopland, line 1: This seems like generic test utilities, should we call it something like `htmlTestUtils` instead?

react-components/tests/tests-utilities/cogs/cogsTestUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+  return element.getAttribute('aria-disabled') === 'true';
+}
+
+export function isSelected(element: HTMLElement): boolean {
+  return element.getAttribute('aria-selected') === 'true';
+}
+
+export function getType(element: HTMLElement): string {
+  return element.getAttribute('type') ?? '';
```
anders-hopland: would it be better to return undefined or null if attribute is not found?

anders-hopland (changes requested):
Some small comments, other than that it looks good :100:

react-components/tests/tests-utilities/cogs/cogsTestUtils.ts
nilscognite: Fixed

react-components/tests/tests-utilities/cogs/cogsTestUtils.ts
nilscognite: Fixed

react-components/tests/tests-utilities/cogs/cogsTestUtils.ts
nilscognite: Fixed

react-components/tests/tests-utilities/cogs/htmlTestUtils.ts
nilscognite: Fixed

PR #5410, opened 2025-10-21
fix(react-components): avoid empty query to DMS byids

haakonflatval-cognite:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-6197

## Description :pencil:

Solves an issue where we would sometimes get a 400 error response from the dms byids endpoint because we sent an empty query.

The test changes are not as large as it seems - it is just restructured sligthly. Hide whitespace to see real changes.


christjt (dismissed):
risk review ok

PR #5426, opened 2025-10-23
test(react-components): Add unit test for some sub components of TreeView (TreeView unit test Part 3)

nilscognite:
#### Type of change

Test

#### Jira ticket :blue_book:

[](https://cognitedata.atlassian.net/browse/BND3D-6172)

https://cognitedata.atlassian.net/browse/BND3D-6172

## Description :pencil:

This unit test is for sub component of the tree view. I have to change some code in these subcomponents, due to easier testing in isolation. So there is some untested code in the calling component. This has bad coverage, but will be made in the next part, and it is not in production (so here is the risk 0)

By unit test

**In practice:**

`yarn storybook`
in the end of the list, you will se the components. There are 2 different folders with a lot of three configurations.


react-components/src/tree-view/view/components/tree-view-load-more.tsx
```diff
@@ -1,40 +1,39 @@
...
         marginTop: horizontalSpacing,
         marginLeft
       }}
-      onClick={() => {
-        onClick(node);
+      onClick={async () => {
+        if (node.loadSiblings !== undefined) {
+          await node.loadSiblings(loader);
+        }
       }}>
```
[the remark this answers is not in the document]
nilscognite: Fixed

react-components/stories/tree-view/advanced-tree-view.stories.tsx
```diff
@@ -115,6 +115,9 @@ export const WithBoldLabels: Story = {
 };
 
 const infoRoot = createSimpleTreeMock({});
+for (const node of infoRoot.getThisAndDescendants()) {
```
nilscognite, line 118: This was missing in the story book in order to get the info icon

react-components/src/tree-view/view/components/tree-view-caret.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { render } from '@testing-library/react';
+import { getIconName, getIconsInContainer } from '#test-utils/cogs/htmlTestUtils';
+import { TreeNode } from '../../model/tree-node';
+import { TreeViewCaret, type TreeViewCaretProps } from './tree-view-caret';
+
+describe(TreeViewCaret.name, () => {
+  test('should not render the caret when it is leaf', () => {
+    const node = new TreeNode();
+
+    const container = renderMe({ node });
```
christjt, line 12: Optional / suggestion:


Personally I would just do this:

```suggestion
    const { container } = render(<TreeViewCaret node={node} />);
```

There is almost no code saved from using a common `renderMe` function. And personally I find it easier to read and understand when the actual implementation is just there. Extending the test becomes harder in the current solution. If I f.ex. want to access the `rerender` function from the `render` utility, I now need to change all tests using the `renderMe` function.

react-components/src/tree-view/view/components/tree-view-load-more.tsx
```diff
@@ -1,40 +1,44 @@
...
         marginLeft
       }}
-      onClick={() => {
-        onClick(node);
+      onClick={async () => {
+        if (node.loadSiblings === undefined) return;
+        try {
+          await node.loadSiblings(loader);
+        } catch (error) {
+          console.error('Failed to load more nodes:', error);
```
christjt: Why do we swallow the error here? I also am not a fan of console.error here, since neither the user, nor we will see this. There isn't any real value in console logging this. I would either remove the try / catch completly and then let the error bubble to global scope, or, if we actually don't care about this erroring, do nothing in the catch and just continue execution as if it was successful.

react-components/src/tree-view/view/components/tree-view-caret.test.tsx
nilscognite: I just followed the same pattern everywhere for those element. I think I will leave it as it is. But I see you point

react-components/src/tree-view/view/components/tree-view-load-more.tsx
nilscognite: I wasn't sure about this. I have not find any rules in the code how we should threat. I have remove it the try block.

PR #5435, opened 2025-10-29
feat: batch loading for DM based 360 image data

pramod-cog:
#### Type of change

Feat  

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-5781

## Description :pencil:
This is part one of above tasks to mitigate 429 error. In this PR, I have added batch loading class for Image 360 files and station data for Data model based 360 images.

Unit test

### DIFF
--- a/viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/get360CdmCollectionsQuery.ts
+++ b/viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/get360CdmCollectionsQuery.ts
@@ -11,16 +11,16 @@ import {
 import { DMInstanceRef } from '@reveal/utilities';
 import { CORE_DM_IMAGE_360_PROPERTIES_LIST } from '../../../cdm/queryProperties';
 
-function createCollectionsQuery(collectionRefs: DMInstanceRef[]) {
+function createCollectionsQuery(instanceReferences: DMInstanceRef[]) {
   return {
     with: {
       image_collections: {
         nodes: {
           filter: {
-            instanceReferences: collectionRefs.map(ref => ref)
+            instanceReferences
           }
         },
-        limit: collectionRefs.length
+        limit: instanceReferences.length
       },
       images: {
         nodes: {


viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,413 @@
...
+  private async getCdmFiles(identifiers: DirectRelationReference[]): Promise<FileInfo[]> {
+    const url = `${this._cogniteSdk.getBaseUrl()}/api/v1/projects/${this._cogniteSdk.project}/files/byids`;
+
+    try {
+      const res = (await this._cogniteSdk.post(url, {
+        data: { items: identifiers.map(id => ({ instanceId: id })) }
+      })) as CoreDmFileResponse;
+
+      return res.data.items;
+    } catch (error: unknown) {
```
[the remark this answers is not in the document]
pramod-cog: Following existing pattern in the viewer

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,413 @@
...
+      if (error instanceof Error) {
+        throw new Error(`Failed to fetch CDM files: ${error.message}`);
+      }
+      throw new Error(`Failed to fetch CDM files: ${JSON.stringify(error)}`);
+    }
+  }
```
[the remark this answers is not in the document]
pramod-cog: This casting is necessary and is correct

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+  nextCursor?: Record<string, string>;
+};
+
+interface FileInfoWithInstanceId extends FileInfo {
+  instanceId?: DirectRelationReference;
+}
+
+// Note: DMS properties are returned as string | number | DMInstanceRef (which is DirectRelationReference)
+// We need to handle this flexibility and convert to the correct types when needed
+type DmsPropertyValue = string | number | DMInstanceRef;
```
haakonflatval-cognite: You could use the `RawPropertyValueV3` type from the SDK instead of this, I think 🤔

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+
+/**
+ * Coordinates batched loading of multiple 360 image collections.
+ * Instead of loading each collection individually (1000 queries for 1000 collections),
+ * it batches them into groups and makes far fewer queries (e.g., 20 queries for 1000 collections).
+ *
+ * Note: DMS API has a limit of 10,000 items per query. With a batch size of 50 collections,
+ * this allows ~200 images per collection on average. If your collections have more images,
```
haakonflatval-cognite: Just so I understand - changing the batch size would require changing the code, right? It sounds like a big shortcoming if we can't load collections with more than 200 images per collection (provided we have a lot of collections) - is there no way we can mitigate this?

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+      if (this._pendingBatch.length >= this._batchSize) {
+        void this.tryExecuteBatch();
+      } else {
+        // Otherwise, wait a bit to accumulate more requests
+        this._batchTimer = setTimeout(() => {
+          void this.tryExecuteBatch();
+        }, this._batchDelay);
+      }
+    });
+  }
```
haakonflatval-cognite: I really like this approach, but I'm worried if there can be some race conditions here.. For instance if one collection ID is added, causing the `pendingBatch` list to go over its limit - there may be a small delay between `void this.tryExecuteBatch()` is being called, and the `pendingBatch` queue being cleared. Which means that there could be another call to `getCollectionDescriptors` adding another ID to the list, which is then picked up by the `executeBatch` call triggered from the previous ID. I guess it's not that serious; we would still process everything, but it can cause batches of length higher than the limit we set

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+ * Note: DMS API has a limit of 10,000 items per query. With a batch size of 50 collections,
+ * this allows ~200 images per collection on average. If your collections have more images,
+ * consider reducing the batch size.
+ */
+export class Cdf360CdmBatchCollectionLoader {
+  private readonly _dmsSdk: DataModelsSdk;
+  private readonly _cogniteSdk: CogniteClient;
+  private readonly _batchSize: number;
+  private _pendingBatch: Array<{
+    identifier: { externalId: string; space: string };
```
haakonflatval-cognite:
```suggestion
    identifier: DMInstanceRef
```

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+      const query = get360CdmCollectionsQuery(collectionRefs);
+
+      const result = await this._dmsSdk.queryNodesAndEdges<CdfImage360CollectionDmQuery>(query);
+
+      // Group results by collection
+      const resultsByCollection = await this.groupResultsByCollection(result);
+
+      // Resolve each request with its specific results
+      batch.forEach(({ identifier, resolve }) => {
+        const key = `${identifier.externalId}:${identifier.space}`;
```
haakonflatval-cognite: We should have an e.g. `createDMKey` function, either in this file or centralized, so that we are sure the key looks the same all places it's created

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+        const descriptors = resultsByCollection.get(key);
+
+        if (descriptors) {
+          resolve(descriptors);
+        } else {
+          // Collection not found
+          console.warn(`Collection not found: ${key}`);
```
haakonflatval-cognite: Could we add more info here to make it easier to debug if it triggers? E.g. `` console.warn(`[${Cdf360CdmBatchCollectionLoader.name}]: ...`); ``

Same for the rest of the `console.warn/error`s in this file

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/get360CdmCollectionsQuery.ts
```diff
@@ -0,0 +1,145 @@
...
+      },
+      images: {
+        sources: [
+          {
+            source: {
+              type: 'view' as const,
+              space: coreDmSpace,
+              externalId: 'Cognite360Image',
+              version: 'v1'
+            },
```
haakonflatval-cognite: We have a `COGNITE_360_IMAGE_SOURCE` you can use here

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/get360CdmCollectionsQuery.ts
```diff
@@ -0,0 +1,145 @@
...
+            properties: [
+              'translationX',
+              'translationY',
```
haakonflatval-cognite: You can use the constant `CORE_DM_IMAGE_360_PROPERTIES_LIST` here

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/get360CdmCollectionsQuery.ts
```diff
@@ -0,0 +1,145 @@
...
+        ]
+      },
+      stations: {
+        sources: [
+          {
+            source: {
+              type: 'view' as const,
+              space: coreDmSpace,
+              externalId: 'Cognite360ImageStation',
+              version: 'v1'
```
haakonflatval-cognite: Could we drag this out into a constant in the `data-providers/src/utilities/constants.ts` file? (Same for the other sources here that are not already in that file)

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/get360CdmCollectionsQuery.ts
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * Creates a DMS query to fetch multiple 360 image collections in a single request.
+ * This is much more efficient than making individual queries per collection.
+ *
+ * @param collectionRefs - Array of collection identifiers to fetch
+ * @returns DMS query that fetches all collections and their images in one request
+ */
+export function get360CdmCollectionsQuery(
+  collectionRefs: Array<{ externalId: string; space: string }>
```
haakonflatval-cognite, line 86: Can we just use type `DMInstanceRef[]` here?

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/types.ts
```diff
@@ -0,0 +1,32 @@
...
+
+// Collection types - batch query uses 'image_collections' (plural)
+export type CollectionInstanceResult = QueryResult['image_collections'][number];
+
+// File response type from CDM files API
+export type CoreDmFileResponse = {
+  data: {
+    items: FileInfo[];
+  };
+};
```
haakonflatval-cognite: Since all of these are only used internally in `Cdf360CdmBatchCollectionLoader`, I think I would prefer moving them there (also, seems like `ImageResult` is not used at all?

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+
+    // Get file descriptors for all images in batch
+    const allFileDescriptors = await this.getFileDescriptorsForBatch(result.images);
+
+    // Group images by their collection
+    const imagesByCollection = new Map<string, ImageInstanceResult[]>();
+
+    result.images.forEach(image => {
+      const collectionRef = image.properties?.cdf_cdm?.['Cognite360Image/v1']?.collection360;
+      if (collectionRef && typeof collectionRef === 'object' && 'externalId' in collectionRef) {
```
haakonflatval-cognite: I think I would recommend using the `isDMIdentifier` type guard here... It will have to be modified to take  `unknown` as argument, but I think that's fine to do here

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+          imagesByCollection.set(key, []);
+        }
+        imagesByCollection.get(key)!.push(image);
+      }
+    });
+
+    // Process each collection
+    result.image_collections.forEach(collection => {
+      const collectionKey = `${collection.externalId}:${collection.space}`;
+      const collectionImages = imagesByCollection.get(collectionKey) || [];
```
haakonflatval-cognite: I would prefer `??` over `||` here

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+
+      if (collectionImages.length === 0) {
+        grouped.set(collectionKey, []);
+        return;
+      }
+
+      const collectionId = collection.externalId;
+      const name = collection.properties?.cdf_cdm?.['Cognite360ImageCollection/v1']?.name;
+      const collectionLabel = typeof name === 'string' ? name : typeof name === 'number' ? String(name) : undefined;
```
haakonflatval-cognite: [Nitpick] I would want a separate utility for extracting this string

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+      );
+
+      grouped.set(collectionKey, historicalSets);
+    });
+
+    return grouped;
+  }
+
+  private async getFileDescriptorsForBatch(images: ImageInstanceResult[]): Promise<Map<string, FileInfo>> {
+    const imageProps = images.map(image => image.properties.cdf_cdm['Cognite360Image/v1']);
```
haakonflatval-cognite, line 185: Just a side-note, I think by adding the source-consts I commented about in the query-file, you will get type safety on property accesses like this

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+    const cubeMapFileIds = imageProps.flatMap(imageProp => {
+      const faces = [imageProp.front, imageProp.back, imageProp.left, imageProp.right, imageProp.top, imageProp.bottom];
+      return faces.filter((face): face is DirectRelationReference => typeof face === 'object' && 'externalId' in face);
```
haakonflatval-cognite: We can use `isDMIdentifier` here as well

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+      } catch (error) {
+        console.error(`Failed to fetch file batch ${i + 1}/${batches.length}:`, error);
+        throw error;
+      }
+    }
+
+    // Create map for quick lookup by externalId:space
+    // DMS DirectRelationReference uses { space: "...", externalId: "..." }
+    // where externalId is a string UUID, not a numeric ID
+    const fileMap = new Map<string, FileInfo>();
```
haakonflatval-cognite: Would also be convenient if the `key` type of the `Map` has an alias (e.g. `` `${ExternalId}:${Space}` ``)

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+      const faces = [props.front, props.back, props.left, props.right, props.top, props.bottom];
+
+      const faceFiles = faces
+        .filter((face): face is DirectRelationReference => typeof face === 'object' && 'externalId' in face)
+        .map(ref => {
+          // Create the same key format used in the file map
+          const key = `${ref.externalId}:${ref.space}`;
+          return fileMap.get(key);
+        })
+        .filter((f): f is FileInfo => f !== undefined);
```
haakonflatval-cognite: I wish we had a `isDefined` type guard here 🥲

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+    const eulerRotation = new Euler(rx, rz, -ry, 'XYZ');
+    const rotation = new Matrix4().makeRotationFromEuler(eulerRotation);
+
+    return translation.multiply(rotation);
+  }
+
+  private async getCdmFiles(identifiers: DirectRelationReference[]): Promise<FileInfo[]> {
+    const url = `${this._cogniteSdk.getBaseUrl()}/api/v1/projects/${this._cogniteSdk.project}/files/byids`;
+
+    try {
```
haakonflatval-cognite: It looks like you can just use `this._cogniteSdk.files.retrieve` here, it supports `instanceId` in input and output.. Or am I missing something?

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+  private createHistoricalImageSets(
+    collectionId: string,
+    collectionLabel: string,
+    images: ImageInstanceResult[],
+    fileDescriptors: FileInfo[][]
+  ): Historical360ImageSet<DMDataSourceType>[] {
+    const imagesGroupedWithFileDescriptors = zip(images, fileDescriptors)
+      .filter(
+        (pair): pair is [ImageInstanceResult, FileInfo[]] =>
+          pair[0] !== undefined && pair[1] !== undefined && pair[1].length === 6
```
haakonflatval-cognite: Disappointed that `zip` doesn't help us with the typing here 🥲

haakonflatval-cognite: [Nitpick] I'm not super-fan of using `zip` on two different inputs into the same function, just assuming that they are of equal length and matching. Personally, I would prefer that `getFileDescriptorsForImages` returned objects or tuples with each image and its corresponding file info directly.

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+    revisionId: Image360RevisionId<DMDataSourceType>,
+    imageProps: ImageResultProperties,
+    fileInfos: FileInfo[]
+  ): Image360Descriptor<DMDataSourceType> {
+    const timestamp = imageProps.takenAt;
+    return {
+      id: revisionId,
+      faceDescriptors: this.getFaceDescriptors(fileInfos),
+      timestamp:
+        typeof timestamp === 'string' ? timestamp : typeof timestamp === 'number' ? String(timestamp) : undefined
```
haakonflatval-cognite: I see we do this a couple of places - can we create a utility for it? (Also, I wonder if it might go away if we get better type safety for the query)

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+  }
+
+  private getFaceDescriptors(fileInfos: FileInfo[]): Image360FileDescriptor[] {
+    return [
+      { fileId: fileInfos[0].id, face: 'front', mimeType: fileInfos[0].mimeType! },
+      { fileId: fileInfos[1].id, face: 'back', mimeType: fileInfos[1].mimeType! },
+      { fileId: fileInfos[2].id, face: 'left', mimeType: fileInfos[2].mimeType! },
+      { fileId: fileInfos[3].id, face: 'right', mimeType: fileInfos[3].mimeType! },
+      { fileId: fileInfos[4].id, face: 'top', mimeType: fileInfos[4].mimeType! },
+      { fileId: fileInfos[5].id, face: 'bottom', mimeType: fileInfos[5].mimeType! }
```
haakonflatval-cognite, line 323: Should we console-warn/throw if one of the file infos don't have a mime type? (Or invalid mime type?)

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+    const groups = groupBy(imagesWithStation, imageResult => {
+      const station = imageResult.image.properties.cdf_cdm['Cognite360Image/v1'].station360;
+      if (station && typeof station === 'object' && 'externalId' in station) {
+        return `${station.externalId}-${station.space}`;
+      }
+      return 'unknown';
+    });
+
+    return Object.values(groups)
+      .concat(imagesWithoutStation.map(p => [p]))
```
haakonflatval-cognite, line 271: Do we need to keep track of the images without station from here?

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,424 @@
...
+          return num;
+        }
+      }
+      throw new Error(
+        `Invalid value for transformation property, expected number or numeric string, got ${JSON.stringify(value)}`
+      );
+    };
+
+    const [x, y, z] = [
+      toNumber(revision.translationX),
```
haakonflatval-cognite: It seems excessive that we need to check whether these fileds are strings when the DMS type states that they are number 🤔 But again, hopefully goes away when you improve the query typing

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,427 @@
...
+import { BatchLoader } from '../../../../utilities/BatchLoader';
+
+// DMS query result type - using batch query structure
+type QueryResult = Awaited<ReturnType<typeof DataModelsSdk.prototype.queryNodesAndEdges<CdfImage360CollectionDmQuery>>>;
+
+type ImageInstanceResult = QueryResult['images'][number];
+type ImageResultProperties = ImageInstanceResult['properties']['cdf_cdm']['Cognite360Image/v1'];
+
+// Collection types - batch query uses 'image_collections' (plural)
+type CollectionInstanceResult = QueryResult['image_collections'][number];
```
haakonflatval-cognite: [Nitpick] Not sure I understand what this comment is trying to help with 🤔

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,427 @@
...
+
+type BatchQueryResult = {
+  image_collections: CollectionInstanceResult[];
+  images: ImageInstanceResult[];
+  stations?: DMInstanceRef[];
```
haakonflatval-cognite: Just wondering - why is this field optional and the others not?

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,427 @@
...
+        accumulatedResults.stations = [...(accumulatedResults.stations || []), ...result.stations];
+      }
+
+      // Update cursor for next page
+      currentCursor = result.nextCursor;
+
+      // Stop if no cursor (backward compat with mocks/older implementations)
+      if (!currentCursor) {
+        break;
+      }
```
haakonflatval-cognite: So, if I understand correctly - we will never get a response without a `nextCursor`

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,380 @@
...
+        accumulatedResults.image_collections = [...accumulatedResults.image_collections, ...result.image_collections];
+      }
+      if (result.images) {
+        accumulatedResults.images = [...accumulatedResults.images, ...result.images];
+      }
+      if (result.stations) {
+        accumulatedResults.stations = [...(accumulatedResults.stations || []), ...result.stations];
+      }
+
+      currentCursor = result.nextCursor;
```
haakonflatval-cognite: Are we sure that this is the correct way of paginating the DMS-queries? I'm not really sure my self, so just double-checking 😅

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,380 @@
...
+      try {
+        const batchFileInfos = await this.getCdmFiles(batch);
+        fileInfos.push(...batchFileInfos);
+      } catch (error) {
+        console.error(`Failed to fetch file batch ${i + 1}/${batches.length}:`, error);
+        throw error;
+      }
+    }
+
+    const fileMap = new Map<string, FileInfo>();
```
haakonflatval-cognite: Can we use `DMInstanceKey` instead of `string` here? (Same with the other `Map`s using DM identifiers as keys)

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.test.ts
```diff
@@ -0,0 +1,282 @@
...
+    const batchLoader = new Cdf360CdmBatchCollectionLoader(dmsSdkMock.object(), cogniteSdkMock.object());
+
+    const results = await Promise.all(
+      collectionIds.map(id => batchLoader.getCollectionDescriptors({ externalId: id, space: 'test_space' }))
+    );
+
+    expect(results).toHaveLength(3);
+    results.forEach((result, idx) => {
+      expect(result.length).toBeGreaterThan(0);
+      if (result.length > 0) {
```
haakonflatval-cognite: Is this if-check necessary?

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.test.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    const result = await batchLoader.getCollectionDescriptors({
+      externalId: 'empty_collection',
+      space: 'test_space'
+    });
+
+    expect(result).toHaveLength(0);
+  });
+
+  test('should batch requests that arrive within delay window', async () => {
```
haakonflatval-cognite: I would maybe argue that this is already tested by the base class and doesn't have to be tested here. But we can keep it if you want

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.test.ts
```diff
@@ -0,0 +1,282 @@
...
+      } else {
+        filesApiMock.setup(instance => instance.retrieve(It.IsAny())).returns(Promise.resolve(filesResponse));
+      }
+    } else if (filesError) {
+      filesApiMock
+        .setup(instance => instance.retrieve(It.IsAny()))
+        .callback(async () => {
+          throw filesError;
+        });
+    }
```
haakonflatval-cognite: It looks like this can simplified by re-ordering the if's:
```
if (filesError) {
 ...
} else if (filesResponse) {
 ...
}
```

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.test.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    const result = await batchLoader.getCollectionDescriptors({
+      externalId: 'nonexistent_collection',
+      space: 'test_space'
+    });
+
+    expect(result).toHaveLength(0);
+  });
+
+  function createMockSdks(
```
haakonflatval-cognite: SdkS? 🧐

haakonflatval-cognite (commented):
Had a few minor comments, but otherwise looks good 👍 Nicely done

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
anders-hopland: Cursors in DMS work like follows:
```
"nextCursor": {
    "cursorForClause1": "lsdfjlsdnf",
    "cursorForClause2": "sdvdlnlre",
    ...
    "cursorForClauseN": "svnlwejfoiewø"
}
```

This means that when you are traversing the knowledge graph you might e.g. want to get all cad nodes connected to all models. If you then query for 10 models at a time, you will get a cursor for both models and cad nodes. In order to use the same result set of models to fetch more cad nodes, you would then only pass the new cursor for the cadnodes, while sending the previous cursor for models.
NB: In DMS, you will get a cursor even if there is no more data to fetch. The DM team has said that if we get less instances than we asked for, we can assume that there is no more data to be fetched

pramod-cog: @anders-hopland This is 360 image data and here with current assignment of nextCursor is correct or should I consider to assign differently?

anders-hopland: I tried to give a broad explaination of how cursoring works. I have now checked out the branch and sifted through the code a bit. For this 36 image query, what I would do is that I would only cursor the 360 images themselves. It does not look like you need to paginate the collections as you will be able to fetch them all in the first query. Teh cursor you pass for sbsequent queries will then be 
```
nextcursor: {
    images: "sdfsdf"
}
```

You do not need to paginate stations as you have a 1-1 relationship between images and stations, and you fetch stations through images

viewer/packages/data-providers/src/utilities/dmsPaginationUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @param results - The array of results from the current page
+ * @param nextCursor - The cursor from the DMS response (object with keys or string)
+ * @param cursorKey - Optional key to extract from the cursor object (e.g., 'images', 'nodes'). If not provided, treats nextCursor as a string.
+ * @param limit - The limit used in the query
+ * @returns The cursor string if pagination should continue, undefined otherwise
+ *
+ */
+export function getDmsPaginationCursor(
+  results: unknown[] | undefined,
+  nextCursor: Record<string, string> | string | undefined,
```
anders-hopland: The API spec states that nextCursor is always an object, so I think that allowing string and undefined is wrong
https://api-docs.cognite.com/20230101/tag/Instances/operation/queryContent

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,377 @@
...
+      if (result.images && result.images.length > 0) {
+        accumulatedResults.images = [...accumulatedResults.images, ...result.images];
+      }
+
+      if (result.stations && result.stations.length > 0) {
+        accumulatedResults.stations = [...(accumulatedResults.stations || []), ...result.stations];
+      }
+
+      // Check if we should continue pagination for images
+      imagesCursor = getDmsPaginationCursor(result.images, result.nextCursor, 'images', MAX_DMS_QUERY_LIMIT);
```
anders-hopland, line 117: I think that it might be just as easy to do this inline instead of generalizing it into a function. In some instances, you might want to look at multiple cursors to determine how to paginate and I think that we need a different abstraction for this

viewer/packages/data-providers/src/utilities/dmsPaginationUtils.ts
pramod-cog: If you check the doc, it states the cursor to be string

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
pramod-cog: Isn't the cursorKey take care of this?
And @haakonflatval-cognite suggested to have a utility function for dms query

viewer/packages/data-providers/src/utilities/dmsPaginationUtils.ts
anders-hopland: Do you have a link to the doc? I see 
image

pramod-cog: Screenshot 2025-11-04 at 15 00 29
??

anders-hopland: But do you intend to take in a nextCursor object or unpack a cursor from the nextCursor object? I think that we should not support both in the same function. What do you think?

viewer/packages/data-providers/src/utilities/constants.ts
```diff
@@ -41,3 +41,21 @@ export const COGNITE_360_IMAGE_SOURCE = {
...
+
+export const COGNITE_360_IMAGE_STATION_SOURCE = {
+  externalId: 'Cognite360ImageStation',
+  space: CORE_DM_SPACE,
+  version: 'v1',
+  type: 'view'
+} as const satisfies ViewReference;
+
+export const BATCH_SIZE = 50;
+export const BATCH_DELAY_MS = 50;
```
haakonflatval-cognite: Should `BATCH_SIZE` (and maybe `BATCH_DELAY_MS`) be defined locally in each implementation of the batch loader instead? I imagine the best batch size may depend on what we are fetching

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/get360CdmCollectionsQuery.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+      },
+      {
+        equals: {
+          property: ['node', 'space'] as const,
+          value: ref.space
+        }
+      }
+    ]
+  }));
```
haakonflatval-cognite: I think you can use the `instanceReferences` filter here, which may help us avoid maxing out the number of and/or clauses in the query

Image

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    const allFileDescriptors = await this.getFileDescriptorsForBatch(result.images);
+
+    const imagesByCollection = new Map<DMInstanceKey, ImageInstanceResult[]>();
+
+    result.images.forEach(image => {
+      const collectionRef = image.properties?.cdf_cdm?.['Cognite360Image/v1']?.collection360;
+      if (isDmIdentifier(collectionRef)) {
+        const key = dmInstanceRefToKey(collectionRef);
```
haakonflatval-cognite: I would have liked this if-statement to be inverted

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,377 @@
...
+      } catch (error) {
+        console.error(`Failed to fetch file batch ${i + 1}/${batches.length}:`, error);
+        throw error;
+      }
+    }
+
+    const fileMap = new Map<DMInstanceKey, FileInfo>();
+
+    fileInfos.forEach(file => {
+      const fileWithInstanceId = file as FileInfoWithInstanceId;
```
haakonflatval-cognite: Is this cast necessary? Can we avoid it?

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.ts
```diff
@@ -0,0 +1,377 @@
...
+
+    const [imagesWithoutStation, imagesWithStation] = partition(imagesGroupedWithFileDescriptors, image => {
+      const station = image.image.properties.cdf_cdm['Cognite360Image/v1'].station360;
+      return !isDmIdentifier(station);
+    });
+
+    const groups = groupBy(imagesWithStation, imageResult => {
+      const station = imageResult.image.properties.cdf_cdm['Cognite360Image/v1'].station360;
+      if (isDmIdentifier(station)) {
+        return `${station.externalId}-${station.space}`;
```
haakonflatval-cognite: Can we use the DM key creation utility here too?

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/Cdf360CdmBatchCollectionLoader.test.ts
```diff
@@ -0,0 +1,245 @@
...
+          }
+        }
+      }
+    }));
+
+    const images = collectionIds.flatMap((collectionId, idx) =>
+      Array.from({ length: 2 }, (_, imageIdx) => ({
+        instanceType: 'node' as const,
+        version: 1,
+        space: 'test_space',
```
haakonflatval-cognite, line 188: We have a collection mock in the `mockGetImage360WithAnnotationsFromRevisionsResponse.ts` - can we re-use it here somehow? You might have to factor it out in a separate test-utils file, but I think I would prefer that over duplicating this giant fixture here

pramod-cog: This would require lot more changes to accommodate to reuse `mockGetImage360WithAnnotationsFromRevisionsResponse`, so I would prefer to keep as it is for now because I use array of collection which require it's images

viewer/packages/data-providers/src/image-360-data-providers/descriptor-providers/datamodels/cdm/get360CdmCollectionsQuery.ts
```diff
@@ -12,16 +12,12 @@ import { DMInstanceRef } from '@reveal/utilities';
...
-    instanceReferences: ref
-  }));
-
   return {
     with: {
       image_collections: {
         nodes: {
           filter: {
-            or: collectionFilters
+            instanceReferences: collectionRefs.map(ref => ref)
```
haakonflatval-cognite: This `.map` seems unnecessary? :eyes:

haakonflatval-cognite (approved):

pramod-cog:
risk review ok

PR #5437, opened 2025-10-31
chore(webpack): add GLSL import tracker loader to track import dependencies in --watch mode 

Strepto:
#### Type of change
Chore 

#### Jira ticket :blue_book:

N/A

## Description :pencil:

When changing shader file dependencies today you need to manually rebuild changes as webpack watch does not track dependencies for glslify imports. This PR adds a custom webpack loader to track imports.

If this change is not needed feel free to dismiss this PR.

Tested in local development, works.

The loader does not mutate the files so there should be no runtime or build-time changes

Start viewer with `yarn build:watch` and change a .glsl file (for example updateFragmentColor.glsl) and see that webpack now will auto-rebuild.

### DIFF
--- a/viewer/glsl-import-tracker-loader.cjs
+++ b/viewer/glsl-import-tracker-loader.cjs
@@ -1,0 +1,83 @@
+const fs = require('fs');
+const path = require('path');
+
+/**
+ * Helper to recursively finds all unique GLSL `import` dependencies in a file.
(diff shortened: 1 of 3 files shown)


christjt:
We migrated to vite, so this PR is not relevant anymore (though the problem still might be :P)

PR #5573, opened 2026-04-14
docs: added high detailed prioritized node details in cad prioritized nodes page

pramod-cog:
#### Type of change

Docs 

#### Jira ticket :blue_book:

## Description :pencil:
Added documentation of High prioritized detailed nodes into `cad-prioritized-nodes.mdx`


danpriori (dismissed):
The text seems well-written and easy to understand.
If I understood correctly, I would only replace the concept of "pattern" with "feature" or "option" because, from what I understand, it's possible to combine these features in the same environment, such as ghosting and blocking sectors at the same time, for example. So, they are not distinct things, right?

pramod-cog:
> The text seems well-written and easy to understand. If I understood correctly, I would only replace the concept of "pattern" with "feature" or "option" because, from what I understand, it's possible to combine these features in the same environment, such as ghosting and blocking sectors at the same time, for example. So, they are not distinct things, right?

Nice catch, updated with `Option`

danpriori (approved):
Nice!

pramod-cog:
risk review ok

PR #5580, opened 2026-04-16
fix: reuse cluster DOM elements when composition unchanged

sarahse:
Hello! I am a frontend developer in the Echo3d team at Equinor. We use Reveal everyday in Echo3d, and lately we have been experiencing some issues with clustering of markers. We have tried to fix this bug several times in Echo3d, but I think there might be some changes needed on Reveal's side in order to get to a proper fix for this issue.

Disclaimer: I'm not sure this is the best solution for this problem, as I'm not that familiar with Reveal, but I thought it was easier to describe the problem and show the bug we have in Echo3d by creating a PR. 

**This is the problem we have: The user needs to click the cluster two times before it is opened and zoomed in. There is also a lot of flickering happening (see attached videos)**

#### Type of change

Bug 

#### Jira ticket :blue_book:

### Problem
Reveal's HtmlOverlayTool.forceUpdate() destroys all cluster DOM elements and recreates them from scratch every frame. When a user clicks a cluster, the element is removed from the DOM between mousedown and mouseup, so the browser sets the click target to the viewer container instead of the cluster element — the click is silently lost. Users must click twice to zoom into a cluster.

This also causes visible flickering as cluster elements are removed and re-added to the DOM every frame during camera movement.

### Solution
Cache composite (cluster) elements keyed by sorted overlay member IDs. When the same overlays are clustered together across frames, reuse the existing DOM element and only update its position. createClusterElementCallback is only called when cluster composition actually changes.

This keeps cluster elements alive in the DOM across frames, so click events fire correctly and flickering is eliminated.

Tested in Echo3d by running Reveal locally.

1. Run Echo3d locally and add enough tags to make a cluster when zooming out.
2. Experiment with clicking the cluster right after the zoom out.
3. Observe visibly less flickering.
4. Observe that you only need to click once to open the cluster.

## Video demo
Before: I need to click twice to open clusters if I click them right after a zoom event.

After: I only need to click the cluster once before it zooms in and opens the cluster.


viewer/packages/tools/src/HtmlOverlay/HtmlOverlayTool.ts
pramod-cog: This comment is now incorrect. With the caching approach, `createClusterElementCallback` is only called when the cluster composition changes, not every frame. Update to reflect the new behavior.

pramod-cog, line 276: After removing an overlay, any `_clusterCache` entry keyed to that element's ID is not invalidated here. The stale cluster composite stays visible in the DOM until the next `forceUpdate()` when `removeStaleClusters()` eventually evicts it. To fix, derive the removed element's ID via `this._elementIds.get(htmlElement)`, then delete any cache entry whose key string contains that ID.

pramod-cog: This push is immediately overwritten by `removeStaleClusters()` (called at line 521), which clears and rebuilds `_compositeOverlays` entirely from `_aliveClusterKeys`. The push here is dead work. `addComposite` should only set visibility/left/top and leave array management to `removeStaleClusters`.

viewer/packages/tools/src/HtmlOverlay/HtmlOverlayTool.ts
```diff
@@ -204,6 +212,8 @@ export class HtmlOverlayTool extends Cognite3DViewerToolBase {
     this._viewer.off('sceneRendered', this._onSceneRenderedHandler);
     this._viewer.off('disposed', this._onViewerDisposedHandler);
     this.clear();
+    this._clusterCache.clear();
+    this._aliveClusterKeys.clear();
```
pramod-cog: `clear()` (line 214) already calls `this._clusterCache.clear()` (line 289) and `this._aliveClusterKeys.clear()` (line 290), so lines 215–216 are redundant. Minor, but safe to remove.

pramod-cog (changes requested):
Thanks for your contribution, we appreciate the fix.
Have some comments added and also if you can address gemini comments will be appreciated. Let me know if you need more details, would be happy to provide.

pramod-cog (dismissed):
LGTM!
@haakonflatval-cognite Do you want to have a look at the PR? If not, we can merge this

sarahse:
Can you await merging? I'd like to test it out once more in echo3d just to double check after the new commits 😊 I'll write here when I have tested it enough!

sarahse:
I tested it out now, and it works well after the latest commit. Thanks for the thorough review process 🙏 If you want, you can trigger gemini again to check the latest commits 😊

pramod-cog:
@sarahse Can you check and address the gemini comment

sarahse:
> @sarahse Can you check and address the gemini comment

Yes, I'll fix it. Many meetings today. But I'll fix it right after lunch!

viewer/packages/tools/src/HtmlOverlay/HtmlOverlayTool.ts
```diff
@@ -259,16 +280,21 @@ export class HtmlOverlayTool extends Cognite3DViewerToolBase {
...
+    // Each remove() call invalidates cluster cache entries referencing the
+    // removed overlay, so by the end of this loop the cluster cache and any
+    // associated composite DOM elements are already cleaned up.
     const overlays = Array.from(this._htmlOverlays.keys());
     for (const element of overlays) {
       this.remove(element);
     }
+    this._compositeOverlays.splice(0);
     this.forceUpdate();
   }
```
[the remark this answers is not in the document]
sarahse: Fixed

viewer/packages/tools/src/HtmlOverlay/HtmlOverlayTool.ts
```diff
@@ -410,10 +436,52 @@ export class HtmlOverlayTool extends Cognite3DViewerToolBase {
...
+      return;
+    }
+    for (const [key, compositeElement] of this._clusterCache.entries()) {
+      if (!key.split(',').includes(`${removedId}`)) {
+        continue;
+      }
+      compositeElement.parentNode?.removeChild(compositeElement);
+      this._clusterCache.delete(key);
+      this._aliveClusterKeys.delete(key);
+    }
```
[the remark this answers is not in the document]
sarahse: Fixed

sarahse:
> > @sarahse Can you check and address the gemini comment
> 
> Yes, I'll fix it. Many meetings today. But I'll fix it right after lunch!

Fixed now.

viewer/packages/tools/src/HtmlOverlay/HtmlOverlayTool.ts
```diff
@@ -450,12 +525,36 @@ export class HtmlOverlayTool extends Cognite3DViewerToolBase {
...
+              // Overlay reached clustering before add()'s deferred ID
+              // assignment ran (or via a path that bypassed it). Assign
+              // lazily so a stable key can still be produced.
+              id = this._nextElementId++;
+              this._elementIds.set(el.htmlElement, id);
+            }
+            return id;
+          })
+          .sort((a, b) => a - b)
+          .join(',');
```
haakonflatval-cognite: [Optional]: I'd prefer this to be its own function (e.g. `createClusterKey`), and likewise with the `split`+`includes`-check on line 458 (i.e. `clusterKeyContainsId`), especially since those two implementations must be kept in sync

haakonflatval-cognite (dismissed):
Had an optional comment, but I'll let it slide if you don't have time to look at it.

Good job 👍

pramod-cog:
risk review ok

viewer/packages/tools/src/HtmlOverlay/HtmlOverlayTool.ts
sarahse: Will fix 👍

sarahse: Fixed now!

haakonflatval-cognite (approved):
LGTM 👍 

Unfortunately, I see we have an issue with the `wasm-pack` package in this repo right now, which is blocking merging. We are looking into if we can get around this somehow, since it's uncertain when a fix will be deployed from `wasm-pack`s end (https://github.com/wasm-bindgen/wasm-pack/issues/1577)

sarahse:
> LGTM 👍
> 
> Unfortunately, I see we have an issue with the `wasm-pack` package in this repo right now, which is blocking merging. We are looking into if we can get around this somehow, since it's uncertain when a fix will be deployed from `wasm-pack`s end ([wasm-bindgen/wasm-pack#1577](https://github.com/wasm-bindgen/wasm-pack/issues/1577))

Ok 👌

pramod-cog:
@sarahse Can you merge the latest master into this branch

pramod-cog:
risk review ok

PR #5585, opened 2026-04-23
fix: floored 360 icon not visible without secondary reference model & Reveal bump to v4.32.3

pramod-cog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:
The Floored 360 icons were not visible when loaded without secondary model (PC or Cad) because `WebGLRenderer` sorts transparent objects which needs each object's `boundingSphere` and `InstancedMesh.boundingSphere` starts as `null` where renderer calls `mesh.computeBoundingSphere()`. As we use instance mesh without `geometry.boundingBox`, the bounding sphere computation fails silently or produces an empty/NaN sphere, with an invalid bounding sphere, the sort depth is wrong and the image 360 disc is sorted behind opaque geometry making it invisible.
This PR fixes the issue by triggering compute bounding box

In examples

- Run [example](https://localhost:3549/?project=3d-test&env=greenfield-3d&modelId=4923307033982497&revisionId=3100284276970593&siteId=hibernia_rs2&space=scene&floorIcons=true)
- Remove the cad model and load the 360 image collection
- Click on any 360 image icon and check floored icon are visible

### DIFF
--- a/viewer/package.json
+++ b/viewer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cognite/reveal",
-  "version": "4.32.2",
+  "version": "4.32.3",
   "description": "WebGL based 3D viewer for CAD and point clouds processed in Cognite Data Fusion.",
   "homepage": "https://github.com/cognitedata/reveal/tree/master/viewer",
   "repository": {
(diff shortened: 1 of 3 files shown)


anders-hopland (approved):
LGTM

anders-hopland:
🦄

PR #5626, opened 2026-05-21
feat(viewer): Fine-tuning the clusters by making them dim down and being invisible when they are occluded.

danpriori:
#### Type of change

Feat  

#### Jira ticket :blue_book:
[BND3D-7039](https://cognitedata.atlassian.net/browse/BND3D-7039)

## Description :pencil:
This PR adds a system for fading and occluding clusters that are behind other clusters and within a certain distance of the camera. The feature also has an effect between collections via a coordinator. The fade range and camera distance parameters can be customized as input.

It was tested in Fusion 3D - Unified 3D using scenes such as Hybrid Mappings with large 360s collections.

- In examples or in Unified-3D with the enableHtmlClusters flag enabled, load a large, well-spaced 360s collection to notice the differences.
- If you have access to Fusion, a suggestion is to use the Hibernia collection or open the Hybrid Mapping scene in 3d-test, which has more than one large Hibernia collection.
- Navigate between the clusters and observe the clusters that are further away have a fade effect and the ones furthest away become invisible. Also observe that the effect also occurs in cross collections.

[screen-capture (39).webm](https://github.com/user-attachments/assets/ac24d11c-b417-4abb-a66d-b6a3ccd24b56)

Also a bit more aggressive in clusterization (more 360s points aggregated into a cluster):

Before:
image

After:
image

[BND3D-7039]: https://cognitedata.atlassian.net/browse/BND3D-7039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


viewer/packages/api/src/public/migration/types.ts
```diff
@@ -40,6 +40,23 @@ export type RenderParameters = {
...
+export type HtmlClusterOptions = {
+  /** Camera distance (world units) at which an occluded cluster begins to fade out. */
+  fadeStartDistance?: number;
+  /** Camera distance (world units) at which an occluded cluster becomes fully invisible. */
+  fadeEndDistance?: number;
+  /** Maximum distance (world units) between icons for them to be grouped into a cluster. */
+  clusterDistanceThreshold?: number;
+  /** Maximum octree depth for clustering. Lower values produce fewer, larger clusters. */
+  maxOctreeDepth?: number;
+};
```
haakonflatval-cognite, line 68: This is nice, but I'm wondering if it's a little too much details for a user to be able to use properly? It also makes it a little harder for us to change the implementation of the HTML clustering in the future, if we must maintain support for all these parameters 🤔 But maybe not critical..

Anyway, if we want to keep them, could we write down the default values for these fields in the TSDoc? It would make it easier for users to tweak them if they find the need

viewer/packages/360-images/src/types.ts
```diff
@@ -44,3 +44,18 @@ export type Image360ClusterIntersectionData<T extends DataSourceType = DataSourc
...
+ */
+export type HtmlClusterRendererOptions = {
+  maxPoolSize?: number;
+  classPrefix?: string;
+  enableHoverAnimations?: boolean;
+  zIndex?: number;
+  clusterFadeStartDistance?: number;
+  clusterFadeEndDistance?: number;
+  clusterDistanceThreshold?: number;
+  maxOctreeDepth?: number;
```
haakonflatval-cognite, line 59: I don't think the line `@module @cognite/reveal`is necessary when we don't expose this type 🤔

viewer/packages/360-images/src/icons/clustering/htmlClusterStyles.ts
```diff
@@ -21,15 +21,15 @@ export function generateClusterStyles(classPrefix: string): string {
...
   border: var(--outer-border) solid #FFFFFF;
   border-radius: 50%;
-  filter: drop-shadow(2px 2px 8px rgba(0, 0, 0, 0.15));
   background: transparent;
   display: flex;
   flex-direction: row;
   justify-content: center;
   align-items: center;
-  transition: transform 0.15s ease-out;
+  transition: filter 0.2s ease-out, transform 0.15s ease-out;
```
haakonflatval-cognite: Should we have `ease-in` here too? I see that overlapped clusters vanish instantly, but fade in when they re-appear, perhaps because that keyword is missing 🤔 

Or, that behavior may be because of what's mentioned in this comment: https://github.com/cognitedata/reveal/pull/5626/changes#r3302975116

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
```diff
@@ -0,0 +1,92 @@
...
+      this._collections.splice(idx, 1);
+    }
+  }
+
+  public dispose(): void {
+    this._onBeforeSceneRenderedEvent.unsubscribe(this._coordinatorHandler);
+    this._collections.length = 0;
+  }
+
+  private _runCoordinator(): void {
```
haakonflatval-cognite: We don't use underscores at the start of private function names in classes in Reveal, I believe (only for private fiels). Same for `computeGlobalOcclusion` below

viewer/packages/360-images/src/icons/IconCollection.ts
```diff
@@ -542,7 +550,6 @@ export class IconCollection {
     const clusteredIcons: ClusteredIcon[] = [];
```
haakonflatval-cognite, line 515: It seems to me that all changes in this file below this point is non-functional, and I'm also not sure if they are for the better (removing comments and renaming some variables).. Is there a specific reason for these changes? 😅

viewer/packages/360-images/src/icons/clustering/HtmlClusterRenderer.ts
```diff
@@ -245,26 +258,57 @@ export class HtmlClusterRenderer {
...
-    const distance = camera.position.distanceTo(this._tempPosition);
-    const projectedSize = Math.max(this._minSize, Math.min(this._maxSize, this._baseSize / Math.max(distance, 1)));
+    return result;
+  }
 
+  private computeFadeOpacity(distance: number): number {
+    if (distance <= this._clusterFadeStartDistance) return 1;
+    if (distance >= this._clusterFadeEndDistance) return 0;
+    const range = this._clusterFadeEndDistance - this._clusterFadeStartDistance;
+    return 1 - (distance - this._clusterFadeStartDistance) / range;
```
haakonflatval-cognite, line 303: Is this math correct? Maybe I'm missing something, but shouldn't opacity be 1 when the distance is sufficiently high (i.e. the clusters are far apart), and 0 when it's sufficiently low (i.e. when the clusters are close)? This code seems to do the opposite 🧐

viewer/packages/360-images/src/icons/clustering/HtmlClusterRenderer.ts
```diff
@@ -245,26 +258,57 @@ export class HtmlClusterRenderer {
...
+      const screenPos = worldToViewportCoordinates(canvas, camera, tempWorldPos, tempProjected).clone();
+
+      result.push({
+        data: clusterData,
+        screenPos,
+        worldPos: tempWorldPos.clone(),
+        distance,
+        projectedSize
+      });
+    }
```
haakonflatval-cognite, line 294: Personally, I would have thought this to be a bit more clearer with `.filter`+`.map`, but not critical

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
```diff
@@ -0,0 +1,92 @@
...
+    this._onBeforeSceneRenderedEvent = onBeforeSceneRendered;
+    this._coordinatorHandler = () => this._runCoordinator();
+    this._onBeforeSceneRenderedEvent.subscribe(this._coordinatorHandler);
+  }
+
+  public onCollectionAdded(collection: HtmlClusterCollection): void {
+    this._collections.push(collection);
+    // Re-subscribe last so this handler always runs after all per-collection handlers
+    this._onBeforeSceneRenderedEvent.unsubscribe(this._coordinatorHandler);
+    this._onBeforeSceneRenderedEvent.subscribe(this._coordinatorHandler);
```
haakonflatval-cognite: Not sure I understand this - why do we need to re-subscribe the `coordinatorHandler`? The event trigger isn't touched at any other point in this function?

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
```diff
@@ -0,0 +1,92 @@
...
+    const occluded = new Set<Overlay3DIcon>();
+    const sorted = [...screenInfos].sort((a, b) => a.distance - b.distance);
+
+    for (let i = 1; i < sorted.length; i++) {
+      const far = sorted[i];
+      for (let j = 0; j < i; j++) {
+        const close = sorted[j];
+        const dx = far.screenPos.x - close.screenPos.x;
+        const dy = far.screenPos.y - close.screenPos.y;
+        const screenDist = Math.sqrt(dx * dx + dy * dy);
```
haakonflatval-cognite: [Optional]: We could write `const screenDist = far.screenPos.distanceTo(close.screenPos);` to shorten this a bit, but not a big deal 😅

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  private _computeGlobalOcclusion(screenInfos: ClusterScreenInfo[]): Set<Overlay3DIcon> {
+    const occluded = new Set<Overlay3DIcon>();
+    const sorted = [...screenInfos].sort((a, b) => a.distance - b.distance);
+
+    for (let i = 1; i < sorted.length; i++) {
+      const far = sorted[i];
+      for (let j = 0; j < i; j++) {
+        const close = sorted[j];
```
haakonflatval-cognite, line 48: I'm not sure how impactful it is, or if it will be worth it, but I'm concerned that this double for-loop can be a bottle-neck. One idea to optimize it, is to sort all clusters by their x-position, which means we can do a `continue;` in the outer loop once we find a cluster that's outside the wanted radius along the x-axis. 

Again, it complicate things, and not sure if it's really a problem that needs solving 🤔

viewer/packages/api/src/api-helpers/Image360ApiHelper.ts
```diff
@@ -179,6 +181,9 @@ export class Image360ApiHelper<DataSourceT extends DataSourceType> {
 
     this._activeCameraManager = activeCameraManager;
     this._onBeforeSceneRenderedEvent = onBeforeSceneRendered;
+    if (iconsOptions?.enableHtmlClusters) {
+      this._htmlClusterCoordinator = new HtmlClusterCoordinator(onBeforeSceneRendered);
+    }
```
haakonflatval-cognite, line 190: I'm not that fond of `HtmlClusterCoordinator` mutating the `onBeforeSceneRendered` event trigger (re-subscribing) upon getting new collections. But I get that the `HtmlClusterCoordinator`'s handler should be run after all the collections'. In my head, it may make more sense to handle all that in this class. For instance, add all collection handlers into a separate `EventTrigger`, and then fire that event trigger upon getting an event on the `onBeforeSceneRendered` event, and, inside the same callback, call `HtmlClusterCoordinator`s handler. 

It does complicate things a little, so I'm not going to mandate it - but do you think it makes sense?

viewer/packages/api/src/api-helpers/Image360ApiHelper.ts
```diff
@@ -215,6 +220,7 @@ export class Image360ApiHelper<DataSourceT extends DataSourceType> {
       preMultipliedRotation
     );
 
+    this._htmlClusterCoordinator?.onCollectionAdded(imageCollection as DefaultImage360Collection<DataSourceT>);
```
haakonflatval-cognite: I believe the cast here is unnecessary

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
```diff
@@ -0,0 +1,92 @@
...
+        allScreenInfos.push(info);
+      }
+    }
+
+    const occluded = this._computeGlobalOcclusion(allScreenInfos);
+
+    for (const collection of this._collections) {
+      collection.applyHtmlClusterOcclusion(occluded);
+    }
+  }
```
haakonflatval-cognite: It's a bit unfortunate that we rely on mutable data from the collections here (`getStagedHtmlClusterScreenInfos`), which is updated from a different place in the code. I'm not sure how to solve this quickly, though. It sounds like we may want to shift all cluster data for all collections in a single place/class, and do the occlusion culling and rendering directly from there. 

But it's a bigger effort - so I don't think it's worth doing here. Maybe good to keep in mind for later, though

viewer/packages/360-images/src/icons/clustering/ClusterRenderingStrategy.ts
haakonflatval-cognite, line 11: I'm a bit sad that this isn't a (discriminated) union type 🥲 But it's far outside the scope of the PR, so I'll keep my woes for myself for now

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      .callback(({ args }) => {
+        capturedCallbacks.push(args[0]);
+      })
+      .setup(e => e.unsubscribe(It.IsAny()))
+      .callback(({ args }) => {
+        const cb = args[0];
+        const idx = capturedCallbacks.indexOf(cb);
+        if (idx !== -1) capturedCallbacks.splice(idx, 1);
+      })
+      .object();
```
haakonflatval-cognite: Is it possible to just use a normal `EventTrigger` object instead of creating a mock? It means you may have to test that certain functions have been called/not called some number of times instead of checking the length of `capturedCallbacks`, but I think that would be more natural, as it tests the behavior of the class, rather than the implementation

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.test.ts
```diff
@@ -0,0 +1,169 @@
...
+});
+
+function createMockIcon(): Overlay3DIcon {
+  return new Mock<Overlay3DIcon>()
+    .setup(i => i.getPosition())
+    .returns(new Vector3())
+    .setup(i => i.getColor())
+    .returns(new Color(1, 1, 1))
+    .object();
+}
```
haakonflatval-cognite, line 88: Could we just use the constructor of `Overlay3DIcon` here instead of using a mock?

viewer/packages/360-images/src/collection/DefaultImage360Collection.ts
```diff
@@ -343,6 +344,14 @@ export class DefaultImage360Collection<T extends DataSourceType> implements Imag
     this._icons.setHoveredClusterIcon(icon);
   }
 
+  public getStagedHtmlClusterScreenInfos(): ClusterScreenInfo[] {
```
haakonflatval-cognite, line 351: Since we're implementing these, should we let `DefaultImage360Collections` implement the `HtmlClusterCollection` interface?

haakonflatval-cognite (commented):
Nice, this is a great improvement on the cluttering from 360-icons! 
 
I had a few comments, but many are optional, and some I don't think will be feasible

viewer/packages/360-images/src/icons/clustering/ClusterRenderingStrategy.ts
danpriori: What do you mean a union type here? Can you be more specific 😅 ?

haakonflatval-cognite: Ah, yes, sorry :sweat_smile: It seems like this type could either be an icon or a cluster, and some of the fields are only relevant if it's a cluster (`clusterIcons`, at least).. So, I was wondering if this should be a union like 
```
ClusteredIconData = 
    | { type: 'icon', icon: Overlay3DIcon, size: number, position: Vector3; sizeScale: number } 
    | { type: 'cluster': icon: Overlay3DIcon, size: number, position: Vector3; sizeScale: number; clusterIcons: Overlay3DIcon[];
};
```
The common fields could probably be put in a separate type too
But anyway, let's not do it in this PR :see_no_evil:

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
danpriori: Actually, the IconCollection, when constructed it subscribes to the same EventTrigger that the coordinator uses.

- Coordinator constructor -->  coordinator subscribes (position 1)
- IconCollection constructor --> _computeClustersEventHandler subscribes (position 2+)
- onCollectionAdded is called --> coordinator re-subscribes (moves to last position).

Without the re-subscription, the coordinator would run before the collections, reading getStagedHtmlClusterScreenInfos() with data from the previous frame, I think. The re-subscription is necessary to avoid that. 

I can rewrite the Coordinator + the HtmlClusterCollection interface  with a method to prepare the screen infos and some other changes in the IconCollection to accomodate it, and then remove the onCollectionAdded method, probably.
Wdyt?

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
danpriori: Yes. But there isn't a very easy solution, perhaps, because it uses data that is affected via events asynchronously.
The coordinator's focus is solely on controlling the detection and coordination of occlusions for the icon collections, which cannot be done acceptably if we do it within each icon collection, since it needs to be done cross-collection as well and not in isolation.

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
danpriori: Ok.  Actually I think this idea will conflicts with the distance sorting. 

If you replace distance sorting with x sorting, we will lose the guarantee that j is always closer to the camera than i. To work correctly, it would also need to check `close.distance < far.distance` in each pair, which adds complexity without necessarily gaining speed 😢 . Unless we have other options on the table that I'm missing here.

In general for clusters, the expected n is relatively small, the x-sweep-line optimization conflicts with the distance sorting that is necessary for correctness. If the number of variables (n) grows in the future, or if we see a significant bottleneck in this area in projects much larger than those tested, such as with Hibernia, then probably one approach would be a 2D spatial index with grids, not just sorting by x.

viewer/packages/360-images/src/icons/clustering/HtmlClusterRenderer.ts
danpriori: I think the math is correct. Maybe you are confusing what `distance ` means here.

`distance` is `camera.position.distanceTo(tempWorldPos)`. The distance from the camera to the cluster in 3D space, not the distance on the screen between two overlapping clusters.

Otherwise the cluster fade effect behavior would do the opposite on the screen 🤔

danpriori: The fade effect only applies to clusters that are being hidden by others closer to the camera.

viewer/packages/360-images/src/icons/clustering/HtmlClusterRenderer.ts
danpriori: Where exactly? You mean replacing the for loop?
Personally, I don't like to put filter + map everywhere as it is less performant, and I don't see too much readability benefits here. 
If you don't object, I prefer to keep it this way 😅

viewer/packages/360-images/src/icons/IconCollection.ts
danpriori: It was just some clean up and cosmetic changes that I think now it looks like over explaining now 😅 . Do you prefer that I keep the old version?

viewer/packages/api/src/api-helpers/Image360ApiHelper.ts
danpriori: Hmmmm. 

I'm not fully sure. What the Image360ApiHelper actually does: managing collection lifecycle, user interactions (hover, click, enter/exit), camera transitions, history, and intersection testing — it's fundamentally an API/interaction helper, not a render orchestrator. The one render subscription it already has (updateHoverStateOnRenderHandler) is justified because it directly serves user interaction (updating hover state from mouse position), which is clearly in scope for this class.

Adding render chain ordering logic here would be a responsibility mismatch I think. Any developer opening Image360ApiHelper looking for interaction or API logic would find render pipeline plumbing with no obvious reason for it to be there, I think. Or did I understood it wrong? 🤔 

However, I think the underlying concern is valid. Actually the re-subscribe trick in onCollectionAdded is implicit and, at some degree, a little fragile, I would say (e.g. a setCullingScheme call after the fact can silently break the ordering guarantee). 

I think a cleaner fix that keeps the logic where it belongs would be to add a prepareHtmlClusterScreenInfos(params) method to the HtmlClusterCollection interface, and have the coordinator call prepare on all collections at the start of its own handler before reading the staged data. 

That way the ordering is self-contained within HtmlClusterCoordinator and doesn't rely on subscriber list position at all, maybe.

Or keep this solution as for now and think more carefully some refactoring taking this new approach into account later. Wdyt?

viewer/packages/api/src/public/migration/types.ts
danpriori: I honestly don't think it's that much, as long as we keep the default values. The idea here is to give the opportunity to expose this in the future, even to Scene Builder for example, where the user could specific visual behavior for the clusters, and I don't foresee much difficulty in maintaining and expanding this if necessary.
Anyhow, I added the defaults in the next commit:

haakonflatval-cognite: Yes, I see :thinking: I'm still leaning towards not exposing this until it's necessary (if it will become necessary). But anyway, not blocking the PR on this

viewer/packages/360-images/src/icons/IconCollection.ts
haakonflatval-cognite: No, it's fine - just wanted to make sure I wasn't missing the point :+1:

viewer/packages/360-images/src/icons/clustering/HtmlClusterRenderer.ts
haakonflatval-cognite: Aaah, I see now :sweat_smile: Yes, you're right, I misunderstood what `distance` was in this case :+1:

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
haakonflatval-cognite: I actually meant to delete this comment in favor of the one I made here: https://github.com/cognitedata/reveal/pull/5626#discussion_r3303759288 :sweat_smile: I'm not sure I understand the method you describe here :thinking: But I think at least it makes sense to keep the icon collection update handlers separate from the cluster coordinator handler somehow

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
haakonflatval-cognite: Yes, it's fine, it's fairly simple to optimize in the future if needed :+1: 

I see the point that you miss the nice ordering-property when we don't sort by the `distance`-field, but I think the performance gained is negligible compared to just checking `distance`-fields of each pair, and sort them by x-value instead. Anyway, indexing with grids could prove to be a better approach in the future anyway, as you suggest :+1:

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
haakonflatval-cognite: (I think it makes more sense to discuss the other comment further down, I'll resolve this)

viewer/packages/api/src/api-helpers/Image360ApiHelper.ts
haakonflatval-cognite: The `Image360ApiHelper` has always been a bit of a mess, to be honest :sweat_smile: It tries to do way too much. 

I think the setup here in the API helper is a little strange to begin with.. It sounds like it would be easier to just expose a method ("update" or "prepareForRender") that would loop through the collections and updated them, and then call the coordinator to coordinate - instead of passing an `EventTrigger` object that may contain callbacks from other parts of the system.

Personally, I think a relatively simple solution would be to implement a method like the one I described above, and just subscribe that function to the `onBeforeSceneRendered` `EventTrigger` (so not exposing it). 

I'm not convinced it's a good idea to leave the responsibility of making the update-calls of the collections to the `HtmlClusterCoordinator`, but again, I'm not fully sure

viewer/packages/api/src/public/migration/types.ts
danpriori: I still think it makes perfect sense to expose them now because we've already planned to test them in client environments that we don't yet have access to, and the PM will likely receive feedback on adjustments, as they already said. This reduces the friction of bumping the Reveal for small feature configuration adjustments that should be adjusted via the app side and not via the library anyway.

viewer/packages/360-images/src/icons/clustering/htmlClusterStyles.ts
danpriori: Yes. It could be ease-in, but what you probably saw was when the cluster is too far from the camera, and in that case, it receives the "none" display, and there's no way to create a transition effect to display none. 
(see the method `HtmlClusterRenderer.applyClusterElementUpdate()` and with the fadeOpacity flag).

It's not good to create effects on clusters that will become completely transparent because they will continue to be allocated in the rendering unnecessarily.

I even tried adding an event listener with a `transitionend `event, and after the transition finishes, displaying "none," but since it's asynchronous and depends on very fast camera movement, the cluster ends up with some flickering effects. I think it's better to leave it this way, and then I'll think more calmly about a more robust way to create a transition to display "none" without this problem.

https://github.com/cognitedata/reveal/pull/5626/commits/273dac45f06ee713b71cb742a2c77a99ee9aabe7

viewer/packages/api/src/api-helpers/Image360ApiHelper.ts
danpriori: https://github.com/cognitedata/reveal/pull/5626/commits/273dac45f06ee713b71cb742a2c77a99ee9aabe7

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
danpriori: ops! Why I forgot that ?😅 🤦‍♂️ 
https://github.com/cognitedata/reveal/pull/5626/commits/273dac45f06ee713b71cb742a2c77a99ee9aabe7

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
danpriori: ops. Old habits die hard 😅 
https://github.com/cognitedata/reveal/pull/5626/commits/273dac45f06ee713b71cb742a2c77a99ee9aabe7

viewer/packages/360-images/src/types.ts
danpriori: https://github.com/cognitedata/reveal/pull/5626/commits/273dac45f06ee713b71cb742a2c77a99ee9aabe7

viewer/packages/api/src/api-helpers/Image360ApiHelper.ts
danpriori: I see. Ok. I made some changes moving the beforeRender subscription out of the Coordinator and making a prepareHtmlCluster method and subscribing it from the api helper, based on my pov of this change. Check if does make more sense now:
https://github.com/cognitedata/reveal/pull/5626/commits/273dac45f06ee713b71cb742a2c77a99ee9aabe7

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.test.ts
danpriori: https://github.com/cognitedata/reveal/pull/5626/commits/273dac45f06ee713b71cb742a2c77a99ee9aabe7

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.test.ts
danpriori: https://github.com/cognitedata/reveal/pull/5626/commits/273dac45f06ee713b71cb742a2c77a99ee9aabe7

viewer/packages/360-images/src/collection/DefaultImage360Collection.ts
danpriori: oh yeah sure
https://github.com/cognitedata/reveal/pull/5626/commits/a299dcba80e2e70360aefce448cf3c679591c507

viewer/packages/360-images/src/icons/clustering/htmlClusterStyles.ts
haakonflatval-cognite: Ah, okay, I see - yes, then I understand that this becomes a bit too complicated to solve properly :sweat_smile:

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
```diff
@@ -0,0 +1,77 @@
...
+  applyHtmlClusterOcclusion(occludedIcons: Set<Overlay3DIcon>): void;
+}
+
+/**
+ * Coordinates HTML cluster occlusion detection across multiple Image360Collections.
+ * Gathers per-frame screen-space data from all collections and computes a single
+ * global occlusion set, then feeds it back to each collection for DOM updates.
+ */
+export class HtmlClusterCoordinator {
+  private readonly _collections: HtmlClusterCollection[] = [];
```
haakonflatval-cognite: [Optional]: We could consider not storing `_collections` here at all, and just give the list of collections in the call to `runCoordinator`. Then we don't need to maintain this duplicate list

viewer/packages/360-images/src/icons/IconCollection.ts
```diff
@@ -125,31 +124,39 @@ export class IconCollection {
     if (this._iconCullingScheme === scheme) return;
 
     this._iconCullingScheme = scheme;
-    this._onBeforeSceneRenderedEvent.unsubscribe(this._activeCullingSchemeEventHandeler);
+
+    if (!this._enableHtmlClusters) {
+      this._onBeforeSceneRenderedEvent.unsubscribe(this._activeCullingSchemeEventHandeler);
+    }
```
haakonflatval-cognite: This function is getting a bit more complex than I'd like, where we have so much conditional logic based on `this._enableHtmlClusters` 🤔 Could we try to just create e.g. an "updateIcons"-function that is triggered  (through the `prepareHtmlClusters` call) from `Image360ApiHelper`, instead of passing the `onBeforeSceneRenderedEvent`-`EventTrigger` into this class? I think it may simplify this quite a bit?

haakonflatval-cognite (commented):
Two more questions, but I think it's good otherwise 👍

viewer/packages/360-images/src/icons/IconCollection.ts
danpriori: Hmm yeah. I've moved the responsibility for managing subscriptions to the API helper and use the updateIcons. I dont like to put this part into the Api Helper but I understand your point I think. 
However, to do that, and to have clean and "centered" code, I had to make changes in other places as well. I also reduced other code involving the HTML cluster flag, which may have made it simpler as well.
https://github.com/cognitedata/reveal/pull/5626/commits/a824fb56afd5ff6716109e7d0fbd3221adf8947f

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
danpriori: Nice catch. 
https://github.com/cognitedata/reveal/pull/5626/commits/12fdbfde34d0529217e68736f7072c9116fbf2e0

haakonflatval-cognite (dismissed):
Nice job! I'm sorry for going a bit overboard with the comments - but I'm happy with the result! LGTM 👍

haakonflatval-cognite (dismissed):
LGTM 👍

gregfagan:
@danpriori can you resolve the conflicts?

haakonflatval-cognite (dismissed):
LGTM 👍

viewer/packages/360-images/src/icons/clustering/htmlClusterStyles.ts
```diff
@@ -61,9 +64,6 @@ export function generateClusterStyles(classPrefix: string): string {
   border-color: #8893CD;
 }
 
-.${classPrefix}-icon.hovered {
-  transform: translate(-50%, -50%) scale(2.1);
-}
```
gregfagan, line 66: was this style removal intentional? it is not clear to me that it's behavior is either now unwanted or implemented elsewhere.

viewer/packages/360-images/src/icons/clustering/HtmlClusterRenderer.ts
```diff
@@ -243,29 +262,61 @@ export class HtmlClusterRenderer {
...
+    const tempProjected = new Vector3();
 
-    const screenPos = worldToViewportCoordinates(canvas, camera, this._tempPosition, this._tempProjectedPosition);
+    for (const clusterData of visibleClusters) {
+      if (!clusterData.isCluster) continue;
+
+      tempWorldPos.copy(clusterData.clusterPosition).applyMatrix4(modelTransform);
+      const distance = camera.position.distanceTo(tempWorldPos);
+      const projectedSize = Math.max(this._minSize, Math.min(this._maxSize, this._baseSize / Math.max(distance, 1)));
+      const screenPos = worldToViewportCoordinates(canvas, camera, tempWorldPos, tempProjected).clone();
```
gregfagan, line 285: How far do y'all (3D team) go to avoid allocating frequently?

`.clone`s here and on L290, and even the temp vec3s above; is this enough of a hot path? could these allocations be moved out?

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
```diff
@@ -0,0 +1,61 @@
...
+  applyHtmlClusterOcclusion(occludedIcons: Set<Overlay3DIcon>): void;
+}
+
+/**
+ * Coordinates HTML cluster occlusion detection across multiple Image360Collections.
+ * Gathers per-frame screen-space data from all collections and computes a single
+ * global occlusion set, then feeds it back to each collection for DOM updates.
+ */
+export class HtmlClusterCoordinator {
+  // Screen-space overlap radius multiplier — matches HtmlClusterRenderer._occlusionFactor
```
gregfagan: comment looks to be out of date

gregfagan:
The style removal is the only one that requires a response, but take a look at the other comments and let me know what you think.

viewer/packages/360-images/src/icons/clustering/htmlClusterStyles.ts
danpriori: Yes, it was intentional. After making style changes to occlusion and opacity via filters and other style adjustments, hovering the mouse over the cluster caused it to increase in size considerably, which was not the intention. So we needed to remove it.

viewer/packages/360-images/src/icons/clustering/HtmlClusterCoordinator.ts
danpriori: oh yeah. Nice catch. outdated comment.
https://github.com/cognitedata/reveal/commit/b6d436207f24071726b95fab8037730151ab03ca

viewer/packages/360-images/src/icons/clustering/HtmlClusterRenderer.ts
danpriori: If I understood the question correctly, it's not a hot path in the sense that it justifies changes, I think, because the bottleneck here is the DOM, not the GC. The tempVecs outside the loop are already in the correct pattern, reused by iteration. And the `.clone()` inside the loop are necessary for correction/mutability.

haakonflatval-cognite (approved):
LGTM 👍

gregfagan:
🦄

PR #5647, opened 2026-06-02
feat: added Image360FaceTextureLoader class for texture loading in 360 visual box

pramod-cog:
#### Type of change

Feat  

#### Jira ticket :blue_book:

Part of https://cognitedata.atlassian.net/browse/BND3D-6886

## Description :pencil:
Added Image360FaceTextureLoader class which will load and apply texture to 360 Image Visualization Box for Jpeg Progressive images

Unit test


viewer/packages/data-providers/src/types.ts
```diff
@@ -162,6 +162,8 @@ export type Image360Face = {
   face: 'front' | 'back' | 'left' | 'right' | 'top' | 'bottom';
   mimeType: 'image/jpeg' | 'image/png';
   data: ArrayBuffer;
+  /** Signed download URL for streaming. When present, data is empty. */
+  downloadUrl?: string;
 };
```
haakonflatval-cognite, line 169: Do you think it's worth using a discriminated union (e.g. `urlOrData: { type: 'url', downloadUrl: string } | { type: 'data', data: ArrayBuffer }`) for these fields?

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+    face: Image360FaceWithUrl,
+    onFirstFaceReady?: () => void,
+    onJpegTypeDetected?: (type: JpegType) => void,
+    abortSignal?: AbortSignal
+  ): Promise<Image360Texture> {
+    return this.streamFaceTexture(face, onFirstFaceReady, onJpegTypeDetected, abortSignal);
+  }
+
+  async loadFromBuffer(face: Image360Face): Promise<Image360Texture> {
+    const blob = new Blob([face.data], { type: face.mimeType });
```
haakonflatval-cognite: Just so I understand how this will be used - will the user create the `Image360Face` object, and then figure out whether it contains a URL or not (e.g. by using the `hasDownloadUrl` typeguard), to determine whether they should call `loadFromUrl` or `loadFromBuffer`? Would it be possible to keep more of that responsibility inside this class? 🤔 In the end, it would be nice if we only had one `load`-function we needed to care about 🤞

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+ * - URL-based streaming (progressive and baseline JPEG detection)
+ * - In-memory buffer decoding
+ */
+export class Image360FaceTextureLoader {
+  private readonly MAX_MOBILE_IMAGE_SIZE = 1024;
+
+  constructor(
+    private readonly _device: DeviceDescriptor,
+    private readonly _textureLoader: TextureLoader,
+    private readonly _requestRedraw: () => void,
```
haakonflatval-cognite: Would it make sense to call this "onUpdate" instead? From this class' perspective, the update doesn't have to trigger a redraw

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+  reject: (reason: unknown) => void;
+};
+
+/**
+ * Handles all texture loading for 360 image cube faces:
+ * - URL-based streaming (progressive and baseline JPEG detection)
+ * - In-memory buffer decoding
+ */
+export class Image360FaceTextureLoader {
+  private readonly MAX_MOBILE_IMAGE_SIZE = 1024;
```
haakonflatval-cognite, line 40: Is this an empirically chosen value, or taken from somewhere else? Should we have a comment on it?

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+      state.texture.needsUpdate = true;
+    }
+    bitmap.close();
+    this._requestRedraw();
+  }
+
+  private async getScaledImageTexture(
+    texture: Texture<HTMLImageElement | HTMLCanvasElement>,
+    imageSize: number
+  ): Promise<Texture<HTMLCanvasElement>> {
```
haakonflatval-cognite: [Optional]: This could be placed outside the class

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+    });
+    if (!response.ok || !response.body) {
+      throw new Error(`Failed to fetch 360 image for face: ${face.face} (status ${response.status})`);
+    }
+    return { body: response.body, contentLength: response.headers.get('content-length') };
+  }
+
+  private createCanvasTextureState(): CanvasTextureState {
+    const canvas = document.createElement('canvas');
+    return { canvas, ctx: canvas.getContext('2d')!, texture: undefined };
```
haakonflatval-cognite: It would be nice if we could crash and burn if `getContext('2d')` fails

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+    let resolve: (value: T) => void = () => {};
+    let reject: (reason: unknown) => void = () => {};
+    const promise = new Promise<T>((res, rej) => {
+      resolve = res;
+      reject = rej;
+    });
+    return { promise, resolve, reject };
+  }
+
+  private async readStream(
```
haakonflatval-cognite: [Optional]: This method and the three methods above could be defined outside the class

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+    body: ReadableStream<Uint8Array>,
+    contentLength: string | null,
+    onJpegTypeDetected: ((type: JpegType) => void) | undefined,
+    onProgressiveScan: (partialBlob: Blob) => Promise<void>,
+    onError: (e: unknown) => void
+  ): Promise<StreamResult> {
+    const size = parseInt(contentLength ?? '0', 10);
+    let fullBuffer: Uint8Array<ArrayBuffer> = new Uint8Array(size > 0 ? size : 2 * 1024 * 1024);
```
haakonflatval-cognite: [Optional]: Should we use `isNaN` to ensure that we were able to parse a valid number?

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+    let bytesReceived = 0;
+    let jpegType: JpegType | 'unknown' = 'unknown';
+    let lastCutpoint = 0;
+
+    const reader = body.getReader();
+    try {
+      while (true) {
+        const { done, value } = await reader.read();
+        if (done) break;
```
haakonflatval-cognite: Could it happen that the last part of the image will not be added, since we return immediately on `done` without putting the `value` in the `fullBuffer`?

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+      state.texture.center.set(0.5, 0.5);
+      state.texture.repeat.set(-1, 1);
+      this._updateFaceTexture(face.face, state.texture);
+      onFirstFaceReady?.();
+      resolveTexture({ face: face.face, texture: state.texture });
+    } else {
+      state.ctx.drawImage(bitmap, 0, 0);
+      state.texture.needsUpdate = true;
+    }
+    bitmap.close();
```
haakonflatval-cognite, line 150: Just so I understand - what does `bitmap.close()` actually do here? 😅

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
```diff
@@ -0,0 +1,268 @@
...
+      async (partialBlob: Blob) => {
+        const bitmap = await createImageBitmap(partialBlob).catch(() => null);
+        if (bitmap)
+          this.applyBitmapToTexture(bitmap, face, textureState, onFirstFaceReady, deferred.resolve, deferred.reject);
+      },
+      deferred.reject
+    );
+
+    await this.finalizeTexture(jpegType, fullBuffer, bytesReceived, face, textureState, onFirstFaceReady, deferred);
+    return deferred.promise;
```
haakonflatval-cognite: I'm not sure whether this `DeferredPromise` concept gives us any benefits here 🤔 I may be wrong, but the way I read this, the user will not get access to the 360 texture at all until all the stages (all the `await`s) in this method are finished. I'll have to test this a bit to be sure

haakonflatval-cognite (commented):
Nice! I like this format a lot better - I made some comments, but not all are that important

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
pramod-cog: `bitmap.close()` explicitly tells the browser to free that pixel memory immediately. Without it, the decoded pixel data stays in memory until the browser decides to grabage collect the `ImageBitmap` object, which can be delayed significantly

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
pramod-cog: This was in `Image360VisualisationBox`

viewer/packages/data-providers/src/types.ts
pramod-cog: This is used in multiple files, providers, util function and test, so it would be more work to accommodate in this PR

viewer/packages/360-images/src/entity/Image360FaceTextureLoader.ts
pramod-cog: No, Stream guarantees that when `done` is true, `value` is always `undefined` for a byte stream. `if (done)` break before processing `value` is correct, there is nothing in `value` to process when the stream is done

haakonflatval-cognite (approved):
This works well on Chrome, but less so on Firefox. However, I'm not sure how to fix it, so I'll approve this PR 👍 

On Firefox, it seems that the images are downloaded fully in sequence, so that it takes a long time before we have enough data to fill all the texture faces, and the camera can move. I don't think this is a regression from earlier behavior, so I suppose it's fine. My current hunch is that Firefox is dealing badly with parallel streaming requests, and cannot provide the streams to the web app properly in parallel like Chrome does.

pramod-cog:
risk review ok

PR #5661, opened 2026-06-09
chore(viewer): remove base catch-all path

christjt:
#### Type of change

Chore 

## Description :pencil:

Clean up path usages, they are strictly not necessary and are only used for d.ts. emit aliasing. So moved to the build step.


pramod-cog (approved):
risk review ok

PR #5664, opened 2026-06-10
chore(viewer): detach CadMaterialManager dependency from CadNode

christjt:
#### Type of change

Chore 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-6524

## Description :pencil:

This PR decouples CadNode and CadMeshManager from CadMaterialManager by moving per-model state directly into CadMaterial.

Changes

ClippingPlanesProvider (new)
- New class in cad-styling that holds a model's clipping planes and fires a changed event when they are updated.
- Replaces the perModelClippingPlanes field and setModelClippingPlanes method on CadMaterialManager.

CadMaterialManager
- CadMaterial now carries a clippingPlanesProvider instead of per-model plane arrays stored in the manager.
- Subscribes to clippingPlanesProvider.changed reactively to update shader uniforms and flag a redraw, instead of requiring callers to go through the manager.
- setModelClippingPlanes removed — callers set cadMaterial.clippingPlanesProvider.clippingPlanes directly.

CadNode
- No longer takes CadMaterialManager as a constructor argument.
- clippingPlanes getter/setter now delegate to cadMaterial.clippingPlanesProvider.
- dispose() now disposes materials, texture builders, and the clipping planes provider directly rather than delegating to the manager.

CadMeshManager
- Takes CadMaterial directly instead of CadMaterialManager + model symbol.
- Textured mesh material creation (addTexturedMeshMaterial) moved inline.

CadManager
- Calls removeModelMaterials on the material manager when a model is removed (fixes a disposal leak).

### DIFF
--- a/viewer/packages/cad-styling/src/ClippingPlanesProvider.ts
+++ b/viewer/packages/cad-styling/src/ClippingPlanesProvider.ts
@@ -14,11 +14,11 @@ export class ClippingPlanesProvider {
     this._changed = new EventTrigger<(clippingPlanes: Plane[]) => void>();
   }
 
-  get clippingPlanes(): Plane[] {
+  getClippingPlanes(): Plane[] {
     return this._clippingPlanes;
   }
 
-  set clippingPlanes(clippingPlanes: Plane[]) {
+  setClippingPlanes(clippingPlanes: Plane[]): void {
     this._clippingPlanes = clippingPlanes;
     this._changed.fire(this._clippingPlanes);
   }
--- a/viewer/packages/cad-styling/src/ClippingPlanesProvider.ts
+++ b/viewer/packages/cad-styling/src/ClippingPlanesProvider.ts
@@ -1,0 +1,37 @@
+/*!
+ * Copyright 2026 Cognite AS
+ */
+
+import { EventTrigger } from '@reveal/utilities';
+import type { Plane } from 'three';
+
+export class ClippingPlanesProvider {
+  private _clippingPlanes: Plane[];
+  private readonly _changed: EventTrigger<(clippingPlanes: Plane[]) => void>;
+
+  constructor() {
+    this._clippingPlanes = [];
+    this._changed = new EventTrigger<(clippingPlanes: Plane[]) => void>();
+  }
+
+  get clippingPlanes(): Plane[] {
+    return this._clippingPlanes;
+  }
+
+  set clippingPlanes(clippingPlanes: Plane[]) {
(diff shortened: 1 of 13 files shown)


viewer/packages/cad-styling/src/ClippingPlanesProvider.ts
```diff
@@ -0,0 +1,37 @@
...
+  private _clippingPlanes: Plane[];
+  private readonly _changed: EventTrigger<(clippingPlanes: Plane[]) => void>;
+
+  constructor() {
+    this._clippingPlanes = [];
+    this._changed = new EventTrigger<(clippingPlanes: Plane[]) => void>();
+  }
+
+  get clippingPlanes(): Plane[] {
+    return this._clippingPlanes;
```
haakonflatval-cognite, line 18: [Optional]: Is there any way we can avoid mutation of the `clippingPlanes` object when returning here? (Not sure if it's worth it.. Even if we construct a new array, the elements are still mutable)

viewer/packages/cad-styling/src/ClippingPlanesProvider.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  set clippingPlanes(clippingPlanes: Plane[]) {
```
haakonflatval-cognite: I personally prefer the traditional form of getters and setters (e.g. `setClippingPlanes`), as to not "hide" the side effect happening behind the scenes. But not blocking the PR for this

haakonflatval-cognite (commented):
Looks good - will have to test a little to be sure, had some comments you can look at in the meantime

viewer/packages/cad-styling/src/ClippingPlanesProvider.ts
christjt: I am worried if this is a breaking change or not. So I would prefer not to handle this change in semantics in this PR. In the previous implementation you then also got a reference to the array. Though in principle I very much agree. Returning a copy would be better.

viewer/packages/cad-styling/src/ClippingPlanesProvider.ts
christjt: Agreed, fixed here: https://github.com/cognitedata/reveal/pull/5664/commits/9c0711f38dbce87c47c9b4778a9eb3db3cbb38b4

haakonflatval-cognite (approved):
LGTM 👍

christjt:
risk review ok

PR #5685, opened 2026-06-18
fix: point-cloud & cad model stopped loading when switching multiple models

pramod-cog:
#### Type of change

Bug 

#### Jira ticket :blue_book:

https://cognitedata.atlassian.net/browse/BND3D-7215

## Description :pencil:
Point cloud data and Cad data stopped loading when switching between models (removing old models and adding new ones). Only 360 image collections were unaffected because they don't use the worker pool.

The Vite migration replaced the custom `setupTransferableMethodsOnMain` worker RPC with `Comlink`. `Comlink.wrap()` installs a permanent message event listener on the worker. Since workers are pooled and reused, each reuse added another listener to the same worker. After some reuses, a single worker had `N` accumulated Comlink listeners.
When the worker asked a response, all `N` listeners received it. The stale listeners from earlier proxies attempted to process messages meant for the current proxy, causing internal Comlink state corruption (`TypeError: Cannot read properties of undefined`). The current proxy's promise never resolved, so `releaseWorker()` in the finally block never executed and the worker was **permanently** stuck.

With a pool of 8 workers, once all 8 were stuck, every subsequent `getWorker()` call blocked forever, no point cloud or CAD model data could load.

This PR fixes the issue

In `examples` and in `fusion` (both `unified-3d` and `3d-management`)

- run `pnpm install && pnpm build` in `viewer` and then run `yarn && yarn start` in `examples`
- Load [example](https://localhost:3549/?project=3d-test&env=greenfield-3d&modelId=6733832742816610&revisionId= 8936381564697305) (which already contains a point-cloud model)
- Now add new new model using using Models GUI. with modelId - `1304266634014089`, revisionId - `271145091694327`
- Both point cloud model should be visible

### DIFF
--- a/viewer/packages/pointclouds/src/potree-three-loader/loading/EptBinaryLoader.ts
+++ b/viewer/packages/pointclouds/src/potree-three-loader/loading/EptBinaryLoader.ts
@@ -86,7 +86,7 @@ export class EptBinaryLoader implements ILoader {
 
   async parse(node: PointCloudEptGeometryNode, data: ArrayBuffer): Promise<ParsedEptData | Error> {
     const autoTerminatingWorker = await EptBinaryLoader.WORKER_POOL.getWorker();
-    const eptDecoderWorker = Comlink.wrap<EptBinaryDecoderWorker>(autoTerminatingWorker.worker);
+    const eptDecoderWorker = autoTerminatingWorker.getComlinkProxy<EptBinaryDecoderWorker>();
     const eptData: EptInputData = {
       buffer: data,
       schema: node.ept.schema,
@@ -99,18 +99,22 @@ export class EptBinaryLoader implements ILoader {
       .filter(objAndBox => objAndBox[1].intersectsBox(node.boundingBox))
       .map(objAndBox => objAndBox[0]);
 
-    const result = await eptDecoderWorker(
-      Comlink.transfer(eptData, [eptData.buffer]),
-      relevantObjects,
-      node.boundingBox.min.toArray(),
-      {
-        min: node.boundingBox.min.toArray(),
-        max: node.boundingBox.max.toArray()
-      }
-    );
-
-    EptBinaryLoader.WORKER_POOL.releaseWorker(autoTerminatingWorker);
-    return result;
+    try {
+      const result = await eptDecoderWorker(
+        Comlink.transfer(eptData, [eptData.buffer]),
+        relevantObjects,
+        node.boundingBox.min.toArray(),
+        {
+          min: node.boundingBox.min.toArray(),
+          max: node.boundingBox.max.toArray()
+        }
+      );
+      return result;
+    } catch (err) {
+      return err as Error;
+    } finally {
+      EptBinaryLoader.WORKER_POOL.releaseWorker(autoTerminatingWorker);
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


viewer/packages/pointclouds/src/potree-three-loader/loading/EptBinaryLoader.ts
```diff
@@ -99,18 +99,22 @@ export class EptBinaryLoader implements ILoader {
...
+        relevantObjects,
+        node.boundingBox.min.toArray(),
+        {
+          min: node.boundingBox.min.toArray(),
+          max: node.boundingBox.max.toArray()
+        }
+      );
+      return result;
+    } catch (err) {
+      return err as Error;
```
danpriori, line 114:
```suggestion
      return err instanceof Error ? err : new Error(String(err));
```

Not crucial, just a suggestion. To be more strict to an actual Error at runtime

viewer/packages/pointclouds/src/potree-three-loader/utils/WorkerPool.ts
```diff
@@ -16,13 +18,27 @@ export class AutoTerminatingWorker {
     return this._wrappedWorker!;
   }
 
+  public getComlinkProxy<T>(): Comlink.Remote<T> {
```
danpriori, line 21: The `getComlinkProxy<T>()` method has an unsound generic. 
This compiles fine today because getComlinkProxy is only called from the EptBinaryLoader and with the correct type. But if another loader or consumer reuses WorkerPool in the future and passes the wrong T, the compiler will not warn and it will silently return the same proxy typed as a completely different interface.
One way to avoid this is to move T to the class level so the type is fixed at construction time, not at each call site. 
Something like: 

```typescript
export class AutoTerminatingWorker<TProxy> {
  private _comlinkProxy: Comlink.Remote<TProxy> | undefined;

  public getComlinkProxy(): Comlink.Remote<TProxy> {
    if (this.terminated || !this._wrappedWorker) {
      throw new Error('Cannot get Comlink proxy for a terminated worker');
    }
    if (!this._comlinkProxy) {
      this._comlinkProxy = Comlink.wrap<TProxy>(this._wrappedWorker);
    }
    return this._comlinkProxy;
  }
}

export class WorkerPool<TWorker extends Worker, TProxy> {
  public getWorker(): Promise<AutoTerminatingWorker<TProxy>> { ... }
}
```

And then we can have something like:

```typescript
// EptBinaryLoader.ts
static readonly WORKER_POOL = new WorkerPool<EptDecoderWorker, EptBinaryDecoderWorker>(8, EptDecoderWorker);
```

But again, as this "hotfix" is urgent and the current code is safe as long as WorkerPool has a single consumer like this, we can for now post a comment to change it later if still make sense.

danpriori (changes requested):
Two comments. They are not crucial based on this context of urgency. But just to keep it in mind in case if does make sense.

viewer/packages/pointclouds/src/potree-three-loader/utils/WorkerPool.ts
pramod-cog: Good suggestion, we can consider this in later PR considering the fix urgency

danpriori (approved):
LGTM

pramod-cog:
risk review ok

PR #5692, opened 2026-06-19
chore(viewer): use happy-dom instead of jsdom and disable test isolation

christjt:
#### Type of change

Chore 

## Description :pencil:

Significantly improve test execution time by using happy-dom instead of js-dom as well as disabling test isolation.

### DIFF
--- a/viewer/packages/api/src/public/migration/Cognite3DViewer.test.ts
+++ b/viewer/packages/api/src/public/migration/Cognite3DViewer.test.ts
@@ -21,7 +21,6 @@ import type { DefaultImage360Collection, Image360ClusterIntersectionData, Image3
 
 import { vi } from 'vitest';
 import type { ResolutionOptions } from './types';
-import { visibility } from 'html2canvas/dist/types/css/property-descriptors/visibility';
 
 const sceneJson = (await import('./Cognite3DViewer.test-scene.json.json', { with: { type: 'json' } })).default;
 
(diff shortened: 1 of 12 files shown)


viewer/packages/api/src/public/migration/Cognite3DViewer.test.ts
```diff
@@ -208,6 +208,12 @@ describe('Cognite3DViewer', () => {
   });
 
   test('beforeSceneRendered and sceneRendered triggers before/after rendering', () => {
+    const computedStyleMock = vi.spyOn(window, 'getComputedStyle').mockImplementation(() => {
+      return {
+        visibility: 'visible',
+        display: 'block'
+      } as Partial<CSSStyleDeclaration> as CSSStyleDeclaration;
```
michalzagojski-cognite, line 215: Nitpick, but isn't it possible to slightly modify `Cognite3DViewer` to accept an optional `getComputedStyle` function in the constructor which would default to `window.getComputedStyle` and would be overridable for testing purposes? Perhaps that way we could avoid having this double cast here?
Also, since this parameter would be optional and would fallback to the desired function, we wouldn't have to implement it in all `Cognite3DViewer` instances across the codebase. 
But this is just an open question - perhaps this is not how things are done in reveal 👍

michalzagojski-cognite (changes requested):
Looks good 👍 Had just one question, lmk what you think

viewer/packages/api/src/public/migration/Cognite3DViewer.test.ts
christjt: I agree with the sentiment here, but Cognite3DViewer is the main entrypoint for Reveal, and this is really not something I want to expose out to the user of Reveal. I would very much like to get rid of this partial casting, but exposing this method out of Reveal is not a trade-off I want to take.

michalzagojski-cognite (approved):
LGTM 👍

christjt:
risk review ok

PR #5697, opened 2026-06-22
chore(viewer): use vite preview server for visual tests

christjt:
#### Type of change

Chore 

## Description :pencil:

Use the vite preview server to host visual tests instead of the vite dev server. This is a lot more performant since the browser will not need to download (and transpile) each individual module, but rather just use a static build.

### DIFF
--- a/viewer/visual-tests/visual-tests.vite.config.ts
+++ b/viewer/visual-tests/visual-tests.vite.config.ts
@@ -29,7 +29,7 @@ function readCdfEnv(): string {
   }
 }
 
-export default defineConfig(_ => {
+export default defineConfig(({ command }) => {
   const open = setTestFixture(process.env.testFixture);
   const cdfEnv = readCdfEnv();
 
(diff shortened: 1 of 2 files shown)


viewer/visual-tests/visual-tests.vite.config.ts
```diff
@@ -55,11 +55,9 @@ export default defineConfig(_ => {
     },
 
     build: {
+      assetsInlineLimit: Infinity,
       outDir: path.resolve(__dirname, 'dist'),
-      sourcemap: 'inline',
-      rollupOptions: {
```
danpriori, line 60: Was the intention behind removing `rollupOptions` here simply to use Vite's default configuration, or was there some other specific reason?

danpriori (approved):
Approved but I have just a question about the `rollupOptions`, which is not a blocker

viewer/visual-tests/visual-tests.vite.config.ts
christjt: So before this PR, build wasn't done at all what so ever. So rollupOptions has no effect. And yes, I can simply rely on vite's default behaviour, which is to look for index.html, which it finds.

christjt:
risk review ok

PR #5715, opened 2026-06-30
chore(viewer): add maxWait for consuming sector geometry

### LINKED ISSUE #5714: suggestion: Max Wait on the Debounced Consume Sectors functionality
Hi, when there are a lot of sectors to load and we have a full cache the debouncedConsumeSectors seem to never "debounce" until after all sectors are loaded. This in some cases causes the entire model to pop in after a few seconds instead of "streaming" in new sectors as they are loaded, and it gives an impression of higher loading times (since there is no visible progress until everything is loaded). The problem is bigger on slower machines.

I believe a maxWait property on the debounce will allow visible loading, and make the model usable while it loads.

https://github.com/cognitedata/reveal/blob/9df2bcaa72903c37a248467edbab8ef0ecabb35d/viewer/packages/cad-geometry-loaders/src/CadManager.ts#L125-L130

Suggestion:

```ts
const debouncedConsumeSectors = batchedDebounce(
  (sectors: ConsumedSector[]) => {
    for (const sector of sectors) {
      consumeNextSector(sector);
    }
    this._cadModelUpdateHandler.reportNewSectorsLoaded(sectors.length);
  },
  this._sectorBufferTime,
  { maxWait: 1_000 } // <--- Arbitrary max wait in ms
);
```

christjt:
#### Type of change

Chore 

Fixes: https://github.com/cognitedata/reveal/issues/5714

## Issue description:
"When there are a lot of sectors to load and we have a full cache the debouncedConsumeSectors seem to never "debounce" until after all sectors are loaded. This in some cases causes the entire model to pop in after a few seconds instead of "streaming" in new sectors as they are loaded, and it gives an impression of higher loading times (since there is no visible progress until everything is loaded). The problem is bigger on slower machines.

`maxWait` property on the debounce will allow visible loading, and make the model usable while it loads."

### DIFF
--- a/viewer/packages/cad-geometry-loaders/src/CadManager.ts
+++ b/viewer/packages/cad-geometry-loaders/src/CadManager.ts
@@ -122,12 +122,16 @@ export class CadManager {
       this.updateTreeIndexToSectorsMap(cadModel, sector);
     };
 
-    const debouncedConsumeSectors = batchedDebounce((sectors: ConsumedSector[]) => {
-      for (const sector of sectors) {
-        consumeNextSector(sector);
-      }
-      this._cadModelUpdateHandler.reportNewSectorsLoaded(sectors.length);
-    }, this._sectorBufferTime);
+    const debouncedConsumeSectors = batchedDebounce(
+      (sectors: ConsumedSector[]) => {
+        for (const sector of sectors) {
+          consumeNextSector(sector);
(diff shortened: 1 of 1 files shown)


haakonflatval-cognite (approved):
LGTM 👍

christjt:
risk review ok

PR #5725, opened 2026-07-17
fix(viewer): add a try-catch block to check if the new endpoint is working/exists when requesting signed files. Fallback if it doesn't

danpriori:
…

#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:
This change adds a check to see if the new endpoint for signed URLs is working in the current project via a dummy request. If it fails, it uses the fallback to the old file request system.

Open a CAD or Point Cloud model in a DM project that doesn't have signed URL enabled. Observe the network traffic. You'll notice that it attempts a dummy request and fails. Therefore, it starts loading the model files using the old method.

It is necessary to test in projects that are DM but are not enabled to use the signed URL endpoint, such as `3d-dms-int-tests-greenfield`.


haakonflatval-cognite (dismissed):
LGTM 👍

viewer/packages/cad-parsers/src/metadata/CadModelMetadataRepository.test.ts
```diff
@@ -132,6 +132,34 @@ describe(CadModelMetadataRepository.name, () => {
 
     expect(result.scene.root.signedUrl).toBe(sectorSignedUrl);
   });
+
+  test('DM model falls back to base URL fetching when loading via signed files fails', async () => {
+    const signedFilesBaseUrl = 'https://api.cognitedata.com/api/v1/projects/myproj/3d/output/files';
```
atussak, line 137: Just to be sure, would this test try to call this url?

atussak (commented):
🦄 
Looks good, just want to verify the test is not trying to call the api 👀

viewer/packages/cad-parsers/src/metadata/CadModelMetadataRepository.test.ts
danpriori: Hey! I think no method with the potential to call HTTP is left without a mock in any of these tests. I also checked the network activity while running the tests to confirm :) 

The URL `https://api.cognitedata.com/api/v1/projects/myproj/3d/output/files` only exists as a comparison value in memory during testing.

atussak (approved):
🦄

haakonflatval-cognite (approved):

PR #5730, opened 2026-07-22
fix(viewer): add missing mechanism to refresh signedURL links when it is expired and update the list and the cache

danpriori:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:
It adds a missing mechanism to refresh signedUrl links when it is expired and update the list and the cache.

Tested in DM and classic projects. For DM projects with signedURL enabled, waited more than 1 hour for the signedURLs expiration in order to force request a new signedURL for each file 

In a DM project with signedURL activated (for example: 3d-dev-quickstart -> Dan scene new output) after loaded the models, wait for 1 hour to let the signedURLs loaded been expired and move the camera to force request those again. Observe the network traffic when the file is denied (error 403) and right after that a new request for a new signedURL link is done and then a new binary file with the new signedURL link is automatically requested it again. 

[screen-capture (42).webm](https://github.com/user-attachments/assets/2cdb9706-63fd-4e27-9ec5-a5d20c697e0a)

### DIFF
--- a/viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
+++ b/viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
@@ -39,8 +39,8 @@ export class CdfModelDataProvider implements ModelDataProvider {
       }
       throw new Error('Could not download binary file');
     });
-    if (isBaseUrlEmpty === true) {
-      if (response.ok === false) {
+    if (isBaseUrlEmpty) {
+      if (!response.ok) {
         throw createHttpErrorFromResponse(response);
       }
       return decompressBinaryResponseBody(response);
@@ -59,7 +59,7 @@ export class CdfModelDataProvider implements ModelDataProvider {
     const response = await this.fetchWithRetry(fileName, { headers, method: 'GET' }).catch(() => {
       throw Error('Could not download signed JSON file');
     });
-    if (response.ok === false) {
+    if (!response.ok) {
       throw createHttpErrorFromResponse(response);
     }
     return parseJsonResponseBody(response);
--- a/viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
+++ b/viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
@@ -1,7 +1,7 @@
 /*!
  * Copyright 2021 Cognite AS
  */
-import type { CogniteClient, HttpRequestOptions } from '@cognite/sdk';
+import { HttpError, type CogniteClient, type HttpRequestOptions } from '@cognite/sdk';
 import type { ModelDataProvider } from '../ModelDataProvider';
 import { DMModelIdentifier } from '../model-identifiers/DMModelIdentifier';
 import type { SignedFileItem } from '../types';
@@ -39,7 +39,10 @@ export class CdfModelDataProvider implements ModelDataProvider {
       }
       throw new Error('Could not download binary file');
     });
-    if (isBaseUrlEmpty) {
+    if (isBaseUrlEmpty === true) {
+      if (response.ok === false) {
+        throw createHttpErrorFromResponse(response);
+      }
       return decompressBinaryResponseBody(response);
     }
     return response.arrayBuffer();
@@ -57,7 +60,7 @@ export class CdfModelDataProvider implements ModelDataProvider {
       throw Error('Could not download signed JSON file');
     });
     if (response.ok === false) {
-      throw new Error(`Signed JSON file request failed with status ${response.status}`);
+      throw createHttpErrorFromResponse(response);
     }
     return parseJsonResponseBody(response);
   }
@@ -135,4 +138,16 @@ export class CdfModelDataProvider implements ModelDataProvider {
     }
     throw error;
   }
+}
+
+function createHttpErrorFromResponse(response: Response): HttpError {
+  const headers: { [key: string]: string } = {};
+  response.headers.forEach((value, key) => {
+    headers[key] = value;
+  });
+  return new HttpError(
+    response.status,
+    { error: { code: response.status, message: response.statusText || 'Signed URL request failed' } },
+    headers
+  );
 }
--- a/viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
+++ b/viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
@@ -4,10 +4,9 @@
 import { HttpError, type CogniteClient, type HttpRequestOptions } from '@cognite/sdk';
 import type { ModelDataProvider } from '../ModelDataProvider';
 import { DMModelIdentifier } from '../model-identifiers/DMModelIdentifier';
-import type { FetchSignedOrClassicOptions, SignedFileItem } from '../types';
+import type { SignedFileItem } from '../types';
 import { stripRestrictedApiGateway } from '../utilities/signedUrlUtils';
 import { decompressBinaryResponseBody, parseJsonResponseBody } from '../utilities/gzipUtils';
-import { SignedUrlRefresher } from '../utilities/signedUrlRefresh';
 import type { ModelIdentifier } from '../ModelIdentifier';
 
 /**
@@ -16,21 +15,10 @@ import type { ModelIdentifier } from '../ModelIdentifier';
 export class CdfModelDataProvider implements ModelDataProvider {
   private readonly client: CogniteClient;
   private authenticationPromise: Promise<string | undefined>;
-  private readonly signedUrlRefresher = new SignedUrlRefresher(this);
 
   constructor(client: CogniteClient) {
     this.client = client;
     this.authenticationPromise = client.authenticate();
-  }
-
-  async getBinaryFileWithRefresher(
-    options: FetchSignedOrClassicOptions & { abortSignal?: AbortSignal }
-  ): Promise<ArrayBuffer> {
-    return this.signedUrlRefresher.getBinaryFileWithRefresher(options);
-  }
-
-  async getJsonFileWithRefresher(options: FetchSignedOrClassicOptions): Promise<unknown> {
-    return this.signedUrlRefresher.getJsonFileWithRefresher(options);
   }
 
   public async getBinaryFile(baseUrl: string, fileName: string, abortSignal?: AbortSignal): Promise<ArrayBuffer> {
--- a/viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
+++ b/viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
@@ -4,9 +4,10 @@
 import { HttpError, type CogniteClient, type HttpRequestOptions } from '@cognite/sdk';
 import type { ModelDataProvider } from '../ModelDataProvider';
 import { DMModelIdentifier } from '../model-identifiers/DMModelIdentifier';
-import type { SignedFileItem } from '../types';
+import type { FetchSignedOrClassicOptions, SignedFileItem } from '../types';
 import { stripRestrictedApiGateway } from '../utilities/signedUrlUtils';
 import { decompressBinaryResponseBody, parseJsonResponseBody } from '../utilities/gzipUtils';
+import { SignedUrlRefresher } from '../utilities/signedUrlRefresh';
 import type { ModelIdentifier } from '../ModelIdentifier';
 
 /**
@@ -15,10 +16,21 @@ import type { ModelIdentifier } from '../ModelIdentifier';
 export class CdfModelDataProvider implements ModelDataProvider {
   private readonly client: CogniteClient;
   private authenticationPromise: Promise<string | undefined>;
+  private readonly signedUrlRefresher = new SignedUrlRefresher(this);
 
   constructor(client: CogniteClient) {
     this.client = client;
     this.authenticationPromise = client.authenticate();
+  }
+
+  async getBinaryFileWithRefresher(
+    options: FetchSignedOrClassicOptions & { abortSignal?: AbortSignal }
+  ): Promise<ArrayBuffer> {
+    return this.signedUrlRefresher.getBinaryFileWithRefresher(options);
+  }
+
+  async getJsonFileWithRefresher(options: FetchSignedOrClassicOptions): Promise<unknown> {
+    return this.signedUrlRefresher.getJsonFileWithRefresher(options);
   }
 
   public async getBinaryFile(baseUrl: string, fileName: string, abortSignal?: AbortSignal): Promise<ArrayBuffer> {
--- a/viewer/packages/data-providers/src/utilities/SignedUrlRefresh.test.ts
+++ b/viewer/packages/data-providers/src/utilities/SignedUrlRefresh.test.ts
@@ -4,9 +4,9 @@
 
 import { vi } from 'vitest';
 import { HttpError } from '@cognite/sdk';
-import { SignedUrlRefresher } from './signedUrlRefresh';
 import type { ModelDataProvider } from '../ModelDataProvider';
 import { createMockModelDataProvider, mockDMModelIdentifier } from '../../../../test-utilities';
+import { SignedUrlRefresher } from './SignedUrlRefresh';
 
 const SIGNED_FILES_BASE_URL = 'https://signed-files.example.com';
 const CURRENT_SIGNED_URL = 'https://signed.example.com/file.bin';
--- a/viewer/packages/data-providers/src/utilities/SignedUrlRefresh.ts
+++ b/viewer/packages/data-providers/src/utilities/SignedUrlRefresh.ts
@@ -30,7 +30,7 @@ function findSignedFileItem(items: SignedFileItem[], fileName: string): SignedFi
   );
 }
 
-type FetchWithRefreshOptions<T> = {
+export type FetchWithRefreshOptions<T> = {
   currentSignedUrl: string | undefined;
   signedFilesBaseUrl: string | undefined;
   modelIdentifier: ModelIdentifier;
@@ -74,30 +74,7 @@ export class SignedUrlRefresher {
       }
     }
 
-    if (signedFilesBaseUrl !== undefined && this._dataProvider.getFileUrlsForModel !== undefined) {
-      const found = await this.refresh(
-        signedFilesBaseUrl,
-        modelIdentifier,
-        fileName,
-        (signedFilesBaseUrl: string, modelIdentifier: ModelIdentifier, fileName: string | undefined) => {
-          assert(
-            this._dataProvider.getFileUrlsForModel !== undefined,
-            'Model data provider does not support signed file fetching'
-          );
-          return this._dataProvider.getFileUrlsForModel(signedFilesBaseUrl, modelIdentifier, fileName);
-        }
-      );
-      if (found === undefined) {
-        throw new Error(`File "${fileName}" not found in signed files response`);
-      }
-
-      if (cacheKey !== undefined) {
-        this._refreshedSignedUrls.set(cacheKey, found.signedUrl);
-      }
-      onUrlRefreshed?.(found);
-      return fetchFn(found.signedUrl);
-    }
-    throw new Error('Model data provider does not support signed file fetching');
+    return this.tryRefresh(signedFilesBaseUrl, modelIdentifier, fileName, cacheKey, fetchFn, onUrlRefreshed);
   }
 
   private resolveCurrentSignedUrl(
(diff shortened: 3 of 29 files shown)


viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
```diff
@@ -39,7 +39,10 @@ export class CdfModelDataProvider implements ModelDataProvider {
       }
       throw new Error('Could not download binary file');
     });
-    if (isBaseUrlEmpty) {
+    if (isBaseUrlEmpty === true) {
+      if (response.ok === false) {
```
haakonflatval-cognite: Both of these are just `boolean`, so can't we just write 
```suggestion
    if (isBaseUrlEmpty) {
      if (!response.ok {
```
?

viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
```diff
@@ -57,7 +60,7 @@ export class CdfModelDataProvider implements ModelDataProvider {
       throw Error('Could not download signed JSON file');
     });
     if (response.ok === false) {
```
haakonflatval-cognite: [Nitpick]
```suggestion
    if (!response.ok) {
```

viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
```diff
@@ -136,3 +139,15 @@ export class CdfModelDataProvider implements ModelDataProvider {
     throw error;
   }
 }
+
+function createHttpErrorFromResponse(response: Response): HttpError {
```
haakonflatval-cognite, line 143: Just so I understand - is this to make the HTTP error visible in the console? Is that something we need? (Including the headers?)

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  private async refresh(
+    signedFilesBaseUrl: string,
+    modelIdentifier: ModelIdentifier,
+    candidates: string[]
+  ): Promise<SignedFileItem | undefined> {
+    const key = `${signedFilesBaseUrl}|${modelIdentifier.sourceModelIdentifier()}|${candidates[0]}`;
+    let promise = this.inFlightRefreshes.get(key);
+    if (promise === undefined) {
+      promise = this.dataProvider.getFileUrlsForModel!(signedFilesBaseUrl, modelIdentifier, candidates[0])
```
haakonflatval-cognite: We don't allow the `!` syntax - perhaps we can instead create a lambda to send here from the calling function, where we know that `getFileUrlsForModel` is defined?

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+/**
+ * Fetches a resource through a signed URL, transparently refreshing the URL and retrying once
+ * if the current URL has expired or is otherwise invalid (HTTP 401/403/404). Concurrent refresh
+ * requests for the same file are de-duplicated.
+ */
+export class SignedUrlRefresher {
+  private readonly dataProvider: ModelDataProvider;
+  private readonly inFlightRefreshes = new Map<string, Promise<SignedFileItem | undefined>>();
```
haakonflatval-cognite: We have the convention of putting `_` as prefix for private fields in Reveal
```suggestion
  private readonly _dataProvider: ModelDataProvider;
  private readonly _inFlightRefreshes = new Map<string, Promise<SignedFileItem | undefined>>();
```

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
```diff
@@ -0,0 +1,85 @@
...
+  return items.find(item =>
+    candidates.some(candidate => item.fileName === candidate || item.fileName.endsWith('/' + candidate))
+  );
+}
+
+type FetchWithRefreshOptions<T> = {
+  currentSignedUrl: string | undefined;
+  signedFilesBaseUrl: string | undefined;
+  modelIdentifier: ModelIdentifier;
+  candidates: string[];
```
haakonflatval-cognite: It's not entirely clear what `candidates` are 👀 Is there a better name we could use, or should we add TSDoc?

viewer/packages/sector-loader/src/GltfSectorLoader.ts
```diff
@@ -114,10 +118,18 @@ export class GltfSectorLoader {
...
-      return this._sectorFileProvider.getBinaryFile('', metadata.signedUrl, abortSignal);
+    const currentSignedUrl = this._refreshedSignedUrls.get(metadata.id) ?? metadata.signedUrl;
+    if (sector.modelIdentifier instanceof DMModelIdentifier && currentSignedUrl !== undefined) {
+      return this._signedUrlRefresher.fetchWithRefresh({
+        currentSignedUrl,
+        signedFilesBaseUrl: sector.signedFilesBaseUrl,
+        modelIdentifier: sector.modelIdentifier,
+        candidates: metadata.sectorFileName ? [metadata.sectorFileName] : [],
+        fetchFn: url => this._dataFileProvider.getBinaryFile('', url, abortSignal),
+        onUrlRefreshed: item => this._refreshedSignedUrls.set(metadata.id, item.signedUrl)
```
haakonflatval-cognite: I know we agreed to do a refactoring later - but this still hurts me too much 🙈 Could we store the refreshed signed URLs inside `signedUrlRefresher`, so that the loaders at least don't need to keep track of those? If we're lucky, that means we can get rid of the `currentSignedUrl` argument and `onUrlRefreshed`, which makes the interface much nicer in my opinion

viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
danpriori: Yes. It is to avoid having to swallow the error and show it with the meaningful info related to the error type.

viewer/packages/pointclouds/src/potree-three-loader/loading/EptBinaryLoader.ts
```diff
@@ -49,11 +52,23 @@ export class EptBinaryLoader implements ILoader {
     if (node.modelIdentifier instanceof DMModelIdentifier) {
       const signedUrl = await this.resolveSignedUrl(node);
       if (signedUrl !== undefined) {
-        return this._dataLoader.getBinaryFile('', signedUrl);
+        const fileName = node.fileName() + this.extension();
+        const filePath = `ept-data/${fileName}`;
+        return this.signedUrlRefresher.fetchWithRefresh({
+          currentSignedUrl: signedUrl,
+          signedFilesBaseUrl: node.signedFilesBaseUrl,
```
haakonflatval-cognite, line 62: On second thought - do we even need to think about refreshing the URL for `EptBinaryLoader`? Isn't it only loaded once, when the point cloud is first loaded anyway? 🤔 

If we that means we can remove the changes in this file, it's very tempting

viewer/packages/pointclouds/src/potree-three-loader/geometry/PointCloudEptGeometryNode.ts
```diff
@@ -305,34 +308,36 @@ export class PointCloudEptGeometryNode implements IPointCloudTreeGeometryNode {
     return map.get(nodeFileName);
   }
 
+  updateSignedFileItem(item: SignedFileItem): void {
+    if ('signedFiles' in this._eptMetadata === false || this._eptMetadata.signedFiles === undefined) return;
```
haakonflatval-cognite: [Nitpick]:
```suggestion
    if (!('signedFiles' in this._eptMetadata) || this._eptMetadata.signedFiles === undefined) return;
```

viewer/packages/pointclouds/src/potree-three-loader/geometry/PointCloudEptGeometryNode.ts
```diff
@@ -305,34 +308,36 @@ export class PointCloudEptGeometryNode implements IPointCloudTreeGeometryNode {
...
+    const items = this._eptMetadata.signedFiles.items;
+    const existingIndex = items.findIndex(existing => existing.fileName === item.fileName);
+    if (existingIndex !== -1) {
+      items[existingIndex] = item;
+    } else {
+      items.push(item);
+    }
+    PointCloudEptGeometryNode._signedFilesCache.delete(this._eptMetadata);
```
haakonflatval-cognite, line 324: Will this delete the entire cache, or how does this work? 🧐

viewer/packages/pointclouds/src/potree-three-loader/geometry/PointCloudEptGeometryNode.test.ts
```diff
@@ -112,6 +113,33 @@ describe(PointCloudEptGeometryNode.name, () => {
...
+        dataProvider,
+        mockDMModelIdentifier,
+        metadata,
+        SIGNED_FILES_BASE_URL
+      );
+
+      await node.getHierarchy(fileName);
+
+      expect(getJsonFileMock).toHaveBeenLastCalledWith('', freshUrl);
+      expect(metadata.signedFiles!.items.find(i => i.fileName === filePath)?.signedUrl).toBe(freshUrl);
```
haakonflatval-cognite:
```suggestion
      expect(metadata.signedFiles?.items.find(i => i.fileName === filePath)?.signedUrl).toBe(freshUrl);
```
🤞

haakonflatval-cognite (commented):
A few more comments - otherwise I think it's all good 👌

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
danpriori: maybe another name. It is basically a 2-tupled/pair which contains the filePath and fileName and it will try to find the item from the signedFileItems list based on the fileName itself or based on the file path in case it is stored like that.

viewer/packages/pointclouds/src/potree-three-loader/geometry/PointCloudEptGeometryNode.ts
danpriori: It deletes for that metadata only for this node. When the getSignedUrlMap is called, it will check if the cache still exist, which is not, then it will make some checks before populate it with the new signedURL again.

viewer/packages/pointclouds/src/potree-three-loader/loading/EptBinaryLoader.ts
danpriori: Unfortunately not, I think. It is being called dynamically while moving the camera to get more bin files within the potree, for example.

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
danpriori: https://github.com/cognitedata/reveal/pull/5730/commits/a5503070613f9d9688369d65c45b4a27a4010fbb

viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
danpriori: https://github.com/cognitedata/reveal/pull/5730/commits/a5503070613f9d9688369d65c45b4a27a4010fbb

viewer/packages/data-providers/src/model-data-providers/CdfModelDataProvider.ts
danpriori: https://github.com/cognitedata/reveal/pull/5730/commits/a5503070613f9d9688369d65c45b4a27a4010fbb

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
danpriori: https://github.com/cognitedata/reveal/pull/5730/commits/a5503070613f9d9688369d65c45b4a27a4010fbb

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
danpriori: https://github.com/cognitedata/reveal/pull/5730/commits/274cbd6e8fba6ae17af62339a3272c63a3fb6b77

viewer/packages/sector-loader/src/GltfSectorLoader.ts
danpriori: I agree, but this is what I think can be done for now. 
https://github.com/cognitedata/reveal/pull/5730/commits/274cbd6e8fba6ae17af62339a3272c63a3fb6b77

viewer/packages/pointclouds/src/potree-three-loader/geometry/PointCloudEptGeometryNode.ts
danpriori: https://github.com/cognitedata/reveal/pull/5730/commits/274cbd6e8fba6ae17af62339a3272c63a3fb6b77

viewer/packages/pointclouds/src/potree-three-loader/geometry/PointCloudEptGeometryNode.test.ts
danpriori: https://github.com/cognitedata/reveal/pull/5730/commits/274cbd6e8fba6ae17af62339a3272c63a3fb6b77

viewer/packages/pointclouds/src/potree-three-loader/loading/EptBinaryLoader.ts
haakonflatval-cognite: I'm confused :sweat_smile: What exactly is being loaded in this file, and what is being loaded in the `PointCloudEptGeometryNode.ts` source file? Isn't one of them loading just the hierarchy information on point cloud load, while the other loads the point cloud geometry data?

viewer/packages/pointclouds/src/potree-three-loader/geometry/PointCloudEptGeometryNode.ts
```diff
@@ -305,34 +308,36 @@ export class PointCloudEptGeometryNode implements IPointCloudTreeGeometryNode {
...
+  updateSignedFileItem(item: SignedFileItem): void {
+    if (!('signedFiles' in this._eptMetadata) || this._eptMetadata.signedFiles === undefined) return;
+    const items = this._eptMetadata.signedFiles.items;
+    const existingIndex = items.findIndex(existing => existing.fileName === item.fileName);
+    if (existingIndex !== -1) {
+      items[existingIndex] = item;
+    } else {
+      items.push(item);
+    }
+    PointCloudEptGeometryNode._signedFilesCache.delete(this._eptMetadata);
```
haakonflatval-cognite, line 324: Hmm, this is a bit problematic, when I think about it - we really should not have mutable static data in a class. If the user has multiple Reveal instances on the same page (like Canvas does), they will share and mutate this data together, which I think is a bad idea in general. Can we just make it a private non-static field?

viewer/packages/data-providers/src/utilities/SignedUrlRefresh.test.ts
haakonflatval-cognite, line 1: Can we rename these files to `SignedUrlRefresher.*ts` instead?

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
```diff
@@ -0,0 +1,142 @@
...
+  return items.find(
+    item =>
+      item.fileName === fileName ||
+      item.fileName === fileBaseName ||
+      item.fileName.endsWith('/' + fileName) ||
+      item.fileName.endsWith('/' + fileBaseName)
+  );
+}
+
+type FetchWithRefreshOptions<T> = {
```
haakonflatval-cognite: This is used in the signature of `fetchWithRefresh`, so it should also be exported

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
```diff
@@ -0,0 +1,142 @@
...
+      const found = await this.refresh(
+        signedFilesBaseUrl,
+        modelIdentifier,
+        fileName,
+        (signedFilesBaseUrl: string, modelIdentifier: ModelIdentifier, fileName: string | undefined) => {
+          assert(
+            this._dataProvider.getFileUrlsForModel !== undefined,
+            'Model data provider does not support signed file fetching'
+          );
+          return this._dataProvider.getFileUrlsForModel(signedFilesBaseUrl, modelIdentifier, fileName);
```
haakonflatval-cognite: An alternative to make this slightly nicer, is perhaps to call the `refresh` method `tryRefresh` instead, and let it do this `if`-check internally, and return whether it succeeded or not

viewer/packages/data-providers/src/utilities/SignedUrlRefresh.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+function findSignedFileItem(items: SignedFileItem[], fileName: string): SignedFileItem | undefined {
+  const fileBaseName = baseName(fileName);
+  return items.find(
+    item =>
+      item.fileName === fileName ||
+      item.fileName === fileBaseName ||
+      item.fileName.endsWith('/' + fileName) ||
+      item.fileName.endsWith('/' + fileBaseName)
```
haakonflatval-cognite, line 29: Just to be sure - do we need to guard against all these cases? Is it not possible for the data loaders (`GltfSectorLoader`, `EptBinaryLoader`, etc.) to send the correct name directly? (I understand this is related to our discussion yesterday, but I'm not convinced that there can be so much unpredictability from the back-end 🤔

haakonflatval-cognite (commented):
A few more comments

viewer/packages/pointclouds/src/potree-three-loader/geometry/PointCloudEptGeometryNode.ts
danpriori: Actually, it’s keyed by the metadata object ref (not by model id or any string id), and each Reveal load creates its own metadata instance. So multiple viewers on the same page don’t share or mutate each other’s cache entries. Also, making it a per-instance field on PointCloudEptGeometryNode wouldn’t fit either: that class is constructed once per octree node, and the cache must stay shared across the whole tree for a single metadata object. A per-node cache would either break invalidation on refresh or duplicate the same map on every node and to move this shared cache up then we would need to move it into the EptLoader, which is not great as well. 
So we can deal with this once we start working on https://cognitedata.atlassian.net/browse/BND3D-7284 in order to concentrate the signed url universe into a more isolated/self-contained space.

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
danpriori: https://github.com/cognitedata/reveal/pull/5730/commits/1802bd250672b4190d44c8f3563596dbd2fc4907

viewer/packages/data-providers/src/utilities/signedUrlRefresh.ts
danpriori: yeah nice catch https://github.com/cognitedata/reveal/pull/5730/commits/1802bd250672b4190d44c8f3563596dbd2fc4907

viewer/packages/data-providers/src/utilities/SignedUrlRefresh.test.ts
danpriori: https://github.com/cognitedata/reveal/pull/5730/commits/1802bd250672b4190d44c8f3563596dbd2fc4907

viewer/packages/data-providers/src/utilities/SignedUrlRefresh.ts
danpriori: We need to do these extra checks right now. I suggest investigating this further when we decouple any remaining involvement of the signed URLs in the loaders and the PointCloudEptGeometryNode in the ticket we will address this in, so we can fix how the loaders receive and handle the strings.

viewer/packages/pointclouds/src/potree-three-loader/loading/EptBinaryLoader.ts
danpriori: EptBinaryLoader loads the binary files.
The GeometryNode, more precisely in the getHierarchy part, loads only the hierarchy. 

In theory, the hierarchy can be loaded only at the beginning, but this cannot be guaranteed, because let's assume that certain models are not loaded at the beginning when opening the scene and are only loaded when activated in the TreeNode, and thus we have a mechanism to load any file related to the model only after activation. Asking the hierarchy file, which also has a signedURL attached to it, to go through the same fetchWithRefresh funnel ensures that the file will be delivered, maintaining resilience.

haakonflatval-cognite (approved):
LGTM 👍

haakonflatval-cognite (commented):
🦄

PR #5780, opened 2026-09-09
fix(pointclouds): grow nearby adaptive points and cap distant splats

msusag:
## Summary
- Raise the default Adaptive point-size maximum from 10 to 32 to reduce gaps during close-up inspection.
- Limit enlargement to nearby points: smoothly transition from the configured maximum at camera-space depth 0.5 to a distant cap of 6 at depth 3 and beyond. Caps are relative to a 1080-pixel drawing-buffer height; depths are world units (metres for metre-based models).
- Reduce oversized distant splats caused by coarse octree spacing, particularly on railings.
- Preserve Fixed and Attenuated defaults, explicit maximum-size overrides across mode changes, and the existing final minimum/maximum clamps. An explicit minimum can override the distant cap.
- Retain the numeric conversion for the example sizing-mode selector so enum comparisons work when switching modes.

## Implementation and scope
- The distance-dependent cap adds only four lines to the existing Adaptive vertex-shader path; there is no additional render pass, render target, or per-point attribute.
- Update the model documentation, material/picking unit coverage, and the Adaptive GPU visual fixture and baseline.

## Local validation
- 15 targeted material and picking unit tests passed.
- AdaptivePointSize, FixedPointCloud, and EdlPointCloud visual tests passed.
- GPU assertions verify coarse-point caps of 6, 6, 19, and 32 at depths 8, 3, 1.75, and 0.5, plus explicit minimum handling and Fixed-size invariance.
- Library/test TypeScript checks, targeted ESLint, example formatting, and whitespace checks passed. Library typechecking and example formatting were rechecked after restoring pointSize to 1.0.
- Inspected near/far views of various point clouds, including walls, floors, pipes, and railings. 

## Limitations
- This is a depth-based cap, not geometric edge detection or density estimation; it cannot eliminate every bulky silhouette or fill every sparse surface.
- Very close views can still expose individual splats and gaps. The finite near cap balances coverage against excessive point growth.
- Thresholds use camera-space depth, not Euclidean camera distance, and assume an appropriate model scale.


msusag:
@pramod-cog, I converted this PR from the last hackaton back to draft. 
There is perhaps only some parts of this PR that makes sense to merge.

PR #5789, opened 2026-09-14
fix: add WebGL Context lost state and skipper, add dispose in styletexture and objects, lru factor

danpriori:
#### Type of change

Bug 

#### Jira ticket :blue_book:

## Description :pencil:
Add Wegbl Context lost state and skipper, add dispose in styletexture and objects, lru factor


viewer/packages/360-images/src/cache/Image360LoadingCache.ts
```diff
@@ -141,6 +142,24 @@ export class Image360LoadingCache<T extends DataSourceType> {
   }
 
   private purgeRevision(entity: Image360Entity<T>, revision: Image360RevisionEntity<T>) {
+    // If the WebGL context is currently lost, calling revision.dispose() /
+    // entity.unloadImage() would tear down textures/materials on a fresh GL
+    // context using stale handles -> INVALID_OPERATION errors and stranded VRAM.
+    // Defer: we still evict from the JS cache so we don't cache stale revisions,
+    // but skip the GL-level teardown until the context is restored.
+    if (isWebGLContextLost()) {
```
haakonflatval-cognite, line 150: This clause seems to do the same as the lines below, excluding the last two lines. Can we rather do a conditional return before the last two lines for the same effect? 

Another thing - will we even have the chance to dispose this revision later at all, even if the GL context recovers? It will be purged from the cache, so won't it effectively be orphaned?

viewer/packages/utilities/src/webglContextState.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Tracks whether the WebGL context is lost; auto-dispose paths (Potree LRU, Image360
+ * cache) should skip work while lost and resume after `webglcontextrestored`.
+ */
+
+let contextLost = false;
+
+export function isWebGLContextLost(): boolean {
+  return contextLost;
+}
```
haakonflatval-cognite, line 14: It makes sense to have this global since the WebGL context is global, but I'm wondering if there is another way to do this that would be more easily to migrate to WebGPU 🤔 It may not be worth it to solve that right now though

viewer/packages/pointclouds/src/potree-three-loader/utils/lru.ts
```diff
@@ -132,7 +139,14 @@ export class LRU {
...
     }
 
-    while (this.numPoints > this.pointBudget * 2) {
+    // Skip while the WebGL context is lost: disposing would delete stale buffer
+    // handles and strand VRAM. Safe to defer since freeMemory() runs every tick.
+    if (isWebGLContextLost()) {
+      return;
+    }
+
+    const threshold = this.pointBudget * LRU_OVERSHOOT_FACTOR;
```
haakonflatval-cognite, line 148: Does this change in ovenshoot factor have a significant impact on memory? I'm worried that it may cause many more network calls in some cases; but maybe it's not as impactful as I fear 😅

haakonflatval-cognite (commented):
Had some suggestions, but otherwise I think this is a good improvement 👍 

I see there are quite heavy usage of comments throughout the new code. Perhaps we can shorten some of the comments a little (I find it a little noisy when there's a lot of comments to read)

PR #5791, opened 2026-09-21
feat: add full screen picking for point cloud

pramod-cog:
#### Type of change

Feat  

#### Jira ticket :blue_book:

Part of [BND3D-7086](https://cognitedata.atlassian.net/browse/BND3D-7086)

## Description :pencil:
Added full frame picking with cache for point cloud to cut down repeated GPU work when the same frame is picked against multiple times in a row (e.g. hover-driven picking or several picks issued before the camera has moved).

Previously, every `PointCloudOctreePicker.pick()` call rendered a small window around the mouse cursor into a pick render target and read it back, a full render + GPU readback per pick, even if nothing in the scene had changed since the last pick. This cause performance issue on lower configured machine on some of the operation/features like mouse-move picking.

In examples and unified-3d

- Load a scene with a point cloud model and 360 image collection in `unified-3d` (`fusion`).
- Enter into 360 image collection and move the mouse repeatedly over the scene without moving the camera, picking should keep returning correct hit points, now served from the cache after the first pick.
- Move/rotate/zoom the camera and confirm subsequent picks still resolve correctly with no visible regression in picking accuracy or responsiveness.

[BND3D-7086]: https://cognitedata.atlassian.net/browse/BND3D-7086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


viewer/packages/pointclouds/src/PointCloudPickingHandler.ts
```diff
@@ -33,6 +33,14 @@ export class PointCloudPickingHandler {
     this._picker.dispose();
   }
 
+  /**
+   * Invalidates the picker's cached full-frame pick buffer. Must be called whenever a new
+   * frame of the scene is rendered.
+   */
+  invalidatePickCache(): void {
```
danpriori, line 40: Where this should be called? I can't find where it has being used 🤔.
Is it should be in Cognite3DViewer right after where we call the render inside the animate?:

```typescript
this.revealManager.render(camera);
this._pointCloudPickingHandler.invalidatePickCache(); // here?
```
Although I don't like it (I don't think it is the best place to put it, tbh), just a question where you were thinking to use it

viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePicker.ts
```diff
@@ -79,4 +163,126 @@ export class PointCloudOctreePicker {
...
+    }
+    if (cache.octrees.length !== octrees.length || !cache.octrees.every((octree, i) => octree === octrees[i])) {
+      return false;
+    }
+    // Nodes may have been unloaded by LOD updates since the cache was built which nulls the
+    // scene node geometry.
+    if (cache.renderedNodes.some(({ node }) => node.sceneNode.geometry === undefined)) {
+      cache.valid = false;
+      return false;
+    }
```
danpriori, line 192: Shouldn't we call the invalidateCache() when the isCacheUsable returns false?

viewer/packages/rendering/src/pointcloud-rendering/PointCloudMaterial.ts
```diff
@@ -215,6 +218,15 @@ export class PointCloudMaterial extends RawShaderMaterial {
...
   }
 
+  get nodeIndexBits(): number {
+    return this.getUniform('nodeIndexBits');
+  }
+  set nodeIndexBits(value: number) {
+    if (value !== this.getUniform('nodeIndexBits')) {
+      this.setUniform('nodeIndexBits', value);
+    }
+  }
```
danpriori, line 228: I would ask to have tests for these new getters/setters, but I understand if you prefer to skip it this time and put it in another PR later, as this file doesn't have the scaffolding yet, which would require hundreds of additional lines of code.  🙈

viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePicker.ts
```diff
@@ -79,4 +163,126 @@ export class PointCloudOctreePicker {
...
+  }
+
+  private async buildCache(
+    camera: Camera,
+    octrees: PointCloudOctree[],
+    params: Partial<PickParams>,
+    width: number,
+    height: number
+  ): Promise<boolean> {
+    const pickState = this.pickState!;
```
danpriori: Better narrowing instead of non-null assertion

viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePickerHelper.ts
```diff
@@ -168,33 +177,35 @@ export class PointCloudOctreePickerHelper {
...
+    nodeIndexOffset: number,
+    maxNodeIndex: number
   ): Points[] {
     const tempNodes: Points[] = [];
     for (let i = 0; i < nodes.length; i++) {
       const node = nodes[i];
+      const nodeIndex = nodeIndexOffset + i + 1;
+      if (nodeIndex > maxNodeIndex) {
+        console.error(`More than ${maxNodeIndex} nodes for pick are not supported.`);
+        continue;
```
danpriori: As this is reaching the limit, it's not better to use "break" instead of "continue" here?

viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePickerHelper.ts
```diff
@@ -101,16 +103,22 @@ export class PointCloudOctreePickerHelper {
...
 
+    // Node index 0 means "no hit" and the all-ones value is rejected by findHit, so the largest usable index is 2^nodeIndexBits - 2.
+    const maxNodeIndex = 2 ** nodeIndexBits - 2;
+    pickMaterial.nodeIndexBits = nodeIndexBits;
+
     const renderedNodes: RenderedNode[] = [];
     for (const octree of octrees) {
-      // Get all the octree nodes which intersect the picking ray. We only need to render those.
-      const nodes = PointCloudOctreePickerHelper.nodesOnRay(octree, ray);
+      // Get all the octree nodes which intersect the picking ray. We only need to render those)
```
danpriori:
```suggestion
      // Get all the octree nodes which intersect the picking ray. We only need to render those
```

viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePicker.ts
```diff
@@ -79,4 +163,126 @@ export class PointCloudOctreePicker {
...
+    return true;
+  }
+
+  private async buildCache(
+    camera: Camera,
+    octrees: PointCloudOctree[],
+    params: Partial<PickParams>,
+    width: number,
+    height: number
+  ): Promise<boolean> {
```
danpriori, line 203: I would suggest to break down this method into smaller pieces/private functions in order to understand the main parts of it, as it is too big and a bit complicated to follow.

viewer/packages/pointclouds/src/PointCloudPickingHandler.ts
danpriori, line 29: Question(out of this scope, but it is related): why we have two picker instances: one here and another one created at PointCloudOctree inside the pick() ?

danpriori (changes requested):
Some suggestions. The rest looks good, although some potential archictetural concerns but it is out of this scope, it seems.
I'm gonna try to test it too. Nice job!

viewer/packages/pointclouds/src/PointCloudPickingHandler.ts
pramod-cog: Existing dead code I guess

viewer/packages/pointclouds/src/PointCloudPickingHandler.ts
pramod-cog: Nice catch

viewer/packages/rendering/src/pointcloud-rendering/PointCloudMaterial.ts
pramod-cog: Will leave it as in, already crossed the limit in the PR, but have added more test in following PR to cover most of the changes

viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePicker.ts
```diff
@@ -184,7 +184,7 @@ export class PointCloudOctreePicker {
     // Nodes may have been unloaded by LOD updates since the cache was built which nulls the
     // scene node geometry.
     if (cache.renderedNodes.some(({ node }) => node.sceneNode.geometry === undefined)) {
-      cache.valid = false;
+      this.invalidateCache();
```
danpriori, line 187: Sorry, I was actually referring to all the `if` conditions that also return `false` (all the others above too), indicating that the cache isn't usable, meaning the cache needs to be invalidated, if that makes sense. If yes, then probably it can be invalidated outside at once, instead.

danpriori (commented):
Sorry, one last comment that might not have been clear. See if it makes sense. Everything else seems to be fine.

viewer/packages/pointclouds/src/potree-three-loader/tree/PointCloudOctreePicker.ts
danpriori: As we have discussed, we souldnt force invalidate the cache in any of these if statements here, as you already reverted this one here. All good then 👍

danpriori (approved):
LGTM

pramod-cog:
risk review ok

